PR #8, opened 2021-05-04
Update burger-sizes.demo.tsx

cstrat:
The title text is incorrect in the demo.

Hovering on the unopened burger should say 'Open' and hovering on the `x` should say 'Close'.


rtivital:
Thanks for Pull Request!

Issue #28, opened 2021-05-05
Add new props to the Hr element


khushal87:
Currently, the `Hr` component has only three props as `variant, size, and color`. We can have some more useful props such as inset(left, right, middle), orientation(vertical, horizontal), subHeader, and width. 

Note: I am considering Hr in this package to be a standard `Divider`

Progress:

- [x] Inset
- [x] Orientation
- [x] subHeader
- [x] subHeaderStyle
- [x] Docs
- [x] Tests
- [x] Stories

khushal87:
Hello admins! I want to work on this issue if it seems relevant to you all. Thank you. @cstrat @rtivital @srezanova

rtivital:
Hi @khushal87 sounds good!
Can you describe what subHeader prop will do?
Also please make sure that component works fine with [Menu](https://mantine.dev/core/menu/) and [Card](https://mantine.dev/core/card/) after the changes (in horizontal orientation only), since these components have some context styles for Hr.

khushal87:
@rtivital Thanks, sure I will check the following. For subheaders, you can refer https://material.io/components/dividers#usage.
Also, can you explain what is the realtime testing method you are using to test the changes in the package components visually, this will improve my workflow.

rtivital:
Yep, subHeader looks great, let's add it.
To get started locally:
- fork this repo
- clone your fork
- install dependencies with `yarn`
- run storybook – `npm run storybook` – we use storybook for local development, add stories that you need to test new features, don't worry on how it looks in storybook – it's not published anywhere
- don't forget to add props descriptions and tests
- after you finish adding new features run `npm run docs` – it will start local documentation server
- Add docs and demos to Hr component
- Done! Run npm test to check if everything is fine and send a PR

Here is a link to contributing guidelines – https://mantine.dev/pages/contribute/ with more in depth review

khushal87:
Thanks a lot, @rtivital. I will soon raise a PR for the same.

rtivital:
@khushal87 sorry, I forgot one thing, when you will create a PR, please request a merge to next-1.1.0 as this feature is planned to release in minor release

cstrat:
> Thank you. @cstrat @rtivital @srezanova

Don't include me in that list, I've just helped fix some typos. Hoping I can contribute more though!

khushal87:
@rtivital sure! 😄

khushal87:
[...]

PR #37, opened 2021-05-07
[Core, Docs] add additional props to Hr component

khushal87:
This PR introduces `new props` to the Hr component. The props are:

- [x] subHeader
- [x] subHeaderProps
- [x] orientation

**Have I added the tests/tested my changes?**
Yes

**Have I updated the docs for the changes?**
Yes

Reference issue: #28

### DIFF
--- a/docs/src/docs/demos/core/Hr/hr.demos.tsx
+++ b/docs/src/docs/demos/core/Hr/hr.demos.tsx
@@ -48,6 +48,30 @@ export function HrColorDemo() {
   );
 }
 
+const insetCode = `import React from 'react';
+import { Hr } from '@mantine/core';
+function Demo() {
+  return (
+    <>
+      <Hr inset={true} />
+      <Hr inset={true} insetType="left" />
+      <Hr inset={true} insetType="right" />
+      <Hr inset={true} insetType="middle" />
+    </>
+  );
+}`;
+
+export function HrInsetDemo() {
+  return (
+    <CodeDemo code={insetCode} language="tsx">
+      <Hr inset={true} />
+      <Hr inset={true} insetType="left" style={{ marginTop: 10 }} />
+      <Hr inset={true} insetType="right" style={{ marginTop: 10 }} />
+      <Hr inset={true} insetType="middle" style={{ marginTop: 10 }} />
+    </CodeDemo>
+  );
+}
+
 const sizesCode = `import React from 'react';
 import { Hr } from '@mantine/core';
 
@@ -73,6 +97,30 @@ export function HrSizesDemo() {
       <Hr size="lg" style={{ marginTop: 10 }} />
       <Hr size="xl" style={{ marginTop: 10 }} />
       <Hr size={10} style={{ marginTop: 10 }} />
+    </CodeDemo>
+  );
+}
+
+const subHeaderCode = `import React from 'react';
+import { Hr } from '@mantine/core';
+function Demo() {
+  return (
+    <>
+      <Hr subHeader="Test" />
+      <Hr inset={true} subHeader="Test" />
+      <Hr subHeader="Test" subHeaderStyle={{ color : "red" }} />
+      <Hr inset={true} insetType="right" subHeader="Test" />
+    </>
+  );
+}`;
+
+export function HrSubHeaderDemo() {
+  return (
+    <CodeDemo code={subHeaderCode} language="tsx">
+      <Hr subHeader="Test" />
+      <Hr inset={true} subHeader="Test" style={{ marginTop: 10 }} />
+      <Hr subHeader="Mantine" subHeaderStyle={{ color: 'red' }} style={{ marginTop: 10 }} />
+      <Hr inset={true} insetType="right" subHeader="Mantine" style={{ marginTop: 10 }} />
     </CodeDemo>
   );
 }
--- a/docs/src/docs/demos/core/Hr/hr.demos.tsx
+++ b/docs/src/docs/demos/core/Hr/hr.demos.tsx
@@ -1,5 +1,5 @@
 import React from 'react';
-import { Hr } from '@mantine/core';
+import { Hr, Badge } from '@mantine/core';
 import CodeDemo from '../../../../components/CodeDemo/CodeDemo';
 
 const code = `import React from 'react';
@@ -117,6 +117,53 @@ export function HrSubHeaderDemo() {
         style={{ marginTop: 10 }}
       />
       <Hr subHeader="Test" subHeaderProps={{ style: { fontSize: 20 } }} style={{ marginTop: 15 }} />
+    </CodeDemo>
+  );
+}
+
+const orientationCode = `import React from 'react';
+import { Hr } from '@mantine/core';
+
+function Demo() {
+  return (
+    <>
+      <Hr orientation="horizontal" />
+      <div
+        style={{
+          display: 'flex',
+          flexDirection: 'row',
+          justifyContent: 'space-evenly',
+          marginTop: 15,
+        }}
+      >
+        <Badge>Light</Badge>
+        <Hr orientation="vertical" />
+        <Badge variant="outline">Outline</Badge>
+        <Hr orientation="vertical" />
+        <Badge variant="filled">Filled</Badge>
+      </div>
+    </>
+  );
+}`;
+
+export function HrOrientationDemo() {
+  return (
+    <CodeDemo code={orientationCode} language="tsx">
+      <Hr orientation="horizontal" />
+      <div
+        style={{
+          display: 'flex',
+          flexDirection: 'row',
+          justifyContent: 'space-evenly',
+          marginTop: 15,
+        }}
+      >
+        <Badge>Light</Badge>
+        <Hr orientation="vertical" />
+        <Badge variant="outline">Outline</Badge>
+        <Hr orientation="vertical" />
+        <Badge variant="filled">Filled</Badge>
+      </div>
     </CodeDemo>
   );
 }
--- a/docs/src/docs/demos/core/Hr/hr.demos.tsx
+++ b/docs/src/docs/demos/core/Hr/hr.demos.tsx
@@ -19,7 +19,7 @@ export function HrBaseDemo() {
   return (
     <CodeDemo code={code} language="tsx">
       <Hr />
-      <Hr variant="solid" style={{ marginTop: 10 }} />
+      <Hr variant="dashed" style={{ marginTop: 10 }} />
       <Hr variant="dotted" style={{ marginTop: 10 }} />
     </CodeDemo>
   );
@@ -44,30 +44,6 @@ export function HrColorDemo() {
       <Hr color="blue" />
       <Hr color="red" style={{ marginTop: 10 }} />
       <Hr color="green" style={{ marginTop: 10 }} />
-    </CodeDemo>
-  );
-}
-
-const insetCode = `import React from 'react';
-import { Hr } from '@mantine/core';
-function Demo() {
-  return (
-    <>
-      <Hr inset={true} />
-      <Hr inset={true} insetType="left" />
-      <Hr inset={true} insetType="right" />
-      <Hr inset={true} insetType="middle" />
-    </>
-  );
-}`;
-
-export function HrInsetDemo() {
-  return (
-    <CodeDemo code={insetCode} language="tsx">
-      <Hr inset={true} />
-      <Hr inset={true} insetType="left" style={{ marginTop: 10 }} />
-      <Hr inset={true} insetType="right" style={{ marginTop: 10 }} />
-      <Hr inset={true} insetType="middle" style={{ marginTop: 10 }} />
     </CodeDemo>
   );
 }
@@ -107,9 +83,21 @@ function Demo() {
   return (
     <>
       <Hr subHeader="Test" />
-      <Hr inset={true} subHeader="Test" />
-      <Hr subHeader="Test" subHeaderStyle={{ color : "red" }} />
-      <Hr inset={true} insetType="right" subHeader="Test" />
+      <Hr
+        subHeader="Test"
+        subHeaderProps={{ style: { textAlign: 'center' }, color: 'red' }}
+        style={{ marginTop: 10 }}
+      />
+      <Hr
+        subHeader="Test"
+        subHeaderProps={{ style: { textAlign: 'end', fontFamily: 'cursive' } }}
+        style={{ marginTop: 10 }}
+      />
+      <Hr 
+        subHeader="Test" 
+        subHeaderProps={{ style: { fontSize: 20 } }} 
+        style={{ marginTop: 15 }} 
+      />
     </>
   );
 }`;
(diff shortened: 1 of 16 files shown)


rtivital:
About inset – currently it feels very complicated, I think it's not necessary as margin can achieve this in single prop + inset is fixed, it is impossible to connect it to layout:
```tsx
<Hr style={{ marginLeft: 70 }} />
// same as 
<Hr inset={true} insetType="left" />
```

rtivital:
First demo does not show dashed variant 
Screen Shot 2021-05-07 at 11 00 02 AM

src/mantine-core/src/Hr/Hr.tsx
```diff
@@ -2,30 +2,69 @@ import React from 'react';
...
+    variant,
+    orientation,
+  });
+
+  if (orientation === 'vertical') {
+    return <div data-mantine-hr className={cx(classes.hrVertical, className)} {...others} />;
+  }
+  return (
+    <div data-mantine-hr className={cx(classes.hr, className)} {...others}>
+      <Text className={cx(classes.subHeader)} style={{ ...subHeaderStyle }}>
```
rtivital, line 63: `{subHeader && <Text />}` – we do not want to render any extra react nodes

khushal87:
> About inset – currently it feels very complicated, I think it's not necessary as margin can achieve this in single-prop + inset is fixed, it is impossible to connect it to layout:
> 
> ```tsx
> <Hr style={{ marginLeft: 70 }} />
> // same as 
> <Hr inset={true} insetType="left" />
> ```

So adding an insetLength can work? Or do you want the removal of the `inset` prop?

src/mantine-core/src/Hr/Hr.tsx
rtivital: `subHeaderStyle -> subHeaderProps` – this will allow more customization

```tsx
<Text {...subHeaderProps} className={cx(classes.subHeader, subHeaderProps?.className)}>
  {subHeader}
</Text>
```

src/mantine-core/src/Hr/Hr.styles.ts
```diff
@@ -17,14 +20,37 @@ export const sizes = {
...
+      borderLeftWidth: getSizeValue({ size, sizes }),
+      borderLeftColor: getThemeColor({
+        theme,
+        color,
+        shade: theme.colorScheme === 'dark' ? 4 : 6,
+      }),
+      borderLeftStyle: variant,
+    }),
+    subHeader: ({ theme, color, inset, insetType }: HrStylesProps) => ({
+      fontSize: '14px',
```
rtivital, line 51: set fontSize from theme on Text component instead

khushal87:
> First demo does not show dashed variant
> Screen Shot 2021-05-07 at 11 00 02 AM

Fixed. Thanks for reporting.

src/mantine-core/src/Hr/Hr.styles.ts
```diff
@@ -17,14 +20,37 @@ export const sizes = {
...
+      color: getThemeColor({ theme, color, shade: theme.colorScheme === 'dark' ? 4 : 6 }),
```
rtivital, line 52: Same with color, it will be in subHeaderProps

rtivital (changes requested):

src/mantine-core/src/Hr/Hr.tsx
khushal87: Yes, I have already fixed it and this will be reflected while I push the divider code. Thanks.

khushal87: Okay I will check about subHeaderProps. Thanks.

src/mantine-core/src/Hr/Hr.styles.ts
khushal87: Okay

src/mantine-core/src/Hr/Hr.styles.ts
khushal87: Okay

rtivital:
> So adding an insetLength can work? Or do you want the removal of the `inset` prop? My purpose of having an inset is that it is very common to the Dividers today, and people don't generally feel adding additional styles unnecessarily to it

No, I think, insetLength will bring even more complexity – it'll be 3 props to set single margin, does not feel right.

khushal87:
> > So adding an insetLength can work? Or do you want the removal of the `inset` prop? My purpose of having an inset is that it is very common to the Dividers today, and people don't generally feel adding additional styles unnecessarily to it
> 
> No, I think, insetLength will bring even more complexity – it'll be 3 props to set single margin, does not feel right.

Okay, will remove it.

src/mantine-core/src/Hr/Hr.tsx
khushal87: What would be the type of SubHeaderProps in this case 🤔

src/mantine-core/src/Hr/Hr.styles.ts
khushal87: This was just a default color, do you want me to remove this? @rtivital

khushal87: I think this would be fine?
```
 color: subHeaderProps.color
        ? subHeaderProps.color
        : getThemeColor({ theme, color, shade: theme.colorScheme === 'dark' ? 4 : 6 }),
```

rtivital: Text component already has color prop, set it with Hr color:

```tsx
<Text color={color} />
```

rtivital:
subHeaderProps will be `Record<string, any>`

khushal87:
> subHeaderProps will be `Record<string, any>`

Wouldn't it be simply `TextProps`?

rtivital:
> Wouldn't it be simply `TextProps`?
> `subHeaderProps?: TextProps;`

This will break component props table

khushal87:
> > Wouldn't it be simply `TextProps`?
> > `subHeaderProps?: TextProps;`
> 
> This will break component props table

Even LoadingOverlay uses LoaderProps. It works fine according to me.

rtivital:
Oh, okay, looks good, I was having lots of issues with prop types generator, I think I accidentally fixed them sometime 😄

khushal87:
@rtivital I was going through the code and then I realized if we could export the StyleProps too from the styles.tsx of the components this will improve the reusability of all the components styles. Suppose I need `subheaderProps` styles to be of type `TextStyleProps` this would get easier. What is your opinion on it? 
We can do it individually or making a file that firstly imports this style from every styles.tsx of component and then export it.

rtivital:
I prefer to keep all props description in open place – component props interface – this way you don't need to jump between files to edit them

khushal87:
> I prefer to keep all props description in open place – component props interface – this way you don't need to jump between files to edit them

What I mean is if we can export `TextStylesProps` someway and then import it in other style files this would add flexibility. For my usecase if I want to apply color here in subHeader then I have to assign the type of subHeaderProps in Hr.styles.tsx to `TextStylesProps` which seems impossible now. Did you understand what I mean?

rtivital:
Yep, I've tried to do so, unfortunately there is no good solution to this. TextStylesProps does not include all Text component props and has required `theme` field

khushal87:
> Yep, I've tried to do so, unfortunately there is no good solution to this. TextStylesProps does not include all Text component props and has required `theme` field

Oh, yes it doesn't have all, need to think some other way around.

rtivital:
`Record<string, any>` is still the best option

khushal87:
> `Record<string, any>` is still the best option

Yes, for the style, let me see if this works.

docs/src/docs/demos/core/Hr/hr.demos.tsx
```diff
@@ -76,3 +76,94 @@ export function HrSizesDemo() {
...
+const subHeaderCode = `import React from 'react';
+import { Hr } from '@mantine/core';
+function Demo() {
+  return (
+    <>
+      <Hr subHeader="Test" />
+      <Hr
+        subHeader="Test"
+        subHeaderProps={{ style: { textAlign: 'center' }, color: 'red' }}
+        style={{ marginTop: 10 }}
```
rtivital, line 89: Do not include margins in examples code, it does not add value, just brings a noise

docs/src/docs/demos/core/Hr/hr.demos.tsx
```diff
@@ -76,3 +76,94 @@ export function HrSizesDemo() {
...
+      <Hr
+        subHeader="Test"
+        subHeaderProps={{ style: { textAlign: 'end', fontFamily: 'cursive' } }}
+        style={{ marginTop: 10 }}
+      />
+      <Hr 
+        subHeader="Test" 
+        subHeaderProps={{ style: { fontSize: 20 } }} 
+        style={{ marginTop: 15 }} 
+      />
```
rtivital, line 100: This example serves no purpose, remove it

docs/src/docs/demos/core/Hr/hr.demos.tsx
```diff
@@ -76,3 +76,94 @@ export function HrSizesDemo() {
...
+        style={{
+          display: 'flex',
+          flexDirection: 'row',
+          justifyContent: 'space-evenly',
+          marginTop: 15,
+        }}
+      >
+        <Badge>Light</Badge>
+        <Hr orientation="vertical" />
+        <Badge variant="outline">Outline</Badge>
```
rtivital, line 141: It's Hr docs, not Badge, remove variants from badge

src/mantine-core/src/Hr/Hr.styles.ts
```diff
@@ -26,6 +28,23 @@ export default createUseStyles(
       borderTopStyle: variant,
       margin: 0,
     }),
+    hrVertical: ({ theme, size, variant, color }: HrStylesProps) => ({
```
rtivital, line 31: className should be `vertical`, not `hrVertical`

rtivital: Rename `hr` class to `horizontal`, add className based on `orientation` prop

src/mantine-core/src/Hr/Hr.styles.ts
```diff
@@ -26,6 +28,23 @@ export default createUseStyles(
...
+      borderLeftColor: getThemeColor({
+        theme,
+        color,
+        shade: theme.colorScheme === 'dark' ? 4 : 6,
+      }),
+      borderLeftStyle: variant,
+    }),
+    subHeader: ({ theme, color, orientation, subHeaderProps }: HrStylesProps) => ({
+      fontSize: theme.fontSizes.xs,
+      color: subHeaderProps?.color
```
rtivital, line 43: styles should not work with `subHeaderProps`, these values are already handled in Text component, this entire class should be removed

src/mantine-core/src/Hr/Hr.tsx
```diff
@@ -2,30 +2,63 @@ import React from 'react';
...
-  color = 'gray',
+  variant = 'solid',
   ...others
 }: HrProps) {
-  const classes = useStyles({ color, variant, size, theme: useMantineTheme(themeOverride) });
-  return <hr data-mantine-hr className={cx(classes.hr, className)} {...others} />;
+  const classes = useStyles({
+    color,
+    size,
+    theme: useMantineTheme(themeOverride),
```
rtivital, line 43: Place theme as first property

src/mantine-core/src/Hr/Hr.tsx
```diff
@@ -2,30 +2,63 @@ import React from 'react';
...
+  });
+
+  return (
+    <div
+      data-mantine-hr
+      className={cx(orientation === 'vertical' ? classes.hrVertical : classes.hr, className)}
+      {...others}
+    >
+      {subHeader && orientation === 'horizontal' && (
+        <Text {...subHeaderProps} className={cx(classes.subHeader, subHeaderProps?.className)}>
```
rtivital, line 56: use the same color for hr and text:
```tsx
<Text color={color} {...subHeaderProps} className={cx(classes.subHeader, subHeaderProps?.className)}>
```

src/mantine-core/src/Hr/Hr.tsx
```diff
@@ -2,30 +2,63 @@ import React from 'react';
...
+
+  /** Applies orientation to the Hr */
+  orientation?: 'horizontal' | 'vertical';
 
   /** Hr height */
   size?: MantineNumberSize;
 
-  /** Hr color */
-  color?: string;
+  /** Applies SubHeader Text to the Hr */
```
rtivital, line 19: Prop description:
```
Adds text after line in horizontal orientation
```

src/mantine-core/src/Hr/Hr.tsx
```diff
@@ -2,30 +2,63 @@ import React from 'react';
...
 
 export const HR_SIZES = sizes;
 
 interface HrProps extends DefaultProps, React.ComponentPropsWithoutRef<'hr'> {
-  /** Hr borderStyle */
-  variant?: 'solid' | 'dashed' | 'dotted';
+  /** Hr color */
+  color?: string;
+
+  /** Applies orientation to the Hr */
```
rtivital, line 13: Prop description:

```
Set line orientation
```

src/mantine-core/src/Hr/Hr.tsx
```diff
@@ -2,30 +2,63 @@ import React from 'react';
...
+  orientation?: 'horizontal' | 'vertical';
 
   /** Hr height */
```
rtivital, line 16: Prop description:

```
Sets height in horizontal orientation and with in vertical
``

**Review by rtivital — CHANGES_REQUESTED** [review]
Great work @khushal87, please fix these issue

File: src/mantine-core/src/Hr/Hr.tsx (hunk 9 above)
Comment (khushal87) [reply src/mantine-core/src/Hr/Hr.tsx:43]:
```
<Text color={color} {...subHeaderProps} className={cx(subHeaderProps?.className)}>
          {subHeader}
        </Text>
```
So finally, this would do fine?

File: src/mantine-core/src/Hr/Hr.styles.ts (hunk 8 above)
Comment (khushal87) [reply src/mantine-core/src/Hr/Hr.styles.ts:43]:
```tsx
<Text color={color} {...subHeaderProps} className={cx(subHeaderProps?.className)}>
  {subHeader}
</Text>
```
So finally, this would do fine?

docs/src/docs/demos/core/Hr/hr.demos.tsx
```diff
@@ -76,3 +76,94 @@ export function HrSizesDemo() {
...
+    <>
+      <Hr subHeader="Test" />
+      <Hr
+        subHeader="Test"
+        subHeaderProps={{ style: { textAlign: 'center' }, color: 'red' }}
+        style={{ marginTop: 10 }}
+      />
+      <Hr
+        subHeader="Test"
+        subHeaderProps={{ style: { textAlign: 'end', fontFamily: 'cursive' } }}
```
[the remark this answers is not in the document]
khushal87: https://developer.mozilla.org/en-US/docs/Web/CSS/text-align well it is, but I have changed it to right.

src/mantine-core/src/Hr/Hr.tsx
rtivital: Just my preference for code to look cleaner
```tsx
const classes = useStyles({
  theme: useMantineTheme(themeOverride),
  color,
  size,
  // ...other
});
```

docs/src/docs/demos/core/Hr/hr.demos.tsx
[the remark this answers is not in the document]
khushal87: I have removed the font-family that would be better ig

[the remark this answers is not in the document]
rtivital: Sorry, never seen that used, it's better to stick to `text-align:right` to avoid these confusions

docs/src/docs/demos/core/Hr/hr.demos.tsx
```diff
@@ -76,3 +76,94 @@ export function HrSizesDemo() {
     </CodeDemo>
   );
 }
+
+const subHeaderCode = `import React from 'react';
+import { Hr } from '@mantine/core';
+function Demo() {
+  return (
+    <>
+      <Hr subHeader="Test" />
```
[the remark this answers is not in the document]
rtivital: Yep, `Subheader` will do the job

docs/src/docs/demos/core/Hr/hr.demos.tsx
khushal87: Text would work? @rtivital

rtivital: You can leave badges but remove variants from them

khushal87: Okay

khushal87:
@rtivital Done.

PR #64, opened 2021-05-11
[core] export props type declarations

zeropaper:
This PR is aimed to export `Props` suffixed type declarations (for components and styles). As follow-up from https://github.com/mantinedev/mantine/discussions/62


zeropaper:
Hi @rtivital!

thanks for the careful review and sorry for the mistakes... (they were really obvious and could have been avoided...).

I still have a little trouble for the `ButtonProps` and `AnchorProps` but I will get them right eventually.

rtivital:
Well, that will do the trick, thanks again for the PR!

PR #73, opened 2021-05-13
[@mantine/hooks] use-media-query: Set initial state based on query

### LINKED ISSUE #72: @mantine/hooks: use-media-query should set initial state based on media query
Package: @mantine/hooks

The `use-media-query` hook should set the initial state based on an initial query of the media query.

This prevents e.g. a jump from light to dark color scheme at mounting of the app.

Rouby:
closes #72


rtivital:
Hi, thanks for pull request!

Unfortunately, that won't work during server side rendering, hook with throw an error (tested with Gatsby production build). There must be another check to see if window is defined.

Screen Shot 2021-05-13 at 11 59 47 AM

PR #84, opened 2021-05-19
Update useLocalStorageValue to return defaultValue

Pernick:
This change immediately sets `defaultValue` to localStorage if provided.


rtivital:
Hi, thanks for pull request, please fix eslint errors and add missing useEffect import so that this PR can be merged

Pernick:
Thanks for the response, I initially created the PR through github only, which is not ideal for spotting these errors.

I commited the changes and also added example in storybook. Feel free to remove it if you have different intentions with it.

PR #92, opened 2021-05-22
update repository field in package.json files

zeropaper:
I noticed that the `repository` field of the package.json files were not always set and thought that it would spare you some time by making a PR.

Cheers!


rtivital:
Yep, looks like I forgot to change that, thanks!

Issue #99, opened 2021-05-24
@mantine/core - Select - issue with defaultValue when using Placeholder


cstrat:
https://codesandbox.io/s/mantine-select-error-mkurq

```
Warning: Use the `defaultValue` or `value` props on <select> instead of setting `selected` on <option>.
```

This warning only appears when using `placeholder`.
See in the codesandbox, if you comment out the placeholder the warning doesn't appear.

rtivital:
Seems like an easy thing to fix, do you want to create a PR for this? If not, I'll resolve it on my own in next minor release

cstrat:
I'll take a look to see if I can provide a fix!

What would the best approach be though?
We need to assign a `value` to the placeholder option and then have that as the `defaultValue` on the parent select box.

Assign a random/arbitrary value? How far do you want to go - do we check to see if that value is in use amongst the options?

rtivital:
Resolved in [1.3.0](https://github.com/mantinedev/mantine/releases/tag/1.3.0)

PR #103, opened 2021-05-25
Select - Possible Simple Fix to Placeholder/defaultValue

cstrat:
I haven't tested this - I am editing from the GitHub GUI not from my VSCode.

However this could have a clash if for some reason the user wanted to use `xx-placeholder-xx` as a value, however unlikely that is.

Not sure what the best solution would be.

### DIFF
--- a/src/mantine-core/src/Select/Select.tsx
+++ b/src/mantine-core/src/Select/Select.tsx
@@ -62,7 +62,7 @@ export function Select({
 
   if (placeholder) {
     options.unshift(
-      <option key="placeholder" value="" selected disabled hidden>
+      <option key="placeholder" value="xx-placeholder-xx" disabled hidden>
         {placeholder}
       </option>
     );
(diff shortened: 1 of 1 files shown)


cstrat:
This is for bug #99

src/mantine-core/src/Select/Select.tsx
```diff
@@ -97,6 +97,7 @@ export function Select({
         rightSectionProps={{ style: { pointerEvents: 'none' } }}
         required={required}
         themeOverride={themeOverride}
+        defaultValue={placeholder ? 'xx-placeholder-xx' : null}
```
rtivital, line 100: We still want to support defaultValue from props, by setting it to null we do not provide this option

cstrat: Oh yeah good point.
Now if the user has both, what do you think should take preference - defaultValue from props, or the placeholder?

rtivital: I guess it should be something like `defaultValue || (placeholder ? 'xx-placeholder-xx'  : null)`

cstrat:
Works as expected here:

image


But not here - for some reason React is selected and not the placeholder. Could be something to do with Storybook.

image

rtivital:
Alright looks fine, it does not work in WrappedSelect example because Select there is controlled and react is set to be the initial value

Issue #143, opened 2021-07-11
Feature Request, highlighting : Ability to highlight multiple terms


ingloriousp:
package: @mantine/core
version:2.0.2
browser/OS: chrome 91, mac OS Big sur

In Highlighting feature:
<img width="844" alt="duplicate value" src="https://user-images.githubusercontent.com/79135076/125192545-16ae5f00-e266-11eb-9c2c-b3c936f15f3e.png">
"this" may be highlighted twice in the image above.

<img width="852" alt="two terms" src="https://user-images.githubusercontent.com/79135076/125192550-1910b900-e266-11eb-8828-74013d18f281.png">
Ability to highlight multiple terms in a single term may be done. i.e. "this" and "one" may be highlighted using above syntax

FYI: huge fan of mantine :)

rtivital:
Hi, thanks for feedback and feature proposition, I do not have capacity to work on this right now, so I'm leaving this to community, fell free to implement this and send a pull request

yoroshikun:
Hi, I'll take a crack at working on this, I'm not certain about the syntax, would you rather have the ability to pass an array to the component or split the string based on certain symbol.

yoroshikun:
Functionality for this has been merged into version 2.1.0~
I will close this

ingloriousp:
Thank you :)


Kind Regards,
Gaurav Bhati

PR #150, opened 2021-07-14
feat: add ability for multiple highlights

yoroshikun:
Attempts to resolve: #143 

Current issues, the search is forward only, so an input of `world\,Hello` would result in 'Hello `world`' (only world being highlighted.

it might be better to use regex for this?

Syntax for multiple is currently `\,` however this can be easily changed.

### DIFF
--- a/src/mantine-core/src/components/Highlight/Highlight.story.tsx
+++ b/src/mantine-core/src/components/Highlight/Highlight.story.tsx
@@ -14,6 +14,9 @@ storiesOf('@mantine/core/Highlight', module)
     <>
       <Highlight highlight="that">Nothing nothing</Highlight>
       <Highlight highlight="Highlight all">Highlight all</Highlight>
+      <Highlight highlight={['Highlight', 'array']}>Highlight array</Highlight>
+      <Highlight highlight="multiple" multiple>Multiple strings will be highlighted with multiple enabled</Highlight>
+      <Highlight highlight={['hello', 'world']} multiple>Even arrays like hello world will be highlighted multiple, hello world!</Highlight>
     </>
   ))
   .add('Themes', () => <>{themes}</>)
--- a/src/mantine-core/src/components/Highlight/Highlight.test.tsx
+++ b/src/mantine-core/src/components/Highlight/Highlight.test.tsx
@@ -43,35 +43,47 @@ describe('@mantine/core/Highlight/highlighter', () => {
   const VALUE = 'Hello, World';
 
   it('highlights start of string', () => {
-    expect(highlighter(VALUE, 'Hell')).toEqual({ start: '', highlighted: 'Hell', end: 'o, World' });
+    expect(highlighter(VALUE, 'Hell')).toEqual([
+      { chunk: 'Hell', highlighted: true },
+      { chunk: 'o, World', highlighted: false },
+    ]);
   });
 
   it('Highlights middle of string', () => {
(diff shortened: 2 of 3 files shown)


ghana7989 (approved):

rtivital:
Hi @yoroshikun, thanks for pull request, unfortunately this is not what issue was about. The logic describe in #143 is to highlight every substring which is the part of children string. For example, for `{ highlight: 'react', children: 'I work with React and my favorite library is React' }` React should be highlighted twice.

The same example in jsx:

```tsx
<Highlight highlight="react">I work with React and my favorite library is React</Highlight>
// Will output something like this:
I work with <mark>React</mark> and my favorite library is <mark>React</mark>
```

yoroshikun:
That's actually fine, I forgot to mention that functionality is something i could put in in the pr.

What I am most unsure about is the search forward bug, but this wont affect the use case you have provided vs the split case it currently deals with.

I can slightly modify the code to handle the repetitive search, but I do think it should be optional to give the most control over the highlighting. maybe a "multi" prop or something

rtivital:
Yep, it would be great if you can work on these props combinations:

- `highlight` – string and an array to highlight multiple strings, e.g. `highlight="this"` and `highlight={['hello', 'there']}`
- `multiple` – highlights each substring if true

yoroshikun:
To avoid the search forward only issue would it be better to use a regex?

rtivital:
> To avoid the search forward only issue would it be better to use a regex?

Cannot tell for sure as I did not research this, but seems like a good solution

yoroshikun:
I have updated this, however this implementation will need a look over for cleanup and acceptable complexity, It completely remakes the highlight component. It does not pass one test, I am not fully knowledgeable about the \t test and will need some guidance to pass it.

rtivital:
I've checked the failing test, it's not the error with `\t` but an issue with trailing whitespace in `highlight` value (`sub` in `findIndices` function). This test will also fail:

```tsx
  it('highlights value with whitespace', () => {
    expect(highlighter(VALUE, 'Hello ')).toEqual([
      { chunk: 'Hello ', highlighted: true, i: 0 },
      { chunk: ', World', highlighted: false, i: Infinity },
    ]);
  });
```

The strategy I would suggest:
- separate logic to highlight single and multiple parts and test them separately
- test `findIndices` function apart from highlighter

Maybe there is an easier way to split string into chunks using regex? I've tried the function replacement API and it turns out to work fine, example (native js) – https://codepen.io/rtivital/pen/YzVNMLR

P.S. It's Indi*ces*, not Indi*cies*, you can add spell check in your editor with this extension – https://github.com/streetsidesoftware/vscode-spell-checker

yoroshikun:
Thanks for the tip with the spell check.

I have pushed a WIP with a rewrite with regex replacement, this its more proof that it does work and is much cleaner.

before asking for merging I will remove the need for passing styles into the parser and have it render directly to react as unsafe is inherently bad.

Currently this passes all tests (If i were to write them out correctly after the above changes) however feel free to ask for more tests to add.

I should be able to finish this neatly by tomorrow just pushing  wip for now.

yoroshikun:
@rtivital I have fixed this up and all tests pass.

Far more readable than my previous attempts. Ready for review

rtivital:
Thanks for your work, highlighter function looks much simpler now and works fine

PR #155, opened 2021-07-15
[@mantine/core] Tabs: add support for `orientation` prop

wyze:
Implementing the item found here: https://trello.com/c/zZYcbrU4/8-mantine-core-tabs-add-vertical-orientation-support

#### Todo

- [x] Make sure that dark color scheme is supported in all variants
- [x] Make sure that all component variants have correct styles
- [x] Make sure that component uses up and down arrows to change current tab instead of right and left arrows
- [x] Change aria-orientation according to tabs orientation
- [x] Add segmented control which changes orientation to colorsConfigurator demo

### DIFF
--- a/src/mantine-core/src/components/Tabs/TabControl/TabControl.styles.ts
+++ b/src/mantine-core/src/components/Tabs/TabControl/TabControl.styles.ts
@@ -10,13 +10,14 @@ interface TabControlStyles {
   theme: MantineTheme;
   reduceMotion: boolean;
   color: string;
+  orientation: 'horizontal' | 'vertical';
 }
 
 export default createMemoStyles({
   tabActive: {},
   tabLabel: {},
 
-  tabControl: ({ theme }: TabControlStyles) => ({
+  tabControl: ({ theme, orientation }: TabControlStyles) => ({
     ...getFontStyles(theme),
     ...getFocusStyles(theme),
     WebkitTapHighlightColor: 'transparent',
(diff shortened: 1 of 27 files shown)


src/mantine-core/src/components/Tabs/TabControl/TabControl.styles.ts
```diff
@@ -28,33 +29,35 @@ export default createMemoStyles({
     padding: [0, theme.spacing.md],
     fontSize: theme.fontSizes.sm,
     cursor: 'pointer',
+    width: orientation === 'vertical' && '100%',
```
rtivital, line 32: Will break jss styles in some cases (cannot remember these case but pretty sure that we cannot use this syntax), change this instead to 

```tsx
width: orientation === 'vertical' ?'100%' : undefined,
```

rtivital:
Thanks for PR! Can you also check and fix documentation about vertical orientation? There may be some mentions about it, for example, here https://mantine.dev/core/tabs/#accessibility-and-usability

src/mantine-core/src/components/Tabs/TabControl/TabControl.styles.ts
wyze: Note: I prefer the ternary syntax over the short-circuit (`&&`) syntax, but most of what I saw was short-circuited. I'll change these up.

rtivital:
Outline and pills variants are not correct
Снимок экрана 2021-07-16 в 09 29 06
Снимок экрана 2021-07-16 в 09 28 59

wyze:
The outline one was working. I think it had something to do with the switcher in the docs not recalculating the styles properly. I've changed how the styles are computed for that and it appears to work as expected now.

rtivital:
Thanks, everything works fine now, great job!

Issue #154, opened 2021-07-15
Controlled inputs shouldn't set their value


Kukkimonsuta:
_Packages:_ @mantine/hooks @mantine/core
_Version:_ 2.0.5
_Browsers:_ Edge 91 (not relevant)
_Codesandbox:_ https://codesandbox.io/s/clever-gagarin-qik7m?file=/src/App.tsx
_Contribution:_ I haven't looked into the hook in detail to consider how the issue should be fixed, however I'm willing to contribute it if instructed to do.

_Description:_
Input components in controlled mode should never set their own value - it needs to always come from outside as the requested change may be blocked and parent component might not render again. As you can see in codesandbox the native selects value correctly never changes (because the handler is omiting the setValue call), but `Select` component does. This applies to all components using `useUncontrolled` as far as I can tell.

rtivital:
Hi, thanks for reporting, this is a bug in uncontrolled hook, it can easily be fixed by adding extra checks, feel free to contribute

PR #157, opened 2021-07-15
use-controlled - fix controlled input behavior

Kukkimonsuta:
Fixes https://github.com/mantinedev/mantine/issues/154

This has turned out to be bit more complex than originally anticipated because Select has own state as well that needs to be handled, however I believe I have succeeded.

The behavior should be backwards compatible.

I'm not certain what the expectations of `onValueUpdate?(value: T): void;` are since it isn't used anywhere and I fail to see possible use cases, however I tried to keep the behavior same.

The hook now returns third value "mode" that indicates whether it's controlled or uncontrolled currently.


rtivital:
Thanks for PR, everything works perfect!

PR #158, opened 2021-07-15
fix checkout on windows

Kukkimonsuta:
Force LF on text files for windows with autocrlf=true. This avoids having issues with lint requiring LF endings. 

Note that for the change to take place people on windows will need to save their changes and rebuild the repository:

**Running commands below remove all local changes**
```
git rm --cached -r . 
git reset --hard
```

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+* text=auto eol=lf


rtivital:
Thanks, will try this out, what's the difference between this and editorconfig? Will this work for editors that do not have editorconfig extension?

Kukkimonsuta:
This controls how git checks out the files. On windows the default (I believe it's default, or at least it used to be and is pretty commonly used) settings is `core.autoclrf=true` meaning all text files checked out from remote are automatically converted to CRLF. This configures git to ignore the global setting and forces to use LF on windows as well when doing checkout.

.editorconfig should be used in addition, as that will control what happens to new files created by editors and not downloaded form remote in git.

rtivital:
Thanks for explanation, I'll try thing out on my pc to see how this behaves

PR #166, opened 2021-07-18
Fix useFocusTrap callback signature

poacher2k:
The typed return value of useFocusTrap has been updated to be consistent with the signature in the `setRef` callback.


rtivital:
Thanks!

PR #173, opened 2021-07-21
[@mantine/hooks] use-fullscreen: add hook

yoroshikun:
Adds the use-fullscreen hook.

Attempts to ensure cross browser compatibility and offer a simple to use elementRef for custom root

### DIFF
--- a/docs/src/demos/hooks/use-fullscreen.tsx
+++ b/docs/src/demos/hooks/use-fullscreen.tsx
@@ -1,0 +1,119 @@
+import React from 'react';
+import { useFullscreen } from '@mantine/hooks';
+import { Button, Center, Image, Paper, Text, useMantineTheme } from '@mantine/core';
+
+const code = `
+import { useFullscreen } from '@mantine/hooks';
+import { Button, Center, Paper, Text, useMantineTheme } from '@mantine/core';
+
+function Demo() {
+  const theme = useMantineTheme();
+  const { toggle, fullscreen } = useFullscreen();
+
+  return (
+    <Paper>
+      <Center>
+        <Button
+          onClick={toggle}
+          style={{
+            backgroundColor: fullscreen ? theme.colors.orange[9] : theme.colors.green[9],
+          }}
+        >
+          <Text style={{ color: theme.white }} weight={700}>
+            {fullscreen ? 'Exit Fullscreen' : 'Enter Fullscreen'}
+          </Text>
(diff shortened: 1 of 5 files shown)


src/mantine-hooks/src/use-fullscreen/use-fullscreen.ts
```diff
@@ -0,0 +1,82 @@
+import { useCallback, useRef, useState, useEffect } from 'react';
+
+function getFullscreenElement(): HTMLElement | null {
+  const fullscreenElement =
+    document.fullscreenElement ||
+    // @ts-expect-error
```
rtivital, line 6: It will be much cleaner to assign document with any type to variable instead of repeating @ts-expect-error each time:

```tsx
const _document = window.document as any;
```

docs/src/demos/hooks/use-fullscreen.tsx
```diff
@@ -0,0 +1,104 @@
...
+    <Paper elementRef={ref} style={{ height: 100 }}>
+      <div
+        style={{
+          height: '100%',
+          width: '100%',
+          display: 'flex',
+          alignItems: 'center',
+          justifyContent: 'center',
+          flexDirection: 'column',
+        }}
```
rtivital, line 24: Usually it's not a good idea to include inessential styles to demo to code – this makes demo cluttered and makes harder to understand features that are currently demonstrated, please remove styles and props (e.g. `size="lg"`) that are not associated with hook behavior from demo  code

rtivital (changes requested):
Thanks for PR! Please fix these two issue so that it can be merged

docs/src/demos/hooks/use-fullscreen.tsx
yoroshikun: For the full screen example to look cleaner i included these styles, as when the component goes into full screen this correctly centers the buttons in the middle of the screen.
But can do!

yoroshikun:
Fixed up the demo and all references to document use window.document for consistency (as well as the type aliasing for clean look)

rtivital:
Alright, code looks good, the only missing thing is one more demo that will showcase usage with custom root element, can you add this? I think great example will be with image

yoroshikun:
Added an extra demo for full screening an image with reference to best practices.

rtivital:
Thanks! Works great

PR #174, opened 2021-07-21
useIdle hook implementation

achmurali:
Hi @rtivital, 

Implementaion for useIdle hook following the signature :
```function useIdle(timeout: number, cb?: () => void): boolean;```

1. added an additional optional callback argument to access the event.

Let me know if it is fine or needs any tweaks, thereafter shall proceed with the documentation.

Thanks.

### DIFF
--- a/src/mantine-hooks/src/use-idle/use-idle.ts
+++ b/src/mantine-hooks/src/use-idle/use-idle.ts
@@ -1,0 +1,30 @@
+import { useState, useEffect, useRef } from 'react';
+
+type eventsType = MouseEvent | KeyboardEvent | TouchEvent | Event;
+
+export function useIdle(timeout: number, cb?: (e?: eventsType) => void) {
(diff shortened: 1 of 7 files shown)

### CODE REVIEW

File: src/mantine-hooks/src/use-idle/use-idle.ts [hunk 1]
```diff
@@ -0,0 +1,30 @@
+import { useState, useEffect, useRef } from 'react';
+
+type eventsType = MouseEvent | KeyboardEvent | TouchEvent | Event;
+
+export function useIdle(timeout: number, cb?: (e?: eventsType) => void) {
```
rtivital, line 5: Remove second argument `cb`, it brings no value, user can subscribe to those events if needed in other part of application

rtivital (changes requested):
Thanks for PR! Please fix this issue and add documentation with demo

rtivital:
Works great, thanks for PR!

PR #175, opened 2021-07-22
Update Autocomplete and Select to allow strings

cstrat:
Update to allow Autocomplete and Select to allow for string input in the data prop.

Also included storybook examples.


cstrat:
I had to create a new Type for the data prop, otherwise there were warnings peppered around the place about `item.value` not existing on a `string`.

If this is good, I can also update the docs.

cstrat:
Ah I don't think this is my actual change 😐
Let me try again.

PR #176, opened 2021-07-22
Autocomplete and Select allow strings

cstrat:
Update to allow data prop to be string for both Autocomplete and Select.

This works by mapping over the `data` and formatting correctly and ensuring all future references to data refer to the `formattedData` variable.

Included storybook for both components.

Needed to create a new Type for the data prop as there were warnings suggesting that `item.value` doesn't exist on a `string`.


cstrat:
Not sure why this pulled #172 in. 😐

Let me know if this looks good and I can update the documentation pages too.

rtivital:
Thanks for PR, everything works fine

cstrat:
I will do another PR on the documentation for the data prop format.
It is a bit confusing if you're just reading the docs to fully understand the format that is required.

rtivital:
No worries, I'm already doing that

PR #179, opened 2021-07-23
[@mantine/hooks] use-scroll-lock: migrate to internal state

yoroshikun:
Migrate to internal state handling.

Changed the API to maintain custom init state handling, please let me know if the naming and API is as intended or not.

Once I know what the API looks like I will update the docs accordingly

### DIFF
--- a/docs/src/components/HomePage/Explore/demos/Hooks.tsx
+++ b/docs/src/components/HomePage/Explore/demos/Hooks.tsx
@@ -6,13 +6,11 @@ import { LockClosedIcon, LockOpen2Icon } from '@modulz/radix-icons';
 import { LinkTitle } from './LinkTitle';
 
 const lockCode = `
-import { useState } from 'react';
 import { useScrollLock } from '@mantine/hooks';
 import { Button } from '@mantine/core';
 
 function Demo() {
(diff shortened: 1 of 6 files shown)


rtivital:
Hi @yoroshikun thanks for PR, I think it will be better if we keep backwards compatibility with previous version and change
```tsx
useScrollLock({ lockOverride: true })
```
to
```tsx
useScrollLock(true)
```

yoroshikun:
I was wondering how to handle the second optional argument this way.
{ disableTouchEvents } was an optional argument in the options passthough.
to access it without passing in a prop for override it would need to look like this.
```
useScrollLock(undefined, { disableTouchEvents: true})
```
as passing the first argument (the override) will make scroll lock controlled externally.

can you think of a more elegant solution?

rtivital:
Well, that's a tricky situation, I guess the best option will be to remove `disableTouchEvents` logic all along, as it does not add any extra value and was added by my mistake.

yoroshikun:
Updated to accommodate the requested changes.

Have also updated the docs page to reflect the changes.

rtivital:
Thanks, works great, tested the backwards compatibility, works as expected

PR #180, opened 2021-07-23
Added 'exact' prop to Highlight component

iDavidB:
Adds a prop called "exact" so that only exact matches of words are highlighted.

Ex. "Highlight is my favorite word containing light" will highlight the word "light" twice in this string, once for "highLIGHT" and once for "light". This prop will make sure only the latter is highlighted.

### DIFF
--- a/docs/src/docs/core/Highlight.mdx
+++ b/docs/src/docs/core/Highlight.mdx
@@ -33,6 +33,12 @@ You can highlight multiple substrings by providing an array of values:
 
 <Demo data={HighlightDemos.multiple} />
 
+## Highlight only exact strings
+
+You can highlight only exactly matching strings by providing the exact prop:
+
+<Demo data={HighlightDemos.exact} />
+
 ## Text props
 
 Highlight accepts the same props as <GatsbyLink to="/core/text/">Text</GatsbyLink> component:
--- a/src/mantine-core/src/components/Highlight/Highlight.test.tsx
+++ b/src/mantine-core/src/components/Highlight/Highlight.test.tsx
@@ -43,14 +43,14 @@ describe('@mantine/core/Highlight/highlighter', () => {
   const VALUE = 'Hello, World';
 
   it('highlights start of string', () => {
-    expect(highlighter(VALUE, 'Hell')).toEqual([
+    expect(highlighter(VALUE, 'Hell', false)).toEqual([
       { chunk: 'Hell', highlighted: true },
       { chunk: 'o, World', highlighted: false },
     ]);
   });
 
   it('Highlights middle of string', () => {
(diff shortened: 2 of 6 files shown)


yoroshikun:
Hi, Could this not be completed by changing 
```
  const re = new RegExp(`(${matcher})`, 'gi');
```
to
```
  const re = new RegExp(`(${matcher})`, `g${exact ? 'i' : ''}`);
```
as this would make the match case insensitive without having to assign dynamic pattern variable

iDavidB:
Hi, 

Yes, I agree that using a default value for the exact param is a nicer API. I will push a change to achieve this.

I actually thought about creating an options object with the options you mentioned and more, however I had some issues stopping me from doing this:
- I couldn't think of a clean way to achieve adding these options to the pattern without creating some very ugly (unreadable) code.
- `\b` at the start of the pattern will only care about whether the string starts with your match, adding it to the end will do the opposite, both sides will do achieve what we have now. I think you can see where my problem lies with this, it'll be nearly impossible to "normalize" this. 

If you have any idea about how to go about this, then i'm willing to try it out.

yoroshikun:
How about just changing the const to a let and mutating it if the option appears.

```
const noPartial = true;
const caseSensitive = true;

let pattern = "(hello|World)";

if (noPartial) {
  pattern = `\\b${pattern}\\b`;
}

const re = new RegExp(pattern, `g${caseSensitive ? "" : "i"}`);
```

I have made an example code-sandbox and its seems to work fine.
https://codesandbox.io/s/highlight-mantine-regex-example-pehj3?file=/src/index.ts

iDavidB:
Yes, that could work but only if these are the only options we end up with. Otherwise this will end up quite unreadable.

yoroshikun:
I am not sure what is unreadable about it, 1 level deep if statement mutation seems readable enough.
What other options do you think would cause a readability nightmare?

As long as the options are additive to the pattern it should remain very readable.

rtivital:
@iDavidB are you planning to finish this?

iDavidB:
@rtivital Yes, I just haven't had time yet. I will probably be getting to it somewhere this week.

iDavidB:
Hi, I updated it with the options we talked about. I don't really have time to add more specific tests for when one option may be false while another is true. I could add this later or if someone has time now, then feel free.

rtivital:
@iDavidB It does not seem to work in documentation, see video, also, please resolve conflicts so that it can be merged

https://user-images.githubusercontent.com/10353856/128611071-3d43a62e-0c01-4c8a-934a-a1de69caa781.mov

rtivital:
@iDavidB @yoroshikun do you want to continue working on this?

PR #182, opened 2021-07-23
[@mantine/hooks] use-cookie: fix document not available in ssr

yoroshikun:
Noticed that build failed for SSR Gatsby. This small change fixes it.


rtivital:
Thanks, haven't seen that yet

PR #183, opened 2021-07-24
useShallowEffect hook implementation.

achmurali:
Hi @rtivital,

Implemented the useShallowEffect hook, shall write the tests and docs once you review this as I am not really clear about shallow comparison for other types and the scope of this hook.
Hope I am not being a hassle 😄 😓.  

Thanks

### DIFF
--- a/docs/src/docs/hooks/use-shallow-effect.mdx
+++ b/docs/src/docs/hooks/use-shallow-effect.mdx
@@ -1,0 +1,83 @@
+---
+group: 'mantine-hooks'
+package: '@mantine/hooks'
+title: 'use-shallow-effect'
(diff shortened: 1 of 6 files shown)


rtivital:
Hi @achmurali, thanks for PR, function that compares values seems to be be very complicated, please use something like this – https://github.com/streamich/fast-shallow-equal/blob/master/index.js

rtivital:
@achmurali please add this so that this PR can be merged:
- move function that used for shallow compare to `src/mantine-hooks/src/utils` forlder
- Add hook documentation

rtivital:
Thanks, looks great!

PR #185, opened 2021-07-27
Feature/strict types in public api

Kukkimonsuta:
Adds strict null annotations to

`assign-ref`
`use-merged-ref`
 - ref value can be always null or whatever type the ref is

`use-uncontrolled`
 - return type needs to be specified otherwise the value type will be coerced to non-null
 - value/default value are optional on inputs, so need to conform to both undefined and null as well as `rule` function
 - the final value and values in callback are always either T or null thanks to finalValue

Note that this is hard to discover and test - we'll have to mostly react to what people find when using the library with strict on.


rtivital:
Thanks, I'll have a look at another hooks and will change them to support strict option if nedded

PR #186, opened 2021-07-28
Use MantineSize for shadow as well

mosch:
...for proper code completion and type saftey


rtivital:
HI @mosch shadows support any valid css box-shadow, e.g.

```tsx
<Paper shadow="1px 2px 4px rgba(0, 0, 0, .25)" />
```

That's why shadow is string and not MantineSize

mosch:
@rtivital it would still be good to have it be `MantineSize | string` to indicate it supports MantineSize.

PR #187, opened 2021-07-28
Add MantineSize to shadow type for better DX

mosch:
See: https://github.com/mantinedev/mantine/pull/186

### DIFF
--- a/src/mantine-core/src/components/Paper/Paper.tsx
+++ b/src/mantine-core/src/components/Paper/Paper.tsx
@@ -8,7 +8,7 @@ export interface PaperProps extends DefaultProps, React.ComponentPropsWithoutRef
   padding?: MantineNumberSize;
 
   /** Predefined box-shadow from theme.shadows (xs, sm, md, lg, xl) or any valid css box-shadow property */
-  shadow?: MantineSize;
+  shadow?: MantineSize | string;
 
   /** Predefined border-radius value from theme.radius or number for border-radius in px */
   radius?: MantineNumberSize;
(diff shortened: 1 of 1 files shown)


rtivital:
Unfortunately this is not how TypeScript works: `'xs' | 'sm' | string` will just resolve to `string` and you will not get any value from adding MantineSize, try in [demo](https://www.typescriptlang.org/play?#code/C4TwDgpgBAglC8UDkBDJUA+yBGSDcAUKJFAMIKyZQDOwATgJYB2A5oQQGYCuTAxsAwD2TKNyYAKMCjooAtgC4yASigBvAlCh0IwLnRFSZswgF8C5seKVA)

Снимок экрана 2021-07-28 в 22 24 51

mosch:
It still helps the IDE doing auto-complete suggestions. It's correct typescript and therefore it doesn't hurt but helps as documentation as well.

rtivital:
> It still helps the IDE doing auto-complete suggestions
Can you provide an example of your IDE competition? VS Code does not show any autocomplete suggestions.

mosch:
Well yeah. I couldn't get the suggestion to work on vscode as well. My bad. Thanks for the exchange.

rtivital:
Hi @mosch, I've found a way to get autocomplete working, it should be
```tsx
type Size = 'xs' | 'sm' | 'md' | 'lg' | 'xl' | (string & {});
```

This will trigger autocomplete in vscode and allow any string as value

mosch:
Whoa! What a weird, great thing. I'll test it.

PR #193, opened 2021-08-05
[docs] Adds rightSection example for TextInput

bravo-kernel:
Adds an example of using `rightSection` to the `TextInput` page. This was one I had to figure out myself so perhaps useful for future visitors. Let me know if you want to change the text.

image


rtivital:
Thanks

Issue #196, opened 2021-08-08
Group component taking up more than available width.


achmurali:
- mantine/core
- 2.2.1
- Google Chrome (92.0.4515.107) - 1920*1080
- Group Component is occupying more space than the available space due to the negative margin applied in the Group root style. - ```margin: (-1 * getSizeValue({ size: spacing, sizes: theme.spacing })) / 2```.
Example - the 404 page is having a horizontal scroll bar because of the negative margin given to Group Component.
- I believe we don't need negative margin and that could fix the mantine dev 404 page, if not lets discuss and I shall do the fix.
- Yes I am more than willing to fix the issue 😄. 

![image](https://user-images.githubusercontent.com/22786016/128623381-0001a87d-5f57-4d48-8876-fee84f89cd16.png)

rtivital:
This is not an issue with Group component – it behaves exactly as planned. Negative margins are required as they negate positive margins of Group's child elements, without them group will have unexpected spacing on all sides.

404 page is broken because of how Gatsby handles the layout – not sure at what time this happened. You can investigate this and wrap it in Container to make it look nicer if you want.

achmurali:
The reason I saw it as an issue with Group component and not with gatsby was because (for my screen of width 1920px) -
- The Navbar is occupying `260px` and the rest of the page should occupy the available `1660px`, but it exceeds the `1660px` width because of the negative margin hence causing a horizontal scroll bar.
- I expected a container (Group component) to occupy the available width.

rtivital:
>The Navbar is occupying 260px and the rest of the page should occupy the available 1660px, but it exceeds the 1660px width because of the negative margin hence causing a horizontal scroll bar.

It's an issue with docs 404 page, not Group component – something's not right there, in normal circumstances Group will not cause scrollbars to appear, here is an example – https://codesandbox.io/s/mantine-template-ts-forked-qggl8?file=/src/App.tsx

achmurali:
> The reason I saw it as an issue with Group component and not with gatsby was because (for my screen of width 1920px) -
> 
> * The Navbar is occupying `260px` and the rest of the page should occupy the available `1660px`, but it exceeds the `1660px` width because of the negative margin hence causing a horizontal scroll bar.
> * I expected a container (Group component) to occupy the available width.

Please check this example - https://codesandbox.io/s/mantine-template-ts-forked-qugr4?file=/src/App.tsx, and let me know if you see a scroll bar on viewing in full-width.

[...]

PR #199, opened 2021-08-09
Group component

achmurali:
- Added an additional `gutter` prop in **Group** component, signifying padding for the root element (Issue #196). 
- Fixed `404` page, added margin between navbar and main content.

image


rtivital:
All right, this change to Group seems reasonable and should not break anything, thanks for working on this.
Can you also make these changes:
- Rename `gutter` prop to `withGutter` (gutter usually assumes some value, e.g. `gutter={12} or gutter="sm"`)
- Wrap 404 page in Container so it does not look so inconsistent with the rest of the docs

achmurali:
Hi @rtivital, have a minor issue ( I think ), because the [splitCamelCase](https://github.com/mantinedev/mantine/blob/master/docs/src/components/Demo/Configurator/controls/BooleanControl.tsx#L11) function in BooleanControl capitalizes the first letter of a string, the mantine docs displays it as `With gutter` and not something like `with Gutter` or `With Gutter`.

Let me know if i should fix this- 
- By either giving an option to specify the explicit string to be displayed.
- or capitalize first letter of every word.

rtivital:
It's correct – `With gutter` is more human readable label than `withGutter`, it does not affect demo code

achmurali:
I am talking about the capitalisation. Currently it is displayed as 'With gutter' and i think 'With Gutter' (capital G ) would look good.

rtivital:
That's Not How We Usually Write Text Or Labels In Application, Look A Little Bit Odd If You Capitalize Each Letter.

Usually capitalization of each word only harms the comprehension. I like the GitLab guide on this topic, seems very reasonable – https://design.gitlab.com/content/punctuation

PR #210, opened 2021-08-18
Added zIndex prop to DatePickerBase

everek:
### Problem
If the DatePicker is used inside a Modal, the DatePicker popover appears behind the Modal, since the zIndex prop of the Popper Component which is used for the DatePicker popover is set to 3.

### Solution
Added zIndex property to DatePickerBase which allows you to modify the z-index attribute of the Popper component. 
The default value is still set to 3.

I don't know if "zIndex" is the correct property name in this case, since it's only for the Popper component and not the DatePicker component itself.


rtivital:
Thanks for PR!
Fixes #211

PR #216, opened 2021-08-19
Fix for Issue#21

### LINKED ISSUE #213: Menu closeOnItemClick option not working in v2.3.0 (Popper related?)
**Package:** @mantine/core
**Version:** 2.3.0
**Browser/OS:** Chrome v92.0.4515.131 on macOS Big Sur

After upgrading to @mantine/core v2.3.0, the `closeOnItemClick` option for the `Menu` component is no longer being respected. If it is set to false, the menu will still close when an item is clicked.

I suspect that it's because that version now wraps the menu with a `Popper` component, since seems like the main difference between v2.2.3 and v2.3.0. As far as I can tell, the option is still being used in the `MenuBody` component (https://github.com/mantinedev/mantine/blob/master/src/mantine-core/src/components/Menu/MenuBody/MenuBody.tsx line 174) so that part seems fine. Unfortunately I don't have much time to continue investigating at the moment.

Since the issue is not present in v2.2.3, I am able to work around this problem by staying on that version.

achmurali:
This PR fixes #213.

### DIFF
--- a/src/mantine-core/src/components/Menu/Menu.tsx
+++ b/src/mantine-core/src/components/Menu/Menu.tsx
@@ -168,7 +168,7 @@ export function Menu({
 
   useWindowEvent('scroll', () => closeOnScroll && handleClose(true));
 
-  const wrapperRef = useClickOutside(() => _opened && handleClose());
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks for taking a look, I've already fixed this issue, it will be released today or tomorrow

PR #217, opened 2021-08-19
Clean use-focus-trap

achmurali:
This PR implements re-factoring in use-focus-trap


rtivital:
Thanks, I guess it's a little bit cleaner now

PR #230, opened 2021-08-30
Feature/responsive grid props

everek:
Here's a first proposal for responsive Columns.

With the changes it's possible to use breakpoint props on each Col and define the span of the column at each breakpoint.

This could look like this:
```JSX
<Col sm={4} lg={6} span={12}>
Content goes here
</Col>
```

**How would the column look?**
* Up until sm it would have a span of 12
* From sm to lg it would have a span of 4
* From lg on it would have a span of 6

Existing Grids **shouldn't** break, as the span prop we've already had will just be used.

There are the usual breakpoints available: 
```JS
breakpoints: {
    xs: 576,
    sm: 768,
    md: 992,
    lg: 1200,
    xl: 1400,
  },
```
The breakpoints are in the _default theme_ and can be _overwritten_.

**How does it work?**
In the Grid component there are inline styles being created for every possible combination, for example `mantine-col-12` or `mantine-col-lg-6`. So if you have multiple Grids on one page, you will have multiple inline styles. This isn't the prettiest solution, but as long as the media query bug in JSS exists, I see no other way.

I adjusted some of the jest tests, since I've switched the spacing of the columns from margin to paddings. I tested a couple setups but would be glad if you could test it further @rtivital. Also I'm not that used to TypeScript, so excuse me if there are any flaws.


rtivital:
Looks great, thanks for PR!
Here are some extra things that should be done:

- Grid component looks very dirty with style calculation logic, can you move it to separate file?
- Docs items seems to no longer show gutter, can you fix that so that it works as it used to? I guess it caused by switching margin to padding.
Снимок экрана 2021-08-30 в 17 54 49

rtivital:
Also can you please add responsive section to documentation where you explain how it works, nothing fancy, something [like this](https://github.com/mantinedev/mantine/blob/next-minor/docs/src/docs/core/SimpleGrid.mdx#breakpoints)

everek:
Good input about moving the style calculation logic to another file. Any idea for the naming of the file?

Didn't think about the docs, will update it!

rtivital:
> Any idea for the naming of the file?
maybe something like `get-responsive-styles.ts`

everek:
I just noticed a bug. If you have multiple Grids on one page and set a different amount of columns on one Grid component, the Column width gets overwritten for all Columns on the page. 

For example we have 

```JSX
<Grid columns={12}>
      <Col span={12}>1</Col>
      <Col span={6}>2</Col>
      <Col span={6}>3</Col>
    </Grid>
```

and

```JSX
<Grid columns={24}>
      <Col span={12}>1</Col>
      <Col span={6}>2</Col>
      <Col span={6}>3</Col>
    </Grid>
```

on one page. The CSS will look like this:

```CSS
.mantine-col-4 {
    flex-basis: calc(33.333333333333336% );
    flex-shrink: 0;
    max-width: calc(33.333333333333336% );
}
````

But the second Grid will overwrite it to this:

```CSS
.mantine-col-4 {
    flex-basis: calc(16.666666666666668% );
    flex-shrink: 0;
    max-width: calc(16.666666666666668% );
}
```

Any idea how I could fix this?
I just noticed this in the docs, locally.

rtivital:
Yes, you will need to generate uuid, I've already built very similar thing in SimpleGrid – https://github.com/mantinedev/mantine/blob/next-minor/src/mantine-core/src/components/SimpleGrid/SimpleGrid.tsx

rtivital:
Also is there a need for `calc` in css styles as it includes only `%`?

everek:
Thank you, I will have a look at it.
You are right, the `calc` is a leftover from before there was spacing. I will remove it.

everek:
I moved the styling logic to a separate file and added the uuid to the css classes. Did some more cleanup.

For the docs:
* Fixed the background of the Columns (added a new div to the ColWrapper in the docs)
* Changed the responsive columns section in the Grid docs
* Added breakpoints section to the extend theme doc

@rtivital do you have some more input for the documentation? I'm not sure if it's clear enough how you achieve responsive columns.

rtivital:
All good, thanks for contribution!

dukesx:
Thank you @rtivital and @everek for this. Looks amazing and good job to you both. Kudos , really appreciate this.

PR #241, opened 2021-09-06
Enable `loading` state for ActionIcon

cstrat:
I might have over-simplified how I inject the `Loader` in place of the Icon when in loading state but seems to work fine in the stories.

### DIFF
--- a/src/mantine-core/src/components/ActionIcon/ActionIcon.story.tsx
+++ b/src/mantine-core/src/components/ActionIcon/ActionIcon.story.tsx
@@ -29,6 +29,13 @@ storiesOf('@mantine/core/ActionIcon', module)
       <Group>{getThemes({ disabled: true, variant: 'transparent' })}</Group>
     </>
   ))
+  .add('Loading', () => (
+    <>
+      <Group>{getThemes({ loading: true })}</Group>
+      <Group>{getThemes({ loading: true, variant: 'filled' })}</Group>
+      <Group>{getThemes({ loading: true, variant: 'transparent' })}</Group>
+    </>
+  ))
   .add('Sizes', () => (
     <>
       <Group>{getThemes({ size: 'xs' }, { style: { width: 12, height: 12 } })}</Group>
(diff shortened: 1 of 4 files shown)


cstrat:
Oh I should have tried merging this with `next-minor` right? Sorry still not used to committing to other repos yet. Getting there...

rtivital:
I thinks this should be done the same way as in Button with overlay (see screenshot), otherwise it will be inconsistent. Can you work on that?
Снимок экрана 2021-09-06 в 09 35 58

cstrat:
Ah I didn't realise button used overlay, must have missed that, let me check and update.

cstrat:
https://github.com/cstrat/mantine/blob/loader-action-icon/src/mantine-core/src/components/Button/Button.tsx#L154-L189

I am a bit lost - I don't know where overlay comes into it, unless it is part of the style/classname?
I don't fully understand how the underlying classes & style system is implemented (`cx(...)` is a bit magic to me).
When I did the PR I did copy heavily from `Button`.

I figured it should look most similar to `disabled` state - just with the loader.

Storybook view:

image

rtivital:
It's in styles – https://github.com/mantinedev/mantine/blob/master/src/mantine-core/src/components/Button/Button.styles.ts#L99-L114

cstrat:
Champion, thanks I will see how I go copying some of that into `ActionIcon` styles.

cstrat:
That wasn't too hard.
It does look better!

image


image


https://github.com/cstrat/mantine/blob/loader-action-icon/src/mantine-core/src/components/Button/Button.tsx#L156
^ this line is confusing me because I can't tell if its typescript or if it is a computed property name?

cstrat:
Did you want the component to keep the colouring of the button but simply fade it? I've over-ridden the theme with grey but that doesn't need to happen.

cstrat:
image

image

rtivital:
Yes, it should look exactly the same as in button

cstrat:
The only issue I have right now is the `filled` style doesn't look right when it flicks to loading state.
I mean it looks similar to `disabled` state, but not like buttons.

I think I'd need to wrap the `Loader` in a div with appropriate styles to achieve the background colouring to match the Icon?

rtivital:
Okay, thanks, I'll have a look and provide feedback later today

rtivital:
Alright, I've checked, it looks good, thanks for contribution!

rtivital:
You can use this feature with version `2.5.0-alpha.1` if you need it immediately, otherwise it will be out with other 2.5.0 features (September 28th or earlier)

cstrat:
No worries, thanks!

PR #246, opened 2021-09-09
[@mantine/dates]: TimeField colon centering

achmurali:
As colon isn't centered but put above the baseline, thought you would consider this.
references - https://medium.com/@amachino/the-secret-of-san-francisco-fonts-4b5295d9a745.

### DIFF
--- a/src/mantine-dates/src/components/TimeInput/TimeField/TimeField.tsx
+++ b/src/mantine-dates/src/components/TimeInput/TimeField/TimeField.tsx
@@ -94,7 +94,13 @@ export function TimeField({
       />
 
       {withSeparator && (
-        <Text size={size} style={{ lineHeight: 1, color: 'inherit' }}>
+        <Text
+          size={size}
+          style={{ lineHeight: 1,
+                    color: 'inherit',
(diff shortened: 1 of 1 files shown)


rtivital:
I does not seem to work fine with all fonts (screenshot from a mac)
Снимок экрана 2021-09-10 в 07 12 01

achmurali:
For a mac the system default San Francisco font centers the colon, but in case of windows that isn't the case - 
 
image

rtivital:
It is also misaligned in your screenshot
Снимок экрана 2021-09-10 в 14 45 48

achmurali:
> It is also misaligned in your screenshot
> Снимок экрана 2021-09-10 в 14 45 48

Yeah, the screenshot i pasted is without any changes (the existing master branch).

rtivital:
Well, we cannot merge this now since it depends on the font: your change does not work on macOS, current does not work on Windows. The only way to solve this is to give colon absolute position and center it manually depending on input size

achmurali:
But I don't think even that was working as I was looking to position the colon with `em`.

PR #248, opened 2021-09-11
[@mantine/core]: add Pagination component

EmilMalanczak:
Hello, I allowed myself to check a little bit planned API - I was modelling a bit on [material-ui](https://material-ui.com/api/pagination/) Pagination API with siblings and boundary because it a bit more intuitive - total number of pages has been even - `active page + 2x siblings amount + 2x boundary amount + 2x dots`. If provided `limit` prop would be odd it would be problematic to keep pagination in the same structure (non-moving/dynamic content). Also `getItemAriaLabel` or something similar to material-ui would be nice to have here as well.

I don't know if more examples are needed in docs since its usage is pretty straightforward. Maybe some async pagination example would be fine to attach.

I've based component logic on hook `usePagination` shall I move it to `hooks` folder? It shares almost the same API as the `Pagination` component does.

look forward to your feedback


src/mantine-core/src/components/Pagination/Pagination.styles.ts
```diff
@@ -0,0 +1,73 @@
...
+    });
+
+    return {
+      display: 'flex',
+      alignItems: 'center',
+      justifyContent: 'center',
+      fontWeight: 600,
+      border: `1px solid ${colors.border}`,
+      color: colors.color,
+      height: '2em',
```
rtivital, line 30: Do not use `em` or other relative values, only px

src/mantine-core/src/components/Pagination/DefaultItem/DefaultItem.tsx
```diff
@@ -0,0 +1,43 @@
...
+        return <NextIcon />;
+      case 'prev':
+        return <PrevIcon />;
+      default:
+        return page;
+    }
+  };
+
+  return (
+    <UnstyledButton
```
rtivital, line 34: It's better to use ActionIcon, it already has :active and :disabled styles

src/mantine-core/src/components/Pagination/Pagination.tsx
```diff
@@ -0,0 +1,129 @@
...
+    initialPage,
+    boundary,
+  });
+
+  return (
+    <nav className={cx(classes.root, className)} style={{ ...style, ..._styles.root }} {...others}>
+      <Item
+        page="prev"
+        onClick={goPrevPage}
+        aria-label="go to previous page"
```
rtivital, line 82: Do not hardcode labels, it cannot be translated or changed this way

src/mantine-core/src/components/Pagination/demos/controlled.tsx
```diff
@@ -0,0 +1,25 @@
+import React, { useState } from 'react';
```
rtivital, line 1: There is no need in controlled demo, it looks and works exactly the same as usage demo, replace it with just a code snippet (see TextInput for reference)

src/mantine-core/src/components/Pagination/icons/DotsIcon.tsx
```diff
@@ -0,0 +1,19 @@
+import React from 'react';
+
+export function DotsIcon(props: React.SVGProps<SVGSVGElement>) {
```
rtivital, line 3: Replace `React.SVGProps<SVGSVGElement>` with `React.ComponentPropsWithoutRef<'svg'>`

src/mantine-core/src/components/Pagination/icons/DotsIcon.tsx
```diff
@@ -0,0 +1,19 @@
...
+  return (
+    <svg
+      width="1.25em"
```
rtivital, line 6: em -> px

src/mantine-core/src/components/Pagination/styles.api.ts
```diff
@@ -0,0 +1,9 @@
+import { PaginationStylesNames } from './Pagination';
+
+export const Pagination: Record<PaginationStylesNames, string> = {
+  root: 'Root element',
+  item: 'Pagination item, applies on page items and arrows',
+  dots: 'Styles applied to dots item between page numbers',
```
rtivital, line 6: Describe element, not actions, here it will be just "Dots element"

rtivital:
Hi @EmilMalanczak thanks for pull request! Component works great. Let's move use-pagination hook to @mantine/hooks, it would be the same as use-queue for notifications system.

src/mantine-core/src/components/Pagination/Pagination.tsx
EmilMalanczak: That's why I've left the proposal for the `getItemAriaLabel ` prop. A11y leaves a warning if I will not leave some info about the button. I can delete it but that will cause some warnings in console after `npm test`

EmilMalanczak:
what is the reason behind using `px` instead of `em`?

rtivital:
We do not use ems, rems and other relative values because it's hard to comprehend and calculate

EmilMalanczak:
Fair enough it's tricky to follow it sometimes. Shall I attach an additional demo with a custom component as a prop?

rtivital:
No, just a documentation section. 
It would be great to include a configurator demo that will showcase `boundary`, `total` and `color` props

src/mantine-core/src/components/Pagination/Pagination.tsx
EmilMalanczak: @rtivital I'm doing fixes to PR. What should I do here?

rtivital: Yep, `getItemAriaLabel` seems like a good solution, do not forget to include it to Accessibility docs section

EmilMalanczak: Sure

EmilMalanczak:
@rtivital I made all of the requested changes except the one with `ActionIcon`. Its adaptation was problematic. Additionally, I've provided a configurator in the docs.

rtivital:
Alright, thanks, looks good

rtivital:
Hi @EmilMalanczak since you are interested in contributing more, I'll provide more feedback after merging and fixing some issues, please review changes that were made – https://github.com/mantinedev/mantine/compare/d20db7a7214b0cb90b6d403ef7c93bd1c583d17f...06a6ae8751cd8617ca0370287af2aacea17ba005

Here are some highlights:
- do not nest `describe`, tests become unreadable, it's much more easier to understand what's happening if each `it` includes separate initialization
- `it` description should read as a finished statement – `it('sets correct initial state', () => {})`, not `it('onChange function correctly handles results')` – `it` should always be considered as part of the sentence 
- Check how component can be used with keyboard only and make sure focus flow is correct – in your implementation dots icon can be focused while click on it does nothing – it should not be interactive
- When designing components prefer neutral colors (`theme.colors.gray` and `theme.white` for light color scheme and `theme.colors.dark` for dark color scheme). Only active elements must have color accent.

Снимок экрана 2021-09-14 в 14 59 53

EmilMalanczak:
Yea I see, thank you for the feedback! 

That code is much cleaner. It's my first open-source project that I contribute to 😄, the project entry threshold is high imho but feedback like that makes it much easier.

I know quite little about testing but I'm working on it so if you have some valuable materials or tips, I will be happy to read them.

PR #254, opened 2021-09-17
Updated demo and removed `withSeconds` prop

cstrat:
I was checking the new components and noticed this demo was previewing `withSeconds` when it wasn't supposed to.


cstrat:
I was also unable to `npm run docs` - not sure if that was just something on my local machine?

rtivital:
Thanks!

PR #255, opened 2021-09-17
[@mantine/hooks]: add use-viewport-size

EmilMalanczak:
I've prepared a simple version of that hook. Is there a need for a debouncing? If so I think there will be a need for the `useDebouncedCallback` hook due to `useDebouncedValue` won't prevent listeners from 'spam' changing the state.

### DIFF
--- a/docs/src/demos/hooks/use-viewport-size.tsx
+++ b/docs/src/demos/hooks/use-viewport-size.tsx
@@ -1,0 +1,44 @@
+import React from 'react';
+import { useViewportSize } from '@mantine/hooks';
+import { Text, Group } from '@mantine/core';
+
+const code = `
+import { useViewportSize } from '@mantine/hooks';
+import { Text, Group } from '@mantine/core';
+
+function Demo() {
+  const { height, width } = useViewportSize();
+
+  return (
+    <Group position="center">
+      <Text>
+        Viewport width: {width}
+      </Text>
+      <Text>
+        Viewport height: {height}
+      </Text>
+    </Group>
+  );
+}
+`;
+
+function Demo() {
+  const { height, width } = useViewportSize();
+
+  return (
+    <Group position="center">
+      <Text>
+        Viewport width: {width}
+      </Text>
+      <Text>
+        Viewport height: {height}
+      </Text>
+    </Group>
+  );
+}
+
+export const useViewportSizeDemo: MantineDemo = {
+  type: 'demo',
+  code,
+  component: Demo,
+};
--- a/docs/src/docs/hooks/use-viewport-size.mdx
+++ b/docs/src/docs/hooks/use-viewport-size.mdx
@@ -1,0 +1,31 @@
+---
+group: 'mantine-hooks'
+package: '@mantine/hooks'
+category: 'dom'
+title: 'use-viewport-size'
+order: 1
+slug: /hooks/use-viewport-size/
+description: 'Get information about current viewport size'
+import: "import { useViewportSize } from '@mantine/hooks';"
+docs: 'hooks/use-viewport-size.mdx'
+source: 'mantine-hooks/src/use-viewport-size/use-viewport-size.ts'
+---
+
+import { useViewportSizeDemo } from '@docs/demos';
+
+## Usage
+
+Hook returns current viewport width and height:
+
+<Demo data={useViewportSizeDemo} />
+
+On the first render (as well as during SSR), both width and height are equal to `0`.
+
+## Definition
+
+```tsx
+function useViewportSize(): Readonly<{
+  height: number;
+  width: number;
+}>;
+```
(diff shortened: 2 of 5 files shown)


src/mantine-hooks/src/use-viewport-size/use-viewport-size.ts
```diff
@@ -0,0 +1,44 @@
+import { useCallback, useState, useEffect } from 'react';
+import { useWindowEvent } from '../use-window-event/use-window-event';
+
+interface DebouncedWindowSizeOptions {
```
EmilMalanczak, line 4: gonna fix the name after review

rtivital: Just remove it, hook should not receive any values, there is no point to do default value

src/mantine-hooks/src/use-viewport-size/use-viewport-size.ts
```diff
@@ -0,0 +1,44 @@
...
+  initialWidth?: number
+  initialHeight?: number
+}
+
+const eventListerOptions = {
+  passive: true,
+};
+
+export const useViewportSize = ({
```
rtivital, line 13: Please use `function` instead of arrow function (consistent with other hooks)

rtivital (commented):
Overall everything is fine, not much to review

EmilMalanczak:
sure, I'm just used to arrow functions. Going to make changes tomorrow

docs/src/demos/hooks/use-viewport-size.tsx
```diff
@@ -0,0 +1,44 @@
...
+    </Group>
+  );
+}
+`;
+
+function Demo() {
+  const { height, width } = useViewportSize();
+
+  return (
+    <Group position="center">
```
rtivital, line 29: There is no need in Text and Group components, just a simple div will do the job

```tsx
<div>Width: {width}, height: {height}</div>
```

docs/src/docs/hooks/use-viewport-size.mdx
```diff
@@ -0,0 +1,31 @@
+---
+group: 'mantine-hooks'
+package: '@mantine/hooks'
+category: 'dom'
+title: 'use-viewport-size'
+order: 1
+slug: /hooks/use-viewport-size/
+description: 'Get information about current viewport size'
```
rtivital, line 8: "Get viewport width and height" – more clear description

src/mantine-hooks/src/use-viewport-size/use-viewport-size.ts
```diff
@@ -0,0 +1,36 @@
...
+  width: number;
+  height: number;
+}> {
+  const [windowSize, setWindowSize] = useState({
+    width: 0,
+    height: 0,
+  });
+
+  const setSize = useCallback(
+    (): void => {
```
rtivital, line 18: Do not add unnecessary types, `void` is redundant

src/mantine-hooks/src/use-viewport-size/use-viewport-size.ts
```diff
@@ -0,0 +1,36 @@
+import { useCallback, useState, useEffect } from 'react';
+import { useWindowEvent } from '../use-window-event/use-window-event';
+
+const eventListerOptions = {
+  passive: true,
+};
+
+export function useViewportSize(): Readonly<{
```
rtivital, line 8: Return type here is also redundant, TypeScript will handle return values correctly, there is no need to make them readonly

src/mantine-hooks/src/use-viewport-size/use-viewport-size.ts
```diff
@@ -0,0 +1,36 @@
...
+  width: number;
+  height: number;
+}> {
+  const [windowSize, setWindowSize] = useState({
+    width: 0,
```
rtivital, line 13: Can we initialize these values in browser in useState instead of useeffect?

```tsx
const [windowSize, setWindowSize] = useState({
  width: typeof window !== 'undefined' ? window.innerWidth : 0,
  height: typeof window !== 'undefined' ? window. innerHeight : 0,
});
```

This way we can save up setting state after component was mounted

EmilMalanczak:
I'll do fixes for all of the PR's this week.

rtivital:
Okay sounds good, also I was thinking maybe you could put them in one single pr to avoid merge conflicts (I guess there could be some in index files)

EmilMalanczak:
There shouldn't be conflicts I think, I was creating branches from the same `next-minor` version. How should I do that properly?

rtivital:
I think that merge conflict can happen when we merge one PR and then second and third will have conflicts with updated next-minor branch, well at least this is what usually happens to me 😃

EmilMalanczak:
Haha 😄 happens to everyone. I'll do the merge locally into next-minor and then I'll create one separate branch with its own PR and I'll close these 3 PR's if that's ok

EmilMalanczak:

PR #256, opened 2021-09-17
[@mantine/hooks]: add use-mouse

EmilMalanczak:
prepared another hook. @rtivital Can you tell me which components should I use for demos in docs? It also applies to the `useViewportSize` hook. I will provide them tomorrow.

### DIFF
--- a/docs/src/demos/hooks/use-mouse/usage.tsx
+++ b/docs/src/demos/hooks/use-mouse/usage.tsx
@@ -1,0 +1,35 @@
+import React from 'react';
+import { useMantineTheme, Text, Code } from '@mantine/core';
+import { useMouse } from '@mantine/hooks';
+
+const code = `
+import { useMantineTheme, Text, Code } from '@mantine/core';
+import { useMouse } from '@mantine/hooks';
+
+function Demo() {
+  const theme = useMantineTheme();
+  const { x, y } = useMouse();
+
+  return (
+    <Text align="center" style={{ marginTop: theme.spacing.sm }}>
+      Mouse coordinates <Code>{\`{ x: \${x}, y: \${y} }\`}</Code>
+    </Text>
+  );
+}`;
+
+function Demo() {
+  const theme = useMantineTheme();
+  const { x, y } = useMouse();
+
+  return (
+    <Text align="center" style={{ marginTop: theme.spacing.sm }}>
+      Mouse coordinates <Code>{`{ x: ${x}, y: ${y} }`}</Code>
+    </Text>
+  );
+}
+
+export const useMouseUsage: MantineDemo = {
+  type: 'demo',
+  code,
+  component: Demo,
+};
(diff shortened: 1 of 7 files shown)


EmilMalanczak:
also, I'm not sure which exact params should be used its used on a custom element 

should that work like this one - [link?](https://codepen.io/MarioGiannini/pen/Mrmwdj?editors=0010)

rtivital:
Remember that hook should work with document element if ref is not present. 
Regarding demos: good demo should showcase features of given hook and include only code that is required for understanding hook behavior, for example, for this hook you will need to create two demos:
- div which displays coordinates when mouse enters it
- regular text with document coordinates

EmilMalanczak:
The actual implementation works in both cases but I'm not sure which args I should take. 

As I will attach a ref to a div (which idk possibly is in the middle of the screen). When I'll have a cursor on the left edge of the div a `x` coordinate should be `0` (like in the example from the link) or should be relative to the `x` position of document total width.

docs/src/demos/hooks/use-mouse/usage.tsx
```diff
@@ -0,0 +1,35 @@
...
+const code = `
+import { useMantineTheme, Text, Code } from '@mantine/core';
+import { useMouse } from '@mantine/hooks';
+
+function Demo() {
+  const theme = useMantineTheme();
+  const { x, y } = useMouse();
+
+  return (
+    <Text align="center" style={{ marginTop: theme.spacing.sm }}>
```
rtivital, line 14: Do not include irrelevant styles in demo code, remove `style={{ marginTop: theme.spacing.sm }}` from demo code and meaning won't change

docs/src/demos/hooks/use-mouse/usage.tsx
```diff
@@ -0,0 +1,35 @@
...
+    </Text>
+  );
+}`;
+
+function Demo() {
+  const theme = useMantineTheme();
+  const { x, y } = useMouse();
+
+  return (
+    <Text align="center" style={{ marginTop: theme.spacing.sm }}>
```
rtivital, line 25: You should leave margin here though for demo to look nice. General rule – remove all irrelevant presentational styles and properties from demo code but leave them in demo itself

docs/src/docs/hooks/use-mouse.mdx
```diff
@@ -0,0 +1,53 @@
+---
+group: 'mantine-hooks'
+package: '@mantine/hooks'
+title: 'use-mouse'
+category: 'dom'
+order: 1
+slug: /hooks/use-mouse/
+description: 'Get mouse coordinates relatively to the viewport'
```
rtivital, line 8: Get mouse position relative to viewport or given element

docs/src/docs/hooks/use-mouse.mdx
```diff
@@ -0,0 +1,53 @@
...
+
+<Demo data={useMouseUsage} />
+
+## API
+
+Hook returns object with ref and x, y mouse coordinates:
+
+```tsx
+const {
+  ref, // -> pass ref to target element - optional in use
```
rtivital, line 28: "pass ref to target element - optional in use" should be "pass ref to target element, if not used document element will be used as target element"

EmilMalanczak:

PR #257, opened 2021-09-18
[@mantine/hooks]: add use-resize-observer

EmilMalanczak:
another hook for an initial review - for now, I don't see any necessary options for it for now except something like debounce timing.  Is basic resizeable textarea with no value as demo will be enough for that hook?

### DIFF
--- a/docs/src/demos/hooks/use-resize-observer.tsx
+++ b/docs/src/demos/hooks/use-resize-observer.tsx
@@ -12,7 +12,7 @@ function Demo() {
 
   return (
     <>
-      <Text align="center" size="sm" style={{ marginBottom: theme.spacing.xs }}>
+      <Text align="center" size="sm">
         Resize current element by moving its left bottom corner
       </Text>
       <Group position="center">
@@ -30,7 +30,7 @@ function Demo() {
           }}
         />
       </Group>
-      <Text align="center" style={{ marginTop: theme.spacing.sm }}>
+      <Text align="center">
         Rect: <Code>{JSON.stringify(rect, null, 2)}</Code>
       </Text>
     </>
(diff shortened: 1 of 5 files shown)


rtivital:
Thanks for all PRs, for this one I cannot provide any feedback as I did not work with ResizeObserver much, will see how it works in documentation demos

EmilMalanczak:
Sure, today I'm planning to provide docs and other necessary things to all PRs

docs/src/demos/hooks/use-resize-observer.tsx
```diff
@@ -0,0 +1,75 @@
...
+            height: 120,
+            border: `1px solid ${theme.colorScheme === 'dark' ? theme.colors.dark[4] : theme.colors.gray[4]}`,
+            backgroundColor:
+              theme.colorScheme === 'dark' ? theme.colors.dark[5] : theme.colors.gray[2],
+            position: 'relative',
+          }}
+        />
+      </Group>
+      <Text align="center" style={{ marginTop: theme.spacing.sm }}>
+        Rect: <Code>{JSON.stringify(rect, null, 2)}</Code>
```
EmilMalanczak, line 65: May I ask for a tip for visualization of that?

rtivital:
Cool, demo is super clean and easy to understand, great work!

src/mantine-hooks/src/use-resize-observer/use-resize-observer.ts
```diff
@@ -0,0 +1,58 @@
...
+    return () => {
+      observer.disconnect();
+
+      if (frameID.current) {
+        cancelAnimationFrame(frameID.current);
+      }
+    };
+  }, [ref.current]);
+
+  return [ref, rect];
```
rtivital, line 57: It will be easier with just `return [ref, rect] as const;` instead of defining return type (it should automatically handle generic type)

src/mantine-hooks/src/use-resize-observer/use-resize-observer.ts
```diff
@@ -0,0 +1,58 @@
+import { MutableRefObject, useEffect, useMemo, useRef, useState } from 'react';
+
+type ObserverRect = Pick<
```
rtivital, line 3: Is there any reason why we cannot use just `DOMRectReadOnly` type?

EmilMalanczak:

PR #261, opened 2021-09-20
Migration to @mantine/styles from react-jss

rtivital:


rtivital:
@EmilMalanczak @achmurali would you please have a look and try things out, maybe you would be able to see some broken styles that I've missed

achmurali:
> @EmilMalanczak @achmurali would you please have a look and try things out, maybe you would be able to see some broken styles that I've missed

Yeah definitely @rtivital . So, I thought I could go over the storybooks of the components and maybe tryout some demos.
Any other ideas are welcome 😅.
 
First thing I noticed is in the docs, the jumbotron isn't centered. Was it deliberate?
image

rtivital:
Yep, I'll be rebuilding home page to be more clear

EmilMalanczak:
I will dig into it tonight. 

Backing to the docs page, I've found that lib some time ago and I think it would be good to use that trick mentioned there for gatsby / next [link](https://github.com/donavon/use-dark-mode#that-flash). 
As I see there is a gatsby plugin for it. I'm highly into dark mode preference and almost every time I open docs I get light mode flash at first render

dukesx:
I see good things @rtivital 👍 , i suppose emotion was, is and should be the next level of CSS in JSS 😄

rtivital:
Well, I've found a good way to implement it without changing the whole project structure, so yes, it's much better now

dukesx:
I am sure that this transition will attract more PRs too, since emotion is alot more loved and easier to use too.

PR #266, opened 2021-09-22
New hooks

EmilMalanczak:
PR containing:
- `use-mouse`
- `use-viewport-size`
- `use-resize-observer`

Please, check this `use-mouse` hook returning params relative to the given element. There's a bug I can't get why happens. Very little miscalculations (around 0.5), there are able to catch in demo


rtivital:
Thanks for putting everything together!

PR #268, opened 2021-09-24
hook/useScrollIntoView

EmilMalanczak:
That's a proposal version of the `useScrollIntoView` hook, as we've talked about on pm:

It works almost the same as `window.scrollIntoView` but it gives us more control over it. An additional problem with the native approach is that the safari doesn't support `scroll-behavior: smooth` which kinda 'teleports' content. 

features:
- allow configuring `duration` of scroll
- allow configuring `easing function` 
- scrolling on `x` and `y` axis
- `onScrollFinish` callback after scroll animation is finished
- scrolling inside any scrollable HTML node
- works at 60fps by using `requestAnimationFrame` under the mask

hook constructor params:
```
{
  target: HTMLElement;
  parent?: HTMLElement;
  onScrollFinish?: () => void;
  duration?: number;
  axis?: 'x' | 'y'
  easing?: (t: number) => number
};
```

hooks returns: 
```
{
    scrollIntoView: () => void;
    cancel: () => void;
}
```

possible is changing the approach to make the `scrollIntoView` method handle taking `parent` and `target` refs instead of doing this in the constructor.

### DIFF
--- a/src/mantine-hooks/src/use-scroll-into-view/use-scroll-into-view.ts
+++ b/src/mantine-hooks/src/use-scroll-into-view/use-scroll-into-view.ts
@@ -22,6 +22,9 @@ type ScrollIntoViewParams = {
 
   /** axis of scroll */
   axis?: 'x' | 'y'
+
+  /** custom mathematical easing function */
+  easing?: (t: number) => number
 };
 
 export function useScrollIntoView({
@@ -30,6 +33,7 @@ export function useScrollIntoView({
   duration = 1.25,
   axis = 'y',
   onScrollFinish,
+  easing = easeInOutQuad
 }: ScrollIntoViewParams) {
   const forceRerender = useForceUpdate();
   const frameID = useRef(0);
@@ -49,7 +53,7 @@ export function useScrollIntoView({
       // easing timing progress
       const t = elapsed / duration;
 
-      const distance = start + change * easeInOutQuad(t);
+      const distance = start + change * easing(t);
 
       setScrollParam({ parent, axis, distance });
 
--- a/src/mantine-hooks/src/use-scroll-into-view/use-scroll-into-view.ts
+++ b/src/mantine-hooks/src/use-scroll-into-view/use-scroll-into-view.ts
@@ -1,0 +1,84 @@
+/* eslint-disable no-param-reassign */
+import { useForceUpdate } from '@mantine/hooks';
+import { useCallback, useRef, useEffect } from 'react';
+
+import { easeInOutQuad } from './utils/ease-in-out-quad';
+import { getRelativePosition } from './utils/get-relative-position';
+import { getScrollStart } from './utils/get-scroll-start';
+import { setScrollParam } from './utils/set-scroll-param';
+
+type ScrollIntoViewParams = {
+  /** target node to be scrolled yo */
+  target: HTMLElement;
+
+  /** scrollable parent node default to document */
+  parent?: HTMLElement;
+
+  /** callback fired after scroll */
+  onScrollFinish?: () => void;
+
+  /** duration of scroll in seconds */
+  duration?: number;
+
+  /** axis of scroll */
+  axis?: 'x' | 'y'
+};
+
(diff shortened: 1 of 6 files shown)


EmilMalanczak:
Tomorrow I will provide example demos

src/mantine-hooks/src/use-scroll-into-view/use-scroll-into-view.ts
```diff
@@ -0,0 +1,88 @@
+/* eslint-disable no-param-reassign */
+import { useForceUpdate } from '@mantine/hooks';
+import { useCallback, useRef, useEffect } from 'react';
+
+import { easeInOutQuad } from './utils/ease-in-out-quad';
+import { getRelativePosition } from './utils/get-relative-position';
+import { getScrollStart } from './utils/get-scroll-start';
+import { setScrollParam } from './utils/set-scroll-param';
+
+type ScrollIntoViewParams = {
```
rtivital, line 10: Replace with interface

rtivital (commented):
Overall everything is clean, thanks for PR!

EmilMalanczak:
@rtivital take a look at the docs I've attached. On Monday I will back to it and try to replace `scrollIntoView` util. 

Please tell me if I should move `parent` and `target` props from the constructor and be provided to `scrollIntoView`

EmilMalanczak:
also at basic usage example, I've spotted an issue related to `fixed` position on the header. It covers up the target element. I did a test, `element.scrollIntoView` API has the same issue.

Would it be a good idea to add additional prop - `space` or `margin` which will default to `0` and would add an extra space to the calculated distance. In this case it would be the height of the docs header element.

rtivital:
Yep, that's a great idea, we already have similar logic in rich text editor sticky header, let's call this prop `offset`

rtivital:
@EmilMalanczak what's the progress on this? Is it ready for review? Do you need help with anything? Can you please merge your branch with 3.0.0 so that we do not have conflicts?

rtivital:
No worries, we will include this hook in 3.1 release then

EmilMalanczak:
@rtivital I fixed all of the current issues I've found so far. 

Recent changes in the hook:
- `offset` prop - the additional distance between nearest edge and element
- `alignment` argument inside `scrollIntoView` - optional target element alignment relatively to parent based on current axis
- fixed issue with spam clicking, now it correctly cancels previous RAF(requestAnimationFrame) and recalculates to correct values

Additionally for `Select` and `MultiSelect` components:
- replace old `scrollIntoView` util
- add little transition on key navigation equal `0.25s` - available to check via their docs

rtivital:
I've checked how it works and there are some hiccups if container is scrolled during animation. Can we cancel scrolling triggered by hook when user starts scrolling?

https://user-images.githubusercontent.com/10353856/137099953-812f9ab6-1d0f-4ced-b390-c9cd14c83dc0.mov

rtivital:
I've also checked multi select – it works great when scroll moves down, but strange when it goes up. Can we do so that dropdown does not scroll if there is no need for that? Overall looks great!

EmilMalanczak:
@rtivital I have finished the current issues. 

Hook uses now `use-reduced-motion`.

Now, by default animation is cancelable but eventually, I left boolean prop `cancelable` to give an opportunity to turn it off. It might come useful in MultiStep forms (a rare case but a real example from my project) or something similar in logic.

Tried to play around with the issue with `Select` component. I've decided to add prop `isList` which prevents content from jumping in lists. eg. Select, Carousel components. 

Current calculations aren't ideal but I think they come from non-integer heights of list items (37,625px I think but not sure) but still works far better than the previous one.

rtivital:
I've checked how it works, here are some comments:
- Let's reduce transition duration in Select and MultiSelect dropdowns to 300ms
- We use ms in all other hooks, change seconds in hook to ms
- Remove old `utils` folder in @mantine/core package (it contains only old scroll into view and can be safely removed)

I'll provide some more code related comments in code review. Overall everything works smooth, great job 🚀

PR #273, opened 2021-09-26
[@mantine/core] Accordion: Added ability to customize icon

rohaanhamid:
Added ability to customize the down icon. This is necessary because the user might want to use a chevron from the icon library of their choice.

I repurposed the icon prop which was currently not being used for anything.


rtivital:
Hi, unfortunately we cannot merge this. This issue was already implemented in next version. Please check up with maintainers via issues or Discord before working on an issue.

Issue #330, opened 2021-10-21
[@mantine/hooks] use-merged-ref: Unfocusing current element


fishmandev:
When entering data into the first input, the focus goes to the second one.
Example repository: https://github.com/fishmandev/react-mantine-demo
OS: Debian 11
Browser: Version 94.0.4606.81 (Official Build) (64-bit)
@mantine/hooks": "^3.0.5

If you uncomment the [line 10](https://github.com/fishmandev/react-mantine-demo/blob/main/src/App.tsx#L10) and comment out [line 11](https://github.com/fishmandev/react-mantine-demo/blob/main/src/App.tsx#L11), everything works correctly.
![carbon](https://user-images.githubusercontent.com/29619660/138300173-15c906ca-2953-4c96-9947-2564bab9bec1.png)

https://user-images.githubusercontent.com/29619660/138301739-0e78a727-e6c5-44e8-9f52-7e335d37f912.mp4

PR #334, opened 2021-10-22
[mantine/hooks] use-focus-trap: fix unfocusing current element

EmilMalanczak:
A fix for [issue#330](https://github.com/mantinedev/mantine/issues/330)

`setRef` function was re-executing with every re-render which was causing a focus on the element with `data-autofocus` attribute. I've attached an internal state if the focus was already trapped, that covers this issue.


Alikhan1364:
Help my

PR #355, opened 2021-10-29
[@mantine/hooks] use-logger: Logging only updated data

fishmandev:
[Repository to reproduce the bug](https://github.com/fishmandev/react-mantine-demo/blob/main/src/App.tsx):
Screenshot_2021-10-29_16-41-05
Screenshot_2021-10-29_16-27-22


fishmandev:
Moved to #369

PR #357, opened 2021-10-29
3.1 fixes

EmilMalanczak:


src/mantine-core/src/components/Modal/Modal.tsx
```diff
@@ -142,7 +148,10 @@ export function MantineModal({
               aria-labelledby={titleId}
               aria-describedby={bodyId}
               aria-modal
-              style={transitionStyles.modal}
+              style={{
+                ...transitionStyles.modal,
+                marginLeft: 'calc(var(--removed-scroll-width, 0px) * -1)',
+              }}
```
EmilMalanczak, line 154: I've used margin because `transform` has its transition duration etc. additionally there would be issues with transitions based on the x-axis. Current solutions might behave 'weird' in dynamic widths like `%` or `vw`. @rtivital do u have any proposal solution for this?

PR #361, opened 2021-11-01
[docs] correct wording

zifeo:


rtivital:
But hook does not work with dom, docs are correct

zifeo:
@rtivital Can you outline what is not working?

rtivital:
Everything's working. I mean that docs were correct, use-form does not work with the dom.

PR #379, opened 2021-11-03
[@mantine/core] NumberInput: Add the ability for the field to be empty

### LINKED ISSUE #375: Allow NumberInput to have the ability to be empty
As mentioned in the Github discussion I'd like to implement the ability for the NumberInput component to have the ability to be empty.

The way I'm proposing to do will introduce a breaking change in the `onChange` prop signature from `onChange?(value: number): void` to `onChange?(value: number | undefined): void`

Basically, I'm thinking that onBlur we just check if the input is empty and if it is then call the `handleValueChange` with undefined and that'll bubble up right. And if it's not an empty input it'll work the same as before/use the same logic as before. 

Would have to change the input from `type="number"` to `type="text"` because some browsers (for example safari) the input's onChange `event.target.value` is `""` if the text inputted into the field isn't deemed a valid number string by the browser (i.e. can't distinguish between the field being empty and the value typed being "1." as far as I know). Switching it to `type="text"` and adding the `inputmode` attribute might not be such a bad thing given the [input type for numbers has a handful of downsides](https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/)

### Discussed in https://github.com/mantinedev/mantine/discussions/356

<div type='discussions-op-text'>

<sup>Originally posted by **TrashPandacoot** October 29, 2021</sup>
Basically in my use case I have a decent amount of number fields that are optional fields that don't have a natural default value. They're usually things that will be entered at a later date by editing an existing record so currently if one of these optional number fields is left blank it's "we don't know what the value is going to be at this time". 

I took a peak at the NumberInput source and I'm pretty sure this behavior I'm describing is not possible with the current implementation. I'm guessing just based on the `onChange` signature it'd be a breaking change in order to do though.

Is adding this sort of behavior something that makes sense for this library? I'd be happy to contribute if it is, although I'd likely need a bit of guidance as I am a little lost trying to follow the current implementation</div>

TrashPandacoot:
fixes #375

### DIFF
--- a/src/mantine-core/src/components/NumberInput/NumberInput.test.tsx
+++ b/src/mantine-core/src/components/NumberInput/NumberInput.test.tsx
@@ -111,4 +111,24 @@ describe('@mantine/core/NumberInput', () => {
 
     expect(spy).toHaveBeenLastCalledWith(12);
   });
+
+  it('returns undefined when input is empty', () => {
+    const spy = jest.fn();
+    const element = mount(
+      <NumberInput value={5} max={10} min={0} step={6} onChange={spy} />
+    );
+
+    const input = element.find('input').at(0);
+    input.simulate('change', { target: { value: '6' } });
+    expect(spy).toHaveBeenLastCalledWith(6);
+    input.simulate('change', { target: { value: '6.q' } });
+    expect(spy).toHaveBeenLastCalledWith(6);
+    input.simulate('change', { target: { value: '.q' } });
+    expect(spy).toHaveBeenLastCalledWith(6);
+    input.simulate('change', { target: { value: '' } });
+    expect(spy).toHaveBeenLastCalledWith(undefined);
+    input.simulate('blur', { target: { value: '' } });
+
+    expect(spy).toHaveBeenLastCalledWith(undefined);
+  });
 });
--- a/src/mantine-core/src/components/NumberInput/NumberInput.test.tsx
+++ b/src/mantine-core/src/components/NumberInput/NumberInput.test.tsx
@@ -65,6 +65,8 @@ describe('@mantine/core/NumberInput', () => {
     element.find('.mantine-NumberInput-controlUp').simulate('mousedown');
     expect(spy).toHaveBeenLastCalledWith(10);
 
+    // 1st decrement should get it to 0, 2nd decrement should get it to -10
+    element.find('.mantine-NumberInput-controlDown').simulate('mousedown');
     element.find('.mantine-NumberInput-controlDown').simulate('mousedown');
     expect(spy).toHaveBeenLastCalledWith(-10);
   });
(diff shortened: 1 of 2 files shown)


rtivital:
Hi @TrashPandacoot thanks for PR! Here are some issues that I've found:
- When I click up/down controls with empty field, it displays `NaN` – instead, it should replace `undefined` value with `min` on first button click or with `0` if min is not set
- When up/down arrows are pressed on keyboard and input has `undefined` value – nothing happens. Instead, it should work the same as with up/down controls
- I can input any string when initial value is undefined, and it will leave it as is. Instead, it should clear field on blur
- `undefined` value does not work with controlled mode – when I force `value` prop to be undefined, internal state does not update (see code below to get an example). You will need to update `useUncontrolled` hook to rule to treat undefined as a valid value. 

```tsx
function Controlled() {
  const [value, setValue] = useState(undefined);
  return (
    <>
      <NumberInput value={value} onChange={setValue} />
      <button type="button" onClick={() => setValue(undefined)}>
        Set undefined
      </button>
      <button type="button" onClick={() => setValue(10)}>
        Set number
      </button>
    </>
  );
}
```

Please fix these issues and make sure that tests are green to merge this PR.

rtivital:
@TrashPandacoot Are you planning to finish your work on this PR? If not please let me know, I'll merge this and finish it myself

TrashPandacoot:
I can get to this Monday. But you're more than welcome to take over and finish this if you'd like it done sooner

rtivital:
That's fine, we are planning to release this as part of 3.2 release, so there is no rush

TrashPandacoot:
Alright, I have some changes I haven't pushed yet, but I found something I don't understand with a few tests that started failing after my changes. Looking at it makes me think that the tests are incorrect and need to be adjusted but I could totally just not be understanding something either with the test or the NumberInput component.

https://github.com/mantinedev/mantine/blob/7414a6724e51dc053519398a149ca1f9d3debe47/src/mantine-core/src/components/NumberInput/NumberInput.test.tsx#L61-L81

In both the tests above, increment is called once and as expected the onChange is last called with the value of 10 (the second is clamped to the max as expected). That part makes sense to me so far. What I don't understand is why decrement is being called once but the expected value is the value I'd expect from decrementing it twice. It looks to me like the expected values after the decrement were set as if the NumberInput was reseting to the initial value before the decrement was called.

I would've expected the last onChange from the first test to be called with a value of 0 (i.e. the state should be 10 after the first increment, so the decrement with a step of 10 should make the state 0 so that should be the last onChange value rather than -10 as the test is written)

I would've expected the last onChange from the second test to be called with a value of 0 (i.e. the state should be 10 after the first increment, so the decrement with a step of 6 should make the state 4 so that should be the last onChange value rather than 0 as the test is currently written)

Is there something I'm not understanding or are the tests setup incorrectly and my changes corrected some unexpected behavior?

rtivital:
Thanks for your work! We will fix tests and other issues

PR #392, opened 2021-11-07
set cursor to pointer for checkbox and radio

MrHBS:
Set cursor to pointer for checkbox and radio


rtivital:
Pointer cursor on checkbox and radio buttons is not a default browser and system behavior – cursor over checkbox and radio is determined by user OS. You can override this by using global styles in your application.

PR #393, opened 2021-11-08
[@mantine/core] Col: Add breakpoint offsets

slbls:
As discussed over Discord, this PR adds offset props for every breakpoint to the `Col` component.


rtivital:
Thanks!

PR #406, opened 2021-11-10
[@mantine/core] Multiselect - maxSelectedValues prop

achmurali:


rtivital:
@achmurali we should make input `readOnly`, not `disabled` when items are at max – currently input is not focused when last item is removed – see video – when I remove last item input is not focused, but when I remove forth item input is focused (correct behavior)

https://user-images.githubusercontent.com/10353856/141177541-700db02e-5c66-422f-9878-14c802a01b89.mov

rtivital:
Still does not feel right, can you make sure that dropdown is opened when last item is removed?

rtivital:
Thanks!

PR #412, opened 2021-11-12
[@mantine/core] MultiSelect - fix dropdown and clearable issues

achmurali:
This PR fixes the following -
- prevent from clearing when disabled
- when all items are selected, removing the last item doesn't open the drop down
- removing items shouldn't close the dropdown but just open it.

### DIFF
--- a/src/mantine-core/src/components/MultiSelect/MultiSelect.tsx
+++ b/src/mantine-core/src/components/MultiSelect/MultiSelect.tsx
@@ -313,6 +313,11 @@ export const MultiSelect = forwardRef<HTMLInputElement, MultiSelectProps>(
       );
     }, [searchValue]);
 
+    useDidUpdate(() => {
+      //using greater than equal to take into account creatable type.
+      if (!disabled && _value.length >= data.length) setDropdownOpened(false);
+    }, [_value]);
+
     const handleItemSelect = (item: SelectItem) => {
       setTimeout(() => {
         clearSearchOnChange && handleSearchChange('');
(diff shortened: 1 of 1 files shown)


src/mantine-core/src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -313,6 +313,11 @@ export const MultiSelect = forwardRef<HTMLInputElement, MultiSelectProps>(
       );
     }, [searchValue]);
 
+    useDidUpdate(() => {
+      //using greater than equal to take into account creatable type.
+      if (!disabled && _value.length >= data.length) setDropdownOpened(false);
```
rtivital, line 318: This won't work for controlled input – dropdown will open when the value is set externally:

```tsx
import { useState } from 'react';
import { MultiSelect } from '@mantine/core';

function Demo() {
  const [value, setValue] = useState(null);
  return (
    <div>
      <MultiSelect data={['hello']} value={value} onChange={setValue} />
      <button type="button" onClick={() => setValue('hello')}>set value</button>
    </div>
  )
}
```

achmurali: Could you please explain, i didn't get the issue?

rtivital: I've edited the above example. Based on `useDidUpdate` hook, it will open dropdown – this should not happen.

rtivital:
Thanks, works fine

PR #419, opened 2021-11-15
[@mantine/hooks] use-idle: Add initialState option to use-idle hook

LucasLemanowicz:
It would useful to be able to set the initialState of `useIdle` (which currently defaults to true).

Currently the following code logs you out the moment you render the component:
```
  // ...

  const shouldLogout = useIdle(logoutSeconds * 1000);

  useEffect(() => {
    if (shouldLogout) {
      logout();
    }
  }, [logout, shouldLogout]);

  // ...
```

With this PR you will be able to set the initial idle state so that it doesn't start off rendering as idle:
```
  // ...

  const shouldLogout = useIdle(logoutSeconds * 1000, { initialState: false });

  useEffect(() => {
    if (shouldLogout) {
      logout();
    }
  }, [logout, shouldLogout]);

  // ...
```

The cleanest way to implement this interface is to move everything after the timeout into an options prop, but unfortunately that is a breaking change so I am open to other ideas.

### DIFF
--- a/docs/src/demos/hooks/use-idle.tsx
+++ b/docs/src/demos/hooks/use-idle.tsx
@@ -31,7 +31,7 @@ function Demo() {
 }
 
 function EventsDemo() {
-  const idle = useIdle(2000, ['click', 'touchstart']);
+  const idle = useIdle(2000, { events: ['click', 'touchstart'] });
 
   return (
     <Group position="center">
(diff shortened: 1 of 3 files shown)


rtivital:
Thanks for PR, please update hook documentation to reflect your changes

rtivital:
Thanks!

PR #425, opened 2021-11-17
Fixed return of null in first render in useLocalStorageValue

### LINKED ISSUE #424: useLocaleStorageValue in @mantine/hooks provides null for first render despite given defaultValue
Using the `useLocalStorageValue` in 3.1.9 has a bug. The first render will always return `null`, because the defaultValue gets set in an `useEffect`.

Reproduction (check the console.logs and change the locale storage key to a non existing one if you missed it):
https://codesandbox.io/s/nice-perlman-fmmtb?file=/src/App.js

the second render will return the `defaultValue`, but this behavior is inconsistent with the typings which imply that a `string` will always be returned if `defaultValue` is given.

I am not sure, but I think changing the return from

```ts
return [value, setLocalStorageValue] as const;
```

to

```ts
return [value || defaultValue, setLocalStorageValue] as const;
```

should be sufficient enough to fix it.

JohannesKlauss:
This fixes #424 
Let me know if this is sufficient enough.

### DIFF
--- a/src/mantine-hooks/src/use-local-storage-value/use-local-storage-value.ts
+++ b/src/mantine-hooks/src/use-local-storage-value/use-local-storage-value.ts
@@ -42,5 +42,5 @@ export function useLocalStorageValue<T extends string>({
     }
   }, [defaultValue, value, setLocalStorageValue]);
 
-  return [value, setLocalStorageValue] as const;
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #432, opened 2021-11-19
Fixes clear when creatable prop is set.

### LINKED ISSUE #423: [Issue] Cannot clear Select value with creatable prop
When the creatable prop is being used for a Select, the value is not clearable.  I've attached a code sandbox below to demo.

Select a value in the select and then click the reset button. 

https://codesandbox.io/s/blue-breeze-ts023?file=/src/App.js

EDIT: It seems like there is a workaround using the `useForceUpdate` utility hook.

achmurali:
Fixes #423.


rtivital:
Thanks!

PR #445, opened 2021-11-21
Feature: Popper without Portal

jvdsande:
This MR follows up the conversation #435, and enables Popper and all Popper-related components to be used without Portals, through an opt-out prop `withinPortal` which defaults to `true`.

I've added a `Without Portal` story to all related components Storybook, and added a section about Portal in the `Popper` documentation.

Tell me if anything is missing or if you see anything wrong in my implementation 👍

### DIFF
--- a/docs/src/docs/core/Popper.mdx
+++ b/docs/src/docs/core/Popper.mdx
@@ -121,6 +121,13 @@ After each placement change `onPlacementChange` which takes current placement as
 />
 ```
 
+## Portal
+
+By default, Popper content will be rendered inside a [Portal](/core/portal) component.
+
+You can opt-out of this behavior through the `withinPortal` prop, setting it to `false`. Other Mantine components
+relying on Popper expose this prop too.
+
 ## TypeScript
 
 ```tsx
--- a/docs/src/docs/core/Popper.mdx
+++ b/docs/src/docs/core/Popper.mdx
@@ -123,7 +123,7 @@ After each placement change `onPlacementChange` which takes current placement as
 
 ## Portal
 
-By default, Popper content will be rendered inside a [Portal](/core/portal) component.
+By default, Popper content will be rendered inside a [Portal](/core/portal/) component.
 
 You can opt-out of this behavior through the `withinPortal` prop, setting it to `false`. Other Mantine components
 relying on Popper expose this prop too.
--- a/src/mantine-core/src/components/Popper/PopperContainer/PopperContainer.tsx
+++ b/src/mantine-core/src/components/Popper/PopperContainer/PopperContainer.tsx
@@ -19,7 +19,7 @@ export function PopperContainer({
   children,
   zIndex = 1,
   className,
-  withinPortal,
+  withinPortal = false,
 }: PopperContainerProps) {
   if (withinPortal) {
     return (
--- a/src/mantine-core/src/components/Popper/PopperContainer/PopperContainer.tsx
+++ b/src/mantine-core/src/components/Popper/PopperContainer/PopperContainer.tsx
@@ -1,0 +1,39 @@
+import React from 'react';
+import { Portal } from '../../Portal';
+
+export interface PopperContainerProps {
+  /** PopperContainer children, for example, modal or popover */
+  children: React.ReactNode;
+
+  /** Root element z-index property */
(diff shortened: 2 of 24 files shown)


src/mantine-core/src/components/Popper/PopperContainer/PopperContainer.tsx
```diff
@@ -0,0 +1,39 @@
...
+  withinPortal?: boolean;
+}
+
+export function PopperContainer({
+  children,
+  zIndex = 1,
+  className,
+  withinPortal,
+}: PopperContainerProps) {
+  if (withinPortal) {
```
jvdsande, line 24: I decided to delegate this conditional rendering inside a dedicated component, rather than making the render function of Popper more complex. Is that ok with you?

src/mantine-core/src/components/Popper/PopperContainer/PopperContainer.tsx
```diff
@@ -0,0 +1,39 @@
...
+      <Portal className={className} zIndex={zIndex}>
+        {children}
+      </Portal>
+    );
+  }
+
+  return (
+    <div className={className} style={{ position: 'relative', zIndex }}>
+      {children}
+    </div>
```
jvdsande, line 35: I case of no-portal, I simply render what would be rendered by the `Portal` component, but outside of Portal itself.

In the end it gets very close to your `disabled` prop on `Portal` proposal I guess, but at least it is not exposed to the end-user that way.

docs/src/docs/core/Popper.mdx
```diff
@@ -121,6 +121,13 @@ After each placement change `onPlacementChange` which takes current placement as
 />
 ```
 
+## Portal
+
+By default, Popper content will be rendered inside a [Portal](/core/portal) component.
```
rtivital, line 126: Please change links to contain a trailing slash (/core/portal/), it's a small issue with Gatsby – it will not highlight link in navbar if it does not have a trailing slash.

src/mantine-core/src/components/Popper/PopperContainer/PopperContainer.tsx
rtivital: It's fine

src/mantine-core/src/components/Popper/PopperContainer/PopperContainer.tsx
```diff
@@ -0,0 +1,39 @@
...
+
+  /** Whether to render the target element in a Portal */
+  withinPortal?: boolean;
+}
+
+export function PopperContainer({
+  children,
+  zIndex = 1,
+  className,
+  withinPortal,
```
rtivital, line 22: Let's set this to false by default to be more explicit

rtivital (commented):
Thanks for a PR! Everything seems fine, but I've completely forgot that we have a DatePickerBase component that also uses Popper – can you also add this option there (should propagate to DatePicker and DateRangePicker components)?

jvdsande:
Ah, my bad, I did a quick Find Everywhere for Popper and saw the ColorInput one, but I missed this one. I'll add it in a few :)

docs/src/docs/core/Popper.mdx
jvdsande: Done :)

src/mantine-core/src/components/Popper/PopperContainer/PopperContainer.tsx
jvdsande: Done in latest commit

jvdsande:
Added the prop in @mantine/dates relevant components :)

rtivital:
@jvdsande can you please fix issue from ci?

jvdsande:
Woops sorry, I thought I ran `yarn test` before commit, but I only ran `yarn jest`. 

Maybe a commit hook could be added to run Prettier?

rtivital:
Thanks for your work! These changes will be out with 3.3 release, if you need them asap I can publish an alpha release for you to get early access.

jvdsande:
Perfect, thanks :) 

No, I think current release frequency will be enough, we can wait a few weeks for this. I'll stick with 3.1.9 for now to start with, and update to 3.3.0 once it's out!

PR #446, opened 2021-11-21
[@mantine/hooks] Fixed scroll lock not working correctly on iOS safari

### LINKED ISSUE #300: @mantine/hooks use-scroll-lock does not prevent scrolling on iOS Safari
Mantine v 3.0.4
iOS Safari 15.0.1

The `use-scroll-lock` hook does not lock scrolling on iOS Safari (testing demo page https://mantine.dev/hooks/use-scroll-lock/).

Here's an up-to-date article that goes into the specific problems and how to properly prevent scrolling in iOS Safari that may be worth referencing for a fix: https://pqina.nl/blog/how-to-prevent-scrolling-the-page-on-ios-safari/

edit: fix link to article

iDavidB:
Closes #300.

This does create one issue, as it completely blocks any touch move input it also means that users can't use pinch to zoom. This could be fixed by checking the amount of touch inputs that are happening so that two finger inputs are allowed, however this would allow scrolling again with two fingers. 

I think it comes down to what we prefer, no touch move input at all, or allow scrolling to work with 2 fingers so that we are able to keep other multi-finger gestures working.


rtivital:
The problem with this approach is that is disables scroll on all elements, while we want to disable scroll only on body element. This is not obvious from the hook name, but if we disable touch events it will make modals and other scrollable elements unscrollable.

https://user-images.githubusercontent.com/10353856/142818098-db8ce31f-ef40-4260-b6a3-5bb1bee440f7.MP4

iDavidB:
I see. Then I too am defeated by this issue

PR #452, opened 2021-11-23
[@mantine/core] Popper: fallback placement adjustments

### LINKED ISSUE #434: Menu/Popover fallback placement broken in 3.2.1
Hello!

I've come across an issue with the latest version of Mantine, where `Menu` and `Popover` fallback placement don't seem to work anymore.

Version information:
- @mantine/core@3.2.1
- @mantine/hooks@3.2.1
- Chromium: 95.0.4638.69

Here is an attached screenshot of the behavior I'm observing:

image

The Menu placement is `end`, the position is `bottom`. With 3.1.x in this situation the placement correctly flipped to `start`, but not anymore.

I've seen than if editing the `Popper` component and adding `bottom-start` and `bottom-end` to the `fallbackPlacements` prop, I get the correct behavior back.

Maybe `fallbackPlacements` should be available on `Menu` and `Popover` and passed down to `Popper` to make this customizable?

Anyway, thanks for the awesome library, it's been a pleasure trying it out lately :)

EmilMalanczak:
`Menu`, `Popover`, `Tooltip` got extended props from `Popper` component - `allowPlacementChange` and `placementFallback`

`Popper` now has a function that by default setups `fallbackPlacements` based on initial position of the `Popper`

fixes #434


rtivital:
@EmilMalanczak this will be released as a patch, can you please make this PR mergable into master branch?

EmilMalanczak:
Sure, but not faster than tonight I'm at work till 6-7pm

rtivital:
@EmilMalanczak I've tried things out – all Popper modifiers were not necessary, it works well without them, can you please check if it work on your side in current master branch?

EmilMalanczak:
@rtivital Now when I've backed to `popper` docs I see that might be true - I possible misunderstood some info. However `onDirectionChange` was required to be able to manage changing direction on `Select` component - reverse order and navigation

EmilMalanczak:
I will move it level up directly to `SelectDropdown` component, should work with no issue - `modifiers` prop still allows to pass it down to `popper` instance

EmilMalanczak:
as it will be closely related to `Select` I think I'll also cover #454 issue but as a next PR to `next-minor`.

rtivital:
Sounds good, I completely forgot about direction change

PR #462, opened 2021-11-27
[@mantine/core] Select: improve keyboard navigation

### LINKED ISSUE #454: Select should remember the selected item index and start from there
- What package has an issue (@mantine/core, mantine/hooks or other)?
**core**
- What package version are you using?
**3.2.2**

The select should keep the state of the index of the selected item once closed.
At this moment, if you select an item, let's say the 4th of a list of 6, you close the select and you open again, the selected item is the 1st of the list.
It's true that the selected item remains with the visuals of selected, but if you press key down or key up, the cursor moves from the first item of the list.

Example:
select

The right behavior would be:

select2

Please, note how after opening the select again, it remembers the position where the cursor was

EmilMalanczak:
fixes #454 

I didn't see any opportunity for similar logic in`MultiSelect` and `Autocomplete` since they work a bit differently

### DIFF
--- a/src/mantine-core/src/components/Select/Select.tsx
+++ b/src/mantine-core/src/components/Select/Select.tsx
@@ -287,6 +287,9 @@ export const Select = forwardRef<HTMLInputElement, SelectProps>(
       );
     }, [inputValue]);
 
+    const getSelectedItemIndex = () =>
+      _value ? filteredData.findIndex((el) => el.value === _value) : -1;
+
     const handleInputKeydown = (event: React.KeyboardEvent<HTMLInputElement>) => {
       typeof onKeyDown === 'function' && onKeyDown(event);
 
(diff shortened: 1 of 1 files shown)


src/mantine-core/src/components/Select/Select.tsx
```diff
@@ -287,6 +287,9 @@ export const Select = forwardRef<HTMLInputElement, SelectProps>(
       );
     }, [inputValue]);
 
+    const getSelectedItemIndex = () =>
```
rtivital, line 290: Is it necessary to create a function, why just not put it in a variable?
```tsx
const selectedItemIndex = _value ? filteredData.findIndex((el) => el.value === _value) : -1
```

EmilMalanczak: hmm good point - tbh no idea why i did not do this as variable

EmilMalanczak:
shall I ignore the failing tests? They are not related to my changes

rtivital:
Yeah, I guess I'll need to work in a separate branch, not next-minor

PR #470, opened 2021-11-30
[@mantine/dates] - Added datesLocale property to MantineProvider & ma…

iDavidB:
…de all date components locale aware


rtivital:
Thanks, look good!

PR #478, opened 2021-12-04
[@mantine/hooks] use-page-leave

EmilMalanczak:
as potential improvement and extendable future we can:

return a boolean variable that indicates if mouse is out of page. 

accept `onReturn` function callback - left a comment with similar implementation


rtivital:
Thanks, looks great!

PR #479, opened 2021-12-04
[@mantine/core] Make direction switch optional on Select

jvdsande:
Implement a `switchDirectionOnFlip` prop for Select, MultiSelect and Autocomplete, allowing to opt out of direction switching.

It simply disables the `onDirectionSwitch` callback, effectively freezing the state into the default direction.


rtivital:
Thanks for PR! It works great

Issue #481, opened 2021-12-05
Drawer z-index fallback to 1000


lrvinye:
- What package has an issue (@mantine/core, mantine/hooks or other)?  @mantine/core
- What package version are you using?  3.3.1
- In which browser and operating system issue occurred? chrome
- If possible please attach link to codesandbox with reproduced issue
- Do you have an idea how to fix the issue?  check source code of Drawer
- Are you willing to participate in fixing this issue (it's totally fine if not) and create a pull request with the fix? yes

when I want to use Select in Drawer ,
I found that the dropdown of Select can not show correctly,
I thought that is cause by z-index
and I read the docs of Select & Drawer
the default z-index of Select and Drawer should be `getDefaultZIndex('popover')`  and  `getDefaultZIndex('modal')`
but actully the z-index of Drawer is `1000` (what i found in source code of Drawer and confirm it in my local test)
so i need to set `1001` to zIndex property of Select so that its dropdown can show correctly in my case

lrvinye:
I have create a PR here #482

rtivital:
Thanks for reporting and contributing! Fix for this issue was released in 3.3.3

PR #482, opened 2021-12-05
[@mantine/core] Drawer: fix Drawer z-index

lrvinye:
the default z-index of Drawer should be `getDefaultZIndex('modal')` rather than `1000`
issue here #481


rtivital:
Good catch, thanks!

PR #488, opened 2021-12-06
[core] Fix default value of ColorSchemeContext

### LINKED ISSUE #487: [core] 3.3.3 Does not build with blitz.js
The CI pipeline for my project fails when trying to update the core package to 3.3.3.

It fails with:

```
./node_modules/@mantine/styles/src/theme/ColorSchemeProvider.tsx:9:67
Type error: Argument of type 'null' is not assignable to parameter of type 'ColorSchemeContextProps'.

   7 | }
   8 | 
>  9 | const ColorSchemeContext = createContext<ColorSchemeContextProps>(null);
     |                                                                   ^
  10 | 
  11 | export function useMantineColorScheme() {
  12 |   const ctx = useContext(ColorSchemeContext);
```

This is my tsconfig.json

```
{
  "compilerOptions": {
    "target": "es5",
    "lib": ["dom", "dom.iterable", "esnext"],
    "baseUrl": "./",
    "allowJs": true,
    "skipLibCheck": true,
    "strict": false,
    "strictNullChecks": true,
    "forceConsistentCasingInFileNames": true,
    "noEmit": true,
    "noUncheckedIndexedAccess": true,
    "esModuleInterop": true,
    "module": "esnext",
    "moduleResolution": "node",
    "resolveJsonModule": true,
    "isolatedModules": true,
    "jsx": "preserve",
    "incremental": true,
    "tsBuildInfoFile": ".tsbuildinfo",
    "experimentalDecorators": true
  },
  "exclude": ["node_modules", "**/*.e2e.ts", "cypress"],
  "include": ["blitz-env.d.ts", "**/*.ts", "**/*.tsx"]
}
```

And I am using Typescript 4.5.2

It is working with core version 3.0.5, but after that it fails for every version.

JohannesKlauss:
This fixes #487 

Let me know if this is sufficient enough or if I should change something.


rtivital:
Hmm, I think that this maybe another issue. Currently, we are including source code in @mantine/styles package, if we exclude it, then the issue will be resolved.

JohannesKlauss:
As mentioned here https://github.com/mantinedev/mantine/issues/487#issuecomment-986659603 I think you are right about that. Closing the PR.

PR #495, opened 2021-12-09
Docs typo fix on MultiSelect page

chrisrudman:
The docs for MultiSelect referenced labelComponent, which is wrong. It should be valueComponent.


rtivital:
Thanks!

PR #497, opened 2021-12-09
[@mantine/core] Select: Add SelectItem value number type

huijiewei:


rtivital:
SelectItem does not support number as value, that's why it's not included in type

PR #511, opened 2021-12-12
Fix typo in `GlobalStyles` fontSize property

MoSheikh:
Just noticed that the invalid property `fontSizes` was being added to the css of the document's `body` element when using `GlobalStyles`. I believe this is intended to be `fontSize` instead.

#### Error in Chrome:
image


rtivital:
Thanks!

MoSheikh:
No problem, thanks for the great library

orel91:
Hi @rtivital , when will this fix be released?

rtivital:
It was released several months ago

orel91:
> It was released several months ago

As of version 3.6.9, the typo is still present: https://github.com/mantinedev/mantine/blob/41729b6aec83bde10cf0596f4c345eca7cec8315/src/mantine-styles/src/theme/MantineProvider.tsx#L62

PR #520, opened 2021-12-15
Change the bug_report Issue Template to a Issue form

ruessej:
I think it is easier to use issue forms instead of issue templates.


rtivital:
Can you convert other issue templates to forms as well?

ruessej:
@rtivital I converted the other forms as well

rtivital:
Alright, let's see it they'll work

PR #526, opened 2021-12-15
[@mantine/core] Skeleton: Add the `count` prop.

Ultra-Instinct-05:
This PR proposes to add a `count` prop to the Skeleton component. This will help in quickly generating Skeleton components without having to manually write that many Skeleton components. The default value is `1` to maintain backwards compatibility.

TODO Checklist
---

- [ ] Storybook demos.
- [ ] Documentation.


rtivital:
This will not work well, in most cases Skeleton should have margins, and it is hard to control it with `count` prop

Ultra-Instinct-05:
No problem ! I probably should have asked in Discord first before working.

Issue #521, opened 2021-12-15
[NumberInput] Support of different decimal separators


ruessej:
- Usage example, including component/hook API: use , instead of a . as decimal separator
- Possible implementation – describe how feature can be implemented: You could specify an object with the separator or an enum where you have fixed formats
- Do you want to contribute this feature and create a pull request? I will try it

rtivital:
What prop will it be to override separator, I do not understand how we can use an object, all I can think of is a string:
```tsx
<NumberInput decimalSeparator="," />
```

ruessej:
Yes you right i though of the thousands separator too

```ts
<NumberInput separators={ decimal: ",", thousands: "." } />
```

but when i think now about it we dont need that separator or?

rtivital:
It seems like a hard thing to do, but feel free to try with your proposed api

jerebtw:
@rtivital is it a problem when all . are replaced with the custom separator?

LucasLemanowicz:
I suggest looking into the built-in JS `Intl.NumberFormat` API: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat

Would provide good extendability beyond just that one specific use case and let you do more with less code.

ruessej:
Okay i will look in to it later

ruessej:
The simplest solution would be to let the user create their own parse and format function.
But I don't know if this is too much for this simple API.

rtivital:
Fixed in 3.5.0

PR #527, opened 2021-12-15
[@mantine/core] NumberInput: Add decimal separator prop

jerebtw:
https://github.com/mantinedev/mantine/issues/521


LucasLemanowicz:
Thoughts on taking advantage of the JS built-in Intl.NumberFormat API? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/NumberFormat

Added benefit that it would allow for other types of formatting (such as currencies), and also provide better integration with other JS libraries.

rtivital:
@LucasLemanowicz the problem with NumberFormat is that we need to convert formatted number back to number value and I do not know a good way to do so

rtivital:
@jerebtw can you please resolve conflicts?

jerebtw:
I resolved the conflicts

ruessej:
I will fix the error later when im at home

jerebtw:
Cleaned version: https://github.com/mantinedev/mantine/pull/535

PR #529, opened 2021-12-16
[@mantine/hooks] use-form: Change UseForm type to return value of use…

MartinMuzatko:
As discussed in Discord, I am proposing this change to make re-use of `use-form` easier in components getting the return value of `useForm` passed.

Example:

``` tsx
const EventEditor = (event: Event) => {
    const eventForm = useForm({ initialValues: event })
    return <>
        <EventEditorFinancialDetails eventForm={eventForm}>
        <EventEditorRegistrationDetails eventForm={eventForm}>
    </>
}

const EventEditorFinancialDetails(props: { eventForm: UseForm<Event> }) => {
    // ...
}

```

Note: `UseForm` is not documented, but could potentially cause breaking changes, when users depend on the type. Please keep that in mind when updating. @rtivital agreed to rename the type.

Thank you!

### DIFF
--- a/src/mantine-hooks/src/use-form/use-form.ts
+++ b/src/mantine-hooks/src/use-form/use-form.ts
@@ -9,17 +9,41 @@ export type UseFormErrors<T> = {
   readonly [P in keyof T]?: React.ReactNode | null;
 };
 
-export interface UseForm<T> {
+interface UseFormInput<T> {
   validationRules?: ValidationRule<T>;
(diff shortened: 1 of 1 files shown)


rtivital:
@MartinMuzatko tests are failing

MartinMuzatko:
I fixed the type error @rtivital

rtivital:
Thanks!

PR #535, opened 2021-12-17
[@mantine/core] NumberInput: Add decimal separator prop

jerebtw:
This is for https://github.com/mantinedev/mantine/issues/521


rtivital:
Thanks!

PR #539, opened 2021-12-18
[@mantine/core] Add Completed component for Stepper

jvdsande:
Hello!

As discussed on Discord, this PR adds a `Stepper.Completed` component (that only accepts children) to the `Stepper` component.

If a `Stepper.Completed` component is found in the `Stepper`'s children, its content will be displayed when no active `Stepper.Step` is found.


docs/src/docs/core/Stepper.mdx
```diff
@@ -83,6 +83,12 @@ Loader will replace step icon. You can configure default loader in [MantineProvi
 
 <Demo data={StepperDemos.loading} />
 
+## Completed step
+
+You can use `Stepper.Completed` component to add a completed step, which content will be shown once all steps are completed.
+
+<Demo data={StepperDemos.completedStep} />
+
```
jvdsande, line 91: I was not sure if we wanted to promote this functionality with its own section, or if the `Usage` section would be enough. In doubt, I added this. I can remove it if it's deemed unnecessary.

scripts/docgen.ts
```diff
@@ -12,6 +12,7 @@ const EXTRA_FILES_PATHS = [
   '../src/mantine-core/src/components/Timeline/TimelineItem/TimelineItem.tsx',
   '../src/mantine-core/src/components/Slider/Slider/Slider.tsx',
   '../src/mantine-core/src/components/Slider/RangeSlider/RangeSlider.tsx',
+  '../src/mantine-core/src/components/Stepper/Step/Step.tsx',
```
jvdsande, line 15: I added this extra file for docgen because otherwise the `Step Component props` section in `Stepper` props tabs is empty.

src/mantine-core/src/components/Stepper/demos/usage.tsx
```diff
@@ -24,9 +24,11 @@ function Demo() {
         <Stepper.Step label="Final step" description="Get full access">
           Step 3 content: Get full access
         </Stepper.Step>
-      </Stepper>
 
-      {active === 3 && <div>Form completed, click back button to get to previous step</div>}
+        <Stepper.Completed>
+          Form completed, click back button to get to previous step
+        </Stepper.Completed>
+      </Stepper>
```
jvdsande, line 31: Updated the usage to use new `Stepper.Completed` instead of manually checking

src/mantine-core/src/components/Stepper/Stepper.tsx
```diff
@@ -122,7 +127,7 @@ export const Stepper: StepperComponent = forwardRef<HTMLDivElement, StepperProps
       return acc;
     }, [] as React.ReactNode[]);
 
-    const content = filteredChildren[active]?.props?.children;
+    const content = filteredChildren[active]?.props?.children ?? completedStep?.props?.children;
```
rtivital, line 130: We should use `active` to check whether to show completed step or current step. If I'm not wrong, with current implementation it will show the completed step if one of the steps does not have content. 

The logic should be simple: `filteredChildren.length - 1 > active ? completed : current step`

jvdsande: Ah yes indeed, I did not think of this use case :+1:

docs/src/docs/core/Stepper.mdx
rtivital: I think we should add this to usage and remove this section

jvdsande: Removed :) 

The usage example has been modified to use the new component already.

src/mantine-core/src/components/Stepper/Stepper.tsx
jvdsande: Fixed, added the condition

rtivital:
Thanks!

PR #543, opened 2021-12-20
[@mantine/core] InputNumber: Add props to increment values on holding step arrows

### LINKED ISSUE #525: [Feature Request] NumberInput holding increment/decrement
### Usage example, including component/hook API

I would like to see the NumberInput component be able to handle holding down on the increment or decrement keys to exponentially speed up the number change. Is this something that might be possible?

### Possible implementation – describe how the feature can be implemented

_No response_

### Do you want to contribute this feature and create a pull request

_No response_

nrayburn-tech:
Add the ability to automatically increment the value of an InputNumber, by clicking and holding the step arrow buttons.

If either `stepIncrementInterval` or `stepIncrementInitialDelay` are not supplied, then the current functionality is retained and there is not any automatic incrementing.

#525 specifically asks for exponential increase, but I think that a single delay value in milliseconds handles most use cases without much complexity.

If providing something other than a single delay value is required, then I think the type for `stepIncrementInterval` should change and the prop can be a function to calculate the rate.  The type would change from `stepIncrementInterval?: number` to `stepIncrementInterval?: number | (count: number) => number` where count is the current increment count and the return value should be the delay to use.
Example function that could then be provided.
```ts
function calcDelay(count: number): number {
  return 100 * count * count;
}
```

Fixes #525

### DIFF
--- a/src/mantine-core/src/components/NumberInput/NumberInput.tsx
+++ b/src/mantine-core/src/components/NumberInput/NumberInput.tsx
@@ -188,11 +188,11 @@ export const NumberInput = forwardRef<HTMLInputElement, NumberInputProps>(
     const shouldUseStepInterval =
       stepIncrementInitialDelay !== undefined && stepIncrementInterval !== undefined;
 
-    const onStepTimeoutRef = useRef<NodeJS.Timeout>(null);
+    const onStepTimeoutRef = useRef<number>(null);
 
     const onStepDone = () => {
       if (onStepTimeoutRef.current) {
-        clearTimeout(onStepTimeoutRef.current);
+        window.clearTimeout(onStepTimeoutRef.current);
         onStepTimeoutRef.current = null;
       }
     };
@@ -209,7 +209,10 @@ export const NumberInput = forwardRef<HTMLInputElement, NumberInputProps>(
       onStepHandleChange(isIncrement);
 
       if (shouldUseStepInterval) {
-        onStepTimeoutRef.current = setTimeout(() => onStepLoop(isIncrement), stepIncrementInterval);
+        onStepTimeoutRef.current = window.setTimeout(
+          () => onStepLoop(isIncrement),
+          stepIncrementInterval
+        );
       }
     };
 
@@ -217,7 +220,7 @@ export const NumberInput = forwardRef<HTMLInputElement, NumberInputProps>(
       event.preventDefault();
       onStepHandleChange(isIncrement);
       if (shouldUseStepInterval) {
-        onStepTimeoutRef.current = setTimeout(
+        onStepTimeoutRef.current = window.setTimeout(
           () => onStepLoop(isIncrement),
           stepIncrementInitialDelay
         );
--- a/src/mantine-core/src/components/NumberInput/NumberInput.tsx
+++ b/src/mantine-core/src/components/NumberInput/NumberInput.tsx
@@ -1,4 +1,4 @@
-import React, { MouseEvent, useState, useEffect, useRef, forwardRef } from 'react';
+import React, { useState, useEffect, useRef, forwardRef } from 'react';
 import { useMergedRef, assignRef, clamp } from '@mantine/hooks';
 import { DefaultProps, ClassNames } from '@mantine/styles';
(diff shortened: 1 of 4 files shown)


src/mantine-core/src/components/NumberInput/NumberInput.tsx
```diff
@@ -175,6 +185,52 @@ export const NumberInput = forwardRef<HTMLInputElement, NumberInputProps>(
       }
     }, [value]);
 
+    const shouldUseStepInterval =
+      stepIncrementInitialDelay !== undefined && stepIncrementInterval !== undefined;
+
+    const onStepTimeoutRef = useRef<NodeJS.Timeout>(null);
```
rtivital, line 191: use `number` instead of `NodeJS.Timeout`, Mantine is built for browser

src/mantine-core/src/components/NumberInput/NumberInput.tsx
```diff
@@ -175,6 +185,52 @@ export const NumberInput = forwardRef<HTMLInputElement, NumberInputProps>(
...
+      } else {
+        decrementRef.current();
+      }
+    };
+
+    const onStepLoop = (isIncrement: boolean) => {
+      onStepHandleChange(isIncrement);
+
+      if (shouldUseStepInterval) {
+        onStepTimeoutRef.current = setTimeout(() => onStepLoop(isIncrement), stepIncrementInterval);
```
rtivital, line 212: To make `setTimeout` work with `number` type, replace it with `window.setTimeout`, same applies to `clearTimeout`

rtivital:
Thanks for a PR, It works great! I think we should change props names:
- `stepIncrementInterval`
- `stepIncrementInitialDelay`

Increment means that it will go up, while it is also responsible for decrementing values. Something like `stepHoldInterval` and `stepHoldDelay`, or maybe you can propose something more descriptive?
Also, I think that adding function to `stepIncrementInterval` is a great idea, do you want to add this feature?

nrayburn-tech:
Your proposed names are better.
The only concern I have with the names is that these delays only apply when holding the mouse down, not when holding the keyboard keys down.  Should the names be more descriptive to indicate this?

I’ll change the type to allow a function as well as tests/docs for using the function.

rtivital:
I think that this feature should be supported with arrow buttons as well instead of restricting it to mouse events only

nrayburn-tech:
When holding an arrow key down, the value already updates multiple times because the “keydown” event will emit while still holding the key.

Are you saying this behavior should be overridden to use the same “stepHold*” property values for updating instead of how the browser handles it now?
If so, should we prevent multiple updates when the “stepHold*” properties are not supplied to align with how the value updates work when using the mouse?

rtivital:
When user holds button, `event.repeat` is true – https://codesandbox.io/s/wizardly-ardinghelli-zjlnn?file=/src/App.tsx
If it sounds too complicated, maybe we should leave it as is, I think it maybe just an edge case to support function `stepIncrementInterval`.

nrayburn-tech:
Updated with the function type for `stepHoldInterval` as well as using the `stepHoldInterval` and `stepHoldDelay` for the `keydown` event.  If the `stepHoldInterval` and `stepHoldDelay` are not supplied, then the previous behavior for `keydown` is retained and the value is stepped for each `keydown` repeat.

rtivital:
Thanks, it works great!

jerebtw:
I get a error when running jest

```cmd
Summary of all failing tests
 FAIL  src/mantine-core/src/components/NumberInput/NumberInput.test.tsx
  ● @mantine/core/NumberInput › steps value with controls on hold mousedown with a function delay

    expect(jest.fn()).toHaveBeenLastCalledWith(...expected)

    Expected: 50
    Received
           3: 30
    ->     4: 40

    Number of calls: 4

      276 |     });
      277 |
    > 278 |     expect(spy).toHaveBeenLastCalledWith(50);
          |                 ^
      279 |   });
      280 |
      281 |   it('steps value with controls on hold keydown and stepHoldInterval', async () => {

      at Object.<anonymous> (src/mantine-core/src/components/NumberInput/NumberInput.test.tsx:278:17)
```

Issue #544, opened 2021-12-20
[Feature Request] Prop to disable transfer all on TransferList


axhl:
### Usage example, including component/hook API

There are instances where "transfer all" makes for bad UX as any form of "undo" then resets also individual transfers that have been made prior to the "transfer all."  This is particularly an issue where transfer all is accidental. Allowing each of the below actions to be individually disabled would increase utility of the Transferlist component. 

<img width="255" alt="Screenshot 2021-12-20 at 12 21 55" src="https://user-images.githubusercontent.com/62632027/146761042-e7d6602b-b204-4de7-9858-153497ba07f9.png">

Thoughts welcome. Thanks. 

### Possible implementation – describe how the feature can be implemented

`disableTransferAll` implemented as a component prop.

axhl:
Closing with thanks to @Ultra-Instinct-05

PR #545, opened 2021-12-20
[@mantine/core] TransferList: Add the `showTransferAll` prop.

Ultra-Instinct-05:
This PR adds the `hideTransferAll` prop, which controls whether the Transfer All buttons (`<<` && `>>`) get's shown. The default value for this prop is false to retain backwards compatibility.

Addresses #544


Ultra-Instinct-05:
In order to be consistent with naming, I am going to change the prop name to `showTransferAll` (consistent with `showSearch`)

rtivital:
Thanks!

PR #554, opened 2021-12-22
[@mantine/core] Text: Fix tsc error

jerebtw:


LucasLemanowicz:
Oops don't know how I missed that. Good catch!

rtivital:
Thanks!

Issue #560, opened 2021-12-22
[Feature Request] Tooltip: Add radius


jerebtw:
### Do you want to contribute this feature and create a pull request

Yes

jerebtw:
That's my only problem with the radius:

![image](https://user-images.githubusercontent.com/49494752/147039406-107d39cc-8996-4be7-a423-bed3863fb0af.png)

rtivital:
I cannot think of a good way to enable arrow for round tooltip, maybe you should set placement to center

jerebtw:
Yes this only happens when you put the placement to the beginning or ending

rtivital:
Fixed in 3.5.0

PR #563, opened 2021-12-22
[@mantine/core] Tooltip: Add the `radius` prop

jerebtw:
Fixes https://github.com/mantinedev/mantine/issues/560

### DIFF
--- a/src/mantine-core/src/components/Popper/Popper.tsx
+++ b/src/mantine-core/src/components/Popper/Popper.tsx
@@ -21,6 +21,9 @@ export interface SharedPopperProps {
   /** Arrow size in px */
   arrowSize?: number;
 
+  /** Arrow distance to the left/right * arrowSize */
+  arrowDistance?: number;
+
   /** Renders arrow if true */
   withArrow?: boolean;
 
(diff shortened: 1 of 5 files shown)


rtivital:
How does it solve the bug with arrow discussed in the issue?

jerebtw:
We still need to find a fix for that problem

rtivital:
I do not think that we should provide a feature that does not work, so this PR can be merged only with a fix

jerebtw:
Okay I will think about a fix

src/mantine-core/src/components/Popper/Popper.tsx
```diff
@@ -21,6 +21,9 @@ export interface SharedPopperProps {
   /** Arrow size in px */
   arrowSize?: number;
 
+  /** Arrow distance to the left/right * arrowSize */
```
jerebtw, line 24: @rtivital is that comment right?

jerebtw:
> I do not think that we should provide a feature that does not work, so this PR can be merged only with a fix

I fixed the issue 👍🏻 I just moved the arrow a little

rtivital:
Thanks, it works great!

PR #567, opened 2021-12-23
[@mantine/core] Make Select re-render the input on label change

jvdsande:
Hello!

This PR fixes the fact that Select's input does not re-render if the label of the selected item changes in the `data` array.

I did not have to change MultiSelect which didn't have the issue. I added a story for both Select and MultiSelect to see it in action.

### DIFF
--- a/src/mantine-core/src/components/Select/Select.tsx
+++ b/src/mantine-core/src/components/Select/Select.tsx
@@ -254,6 +254,12 @@ export const Select = forwardRef<HTMLInputElement, SelectProps>(
       }
     }, [_value]);
 
+    useEffect(() => {
+      if (selectedValue && (!searchable || !dropdownOpened)) {
+        handleSearchChange(selectedValue.label);
+      }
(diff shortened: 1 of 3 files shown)


src/mantine-core/src/components/Select/Select.tsx
```diff
@@ -254,6 +254,12 @@ export const Select = forwardRef<HTMLInputElement, SelectProps>(
       }
     }, [_value]);
 
+    useEffect(() => {
+      if (selectedValue && (!searchable || !dropdownOpened)) {
+        handleSearchChange(selectedValue.label);
+      }
+    }, [selectedValue?.label]);
```
jvdsande, line 261: I targeted the effect trigger to directly the currently selected item's label. This way we avoid any unnecessary re-renders (such as when the data array reference has changed but the actual data is the same, or when another label not currently selected changes)

rtivital:
Thanks!

Issue #562, opened 2021-12-22
[@mantine/core] NumberInput: Jest error


jerebtw:
### What package has an issue

@mantine/core

### Describe the bug

I get a error when running jest (On dev branch)

```cmd
Summary of all failing tests
 FAIL  src/mantine-core/src/components/NumberInput/NumberInput.test.tsx
  ● @mantine/core/NumberInput › steps value with controls on hold mousedown with a function delay

    expect(jest.fn()).toHaveBeenLastCalledWith(...expected)

    Expected: 50
    Received
           3: 30
    ->     4: 40

    Number of calls: 4

      276 |     });
      277 |
    > 278 |     expect(spy).toHaveBeenLastCalledWith(50);
          |                 ^
      279 |   });
      280 |
      281 |   it('steps value with controls on hold keydown and stepHoldInterval', async () => {

      at Object.<anonymous> (src/mantine-core/src/components/NumberInput/NumberInput.test.tsx:278:17)
```

### In which browser did the problem occur

Jest

nrayburn-tech:
I am pretty sure the tests passed in CI and they are passing for me as well.  Using the `next-minor` branch.

I wonder if the use of timers is causing some flakiness?

jerebtw:
Probably because sometimes I get an error and sometimes not

jerebtw:
I also received another error message after running jest a few times:

```cmd
 FAIL  src/mantine-core/src/components/NumberInput/NumberInput.test.tsx
  ● @mantine/core/NumberInput › steps value with controls on hold mousedown with a millisecond delay value

    expect(jest.fn()).toHaveBeenLastCalledWith(...expected)

    Expected: 40
    Received
           2: 20
    ->     3: 30

    Number of calls: 3

      249 |     });
      250 |
    > 251 |     expect(spy).toHaveBeenLastCalledWith(40);
          |                 ^
      252 |   });
      253 |
      254 |   it('steps value with controls on hold mousedown with a function delay', async () => {

      at Object.<anonymous> (src/mantine-core/src/components/NumberInput/NumberInput.test.tsx:251:17)
```

rtivital:
I do not have any issues that you are referring to in this PR, please verify that your fork is synced with the main repo.

jerebtw:
My fork is synced with the main repo and I get this error only sometimes. (Not every run will show this fails)

rtivital:
[...]

PR #572, opened 2021-12-24
[@mantine/core] Progress: Add several props

Ultra-Instinct-05:
This PR proposes to add the following props to `Progress`.

1. Adds the `isIndeterminate` prop for striped, animated progress bars with `value = 100`. If this is set, the `value` prop is ignored. It's `false` by default.
2. Adds the `animated` prop for animating striped progress bars. It's `false` by default.
3. Adds the `label` prop which is a `string` that can be used to set any text within the progress bar. Useful if you want to indicate the current value in the progress bar (or any other form of metric related to `value`)
4. Adds the `showLabel` prop that controls whether the `label` get's shown. It defaults to `false`.

The tests seem to be failing but it seems to be related to #562


src/mantine-core/src/components/Progress/Progress.test.tsx
```diff
@@ -23,7 +23,12 @@ describe('@mantine/core/Progress', () => {
   itSupportsSx(Progress, defaultProps);
   itSupportsRef(Progress, defaultProps, HTMLDivElement);
   checkAccessibility([mount(<Progress value={80} aria-label="test-progress" />)]);
-  itSupportsStylesApi(Progress, defaultProps, Object.keys(ProgressStylesApi), 'Progress');
+  itSupportsStylesApi(
```
rtivital, line 26: Please remove this test entirely

rtivital:
Please also do this to merge this PR:
- Remove `showLabel` prop, it should detect label automatically based on `label` prop
- Rename `isIndeterminate` to `indeterminate` and `animate` to `animated`
- Add `indeterminate` and `animate` props to configurator
- Create new separate demo for new `label` prop and add it to documentation

src/mantine-core/src/components/Progress/Progress.test.tsx
Ultra-Instinct-05: Are you saying to remove the `itSupportsStylesApi` test entirely for `Progress` (why ?) or just the changes _I made to it_ to pass the tests ? (I am still coming to terms on understating the test suite, but I am assuming this has something to do with the generated Mantine classes on components).

The rest of the below mentioned changes will be taken care.

Ultra-Instinct-05:
Regarding `animate`, I feel it should remain `animate` (& not `animated`) to stay consistent with similar props elsewhere (Like the recently merged PR for `animate` for `Skeleton`), since both are booleans. It may confuse people later on that very similar props are named differently across components

src/mantine-core/src/components/Progress/Progress.test.tsx
jerebtw: `itSupportsStylesApi` was completely removed see https://github.com/mantinedev/mantine/commit/4b4943af00e6ecbfbb49030abed6354f2989314d

Ultra-Instinct-05: Hmm.. Okay. Wasn't aware it was removed for all components.

rtivital:
Thanks!

jerebtw:
@Ultra-Instinct-05 could you please look in to the animation prop again

Because if you set the `size` to xl the progress animation looks buggy

yDyjY8LJBh

PR #575, opened 2021-12-24
[@mantine/core] Skeleton: Add the `showAnimation` prop.

Ultra-Instinct-05:
This commit proposes to add a `showAnimation` prop to the `Skeleton` component. This can be used to disable skeleton animation. It's default value is `true` to maintain backwards compatibility.

References:
-------------
1. https://mui.com/api/skeleton/
2. https://ant.design/components/skeleton/#API


rtivital:
Please rename prop to `animate`

PR #577, opened 2021-12-24
[docs] Change resize handle location from left to right

Eldemarkki:
The resize handle is not located on the bottom left corner, but on the right.

Location on Firefox:
kuva

Location on Chrome:
kuva

Location on Edge:
kuva


rtivital:
Thanks!

Issue #581, opened 2021-12-24
[Feature Request] 12 hour format for TimeInput comopnent


jerguslejko:
### Usage example, including component/hook API

The `<TimeInput />` component only allows to input a timestamp in a 24-hour format. This is not ideal for US-based users who use a 12-hour format.

The idea here is to support entering a 12-hour format while also being able to specify `am`/`pm`. 

### Possible implementation – describe how the feature can be implemented

The API could be:

```tsx
<TimeInput format="12" /> // this implicitly shows the AM/PM selector
<TimeInput format="24" />
```

### Do you want to contribute this feature and create a pull request

_No response_ - I'd love to help out! Just not sure I can actually implement this in a nice way

rtivital:
How would you advise implementing am/pm selector? Should it be select or text input?

jerebtw:
This could work

![image](https://user-images.githubusercontent.com/49494752/147383581-2ec4ca98-eee9-468b-94e7-98b5a68d5d81.png)

jerguslejko:
yeah something like that would be great @jerebtw! showing the selector if `format == 12`, hiding it if `format == 24`. there should also be some validation (not allowing to input `18:20` if `format == 12`)

rtivital:
Added in 3.6.0

Issue #580, opened 2021-12-24
[@mantine/dates] TimeInput: not possible to set empty value


jerguslejko:
### What package has an issue

@mantine/dates

### Describe the bug

When setting the `value` prop of `<TimeInput />` to `undefined` (eg. `<TimeInput value={undefined} />`, the time value is still displayed (it uses the current timestamp by default).

My use-case: I have `<TimeInput />` as a form input. The default value is `undefined` as I want the user to fill it in. However, the user always sees a timestamp so they don't fill in the field assuming the current timestamp is being used. Under the hood however, the value is still `undefined`.

### In which browser did the problem occur

Chrome

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/competent-sammet-t6mje?file=/src/App.tsx

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

jerebtw:
The sandbox is empty

![image](https://user-images.githubusercontent.com/49494752/147519328-be3625b3-f3ca-4df7-b1f9-9e3d2d2d5822.png)

jerguslejko:
whops, fixed. thanks @jerebtw

jerebtw:
Is it important for you to have empty inputs, or can it just be `00:00`?

rtivital:
It already can be set to 00:00, I think providing an option to have a blank field is common case, so we should implement this

jerebtw:
@rtivital can I try this?

jerebtw:
Should it only be possible to leave the input blank at the beginning, or should the user also be able to clear the input?

jerguslejko:
@jerebtw for my current use-case, "just at the beginning" works just fine but in general, it'd be great to support a `clearable` prop

rtivital:
Alright, so what we can do about it:
- Allow input to be empty initially
- Add `clearable`prop, that will show clear button
- When value is filled, it can be cleared only with clear button 

@jerebtw you can work on it, you can do it in the same PR as your `format` on time input to avoid conflicts

PR #586, opened 2021-12-27
[@mantine/dates] TimeInput: Add `format` and `clearable` prop

jerebtw:
Fix for https://github.com/mantinedev/mantine/issues/581 and https://github.com/mantinedev/mantine/issues/580


rtivital:
Thanks for a PR! Select does not feel right in this case. It should be replaced with a regular input like every other field withing time inputs:
- When user presses up/down arrows on this field, it switches between am/pm
- When user types letter `a`, it switches to am (in case of range input: focus should be moved to next range hours field)
- When user types letter `p`, it switches to pm (in case of range input: focus should be moved to next range hours field) 
- When user finishes entering last number field (minutes or seconds) am/pm field should be focused
- Icon near am/pm should be removed

Currently, the flow of entering range is broken – there is not a good way to enter time in 12h format with just keyboard:

https://user-images.githubusercontent.com/10353856/147450732-2d5ac355-d7cf-440f-9694-9f5791702a34.mov

jerebtw:
Okay I will work on it. (NativeSelect would probably also work)

rtivital:
Yeah, maybe select will work for that, just don't use NativeSelect for that, it does not have any useful styles or features, `select` html element will work better in this case

jerebtw:
I corrected the keyboard flow and changed the component from Select to NativeSelect because NativeSelect was easier for me to work with. My problem is when I use the variant of NativeSelect the options also have no style. So when you are in dark mode it looks really bad. (see image)

image

rtivital:
Select element does not feel right either when working with keyboard only – there is no indication that it is focused, and it is not clear what the user should do to select it. I guess there is no other option rather than using regular text input for this like I've described in the first comment. 

https://user-images.githubusercontent.com/10353856/147537183-2e5d789a-59fa-415a-a5d3-64561fb3f5c4.mov

jerebtw:
@rtivital can you please check this. Thanks

rtivital:
@jerebtw can you please resolve conflicts?

jerebtw:
I resolved the conflicts and fix the code

jerebtw:
@rtivital is this new hook okay? (https://github.com/mantinedev/mantine/pull/586/commits/f995808a1246a47707f2be26d3b063cb06229ae0)

jerebtw:
Cleanup version: https://github.com/mantinedev/mantine/pull/607

PR #596, opened 2021-12-30
[@mantine/rte] RichTextEditor: Add the `radius` prop.

Ultra-Instinct-05:
This PR adds the `radius` prop to `RichTextEditor`.


rtivital:
Thanks!

PR #597, opened 2021-12-30
[@mantine/hooks] use-state-validator: init hook

achmurali:
This PR implements the use-state-validator hook.

yet to add documentation, shall add it once reviewed

### DIFF
--- a/src/mantine-hooks/src/use-state-validator/use-state-validator.test.ts
+++ b/src/mantine-hooks/src/use-state-validator/use-state-validator.test.ts
@@ -1,0 +1,43 @@
+/* eslint-disable no-console */
+import { act, renderHook } from '@testing-library/react-hooks';
+import { useStateValidator } from './use-state-validator';
+
+describe('@mantine/hooks/use-state-validator', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
+  });
+
+  it('returns correct initial state when valid', () => {
+    const hook = renderHook(() => useStateValidator('init', (val) => typeof val === 'string'));
+    expect(hook.result.current[0]).toBe('init');
+    expect(hook.result.current[2]).toBe(true);
+  });
+
+  it('returns correct initial state when invalid', () => {
+    const hook = renderHook(() => useStateValidator('init', (val) => typeof val === 'number'));
+    expect(hook.result.current[0]).toBe(undefined);
+    expect(hook.result.current[2]).toBe(false);
+  });
+
+  it('returns correct state on valid onChange', () => {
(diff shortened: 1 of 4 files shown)


src/mantine-hooks/src/use-state-validator/use-state-validator.ts
```diff
@@ -0,0 +1,18 @@
+import { useState } from 'react';
+
+export function useStateValidator<T>(
+  initialValue: T | null | undefined,
+  validator: (val: T | null | undefined) => boolean
+) {
+  let isValid = validator(initialValue);
+  const [valid, setValid] = useState<boolean>(isValid);
+  const [value, setValue] = useState<T | null | undefined>(isValid ? initialValue : undefined);
```
rtivital, line 9: initialValue should always be set, no matter of validation state, same in handleChange function

src/mantine-hooks/src/use-state-validator/use-state-validator.test.ts
```diff
@@ -0,0 +1,43 @@
+/* eslint-disable no-console */
+import { act, renderHook } from '@testing-library/react-hooks';
+import { useStateValidator } from './use-state-validator';
+
+describe('@mantine/hooks/use-state-validator', () => {
+  beforeEach(() => {
+    jest.clearAllMocks();
```
rtivital, line 7: Why do we need to clear mocks?

src/mantine-hooks/src/use-state-validator/use-state-validator.test.ts
```diff
@@ -0,0 +1,43 @@
+/* eslint-disable no-console */
```
rtivital, line 1: Redundant no-console

src/mantine-hooks/src/use-state-validator/use-state-validator.ts
achmurali, line 9: I have used `useState` for managing state, should I change it to useRef and push updates manually using useForceUpdate and etc?

rtivital: No, it should be use-state

jerebtw:
I have added such a hook in another pr https://github.com/mantinedev/mantine/pull/586

https://github.com/mantinedev/mantine/pull/586/commits/f995808a1246a47707f2be26d3b063cb06229ae0

rtivital:
@jerebtw please discuss in Discord before adding anything new, the hook that you've added is not the same and works differently, I would prefer the implementation of @achmurali because we've intended to add this for some time

jerebtw:
Ohh okay, I needed that feature for the am/pm so added a hook. And I think we can add both because my hook is quit different from this.

Issue #592, opened 2021-12-29
Callback `onClose` is not called when modal window opened by modal manager is dismissed


vlktomas:
### What package has an issue

@mantine/modals

### Describe the bug

From version 3.4.1 the `onClose` callback is not called when you close modal opened by modal manager:

```tsx
modals.openConfirmModal({
    title: 'Confirm modal',
    onClose: () => console.log('not called when "x" is clicked'),
})
```

Probably related to https://github.com/mantinedev/mantine/issues/588

### In which browser did the problem occur

Google Chrome

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

rtivital:
Thanks for reporting! If you want, you can submit a PR with a fix. This issue will be fixed in next minor release.

rtivital:
Fixed in 3.5.0

PR #601, opened 2021-12-30
[@mantine/modals] Fix onClose callback not called in closeAll function

vlktomas:
This PR fixes https://github.com/mantinedev/mantine/issues/592


rtivital:
Thanks!

PR #606, opened 2022-01-01
[@mantine/core] Select: allowDeselect prop

achmurali:
This PR adds `allowDeselect` prop to Select component.

Features - 
- add deselectable feature.
- making sure the dropdown doesn't close on deselecting

### DIFF
--- a/src/mantine-core/src/components/Select/Select.tsx
+++ b/src/mantine-core/src/components/Select/Select.tsx
@@ -185,7 +185,7 @@ export const Select = forwardRef<HTMLInputElement, SelectProps>(
       zIndex = getDefaultZIndex('popover'),
       name,
       dropdownPosition,
-      allowDeselect = false,
+      allowDeselect,
       ...others
     }: SelectProps,
     ref
@@ -206,7 +206,8 @@ export const Select = forwardRef<HTMLInputElement, SelectProps>(
       cancelable: false,
       isList: true,
     });
-    const isDeselectedValue = useRef<boolean>(false);
+
+    const isDeselectable = allowDeselect === undefined ? clearable : allowDeselect;
 
     const setDropdownOpened = (opened: boolean) => {
       _setDropdownOpened(opened);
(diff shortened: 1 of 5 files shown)


src/mantine-core/src/components/Select/Select.tsx
```diff
@@ -202,6 +206,7 @@ export const Select = forwardRef<HTMLInputElement, SelectProps>(
       cancelable: false,
       isList: true,
     });
+    const isDeselectedValue = useRef<boolean>(false);
```
rtivital, line 209: There is no need to keep deselected state, dropdown should close when item is clicked – if user deselects value it means that the user does not want it and thus dropdown should be closed

src/mantine-core/src/components/Select/demos/deselectable.tsx
```diff
@@ -0,0 +1,40 @@
...
+<Select
+  label="Your favorite framework/library"
+  placeholder="Pick one"
+  data={[
+    { value: 'react', label: 'React' },
+    { value: 'ng', label: 'Angular' },
+    { value: 'svelte', label: 'Svelte' },
+    { value: 'vue', label: 'Vue' },
+  ]}
+  allowDeselect
```
rtivital, line 14: Do not put boolean props after large data prop, it should be:
```tsx
<Select
  allowDeselect
  label="Your favorite framework/library"
  placeholder="Pick one"
  data={[
    { value: 'react', label: 'React' },
    { value: 'ng', label: 'Angular' },
    { value: 'svelte', label: 'Svelte' },
    { value: 'vue', label: 'Vue' },
  ]}
/>
```

rtivital (changes requested):
If Select is clearable, `allowDeselect` should be set to `true` by default unless `allowDeselect` is set to `false` explicitly:

```tsx
<Select /> // -> user CANNOT deselect
<Select clearable /> // -> user CAN deselect
<Select clearable allowDeselect={false} /> // -> user CANNOT deselect
```

rtivital:
Thanks, great job!

PR #608, opened 2022-01-01
Fixed useHash not getting initial value

jacob-j:
The hashchange event is not firing on initial load, adding window.location.hash directly to the useState would solve this.
Tried it locally and these changes make it work.

Let me know if something is wrong.

### DIFF
--- a/src/mantine-hooks/src/use-hash/use-hash.ts
+++ b/src/mantine-hooks/src/use-hash/use-hash.ts
@@ -4,7 +4,7 @@ import { useWindowEvent } from '../use-window-event/use-window-event';
 const browser = typeof window !== 'undefined';
 
 export function useHash(): readonly [string | null, (hash: string) => void] {
-  const [hash, setHashValue] = useState<string>(browser ? window.location.hash : "");
+  const [hash, setHashValue] = useState<string>(browser ? window.location?.hash : "");
 
   const setHash = (value: string) => {
     window.location.hash = value;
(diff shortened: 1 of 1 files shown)


src/mantine-hooks/src/use-hash/use-hash.ts
```diff
@@ -2,7 +2,7 @@ import { useState } from 'react';
 import { useWindowEvent } from '../use-window-event/use-window-event';
 
 export function useHash(): readonly [string | null, (hash: string) => void] {
-  const [hash, setHashValue] = useState<string>(null);
+  const [hash, setHashValue] = useState<string>(window.location.hash);
```
rtivital, line 5: It will not work, window object is not accessible during ssr, thus hook will throw an error

jacob-j: Oh wasn't thinking about backend rendering... maybe we can run the set function under an effect?

rtivital: Hook must check the environment, see any other hook that uses window, e.g. https://github.com/mantinedev/mantine/blob/master/src/mantine-hooks/src/use-viewport-size/use-viewport-size.ts

jacob-j: Sorry for disappearing, I accidentally spilled an entire cup of coffee on my mac :(
I added the check for window object, and also added optional chaining if needed.

Would this work?

jacob-j: Compile failed due to prettier, I am currently laptop-less, so can only do changes in the browser...
Feel free to close this one while my laptop is getting repaired

rtivital:
Fixed in 3.6.1

PR #610, opened 2022-01-02
[@mantine/core] Alert: Add a `variant` prop.

Ultra-Instinct-05:
This PR proposes to add an `outline` prop to the `Alert` component to show an outline instead of filling the Alert container with the chosen color.

Reference: https://mui.com/components/alert/#outlined

### DIFF
--- a/src/mantine-core/src/components/Alert/Alert.styles.ts
+++ b/src/mantine-core/src/components/Alert/Alert.styles.ts
@@ -1,69 +1,94 @@
-import { createStyles, MantineColor, MantineNumberSize } from '@mantine/styles';
+import {
+  createStyles,
+  MantineColor,
+  MantineNumberSize,
+  getSharedColorScheme,
+} from '@mantine/styles';
 
 interface AlertStyles {
   color: MantineColor;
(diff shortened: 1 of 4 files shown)


rtivital:
That's not how variants are added, see Badge component as reference. Since you've already added an outline variant, please also add "filled" variant as well, see Badge for variants names

Ultra-Instinct-05:
I have now modified to add a `variant` prop that takes 3 possible values, `light`, `filled` & `outline`. It will be `light` by default to maintain the previous look.

One thing to note is that previously for `Alert` (https://mantine.dev/core/alert/), in dark mode, the color was not being applied to the background (only to the `title` & `icon`). With subsequent addition of `variant`, this is also fixed.

I have not added `gradient` because I have not seen any use of a `gradient` alert used in the wild. So I don't think that option will be used.

rtivital:
Please do not change design of `light` version of Alert, it should ramain as is

Ultra-Instinct-05:
Made the necessary changes to keep the original `light` version of Alert.

rtivital:
Thanks!

PR #617, opened 2022-01-03
Fix typo

icflorescu:
Should be *if os **cannot** be identified* instead of *if os **can** be identified*.


rtivital:
Thanks!

icflorescu:
You're welcome. And congrats on your outstanding work!

PR #621, opened 2022-01-04
[@mantine/core] TransferList: Fix border issues.

Ultra-Instinct-05:
Reported by user **David** on discord. https://discord.com/channels/854810300876062770/861521000985264138/927839935690604604

This PR fixes a situation where the border radius of the input search and action icons were interfering with the overall border radius of the transfer list body.

So the top left & right border radii of input & icons needs to match that of the body for the radii to be uniform.


rtivital:
Thanks!

PR #630, opened 2022-01-07
[@mantine/core] Image: fix possible permanent placeholder after hydration

### LINKED ISSUE #629: Race condition between hydration and image load event
### What package has an issue

@mantine/core

### Describe the bug

When an Image with placeholder is server-side rendered, the component displayed in browser can be hydrated after load event and in this case will keep displaying placeholder.

Any

_No response_

Yes

Yes

### Possible fix

Looking at this issue's comment https://github.com/facebook/react/issues/15446#issuecomment-484709466

I could implement a fix for Image with a useEffect

```typescript
useEffect(() => {
  const { current } = internalImgRef;
  if (current?.complete) {
    setError(current.naturalHeight === 0);
    setLoaded(current.naturalHeight !== 0);
  }
}, [src])
```

Stouffi:
This PR fixes #629


.nvmrc
```diff
@@ -1 +1 @@
-v14.9.0
+v14.17.0
```
rtivital, line 1: Is there any reason to change node version?

rtivital:
Thanks for a PR, have you tested it to work in ssr environment, e.g. in [next template](https://github.com/mantinedev/mantine-next-template)?

.nvmrc
Stouffi: I was stuck with an error stating that a dependency required node >= 14.17 during my first `yarn install`

rtivital: Alright, then let's leave it in 14.17 then

Stouffi:
@rtivital I encountered the issue with my stack which is viteJS + SSR and with the fix, the problem with placeholder disappeared.

rtivital:
Okay, I'll run some more tests with it then

Stouffi:
Oh btw, did you mean testing **manually** with other SSR environment? I couldn't figure out how to add a test with an html string and an hydration using `@testing-library/react` in `src/mantine-core/src/components/Image/Image.test.tsx`

rtivital:
Yes, I do not think it is possible to check it within jest-dom environment, so it should be checked manually at least in Next.js

rtivital:
I've checked with next, it works fine

PR #632, opened 2022-01-07
[@mantine/core] Make Accordion heading-level customizable

jvdsande:
As discussed on Discord, allow setting the level for Accordion Item headings.

It still defaults to three, in order not to be breaking.


rtivital:
Thanks!

PR #636, opened 2022-01-08
[@mantine/core] AppShell: Update docs about wrapped components

LucasLemanowicz:
Due to how the `AppShell` component is architected, I've noticed that if you are setting a `height` on a `Header` component or a `width` on a `Navbar` component, you need to explicitly set it on the wrapped component, otherwise the layout breaks. It's not enough to just do `{...remainingProps}`


rtivital:
Thanks!

PR #637, opened 2022-01-08
[@mantine/core] AppShell: Fix responsive layout bug

LucasLemanowicz:
Before applying `100vw`:
Screen Shot 2022-01-08 at 2 20 55 PM

After applying `100vw` (also had to add a div with `margin-right: 16px`, but at least it works here):
Screen Shot 2022-01-08 at 2 21 31 PM

Can also consider adding `height: 100vh`, but I'm not an expert at CSS and don't want to fix something that doesn't seem broken.


rtivital:
Thanks!

PR #648, opened 2022-01-11
[@mantine/core] Divider: Fixed labels demo when `RTL` is enabled

jerebtw:
Before this PR:

image

After this PR:

image


rtivital:
Thanks!

PR #668, opened 2022-01-14
[docs] Fix props type in `CustomNavbar` and `CustomHeader` examples

arthurgeek:
Hello,

Since your `CustomHeader` and `CustomNavbar` contains it's own children that's not passed as an argument when using in the example, we need to change the props type to remove the children. We could also make the `children` as an optional argument inside `NavbarProps` or `HeaderProps`, but that's probably not the best idea.


rtivital:
Thanks!

PR #669, opened 2022-01-14
[docs] Fix incorrect import in Dark theme docs

arthurgeek:
Hello,

This fix an example where `useHotKeys` is used but `useWindowEvent` is imported.


rtivital:
Thanks!

PR #670, opened 2022-01-15
Fix(docs): Idle demos

benyou1969:
Copy / paste is not working cause `idleStatus` should be renamed to `idle`


rtivital:
Thanks!

PR #672, opened 2022-01-15
[@mantine/core] Transferlist: Apply filter on moveAll

MiguelKorn:
As the title says. With these changes, the filter will be applied on the items when moveAll is triggered.  Also added a prop to disable this functionality and move all items even if you have a filter.


rtivital:
Thanks for a PR, I do not think that I'll have enough time to maintain changes to TransferList component, as I prefer to focus on other components.

PR #695, opened 2022-01-18
[@mantine/core] TransferList: Fix radius

jerebtw:


jerebtw:
For master branch https://github.com/mantinedev/mantine/pull/695

PR #707, opened 2022-01-19
[@mantine/hooks] use-scroll-lock: Fix not working on iOS

mikirejf:
The PR includes:
- Fix for #300.
- Typing improvements.
- Documentation edits with the new type definition.
- Some cleanup.

Tested on:
- iPhone XS (14.3), 
- simulated iPhone 13 Pro (15.0), 
- simulated iPhone 8 (15.0), 
- simulated iPhone 8 (13.4), 
- desktop browsers.

The current solution leverages `usePreventScroll()` hook from `@react-aria/overlays`. The code is very well documented and it explains and handles all of the different intricacies of scroll prevention on iOS: https://github.com/adobe/react-spectrum/blob/main/packages/%40react-aria/overlays/src/usePreventScroll.ts

If this solution isn't feasible, we can still go with the `position: fixed;` solution as mentioned in #300.

### DIFF
--- a/docs/src/docs/hooks/use-scroll-lock.mdx
+++ b/docs/src/docs/hooks/use-scroll-lock.mdx
@@ -28,8 +28,12 @@ However if you use elements with fixed position you will need to add right paddi
 ## Definition
 
 ```tsx
-function useScrollLock<T extends HTMLElement = HTMLDivElement>(
+interface useScrollLockOptions {
+  disableBodyPadding: boolean;
+}
+
+function useScrollLock(
   lock?: boolean,
-  options?: { disableBodyPadding?: boolean }
+  options?: useScrollLockOptions
 ): readonly [boolean, Dispatch<SetStateAction<boolean>>];
 ```
--- a/src/mantine-hooks/src/use-scroll-lock/use-prevent-scroll.ts
+++ b/src/mantine-hooks/src/use-scroll-lock/use-prevent-scroll.ts
@@ -1,0 +1,303 @@
+/* eslint-disable consistent-return */
+/* eslint-disable no-param-reassign */
+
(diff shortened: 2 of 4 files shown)


LucasLemanowicz:
Nice work fixing this!

Just a note that just the `@react-aria/overlays` dependency is 2.5x larger in size than the entire `@mantine/hooks` library, and doesn't seem very tree shakeable (scroll down to Export Analysis): https://bundlephobia.com/package/@react-aria/overlays@3.7.3

mikirejf:
Ok, this is an issue then. I'll check what causes it to not tree-shake properly, otherwise, I can extract the relevant parts of the code and give credits/link to the original.

mikirejf:
@LucasLemanowicz 
I dug through the issues on the bundlephobia's repo and found out that the tool doesn't report the correct three-shaked size for certain packages (https://github.com/pastelsky/bundlephobia/issues/364). I also bundled `import { usePreventScroll } from '@react-aria/overlays';` in a sample app using webpack and it tree-shaked correctly. 

However, when I was inspecting the final bundle, I noticed that there were some i18n jsons included that weren't removed from the bundle, contributing to the 50% of the final size. Going the extra mile to inspect the bundle and "manually" remove the jsons (if it's even possible) would be too much to ask for the end users of mantine. In light of that, I'm going to extract a custom version of `usePreventScroll()`. That will also remove the `useLayoutEffect` "hacks" and make the code more streamlined.

LucasLemanowicz:
My bad, thanks for pointing that out about Bundlephobia.

mikirejf:
I made all the changes I intended so the PR is ready to be re-reviewed. I also re-tested the solution.

The main changes in this iteration:
- Extracted the relevant code from `@react-aria/overlays`. My main question here is if I conformed to the license of the package? I gave credit and linked to the original file in a comment.
- Used `useIsomorphicEffect` instead of `useLayoutEffect`.

melloware (approved):

rtivital:
I think it's time to let it go. @mikirejf feel free to reopen if you decide to finish your PR

Issue #706, opened 2022-01-19
[@mantine/dates] DatePicker should allow selecting date when Enter is pressed


LucasLemanowicz:
### What package has an issue

@mantine/dates

### Describe the bug

When `allowFreeInput={true}`, the `DatePicker` should allow a user to select the selected date when they press the `Enter` key.

Repro: 
1. https://codesandbox.io/s/stoic-darwin-2ztdh?file=/src/App.tsx
2. Start typing in a date (ex: `Jan 19, 2022`)
3. Press `Enter`

Expect: Jan 19, 2022 to be selected and the Dropdown to close (same functionality as if you pressed Tab)
Actual: Does nothing

### In which browser did the problem occur

Chrome

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/stoic-darwin-2ztdh?file=/src/App.tsx

rtivital:
Yep, it seems reasonable, we should add this

PR #708, opened 2022-01-19
[@mantine/dates] DatePicker: Allow selecting dates with Enter key

LucasLemanowicz:
Fixes https://github.com/mantinedev/mantine/issues/706


rtivital:
Thanks! It will be out with next patch

Issue #698, opened 2022-01-18
TimeInput doesn't trigger onChange handler


jerguslejko:
### What package has an issue

@mantine/dates

### Describe the bug

When I do

```tsx
<TimeInput onChange={(value) => alert(value)} />
```

the `onChange` handler is never triggered. I'd expect the `onChange` handler to be called on `blur` and when pressing `return`

### In which browser did the problem occur

chrome

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/pensive-hugle-0mlk2?file=/src/App.tsx

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

rtivital:
Thanks for reporting, it will be fixed in next patch

PR #710, opened 2022-01-20
[@mantine/dates] TimeInput: Call onChange when input is changed

jerebtw:
Fixes https://github.com/mantinedev/mantine/issues/698


rtivital:
Thanks!

PR #711, opened 2022-01-20
[@mantine/core] TransferList: Fix radius

jerebtw:
Fixes https://github.com/mantinedev/mantine/issues/722

Before this PR:

image

After this PR:

image


rtivital:
Thanks!

Issue #714, opened 2022-01-21
[@mantine/dates] TimeInput with useForm: Maximum update depth exceeded


jerguslejko:
### What package has an issue

@mantine/dates

### Describe the bug

This issue appeared after https://github.com/mantinedev/mantine/issues/698 was fixed.

When I use `<TimeInput />` with `useForm`, after entering a digit into `<TimeInput />`, the code crashes with:

```
index.js:27 Warning: Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either doesn't have a dependency array, or one of the dependencies changes on every render.
    at eval (https://x53x9.csb.app/node_modules/@mantine/dates/esm/components/TimeInput/TimeInput.js:62:27)
    at MantineProvider (https://x53x9.csb.app/node_modules/@mantine/styles/esm/theme/MantineProvider.js:99:34)
    at App (https://x53x9.csb.app/src/App.tsx:22:35) 
    in @mantine/dates/TimeInput (created by App)
    in App
```

You can see this in action on the codesandbox link provided below.

### In which browser did the problem occur

chrome

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/vibrant-fire-x53x9?file=/src/App.tsx:490-495

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

rtivital:
Thanks for reporting, we'll try to resolve the issue in one of the next patches, meanwhile you can wrap your change handler with useCallback to avoid this issue – https://codesandbox.io/s/suspicious-tdd-k0x6q?file=/src/App.tsx

rtivital:
@jerebtw would you like to fix it?

PR #721, opened 2022-01-21
[@mantine/dates] TimeInput: Fix infinite loop of onChange calls

jerebtw:
Fixes https://github.com/mantinedev/mantine/issues/714


rtivital:
Thanks!

PR #607, opened 2022-01-01
[@mantine/dates] TimeInput: Add `format` and `clearable` prop

jerebtw:
Fix for https://github.com/mantinedev/mantine/issues/581 and https://github.com/mantinedev/mantine/issues/580

This also can add a new hook (if wanted) called `use-valid-state`. (It's similar to useState but accepts a `validation` function and returns the last valid value, a set value function and the current value)

PR #731, opened 2022-01-23
[@mantine/dates] TimeInput: Fixed non-qwerty keyboard support

jerebtw:
Fixes https://github.com/mantinedev/mantine/pull/607#issuecomment-1019527356

### DIFF
--- a/src/mantine-dates/src/components/TimeInputBase/AmPmInput/AmPmInput.tsx
+++ b/src/mantine-dates/src/components/TimeInputBase/AmPmInput/AmPmInput.tsx
@@ -39,11 +39,11 @@ export const AmPmInput = forwardRef<HTMLInputElement, AmPmSelectProps>(
         onChange(value === 'am' ? 'pm' : 'am', false);
       }
 
-      if (event.key === 'p') {
+      if (event.key === 'p' || event.nativeEvent.code === 'KeyP') {
         onChange('pm', true);
       }
 
-      if (event.key === 'a') {
+      if (event.key === 'a' || event.nativeEvent.code === 'KeyA') {
         onChange('am', true);
       }
     };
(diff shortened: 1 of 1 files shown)


src/mantine-dates/src/components/TimeInputBase/AmPmInput/AmPmInput.tsx
```diff
@@ -34,16 +34,16 @@ export const AmPmInput = forwardRef<HTMLInputElement, AmPmSelectProps>(
     };
 
     const handleKeyDown = (event: React.KeyboardEvent<HTMLInputElement>) => {
-      if (event.nativeEvent.code === 'ArrowUp' || event.nativeEvent.code === 'ArrowDown') {
+      if (event.key === 'ArrowUp' || event.key === 'ArrowDown') {
```
rtivital, line 37: I think we should leave `event.nativeEvent.code` for arrows since they are universal for all languages and code is more safe

jerebtw: Okay, I have undone the change

Successahead (approved):

src/mantine-dates/src/components/TimeInputBase/AmPmInput/AmPmInput.tsx
```diff
@@ -39,11 +39,11 @@ export const AmPmInput = forwardRef<HTMLInputElement, AmPmSelectProps>(
         onChange(value === 'am' ? 'pm' : 'am', false);
       }
 
-      if (event.nativeEvent.code === 'KeyP') {
+      if (event.key === 'p') {
```
rtivital, line 42: As I think about it more, it won't hurt if we use both `event.nativeEvent.code === 'KeyP'` and `event.key === 'p'`:
- some languages do not have Latin letters, e.g. in Russian letter `а` is not the same as English `a`, (`a` === `а` in other words)
- When this is the case, there will be a fallback for `KeyA`, for example, when user forgets to switch input 
- It should not hurt other languages

What do you think, maybe any objections?

jerebtw: Yes, I think that's a great idea.

And even though a different key might work for other keyboard layouts, that shouldn't be a problem.

rtivital:
Thanks!

PR #732, opened 2022-01-23
Proper hover background color for ActionIcon in Dark Theme

albertbronsky:
This fixes the issue of `ActionIcon` having black-on-black background color when using Dark Theme, which is bad for accessibility as users can't see the hover state well. 

This issue can be observed in [ActionIcon/Color docs](https://mantine.dev/core/action-icon/#color): switch to Dark Theme and try to hover any of the icons in top row. You'll see that all of them have same black background which is barely visible and none of them have parent's color tint.

Before:

After:

The new behavior makes it clearly visible to the user and respects the parent's color by having a subtle tint (in line with Mantine's Light Theme implementation).

### DIFF
--- a/src/mantine-core/src/components/ActionIcon/ActionIcon.styles.ts
+++ b/src/mantine-core/src/components/ActionIcon/ActionIcon.styles.ts
@@ -46,7 +46,7 @@ function getVariantStyles({ variant, theme, color }: GetVariantStyles) {
           ? {}
           : {
               backgroundColor:
-                theme.colorScheme === 'dark' ? theme.colors.dark[8] : theme.fn.themeColor(color, 0),
+                theme.colorScheme === 'dark' ? theme.fn.rgba(theme.fn.themeColor(color, 8), 0.15) : theme.fn.themeColor(color, 0),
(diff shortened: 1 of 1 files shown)


rtivital:
This change will impact the following components: Modal, Drawer, Alert, Popover, Dialog, MultiSelect, Notification, Select, Autocomplete, DatePicker, DateRangePicker, TimeInput, TimeRangeInput. I think that the change to hover background color would feel improper for at least some of those components, and it may result into unwanted breaking changes that are hard to detect. So, I would say that hover effect should not be changed.

albertbronsky:
> This change will impact the following components: Modal, Drawer, Alert, Popover, Dialog, MultiSelect, Notification, Select, Autocomplete, DatePicker, DateRangePicker, TimeInput, TimeRangeInput. I think that the change to hover background color would feel improper for at least some of those components, and it may result into unwanted breaking changes that are hard to detect. So, I would say that hover effect should not be changed.

Is it possible to make this change only for standalone ActionIcon?

rtivital:
No, that's not possible

PR #733, opened 2022-01-23
Add `disabled` prop to SegmentedControl

LucasLemanowicz:
`disabled={false}`:
Screen Shot 2022-01-23 at 3 17 44 PM

`disabled={true}`:
Screen Shot 2022-01-23 at 3 17 31 PM

### DIFF
--- a/src/mantine-core/src/components/SegmentedControl/SegmentedControl.styles.ts
+++ b/src/mantine-core/src/components/SegmentedControl/SegmentedControl.styles.ts
@@ -131,6 +131,12 @@ export default createStyles(
         },
       },
 
+      disabled: {
+        '&, &:hover': {
+          color: theme.colors.gray[5],
+        },
+      },
+
       active: {
         boxSizing: 'border-box',
         borderRadius: theme.fn.size({ size: radius, sizes: theme.radius }),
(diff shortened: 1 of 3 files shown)


src/mantine-core/src/components/SegmentedControl/SegmentedControl.styles.ts
```diff
@@ -131,6 +131,12 @@ export default createStyles(
         },
       },
 
+      disabled: {
+        '&, &:hover': {
+          color: theme.colors.gray[5],
```
LucasLemanowicz, line 136: @rtivital not sure if there's a more generic way here to support custom themes?

rtivital:
`disabled` class is missing in Styles API description

rtivital:
I think you've forgotten about dark color scheme, disabled component does not differ much and in some places has even more contrast than not disabled one

LucasLemanowicz:
ok, which one of these do you prefer?

Enabled SegmentControl in dark mode:
segment-dark

Disabled (Gray 5):
segment-dark-disabled-gray5

Disabled (Gray 6):
segment-dark-disabled-gray6

Disabled (Gray 7):
segment-dark-disabled-gray7

rtivital:
@LucasLemanowicz dark theme should not use gray colors, it can use only `theme.colors.dark`

rtivital:
I'll finish it myself, don't want to bother you too much, thanks for a PR!

PR #774, opened 2022-01-31
[@mantine/core] add formatter and parser to number input

mmahalwy:


jerebtw:
Would be good if you could add tests and a demo to the docs (something like this: https://mantine.dev/core/number-input/#decimal-separator)

src/mantine-core/src/components/NumberInput/NumberInput.tsx
```diff
@@ -88,6 +94,8 @@ export const NumberInput = forwardRef<HTMLInputElement, NumberInputProps>(
       styles,
       size,
       rightSection,
+      formatter,
```
rtivital, line 97: Instead of conditionally applying formatter, create a default one and make it as default prop, same for parser

rtivital (changes requested):
No need to add demos and docs, I'll add them myself, but please create a story that will showcase this feature

src/mantine-core/src/components/NumberInput/NumberInput.tsx
mmahalwy: @rtivital  that works for formatter, but what about parser? notice we `return` the parsed output: 
https://github.com/mantinedev/mantine/pull/774/files#diff-072fc6263e6e304bed6a5dbf188a6470c19309741aa4cc63462fa94ee01c84dcR147

mmahalwy:
cool, if yall are good with this implementation, ill add story and tests

src/mantine-core/src/components/NumberInput/NumberInput.tsx
rtivital: I do not see why this is an issue for having default parser, just restructure code to incorporate it

mmahalwy:
@rtivital ptal

rtivital:
Please remove console.log from the code and resolve conflicts

mmahalwy:
@rtivital strange, i cant figure out what the conflicts are and pull rebase doesnt have any :S

rtivital:
Please make sure that your branch is in sync with the main repo

mmahalwy:
going to close this PR and open a new one.

PR #775, opened 2022-01-31
[@mantine/dates] Improvements for time inputs

mattrunyon:
- 12-hour times are displayed from 01-12 instead of 00-11
- Fixed bug with advancing to next field when entering 00, 01, etc.
- Fixed missing onUpdate calls in certain scenarios
- Add format and clearable to TimeInput configurator
- Fix value prop not overriding TimeInput and TimeRangeInput internal state
- Change TimeRangeInput to use 2 TimeInput fields
- onChange can be called with null to indicate cleared state

### DIFF
--- a/src/mantine-dates/src/components/TimeInput/TimeInput.test.tsx
+++ b/src/mantine-dates/src/components/TimeInput/TimeInput.test.tsx
@@ -1,5 +1,6 @@
 import React from 'react';
 import { render, screen } from '@testing-library/react';
+import userEvent from '@testing-library/user-event';
 import {
   checkAccessibility,
   itSupportsSystemProps,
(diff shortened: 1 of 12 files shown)


rtivital:
@mattrunyon is this ready for review? Can you please resolve conflicts?

mattrunyon:
@rtivital I've fixed the conflicts (haven't pushed yet). There are some issues w/ the controlled story, but that is from before my changes

It seems the value prop isn't used after the initial render

rtivital:
Yep, that's a known bug

rtivital:
Thanks! It's not ideal for controlled input because all inputs get filled when the first character is typed. But I guess this is a limitation that is not possible to solve easily,

PR #778, opened 2022-01-31
Fix for overflow of Paper children elements

albertbronsky:
This fixes the issue of children elements (like the `Table` component with rows on hover in the attached video) overflowing the corner radius of`Paper` parent.

https://user-images.githubusercontent.com/44101337/151796010-c636cb93-5707-4476-aafb-0c8055b4ed48.mov


rtivital:
Paper is designed to have visible overflow, this changes mostly likely will impact the majority of applications using Mantine, including most of Mantine components that have dropdowns. If you need Paper with hidden overflow, use Card component instead.

PR #788, opened 2022-02-01
Fix #787: [@mantime/hooks] useLocalStorageValue allow any object

### LINKED ISSUE #787: use-local-storage-value should support JSON values
### What package has an issue

@mantine/hooks

### Describe the bug

Currently the `useLocalStorageValue` enforces that the object must extend `String`.

```typescript
useLocalStorageValue<T extends string>
```

This way you can't use complex objects or primitives to store data which is a common use case such as...

```typescript
const [valueBoolean, setBoolean] = useLocalStorageValue<boolean>({
    key: "my-boolean",
    defaultValue: true
  });
```

ALL

https://codesandbox.io/s/agitated-gauss-8gk5r?file=/src/App.tsx

Yes

Yes

### Possible fix

I am switching from my own personal local storage hook to Mantine Hooks and my old values I always store and retrieve as JSON makes this the most flexible it can be and allows users to convert directly from `useState`.

melloware:


rtivital:
This change will break applications that already use use-local-storage hook: `JSON.parse('some-string')` will throw an error. The cleanest way will be to create a new hook that will work with JSON data (`use-local-storage-json-value`).

melloware:
You are probably right.

PR #789, opened 2022-02-01
Fix #787: [@mantime/hooks] useLocalStorageJsonValue to allow any object

### LINKED ISSUE #787: use-local-storage-value should support JSON values

melloware:
Just starting the conversation here.  I basically copied the other one and trying to manipulate it to get it to work.

### DIFF
--- a/src/mantine-hooks/src/use-local-storage-json-value/use-local-storage-json-value.story.tsx
+++ b/src/mantine-hooks/src/use-local-storage-json-value/use-local-storage-json-value.story.tsx
@@ -1,0 +1,20 @@
+import React from 'react';
+import { Meta } from '@storybook/react';
+import { useLocalStorageJsonValue } from './use-local-storage-json-value';
+
+export default {
+  title: '@mantine/hooks/use-local-storage-json-value',
+} as Meta;
+
+export function Example() {
+  const [value, setValue] = useLocalStorageJsonValue<string>({
+    key: '@mantine/localStorage/val',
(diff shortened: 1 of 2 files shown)


melloware:
Definitely not a TypeScript expert so not sure what the error means or how to fix it :)

```
Error: src/mantine-hooks/src/use-local-storage-json-value/use-local-storage-json-value.ts(21,26): error TS2349: This expression is not callable.
  Not all constituents of type '((prevState: T) => T) | (T & Function)' are callable.
    Type 'T & Function' has no call signatures.
Error: Process completed with exit code 2.
```

melloware:
For now I removed that ability to call a function.

rtivital:
JSON serialization was added in 4.0.3 by default – https://mantine.dev/hooks/use-local-storage/#serializedeserialize-json

PR #796, opened 2022-02-02
[docs] Fix AuthenticationForm path in use-form documentation

adepreis:
Update a code example link after mantine/demos migration


rtivital:
Thanks!

PR #798, opened 2022-02-03
[@mantine/core] Tooltip: Add "openDelay" prop

### LINKED ISSUE #797: [Feature Request] Add "openDelay" prop to the Tooltip component
### Usage example, including component/hook API

It would be nice if the `Tooltip` component had an additional `openDelay` prop that functioned exactly like the existing `delay` prop, but for the opening. [Here is an example of it in Chakra](https://chakra-ui.com/docs/overlay/tooltip#more-examples).

### Possible implementation – describe how the feature can be implemented

We can add an `openDelay` prop to the `Tooltip` component and modify the `handleOpen` function to behave like the existing `handleClose` function.

### Do you want to contribute this feature and create a pull request

Yes

captchafree:
Closes #797

### DIFF
--- a/src/mantine-core/src/components/Tooltip/Tooltip.test.tsx
+++ b/src/mantine-core/src/components/Tooltip/Tooltip.test.tsx
@@ -80,6 +80,23 @@ describe('@mantine/core/Tooltip', () => {
     expect(screen.queryAllByText('test-tooltip')).toHaveLength(0);
   });
 
+  it('supports open delay', async () => {
+    jest.useFakeTimers();
+
+    await renderWithAct(
+      <Tooltip {...defaultProps} openDelay={500}>
+        <button type="button">test-target</button>
+      </Tooltip>
+    );
+    expect(screen.queryAllByText('test-tooltip')).toHaveLength(0);
+    await actAsync(() => userEvent.hover(screen.getByRole('button')));
+    expect(screen.queryAllByText('test-tooltip')).toHaveLength(0);
+    await actAsync(() => userEvent.unhover(screen.getByRole('button')));
+    expect(screen.queryAllByText('test-tooltip')).toHaveLength(0);
+    await actAsync(() => jest.runAllTimers());
+    expect(screen.queryAllByText('test-tooltip')).toHaveLength(0);
+  });
+
   it('supports close delay', async () => {
     jest.useFakeTimers();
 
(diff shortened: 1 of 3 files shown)


src/mantine-core/src/components/Tooltip/Tooltip.test.tsx
```diff
@@ -80,11 +80,28 @@ describe('@mantine/core/Tooltip', () => {
...
+    );
+    expect(screen.queryAllByText('test-tooltip')).toHaveLength(0);
+    await actAsync(() => userEvent.hover(screen.getByRole('button')));
+    expect(screen.queryAllByText('test-tooltip')).toHaveLength(0);
+    await actAsync(() => userEvent.unhover(screen.getByRole('button')));
+    expect(screen.queryAllByText('test-tooltip')).toHaveLength(0);
+    await actAsync(() => jest.runAllTimers());
+    expect(screen.queryAllByText('test-tooltip')).toHaveLength(0);
+  });
+
```
rtivital, line 99: This test does not check that tooltip was rendered with delay, instead it checks that it was never there, it's meaningless

src/mantine-core/src/components/Tooltip/Tooltip.tsx
```diff
@@ -99,21 +103,30 @@ export const Tooltip = forwardRef<HTMLDivElement, TooltipProps>(
     ref
   ) => {
     const { classes, cx } = useStyles({ color, radius }, { classNames, styles, name: 'Tooltip' });
-    const timeoutRef = useRef<number>();
+    const [openTimeoutRef, closeTimeoutRef] = [useRef<number>(), useRef<number>()];
```
rtivital, line 106: Needless destructuring, please create two separate variables for two refs:
```tsx
const openTimeoutRef = useRef<number>();
const closeTimeoutRef = useRef<number>();
```

src/mantine-core/src/components/Tooltip/Tooltip.tsx
```diff
@@ -99,21 +103,30 @@ export const Tooltip = forwardRef<HTMLDivElement, TooltipProps>(
...
     const [_opened, setOpened] = useState(false);
     const visible = (typeof opened === 'boolean' ? opened : _opened) && !disabled;
     const [referenceElement, setReferenceElement] = useState(null);
 
     const handleOpen = () => {
-      window.clearTimeout(timeoutRef.current);
-      setOpened(true);
+      window.clearTimeout(closeTimeoutRef.current);
```
rtivital, line 112: Both timeout should also be cleared when component unmounts.

rtivital (changes requested):

rtivital:
Works great, thanks!

PR #826, opened 2022-02-11
[@mantine/core] Added inputMode for NumberInput

dmitriifrlv:
These changes add `inputMode` to [NumberInput](https://mantine.dev/core/number-input/) component. 

If `step` and `precision` properties are set, `inputMode` is set to 'decimal' and on mobile devices, a keyboard looks like:
IMG_7429

Without properties  `step` and `precision` `inputMode` is set to 'decimal' and keyboard looks like:

IMG_7430


rtivital:
Thanks!

PR #845, opened 2022-02-17
Allow override of position property in Portal

knpwrs:
I need a `Portal` with a `fixed` position, it would be nice to be able to accomplish this directly!


src/mantine-core/src/components/Portal/Portal.tsx
```diff
@@ -10,14 +11,23 @@ export interface PortalProps {
   /** Root element z-index property */
   zIndex?: number;
 
+  /** Root element position property */
+  position?: Property.Position;
```
rtivital, line 15: Please change type to string instead

knpwrs: All set.

rtivital:
Thanks!

PR #849, opened 2022-02-17
[@mantine/styles] MantineProvider: fix incorrect fontSize naming for body font size theme inheritance

DrummerHead:
At https://github.com/mantinedev/mantine/blob/master/src/mantine-styles/src/theme/MantineProvider.tsx#L62

`fontSizes` is transformed into `font-sizes` which is not the right CSS property name. 

With this change it is now `fontSize` which will transform into `font-size`

Cheers!


rtivital:
Thanks!

PR #853, opened 2022-02-18
[@mantine/hooks] use-interval: Add unit tests

createdbymahmood:
Hi, This PR adds some unit tests on [use-interval](https://github.com/mantinedev/mantine/blob/master/src/mantine-hooks/src/use-interval/use-interval.ts) hook.


rtivital:
Thanks!

Issue #855, opened 2022-02-18
The Notification docs page has a typo


Theboiboi8:
### Link to the page where something nasty is located

https://mantine.dev/others/notifications/#notification-props

### Exact quote of what is wrong

in the `id` section of the `Notification Props` page there is a typo in the word `and`. In the page it is spelled as `adn` and the word `by` slightly later in the sentence is spelled as `be`:
<img width="533" alt="image" src="https://user-images.githubusercontent.com/93494094/154719979-508889cb-ccd4-4297-bf1c-1e4684540396.png">

### Are you willing to create pull request with the fix?

No

rtivital:
Thanks for reporting!

rtivital:
Fixed in #856, will be out with 4.0

PR #861, opened 2022-02-19
docs: added `pnpm` install guide and fix typo

dammy001:


dammy001:
closes #855

rtivital:
Thanks for a PR, but we do not need to display every existing package manager, yarn and npm is enough.

dammy001:
@rtivital alright

dammy001:
@rtivital i get your point. I can see that if we need to add pnpm, it will be added to a whole lot of pages.

PR #865, opened 2022-02-19
[@mantine/core] Avatar: fix src typings

### LINKED ISSUE #864: Avatar typings don't accept src={null}
### What package has an issue

@mantine/core

### Describe the bug

The [example](https://mantine.dev/core/avatar/) shows a component `<Avatar src={null}>...</Avatar>`. However, the TypeScript typings only allow `string | undefined`, not `null`.

N/A

_No response_

Yes

Yes

### Possible fix

Change https://github.com/mantinedev/mantine/blob/4339b8ab07207bb6c04062f49153d7b631a9044f/src/mantine-core/src/components/Avatar/Avatar.tsx#L18 to `alt?: string | null`;

canac:
Fixes #864


rtivital:
Thanks!

PR #868, opened 2022-02-19
[@mantine/dates] DateRangePicker: Fix locale

finico:
There is no related issue (at least I can't find it)
I found this defect during development


rtivital:
Thanks!

PR #873, opened 2022-02-20
[@mantine/core] Carousel: init Component

achmurali:
This PR adds Carousel component.

@rtivital , I kind of need your help with setting the aria labels and roles and documentation.

PS - added a story just for testing, shall remove it after final review

### DIFF
--- a/src/mantine-core/src/components/Carousel/Carousel.story.tsx
+++ b/src/mantine-core/src/components/Carousel/Carousel.story.tsx
@@ -1,0 +1,36 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { Carousel } from './Carousel';
+//import { Image } from '../Image';
+
+storiesOf('@mantine/core/Carousel/stories', module).add('Test Story', () => (
+  <Carousel
+    style={{
+      maxWidth: 400,
+    }}
+    height={200}
+  >
+    <Carousel.Item>
+      <div
+        style={{
+          width: 400,
+          height: 200,
+          backgroundColor: '#FFF661',
+          textAlign: 'center',
+          verticalAlign: 'center',
+        }}
+      />
+    </Carousel.Item>
+    <Carousel.Item>
+      <div
+        style={{
+          width: 400,
+          height: 200,
+          backgroundColor: '#FF9FBB',
+          textAlign: 'center',
+          verticalAlign: 'center',
+        }}
+      />
+    </Carousel.Item>
+  </Carousel>
+));
--- a/src/mantine-core/src/components/Carousel/Carousel.tsx
+++ b/src/mantine-core/src/components/Carousel/Carousel.tsx
@@ -1,0 +1,203 @@
+import {
+  MantineNumberSize,
+  PolymorphicComponentProps,
+  PolymorphicRef,
+  useMantineDefaultProps,
+  ClassNames,
+} from '@mantine/styles';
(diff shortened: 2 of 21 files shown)


src/mantine-core/src/components/Carousel/Carousel.story.tsx
```diff
@@ -0,0 +1,36 @@
+import React from 'react';
+import { storiesOf } from '@storybook/react';
+import { Carousel } from './Carousel';
+//import { Image } from '../Image';
+
+storiesOf('@mantine/core/Carousel/stories', module).add('Test Story', () => (
```
rtivital, line 6: Do not delete the stories, they are a great source for testing

src/mantine-core/src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+  /** Change navigation indicator component */
+  indicatorButtonComponent?: React.FC<CarouselControlProps>;
+}
+
+export type CarouselProps<C> = C extends React.ElementType
+  ? PolymorphicComponentProps<C, _CarouselProps>
+  : never;
+
+type CarouselCompoent = (<C = 'div'>(props: CarouselProps<C>) => React.ReactElement) & {
```
rtivital, line 64: I think there is no need for Carousel to be a polymorphic component. It's fine to render just a div as a root element.

src/mantine-core/src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,203 @@
...
+      classNames,
+      withControls,
+      styles,
+      ...others
+    } = useMantineDefaultProps('Carousel', defaultProps, props);
+    const { classes, cx } = useStyles(
+      { scrollOverflow, axis, height, padding, draggable },
+      { classNames, styles, name: 'Carousel' }
+    );
+    const _children = Children.toArray(children);
```
rtivital, line 114: No need to convert children to array, filterChildrenByType will work with raw children

src/mantine-core/src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,203 @@
...
+
+    useEffect(() => {
+      if (!embla) return;
+      onSelect();
+      setScrollSnaps(embla.scrollSnapList());
+      embla.on('select', onSelect);
+    }, [embla, setScrollSnaps, onSelect]);
+
+    return (
+      <Paper
```
rtivital, line 155: Do not wrap Carousel in Paper, use a plan Box component instead.

src/mantine-core/src/components/Carousel/Carousel.tsx
```diff
@@ -0,0 +1,203 @@
...
+        className={cx(classes.root, className)}
+        radius={radius}
+        padding={padding}
+        ref={ref}
+        component={component as any}
+        {...others}
+      >
+        <div ref={emblaRef} className={classes.viewport}>
+          <Box<any> component="div" className={classes.container}>
```
rtivital, line 164: No need to set box type and component, should be just `<Box className={classes.container}>`, or maybe replace it with a div, since no Box props are being used.

rtivital (commented):
Thanks for a PR, great job! I've left some initial review comments. Also since Carousel depends on an external package we cannot make it a part of `@mantine/core` library, it should be in a separate package called `@mantine/carousel`

PR #875, opened 2022-02-20
Update use-debounced-value.mdx

leochoo:
`const [debounced, cancel] = useDebouncedValue(value, 1000);`
Corrected the description of this line of "minute" to "second"  (1000ms = 1s)


rtivital:
Thanks!

PR #888, opened 2022-02-22
Update AppShell.mdx

na0495:
I was reading the documentation and I notice that Header component is missing.


na0495:
When will the version 4.0 be available @rtivital ?

Issue #508, opened 2021-12-12
[Feature Request] Radius on TransferList props


jerebtw:
Same as https://github.com/mantinedev/mantine/issues/491

Ultra-Instinct-05:
@rtivital I believe I can work on this. I assume by `radius` prop, this would change the radius of the `TransferListBody` ?

rtivital:
Yes, please make sure that items are not overflowing during hover

Ultra-Instinct-05:
So I was able to make the `radius` prop work a bit, but for the `ActionIcon`, I am a bit lost on how to proceed for that. Both the right & left transfer component make use of a single `ActionIcon`. The `ActionIcon` takes a `radius` prop but how do I split that out into a `borderTopLeftRadius` & a `borderTopRightRadius` ?

Was able to make it for the search because the right & left make use of different `RenderList` so I was able to split the `radius` into `leftRadius` & `rightRadius` to achieve it.

![image](https://user-images.githubusercontent.com/34304644/145797493-507ea974-ab9a-439e-b206-52ac7dd5f743.png)

rtivital:
You can use styles to Add specific border radius to ActionIcon

Ultra-Instinct-05:
Yes, but then the `ActionIcon` would need to take 2 kinds of radii props, one for 3 corners and one to respect the `radius` of `TransferList`. Is that fine (at least, that's the only solution I can think of) ? Or maybe there is an "elegant" solution I am missing.

rtivital:
I've decided that radius on TransferList can be handled better with Styles API, and we should not add radius prop for that

PR #892, opened 2022-02-23
[@mantine/core] TransferList: Add radius prop

jerebtw:
Fixes https://github.com/mantinedev/mantine/issues/508


rtivital:
Thanks!

PR #901, opened 2022-02-25
[docs] Fixed typo on Notifications document

yethranayeh:
from: ... update adn remove notification, be default id generates randomly
to : ... update and remove notification, by default id is randomly generated


rtivital:
Thanks!

PR #902, opened 2022-02-26
[docs]  Missing grow prop in example

IAmLuisJ:
Fix: Typo in Navigation.Section example, missing the grow prop that is mentioned in the comment. If I should include all the generated docs in the PR please let me know, thank you!


rtivital:
Thanks!

PR #903, opened 2022-02-26
[docs] Scroll active sidebar item into view on navigate

mattrunyon:
This was already partially implemented, but it only worked on initial page loads.

Now on any navigation the active item in the navbar will scroll into view, but only if needed. This change also expands the navbar section if needed so the item can be visible (e.g. Collapse the hooks section of navbar then navigate to a hooks doc page via search)

Tested w/ clicking on pages, navigating via search, and navigating via links on page (e.g. "Extend Theme" page link to "TypographyStylesProvider")


rtivital:
Thanks!

PR #913, opened 2022-03-01
[@mantine/hooks] - Renamed HokeyItem type to HotkeyItem | Updated docs to match type definition

iDavidB:
This may be a breaking change for people who import the old `HokeyItem` type, but it should be a manageable change.


rtivital:
I think that already may be fixed in dev branch

PR #924, opened 2022-03-03
[@mantine/core] clearer typescript error for polymorphic components

jvdsande:
As discussed on Discord, this PR changes the following structure:

```ts
C extends React.ElementType ? PolymorphicComponentProps<C, SomeProps>  : never
```

to

```ts
C extends React.ElementType ? PolymorphicComponentProps<C, SomeProps>  : (SomeProps & { component: React.ElementType })
```

This allows for clearer error messages when providing a wrong value as `component`: instead of having all props failing, now only the `component` prop fails.

Furthermore it moves the conditional typing logic inside `PolymorphicComponentProps` directly to avoid duplicating it everywhere.

Note: one use case was not discussed on Discord, it's using the "generic" structure:

```ts
<Avatar<wrong> size="sm" />
```

Then the error is a bit less straightfoward, because it starts by saying that `component` prop is missing while required, then when we change to:

```ts
<Avatar<wrong> size="sm" component="wrong" />
```

The error moves to "component" and says "Type '"wrong"' is not assignable to type 'ElementType<any>'"

I'm not quite sure why the docs tells users to use this structure though since simply providing the `component` props makes TS infer correctly the other props too. Using the generic component instance structure is redundant in this situation.

### DIFF
--- a/src/mantine-core/src/components/ActionIcon/ActionIcon.tsx
+++ b/src/mantine-core/src/components/ActionIcon/ActionIcon.tsx
@@ -38,9 +38,7 @@ interface _ActionIconProps extends DefaultProps<ActionIconStylesNames> {
   loading?: boolean;
 }
 
-export type ActionIconProps<C> = C extends React.ElementType
-  ? PolymorphicComponentProps<C, _ActionIconProps>
-  : never;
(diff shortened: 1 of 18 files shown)


rtivital:
Thanks for PR!

Issue #925, opened 2022-03-03
Nested Accordions icons do not rotate


Chandher05:
### What package has an issue

@mantine/core

### Describe the bug

When Accordions are nested, the icons are not behaving the right way. 
The nested Accordions also have the rotated Icon, when the base Accordion is open. 

### In which browser did the problem occur

Brave (Chromium based), Firefox

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/brave-montalcini-u0glf6

### Do you know how to fix the issue

No

PR #926, opened 2022-03-03
[@mantine/core]: Accordion: Fix rotation behavior of nested accordion icons

zhengissov:
Fixes rotation behavior of nested accordion icons
Resolves https://github.com/mantinedev/mantine/issues/925


rtivital:
Thanks!

PR #933, opened 2022-03-04
[@mantine/dates] Improve the UX of the TimeInput component

captchafree:
This PR aims to improve the UX of the TimeInput component by advancing the user's focus to the next input field when we can infer their intent. If we can't deduce anything useful, then it falls back to the existing logic. The additional logic is only used when entering times that require a leading zero (e.g. `1:45`).

Here are a few examples demonstrating the difference:

### Example 1

**TimeInput props**: 12 hour time format without seconds
**Desired time (What the user is trying to enter)**: `1:30`
**User keystrokes (What the user types in the input)**: `1`, `3`, `0`

**Result (with current implementation)**: `12:00` (incorrect)
**Result (with this PR)**: `1:30` (correct)

### Example 2

**TimeInput props**: 24 hour time format without seconds
**Desired time (What the user is trying to enter)**: `2:45`
**User keystrokes (What the user types in the input)**: `2`, `4`, `5`

**Result (with current implementation)**: `23:05` (incorrect)
**Result (with this PR)**: `2:45` (correct)

### Example 3

**TimeInput props**: 12 hour time format without seconds
**Desired time (What the user is trying to enter)**: `12:30`
**User keystrokes (What the user types in the input)**: `1`, `6` (incorrect keystroke), `3`, `0`

**Result (with current implementation)**: `12:30` (correct)
**Result (with this PR)**: `12:30` (correct)

### DIFF
--- a/src/mantine-dates/src/components/TimeInput/TimeInput.tsx
+++ b/src/mantine-dates/src/components/TimeInput/TimeInput.tsx
@@ -176,6 +176,7 @@ export const TimeInput = forwardRef<HTMLInputElement, TimeInputProps>(
       min: format === '12' ? 1 : 0,
       max: format === '12' ? 12 : 23,
       nextRef: minutesRef,
+      nextMax: 59,
     });
 
     const handleMinutesChange = createTimeHandler({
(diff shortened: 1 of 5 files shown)


rtivital:
Thanks!

PR #934, opened 2022-03-04
[@mantine/tests] conformance test cases type safety

createdbymahmood:
Hi,
Recently I noticed that conformance test cases props aren't safe that much, so I decided to make this PR to add this option to those.

```tsx
itSupportsSx(NotificationContainer, 
  { /* all NotificationContainer component props will be suggested  */ }
);
```


rtivital:
Thanks!

PR #937, opened 2022-03-05
[docs] Fix typo in useMemo

elmaxe:
The code example is using `{}` instead of `[]` as a dependency list to `useMemo`


rtivital:
Thanks!

PR #948, opened 2022-03-06
[docs] Fixed typo in Drawer.mdx

binoy638:
Fixed a type in Drawer docs which was causing the drawer to cover the whole screen.


rtivital:
Docs are correct, this example illustrates that although 500vh value is provided Drawer will not be larger than 100% of the screen

PR #962, opened 2022-03-09
added style for span

chenchunaidu:


rtivital:
Thanks for a PR, this is not a good solution for fixing this issue.

PR #983, opened 2022-03-11
return new value from onCreate

### LINKED ISSUE #980: Adding new item on Select doesn't work when used object as data
### What package has an issue

@mantine/core

### Describe the bug

When we make `Select` creatable and pass data as objects in the expected schema and when we add a new item and click outside Select becomes empty but if we select the same item again then it can be seen selected. If we test this with only strings data then there is no problem. 

Google Chrome

https://codesandbox.io/s/hardcore-vaughan-g9d580?file=/src/App.js

No

Yes

### Possible fix

_No response_

DillonStreator:
Fixes #980 

This is a proposal to allow the onCreate prop of the Select component to respect a return value.
Currently, the default behavior of onCreate is assuming that the created value is the same as what was inputted.

The use case is using onCreate to specify a custom value that may or may not be derived from the incoming value from input.

### DIFF
--- a/src/mantine-core/src/components/Select/Select.story.tsx
+++ b/src/mantine-core/src/components/Select/Select.story.tsx
@@ -119,7 +119,53 @@ function DynamicLabels(props: Partial<SelectProps>) {
   );
 }
 
+function CustomValueOnCreate() {
+  const [objectOptions, setObjectOptions] = React.useState([
+    {
+      label: 'AA',
+      value: 'aa',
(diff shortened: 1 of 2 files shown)


DillonStreator:
@rtivital can you please provide some context into why this is being close?

rtivital:
Because you've abandoned your draft PR for about a month with broken tests. If you want to continue work on it, feel free to reopen.

PR #984, opened 2022-03-11
[docs] Fix navbar scroll on height change

mattrunyon:
Fixes the issue where the navbar would scroll the navbar item into view if the window was resized


rtivital:
Thanks!

PR #998, opened 2022-03-12
[@mantine/styles] - Fix primaryColor not being a keyof MantineThemeColors

iDavidB:
This improves the type for strict mode users so that they no longer get a "possibly undefined" error when using it.


rtivital:
Will this work with custom colors definition?

iDavidB:
> Will this work with custom colors definition?

Yes it should work fine

rtivital:
Great, thanks!

PR #1002, opened 2022-03-13
Allow 3 character hex strings for mantine colors

SmitharyZach:
Issue
Using a valid shorthand hex code value for a shade in mantine theme was not being converted to the correct color. For example, #ccc was returning black and not the correct color

Solution
- Update the isHexColor function to accept shorthand hex codes
- if hex is short hand, convert to the long version to be correctly converted to rgba


SmitharyZach:
@rtivital this is ready for review if you have any time to look over it

rtivital:
Thanks!

Issue #1020, opened 2022-03-17
[Feature Request] Return type for usePagination


jimmyruann:
### Usage example, including component/hook API

Adding a return type for `usePagination` for reusability.

```
const BookingPage = () => {
    const pagination = usePagination({ total: 10, initialPage: 1 });
    return (
        <>
            <ControlPannel pagination={pagination} />
            // ...
        </>
    );
};
const ControlPannel = ({ pagination }: { pagination: UsePagination }) => {
    return (
        <>
            <Button onClick={pagination.previous()}>Previous</Button>
            // ...
            <Button onClick={pagination.next()}>Next</Button>
        </>
    );
};
```

### Possible implementation – describe how the feature can be implemented

Add return type `UsePagination` to use-pagination hook

### Do you want to contribute this feature and create a pull request

Yes

PR #1021, opened 2022-03-17
[@mantine/hooks] add reusable return type for hooks

jimmyruann:
#1020


rtivital:
Types are not exported from index.ts, and thus they will not be accessible from `@mantine/hooks` package

jimmyruann:
Thank you for that. Will fix that up.

rtivital:
I do not think that most of these hooks require types export, maybe some state hooks can benefit from that, but adding these types to all of the hooks will hurt readability and make maintenance more complex.

jimmyruann:
Thank you for the feedback @rtivital . I will reassess the each hooks and create a new PR.

PR #1029, opened 2022-03-17
[@mantine/core] MultiSelect: fix maxSelectedValues for controlled input

### LINKED ISSUE #1028: MultiSelect doesn't allow to select more items (when having maxSelectedValues)
### What package has an issue

@mantine/core

### Describe the bug

You can select values from the dropdown when having a `MultiSelect` component with the `maxSelectedValues` prop. When reached the limit, a React ref (`valuesOverflow`) is set to true, and the dropdown does not appear anymore.

The problem is when you update any of the following props: `value` or `maxSelectedValues`, the React ref is not recalculated, and you can not keep selecting items (if the `value` updated to a lower length or the `maxSelectedValues` is increased).

Brave, Chrome

https://codesandbox.io/s/unruffled-orla-mtl75y?file=/src/App.js

Yes

Yes

### Possible fix

Update the React ref valuesOverflow when the value or the `maxSelectedValues` props are updated.

achmurali:
Fixes #1028


rtivital:
@achmurali please use master branch as target

rtivital:
Thanks!

PR #1035, opened 2022-03-18
[docs] Add pnpm in guide to install dependencies

morizkay:


rtivital:
No, thanks, we do not need that

PR #1038, opened 2022-03-18
[@mantine/hooks] fix falsy check for useLocalStorage reset to default value

### LINKED ISSUE #1037: useLocalStorage `defaultValue: true` doesn't work but it does with `false`.
### What package has an issue

@mantine/hooks

## The bug
`useLocalStorage` doesn't work if the default value is `true`. However, it does work if the default value is `true`. 

@jvdsande  said in discord, "Turns out as of now, `useLocalStorage` will not work correctly for storing boolean values, because if the value is ever falsy, it will revert back to the default value`

If this is an intended feature, it would be useful to put this in the documentation for other people who are desire to do the same thing as me.

## The goal / background
I was trying to make a switch for enabling calculations in my program with `useLocalStorage` so the boolean could persist if the user refreshed the page. I didn't want to use a string because it was an enable/disable setting.

Chrome

https://codesandbox.io/s/uselocalstorage-quirk-nnc1zs?file=/src/App.tsx

_No response_

_No response_

### Possible fix

In discord, @jvdsande said, "Depending on how @rtivital sees this (bug or 'feature'), it can be fixed by changing the falsiness check by a strict undefined check on the value."

jvdsande:
Fixes #1037

### DIFF
--- a/src/mantine-hooks/src/use-local-storage/use-local-storage.story.tsx
+++ b/src/mantine-hooks/src/use-local-storage/use-local-storage.story.tsx
@@ -33,3 +33,22 @@ export function SerializeJson() {
     </div>
   );
 }
+
+export function SerializeBoolean() {
+  const [value, setValue] = useLocalStorage<boolean>({
+    key: '@mantine/localStorage/val',
+    defaultValue: true,
+  });
+
+  return (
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #1067, opened 2022-03-20
[@mantine/hooks] use-list-state: fix index undefined issue

### LINKED ISSUE #1018: index parameter in applyWhere helper in use-list-state hook
### What package has an issue

@mantine/hooks

### Describe the bug

In [use-list-state](https://github.com/mantinedev/mantine/blob/01b6be15a7b97f788b5298772166aa0a09dc4643/src/mantine-hooks/src/use-list-state/use-list-state.ts#L80), in `applyWhere` helper:
```ts
const applyWhere = (
    condition: (item: T, index?: number) => boolean,
    fn: (item: T, index?: number) => T
  ) => setState((current) => current.map((item) => (condition(item) ? fn(item) : item)));
```

`index` is always `undefined` in `condition` and `fn` functions, could it be retrieved from the `map` and passed to both of them? (see possible fix below)

Moreover, `index` should not be optional I think in the typing (same thing for the `apply` helper).

All

_No response_

Yes

No

### Possible fix

```ts
const applyWhere = (
    condition: (item: T, index: number) => boolean,
    fn: (item: T, index: number) => T
  ) => setState((current) => current.map((item, index) => (condition(item, index) ? fn(item, index) : item)));
```

achmurali:
Fixes #1018 .

### DIFF
--- a/src/mantine-hooks/src/use-list-state/use-list-state.ts
+++ b/src/mantine-hooks/src/use-list-state/use-list-state.ts
@@ -77,7 +77,10 @@ export function useListState<T>(initialValue: T[] = []): UseListState<T> {
   const applyWhere = (
     condition: (item: T, index?: number) => boolean,
     fn: (item: T, index?: number) => T
-  ) => setState((current) => current.map((item) => (condition(item) ? fn(item) : item)));
+  ) =>
+    setState((current) =>
+      current.map((item, index) => (condition(item, index) ? fn(item, index) : item))
+    );
 
   return [
     state,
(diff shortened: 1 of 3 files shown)


src/mantine-hooks/src/use-list-state/use-list-state.ts
```diff
@@ -77,7 +77,10 @@ export function useListState<T>(initialValue: T[] = []): UseListState<T> {
   const applyWhere = (
     condition: (item: T, index?: number) => boolean,
```
rtivital, line 78: Here the index should not be optional, the type should be:
`condition: (item: T, index: number) => boolean,`

Same with type definition on line 12

rtivital (changes requested):

rtivital:
Thanks!

PR #1081, opened 2022-03-23
[@mantine/demos] Badge: Set initial and default value as light

ezgitetik:
Default value of "[@mantine/core] Badge" component is "light" as configured [here](https://github.com/mantinedev/mantine/blob/master/src/mantine-core/src/components/Badge/Badge.tsx#L55).

But, at the "Badge" configurator docs, initial value and default value are set as "filled". This PR includes fix of this issue.


rtivital:
Thanks!

PR #1088, opened 2022-03-24
Add documentation about dayjs plugins

DanielGibbsNZ:
Not being familiar with `dayjs` I spent the better part of an hour trying to figure out why the date parsing wasn't working when using `allowFreeInput` on a `DatePicker`.

I've updated the docs to include the relevant details to get this to work.

### DIFF
--- a/docs/src/docs/dates/date-picker.mdx
+++ b/docs/src/docs/dates/date-picker.mdx
@@ -107,6 +107,14 @@ function Demo() {
 `dateParser` function should always return Date object.
 If parsed date is invalid when input is blurred value will be restored to last valid value.
 
+Note that if you use the default dayjs parser, you will need to import and configure the
+`customParseFormat` dayjs plugin:
+
+```tsx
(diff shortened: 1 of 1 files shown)


docs/src/docs/dates/date-picker.mdx
```diff
@@ -107,6 +107,9 @@ function Demo() {
 `dateParser` function should always return Date object.
 If parsed date is invalid when input is blurred value will be restored to last valid value.
 
+Note that if you use the default dayjs parser, you will need to import and configure the
```
rtivital, line 110: Documentation should be included on DatePicker page, not index page, other components do not have dates parsing

DanielGibbsNZ: Ok, so I should remove the details about plugins entirely from the "Getting started" page and just put them inline here?

rtivital: yes

PR #1089, opened 2022-03-24
Pass Modal transitionDuration prop to GroupedTransition (fixes #1008)

### LINKED ISSUE #1008: Modal causes "Can't perform a React state update on an unmounted component" warnings when using Testing Library
### What package has an issue

@mantine/core

### Problem

When writing Testing Library tests for an app using the `Modal` component, I get `Can't perform a React state update on an unmounted component` warnings when I have multiple tests in the file.

This can be demonstrated with the following basic app:

**App.tsx**

```
import React from 'react';
import { Button, Modal, Title } from '@mantine/core';

  const [opened, setOpened] = React.useState(false);

  const title = <Title order={1}>Modal title</Title>
  return (
    <div>
      <Button onClick={() => setOpened(true)}>Open modal</Button>
      <Modal transitionDuration={0} trapFocus={false} opened={opened} closeButtonLabel="Close" title={title} onClose={() => setOpened(false)}>Modal contents</Modal>
    </div>
  );
}
```

**App.test.tsx**

```
import React from 'react';
import { render, screen, waitFor, within } from '@testing-library/react';
import userEvent from '@testing-library/user-event';
import App from './App';

test('it opens a modal when the button is clicked', () => {
  render(<App />);

  const button = screen.getByRole('button', { name: 'Open modal' });
  userEvent.click(button);

  const modal = screen.getByRole('dialog');
  const modalTitle = within(modal).getByRole('heading', { name: 'Modal title' });
  const modalContents = within(modal).getByText('Modal contents');
  expect(modal).toBeInTheDocument();
  expect(modalTitle).toBeInTheDocument();
  expect(modalContents).toBeInTheDocument();
});

test('it closes the modal when the close button is clicked', async () => {
  render(<App />);

  const button = screen.getByRole('button', { name: 'Open modal' });
  userEvent.click(button);

  const modal = screen.getByRole('dialog');
  const closeButton = within(modal).getByRole('button', { name: 'Close' });
  userEvent.click(closeButton);

  await waitFor(() => expect(screen.queryByRole('dialog')).not.toBeInTheDocument());
});
```

### Underlying cause

* `Modal` uses `GroupedTransition`
* `GroupedTransition` uses the `useTransition` hook
* The `useTransition` hook uses `setTimeout` **when the transition duration is not 0**

The problem here is that when the `transitionDuration` prop of a `Modal` is set to 0, that isn't passed on to the `GroupedTransition` (and therefore `useTransition`), causing a `setTimeout` to be used, even when there are no transitions.

JSDOM 16.6.0

_No response_

Yes

Yes

### Possible fix

My suggested fix is to pass the following props to `GroupedTransition` from inside `Modal`:

```
duration={transitionDuration}
exitDuration={transitionDuration}
```

This means that when setting the `transitionDuration` prop of `Modal` to `0` there will be no `setTimeout` used by the `useTransition` hook.

If there's a better solution for this, please let me know.

DanielGibbsNZ:
I've implemented my proposed solution to #1008. Please let me know if there's any feedback or better ways to solve this.

### DIFF
--- a/src/mantine-core/src/components/Modal/Modal.tsx
+++ b/src/mantine-core/src/components/Modal/Modal.tsx
@@ -179,6 +179,8 @@ export function Modal(props: ModalProps) {
         onExited={() => lockScroll(false)}
         onEntered={() => lockScroll(true)}
         mounted={opened}
+        duration={transitionDuration}
+        exitDuration={transitionDuration}
         transitions={{
           modal: { duration: transitionDuration, transition },
           overlay: {


rtivital:
Thanks!

PR #1101, opened 2022-03-26
[@mantine/core]: set Default Value 0

manansharma18:
Sets the default value to 0.


manansharma18:
Hey @rtivital , may be I misunderstood the issue. But this looks a correct fix to me to set the default value to 0.

rtivital:
No, it is not a valid fix. It will only work for the exact use case described in the issue with `0`, but would not resolve to problem with setting value if it is undefined.

manansharma18:
ok, can we virtually meet to understand the issue. I do not understand it fully.

PR #1117, opened 2022-03-28
Use useCallback for notification functions

knpwrs:
Currently it is not possible to show a notification from `useEffect` because the `showNotification` function is not referentially equal between renders which results in an infinite loop:

```ts
const { showNotification } = useNotifications();

useEffect(() => {
  showNotification(/* */); // Over and over and over and over again
}, [showNotification, notificationCondition]);
```

Removing `showNotification` triggers `eslint-plugin-react-hooks`. One possible workaround is this:

```ts
const notifications = useNotifications();
const showNotification = useCallback(notifications.showNotification, []);

useEffect(() => {
  showNotification(/* */); // All better :)
}, [showNotification, notificationCondition]);
```

That said, I think it would be nice to make the notification functions referentially equal to begin with.

### DIFF
--- a/src/mantine-notifications/src/NotificationsProvider/use-notifications-state/use-notifications-state.ts
+++ b/src/mantine-notifications/src/NotificationsProvider/use-notifications-state/use-notifications-state.ts
@@ -1,3 +1,4 @@
+import { useCallback } from 'react';
 import { useQueue, randomId } from '@mantine/hooks';
 import { NotificationProps } from '../../types';
 
@@ -7,7 +8,7 @@ export default function useNotificationsState({ limit }: { limit: number }) {
     limit,
   });
 
-  const showNotification = (notification: NotificationProps) => {
+  const showNotification = useCallback((notification: NotificationProps) => {
     const id = notification.id || randomId();
 
     update((notifications) => {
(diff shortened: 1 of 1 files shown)


src/mantine-notifications/src/NotificationsProvider/use-notifications-state/use-notifications-state.ts
```diff
@@ -7,7 +8,7 @@ export default function useNotificationsState({ limit }: { limit: number }) {
     limit,
   });
 
-  const showNotification = (notification: NotificationProps) => {
+  const showNotification = useCallback((notification: NotificationProps) => {
     const id = notification.id || randomId();
 
     update((notifications) => {
```
knpwrs, line 14: Why doesn't this produce a warning in ESLint? Does `eslint-plugin-react-hooks` know that `update` is okay here?

rtivital:
This will not be required. In 4.1 notifications will be migrated to event-based function which will be exported from package, hook will become obsolete.

rtivital:
You can find all information in next-minor branch. No this change will not be breaking.

knpwrs:
Great, I'll close this PR. Is there an ETA for 4.1?

PR #1119, opened 2022-03-28
[@mantine/hooks] use-hover: Add tests

mctessi:
Hi!
I created some simple tests to verify the use-hover functionality! 
Every feedback is deeply appreciated and there are other test files waiting to be commited! 
Happy coding


src/mantine-hooks/src/use-hover/use-hover.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+      {hovered ? 'true' : 'false'}
+    </div>
+  );
+};
+
+describe('@mantine/hooks/use-hover', () => {
+  render(<Target />);
+  const target = screen.getByTestId('target');
+  it('changes "hovered" on mouseenter/mouseleave correctly', () => {
+    expect(target).toHaveTextContent(/false/i);
```
rtivital, line 19: regex is not required, use strings instead.

src/mantine-hooks/src/use-hover/use-hover.test.tsx
```diff
@@ -0,0 +1,27 @@
...
+
+  return (
+    <div data-testid="target" ref={ref}>
+      {hovered ? 'true' : 'false'}
+    </div>
+  );
+};
+
+describe('@mantine/hooks/use-hover', () => {
+  render(<Target />);
```
rtivital, line 16: render inside `it` function call instead.

src/mantine-hooks/src/use-hover/use-hover.test.tsx
mctessi: Thanks for the feedback! I have updated my request...

src/mantine-hooks/src/use-hover/use-hover.test.tsx
mctessi: Thanks for the feedback! I have updated my request...

PR #1130, opened 2022-03-29
[@mantine/dates] Calendar: Added render day and weekday labels

trevlar:
Adds a `renderDay` prop to the Calendar which is passed to the Day component that allows a day to be rendered using a custom component.  
Added a weekday label to allow custom labels. Typical in calendars is the use of three-letter weekdays as the size of a calendar grows. Whereas the default in Mantine is two-letter weekdays which works well for a calendar date picker though not so much as an event calendar.

### DIFF
--- a/src/mantine-dates/src/components/Month/Day/Day.tsx
+++ b/src/mantine-dates/src/components/Month/Day/Day.tsx
@@ -25,7 +25,7 @@ export interface DayProps
   firstInMonth: boolean;
   focusable?: boolean;
   hideOutsideDates?: boolean;
-  calendarEvent?(date: Date): React.ReactNode | null;
+  renderDay?(date: Date): React.ReactNode;
 }
 
 export const Day = forwardRef<HTMLButtonElement, DayProps>(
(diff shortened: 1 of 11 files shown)


src/mantine-dates/src/components/Month/Day/Day.tsx
```diff
@@ -79,7 +81,7 @@ export const Day = forwardRef<HTMLButtonElement, DayProps>(
           className
         )}
       >
-        {value.getDate()}
+        {renderDay ? renderDay(value) : value.getDate()}
```
rtivital, line 84: `typeof renderDay === 'function' ? renderDay(value) : value.getDate()`

src/mantine-dates/src/components/Month/Month.tsx
```diff
@@ -98,6 +98,10 @@ export interface MonthProps
 
   /** Called when keydown event is registered on day */
   onDayKeyDown?(payload: DayKeydownPayload, event: React.KeyboardEvent<HTMLButtonElement>): void;
+
+  renderDay?(date: Date): React.ReactNode;
+
+  weekdayLabelFormat?: string;
```
rtivital, line 104: Missing props descriptions

rtivital (commented):
renderDay should be supported in Calendar, RangeCaledar, Month, DatePicker, DateRangePicker components. Currently it is available only in Caledar, RangeCalendar and Month.

rtivital:
Thanks!

PR #1149, opened 2022-04-01
[@mantine/core] PasswordInput: Make input respect theme default radius

iammathew:
Previously the PasswordInput did not respect the default radius set in the theme, due to overwriting it with its own default props.


rtivital:
Thanks!

PR #1156, opened 2022-04-01
[@mantine/prism] Fix prism number spacing

auronsan1st:
use min-width of max lines 65535 like VSCode.

user can easy override with styles. if they want smaller space caused only small lines code

### DIFF
--- a/src/mantine-prism/src/Prism.styles.ts
+++ b/src/mantine-prism/src/Prism.styles.ts
@@ -42,8 +42,9 @@ export default createStyles((theme, { colorScheme, native }: PrismStylesParams) 
   lineNumber: {
     color: colorScheme === 'dark' ? theme.colors.dark[4] : theme.colors.gray[4],
     textAlign: 'right',
-    paddingRight: theme.dir === 'ltr' ? theme.spacing.xl : undefined,
-    paddingLeft: theme.dir === 'rtl' ? theme.spacing.xl : undefined,
+    minWidth: 43,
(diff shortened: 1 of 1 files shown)


src/mantine-prism/src/Prism.styles.ts
```diff
@@ -42,8 +42,9 @@ export default createStyles((theme, { colorScheme, native }: PrismStylesParams)
   lineNumber: {
     color: colorScheme === 'dark' ? theme.colors.dark[4] : theme.colors.gray[4],
     textAlign: 'right',
-    paddingRight: theme.dir === 'ltr' ? theme.spacing.xl : undefined,
-    paddingLeft: theme.dir === 'rtl' ? theme.spacing.xl : undefined,
+    minWidth: 43,
```
rtivital, line 45: Will it fix the issue for triple digits, e.g. `122`?

auronsan1st: hmm nope if like that. need use margin

auronsan1st: Screen Shot 2022-04-02 at 04 00 17
```
 width: 43,
    marginRight: theme.dir === 'ltr' ? theme.spacing.xs : undefined,
    marginLeft: theme.dir === 'rtl' ? theme.spacing.xs : undefined,
   
   ```

auronsan1st: if just support three lines number, only need 25
Screen Shot 2022-04-02 at 10 16 14

rtivital:
Great, thanks!

PR #1164, opened 2022-04-02
[@mantine/core] Export Saturation component

u3u:


rtivital:
No, this component should not be exported, it relies on hsva color format that cannot be used in any other scenario

u3u:
Can be used with tinycolor when deep customization is required

rtivital:
Yes, and it will require additional documentation and maintenance, this is not something I would want to do, Saturation should remain an internal component.

Issue #1086, opened 2022-03-24
Rich text editor. Pasting over highlighted text deletes old text and doesn't paste new text.


runthis:
### What package has an issue

@mantine/rte

### Describe the bug

Highlight some text in the rich text editor, paste new text over it. The old text is deleted and the new text is not added. Unable to replicate using react quill without Mantine.

### In which browser did the problem occur

Chrome

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/nice-proskuriakova-iqbrl8?file=/src/App.tsx

### Do you know how to fix the issue

No

### Possible fix

Unsure if related: https://stackoverflow.com/questions/59825450/react-quill-custom-image-handler-module-causing-typing-issues-with-the-editor

rtivital:
Fixed in 4.1.3

PR #1174, opened 2022-04-03
[@mantine/rte] fix rte copy paste on highlight

auronsan1st:
copy styles from demo react quill https://codesandbox.io/s/6x93pk4rp3?file=/index.js:70-80

Screen Shot 2022-04-03 at 23 57 48

display none make useStyles re-render and caused racing setValue. onChange value from paste and onChange from re-render

https://github.com/mantinedev/mantine/issues/1086


rtivital:
Thanks!

PR #1184, opened 2022-04-04
Fix DatePicker when type="date"

knpwrs:
When an input has `type="date"` the `value` read/written to the dom node is a string in the format `YYYY-MM-DD` (the format actually rendered is determined by the browser), so when someone sets `type="date"` on the `DatePicker` component and it attempts to set the value in the dom the value doesn't actually get rendered without a properly set formatting string. Having `type="date"` is preferential when using a framework that ecourages form submissions such as Remix since otherwise the `input` will submit locale-formatted values rather than a timestamp or ISO date (whereas with `type="date"` the value is ISO-formatted).

Additionally, when `type="date"`, there is a browser-supplied clear button, resulting in two clear buttons being rendered. There is no reliable cross-browser way to hide this button, so this PR simply hides mantine's close button when `type="date"`.

### DIFF
--- a/src/mantine-dates/src/components/DatePicker/DatePicker.tsx
+++ b/src/mantine-dates/src/components/DatePicker/DatePicker.tsx
@@ -102,12 +102,13 @@ export const DatePicker = forwardRef<HTMLButtonElement, DatePickerProps>(
       onDropdownOpen,
       hideOutsideDates,
       hideWeekdays,
+      type,
       ...others
     } = useMantineDefaultProps('DatePicker', defaultProps, props);
 
     const theme = useMantineTheme();
     const finalLocale = locale || theme.datesLocale;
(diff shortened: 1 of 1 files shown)


rtivital:
I do not see why type date should be set by anyone, it only causes problems. If date picker is used as described in documentation, this problem will never occur.

knpwrs:
Like I said, when doing a vanilla form submission (as encouraged in frameworks like Remix) the input value is submitted, which in the case of the Mantine date picker is locale-formatted. Using `type="date"` makes the value received by the server consistent (ISO).

The workaround is to set the format to `YYYY-MM-DD` and hide the clear button. If it's preferable then the workaround can be documented rather than behavior changed.

rtivital:
Thanks!

PR #1197, opened 2022-04-07
[@mantine/core] Autocomplete: make dropdownPosition default to flip

ikapta:
The default dropdownPosition is bottom, it not flips to above when there is not enough space. The Select component dropdownPosition default also is flip, so I did this.

another, i saw the default limit is 5，but i want show large data，when I set limit to 30, the flip is also not work, so I also
need to set maxDropdownHeight，thus Autocomplete SelectDropdown will same as Select component.

I hope I did the right thing, Thanks.

https://codesandbox.io/s/adoring-merkle-bffpbm?file=/src/App.tsx


rtivital:
Thanks!

PR #1206, opened 2022-04-08
[docs] update package name in `get started with Remix` section

BWsix:
Seems like remix updated their package
https://github.com/remix-run/remix/blob/main/examples/mantine/app/entry.server.tsx


rtivital:
Thanks!

PR #1215, opened 2022-04-09
[@mantine/core] Notification: fix text so large not hidden

ikapta:
while text so large it not hidden.

ps: I am worried about if I should removing the `disallowClose` prop in NotificationStylesParams,  maybe others custom Notification styles  might use this property?

rk1NkYf2DX


rtivital:
Thanks!

PR #1223, opened 2022-04-10
Fix typo in Stepper demo

youssefm:
This typo is present in the docs too:
image

https://mantine.dev/core/stepper/


rtivital:
Thanks!

PR #1228, opened 2022-04-11
[@mantine/core] StepperProgress: New component

develowl:
Creation of a new Stepper component that has progressive milestones, with a multiple content to be shown before get to the next step.


rtivital:
I do not think that we should add such component as a part of `@mantine/core` library as Stepper component already covers the majority of use cases and StepperProgress component is very situational (I've never seen anything like it on a real website).

If you need this component for your work, I suggest you to publish a separate package that depends on `@mantine/core`.

develowl:
I understand, I'll do that you suggested. See you!

Issue #1211, opened 2022-04-09
MultiSelect does not respect width


PegasisForever:
### What package has an issue

@mantine/core

### Describe the bug

`MultiSelect` does not respect its `width` prop. The search input area is always very small (try typing in the input, it will clip) and the border always occupies full width.

### In which browser did the problem occur

every browser

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/kind-kirch-ojgr82?file=/src/App.js

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

No

megancooper:
The `Multiselect` component does not have a `width` prop. You can [check the list of props it has here](https://mantine.dev/core/multi-select/?t=props).
You can pass in a class or use `createStyles` to get component to respect a certain `width`:

```
import "./styles.css";
import { MultiSelect, createStyles } from "@mantine/core";
import { useState } from "react";

const useStyles = createStyles((theme, _params, getRef) => ({
  multiselect: {
    width: 200
  }
}));

export default function App() {
  const { classes } = useStyles();
  const [searchStr, setSearchStr] = useState("");
  return (
    <div>
      <MultiSelect
        searchable
        className={classes.multiselect}
        data={["a", "b", "c"]}
        value={["a"]}
        onSearchChange={setSearchStr}
      />
    </div>
  );
}
```

PegasisForever:
I see, I get confused because in the typescript definition it has the `width` prop.

PR #1234, opened 2022-04-12
Fix MultiSelect

PegasisForever:
- Fix search input width #1211 
- Deep compare `value` prop
  sometimes the `value` prop is generated on each render (e.g. `value={users.map(it => it.name)}`) which causes [useDidUpdate](https://github.com/mantinedev/mantine/blob/a51bf44a51347917b797ce5b8f573900a8a780a2/src/mantine-core/src/components/MultiSelect/MultiSelect.tsx#L293) to run at every render


PegasisForever:
wait, I found out directly using `value` in useEffect doesn't work

PR #1235, opened 2022-04-12
[@mantine/core] MultiSelect: Fix search input width

PegasisForever:
Related: #1211


rtivital:
Are you sure it will fix it? I haven't checked, but input has width: 60

PegasisForever:
Which example are you using? I checked it on storybook and its working great.

PegasisForever:
Oh i should prabably remove the width line since it is redundent now

rtivital:
I guess it should be min-width instead.

rtivital:
Thanks!

PR #1247, opened 2022-04-14
Option to clear values in select and multiselect when key press escape

ahweis:
Makes it possible to clear the value or all values when the drop down is closed and the escape key is pressed. Requires the clearable property to also be true

- This allow for keyboard interaction only on a select that is also clearable instead of needing mouse click to clear its content.


rtivital:
This is not a common use case and I consider that to be a bad UX – user does not expect that escape key will clear value – it is not what escape key is for in system ui and other ui libraries.

Issue #1246, opened 2022-04-14
Next.JS SSR not working even after setting up same as docs


maks112v:
### What package has an issue

@mantine/core
@mantine/next

### Describe the bug

Nextjs SSR doesn't work even after following config: https://mantine.dev/theming/next/

Project on the lastest version of all packages.

The issue only seems to happen when navigation to another screen using next/link but on direct load the issue doesn't exist

![CleanShot 2022-04-13 at 22 34 44](https://user-images.githubusercontent.com/11082252/163302609-b635dcc2-c6a3-4eb2-a19e-d8d44648d462.gif)

<img width="641" alt="CleanShot 2022-04-13 at 21 48 12@2x" src="https://user-images.githubusercontent.com/11082252/163298032-d2ae3fc7-ca93-47e4-b6f7-6c0ea8cd75c5.png">

There are no errors returned in console.

### In which browser did the problem occur

Chrome

rtivital:
1. Delete node_modules folder
2. Delete .next folder
3. Delete package-lock.json/yarn.lock
4. Install dependencies

If that does not fix the issue, use template to get started, it is known to not have any issues.
https://github.com/mantinedev/mantine-next-template

maks112v:
I tried the steps above and it still didn't work the createStyles that are not loaded initially are still broken. I tried to also downgrade to version 17 of react and still had the same issues.

rtivital:
Well, I do not have any other ideas, I can only propose to use template.

fuzunspm:
@maks112v could you find any solution?

maks112v:
@fuzunspm  Yep in my case it was that I didn't have the full _document file. The documentation was fixed in the pr above ^
```
import { createGetInitialProps } from '@mantine/next';
import Document, { Head, Html, Main, NextScript } from 'next/document';

const getInitialProps = createGetInitialProps();

export default class _Document extends Document {
  static getInitialProps = getInitialProps;

  render() {
    return (
      <Html>
        <Head />
        <body>
          <Main />
          <NextScript />
        </body>
      </Html>
    );
  }
}
```

fuzunspm:
thanks for the reply. This worked and not worked. It can only display styles after a refresh and modals are still un-styled

devtim123:
For me the problem was solved by pinning the versions of following modules


```
{
  "dependencies": {
    "next": "12.1.4"
  },
  "devDependencies": {
    "@types/node": "17.0.34",
[...]

PR #1253, opened 2022-04-14
[docs] Adding more content to nextjs _document

maks112v:
Adding docs to prevent confusion when working with nextjs.

Solves: https://github.com/mantinedev/mantine/issues/1246


rtivital:
Does these changes fix your issue?

maks112v:
Yep the deployment works perfectly. I copied the _document without looking up nextjs docs and it worked on development but not production so thought I would fix the docs to prevent farther confusion.

rtivital:
Okay, thanks! I'll review that and add to the docs

rtivital:
Thanks!

Issue #1100, opened 2022-03-25
[@mantine/form] zod-resolver doesn't handle asynchronous validation


Punie:
### What package has an issue

@mantine/form

### Describe the bug

First of all, thank you for the amazing work done on Mantine 4.0 🙏 

I my current project, I am making heavy use of form validation through `zod` and in some cases I have a need for asynchronous validation, eg:
```ts
const signupSchema = z
  .object({
    username: z
      .string()
      .refine(
        value =>
          axios
            .post('/api/check_unique_username', { username: value })
            .then(() => true)
            .catch(() => false),
        {
          message: 'Username already exists',
        },
      ),
  });
```

However, this doesn't work with the current implementation of `zodResolver` because it uses the `safeParse` method (see [here](https://github.com/mantinedev/mantine/blob/master/src/mantine-form/src/resolvers/zod-resolver/zod-resolver.ts#L23)). The `zod` library does handle async validation but you need to call the [`safeParseAsync`](https://github.com/colinhacks/zod#safeparseasync) method for it to work.

### In which browser did the problem occur

All of them

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

rtivital:
You can send a PR to the `master` branch

maticrivo:
Is anyone working on this? i see https://github.com/mantinedev/mantine/pull/1255 is closed..

PR #1255, opened 2022-04-15
[@mantine/form] add asyncResolver on use-form

auronsan1st:
https://github.com/mantinedev/mantine/issues/1100

to check promise need call function ex schema(null) I think there will be performance loss on other resolver ex: yup, zoi. so I decided just use asyncResolver props.

my suggest to make not bigger code should split between async and sync so no hidden bug will occur.
the schema types depend on many place. so I can't change to use Promise<schema>. for now I just bypass it by any when init useForm

### DIFF
--- a/src/mantine-demos/src/demos/form/Form.demo.zod.tsx
+++ b/src/mantine-demos/src/demos/form/Form.demo.zod.tsx
@@ -66,8 +66,7 @@ const schema = z.object({
 
 function Demo() {
   const form = useForm({
-    asyncResolver: true,
-    schema: zodResolver(schema) as any,
+    schema: zodResolver(schema),
     initialValues: {
       name: '',
       email: '',
--- a/src/mantine-demos/src/demos/form/Form.demo.zod.tsx
+++ b/src/mantine-demos/src/demos/form/Form.demo.zod.tsx
@@ -66,7 +66,8 @@ const schema = z.object({
 
 function Demo() {
   const form = useForm({
-    schema: zodResolver(schema),
+    asyncResolver: true,
+    schema: zodResolver(schema) as any,
     initialValues: {
       name: '',
       email: '',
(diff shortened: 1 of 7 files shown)


src/mantine-form/src/resolvers/async-test-resolver.ts
```diff
@@ -0,0 +1,97 @@
+import { renderHook, act } from '@testing-library/react-hooks';
```
rtivital, line 1: All tests files should be in tests folder.

src/mantine-demos/src/demos/form/Form.demo.zod.tsx
```diff
@@ -66,7 +66,8 @@ const schema = z.object({
 
 function Demo() {
   const form = useForm({
-    schema: zodResolver(schema),
+    asyncResolver: true,
+    schema: zodResolver(schema) as any,
```
rtivital, line 70: API does not fell nice. Can you make it as before `schema: zodResolver(schema)`, without `asyncResolver` field and `as any`?

rtivital (commented):
Currently, I do not really like what I'm seeing, async api feels very strange and unintuitive.

src/mantine-form/src/resolvers/async-test-resolver.ts
auronsan: it's test helper async version of `src/mantine-form/src/resolvers/test-resolver.ts`

auronsan:
> Currently, I do not really like what I'm seeing, async api feels very strange and unintuitive.

yeah I feel same. but if can need support it right. actually if https://github.com/mantinedev/mantine/issues/1100
they can check theirselves and throw error when trigger on submit

rtivital:
Alright, I think this issue will require more investigation before it can be merged, getting the api right for these use cases seems to be tricky

PR #1264, opened 2022-04-16
[@mantine/core] SegmentedControl: remove initial active tab translation animation

achmurali:
#1239 ... assuming we don't want initial transition on render, have done this. If it is to be avoided via a prop, shall do it


rtivital:
It's much better than it was before! There is still one issue remaining that is related to that – when page is refreshed, the active item is not highlighted until component mounts (see video). So, I'm proposing that we should add styles to the active item that match styles of the floating indicator and remove them after mount.

How to test this:
- Make changes to component
- `npm run build @mantine/core`
- `npm run docs:build && npm run docs:serve`
- Open SegmentedControl page and refresh it

P.S. do not forget about different supported colors.

https://user-images.githubusercontent.com/10353856/163685728-0c16195a-a2ea-4d51-802f-4df66ea43685.mov

rtivital:
Works perfect, thanks!

PR #1267, opened 2022-04-16
[core] Omit "*" from label accessible names

apoco:
Omit the required "*" from the accessible name of labels by using `aria-hidden` on the span.


rtivital:
Thanks!

PR #1268, opened 2022-04-16
[@mantine/rte] add args for onChange callback

Jawell:
Hi! I see that there are not enough args. Can we add them?

### DIFF
--- a/src/mantine-rte/src/components/RichTextEditor/RichTextEditor.tsx
+++ b/src/mantine-rte/src/components/RichTextEditor/RichTextEditor.tsx
@@ -1,5 +1,6 @@
 import React, { useMemo, useRef, useEffect, forwardRef } from 'react';
 import Editor, { Quill } from 'react-quill';
+import type { Delta, Sources } from 'quill';
 import 'quill-mention';
 import {
   DefaultProps,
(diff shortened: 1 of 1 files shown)


src/mantine-rte/src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -44,10 +45,10 @@ export interface RichTextEditorProps
   extends DefaultProps<RichTextEditorStylesNames>,
     Omit<React.ComponentPropsWithoutRef<'div'>, 'onChange'> {
   /** HTML content, value not forced as quill works in uncontrolled mode */
-  value: string;
+  value: string | Delta;
```
rtivital, line 48: We cannot change value type, it will be breaking. Adding types to `onChange` is fine.

Jawell: But this is input value. This is just extend types which you can provide to RTE component. Also, we leave `string` type

rtivital: Ah, yes, that's true, sorry, I was thinking of something else

rtivital:
Thanks!

PR #1269, opened 2022-04-16
[@mantine/core] Alert: a11y fixes

apoco:
- Add `alert` role to `Alert` components
- Use the title to label alerts
- Use the body to describe alerts


apoco:
I wasn't sure if you had something pre-React-18 in place to get a unique id so that `aria-labeledby` can work without an alert having an ID.

rtivital:
Thanks!

PR #1270, opened 2022-04-17
[@mantine/core] Modal: add id to root element

apoco:
The `id` prop passed to `<Modal/>` is not being added to the DOM.


src/mantine-core/src/components/Modal/Modal.tsx
```diff
@@ -193,6 +193,7 @@ export function Modal(props: ModalProps) {
         {(transitionStyles) => (
           <Box className={cx(classes.root, className)} {...others}>
             <div
+              id={baseId}
```
rtivital, line 196: You are adding id to the inner element, not to the root. It should be added to the Box component above.

apoco: :+1: updated.

rtivital:
Thanks!

PR #1274, opened 2022-04-17
[docs] Update dynamic RTL section

serafy0:
adds dynamic document direction to the docs, similar to https://github.com/mantinedev/mantine/blob/master/docs/src/components/Layout/Layout.tsx#L53


rtivital:
Thanks!

PR #1281, opened 2022-04-19
[@mantine/core] Segmented control : use isomorphic effect hook

achmurali:


rtivital:
Please target PR towards master branch without extra commits, it should be in the next patch

PR #1283, opened 2022-04-19
[@mantine/core] Fix "sticky hover effect" on touch device

alanko0511:
Problem: when clicking on components with the hover effect included, the effect stays on.

To fix this, util is added as a polyfill to return style that `hover` effect will act as `active` effect on touch devices, and used this util to components with `hover` effect.


src/mantine-core/src/components/Accordion/AccordionItem/AccordionItem.styles.ts
```diff
@@ -1,4 +1,5 @@
 import { createStyles } from '@mantine/styles';
+import { applyHoverStyleForTouch } from '../../../utils';
```
rtivital, line 2: I think this function should be a part of `theme.fns`, usage example:

```tsx
...theme.fns.hover({ backgroundColor: 'red' })
```

(No additional documentation required)

This way function will be accessible in other packages as well.

rtivital (commented):
Thanks for the PR! Please add function to theme.fns and also update components in `@mantine/dates` package that have hover effects

src/mantine-styles/src/theme/functions/fns/hover/hover.ts
```diff
@@ -0,0 +1,12 @@
+import type { CSSObject } from '../../../../tss';
+
+export function hover(hoverStyle: any) {
```
rtivital, line 3: CSSObject should be here instead of any

src/mantine-styles/src/theme/types/MantineTheme.ts
```diff
@@ -36,6 +36,7 @@ interface MantineThemeFunctions {
   radius(size: MantineNumberSize | (string & {})): string | number;
   variant(payload: VariantInput): VariantOutput;
   primaryShade(colorScheme?: ColorScheme): Shade;
+  hover(hoverStyle: any): any;
```
rtivital, line 39: Same here `hover(hoverStyle: CSSObject): any;`

rtivital:
Thanks!

PR #1287, opened 2022-04-20
[@mantine/modals] Add new value to Record type ConfirmLabels

sevgeek:
Replace for `confirm` and `cancel` labels type `string` to `ReactNode`. 

This change will allow to place all valid values of the type `ReactNode` inside the confirm and cancel buttons confirm modal component.

#1286


rtivital:
Thanks!

PR #1303, opened 2022-04-22
[@mantine/core] SegmentedControl: Add disabled property support for individual options.

zhulien-ivanov:
Add `disabled` property support for individual options.
Add missing `not-allowed` value for `cursor` for `disabled` state of control and individual options.


rtivital:
Please fix errors

rtivital:
Thanks!

PR #1308, opened 2022-04-23
[docs] use-element-size: Update source reference.

zhulien-ivanov:
`use-element-size` hook is not exported into its own file and is actually located inside `use-resize-observer`'s source.


rtivital:
Thanks!

PR #1310, opened 2022-04-23
[docs] Fix incorrect source references for hooks.

zhulien-ivanov:
Fix missing `.ts` extension for:
1. `use-element-size`.
2. `use-event-listener`.


zhulien-ivanov:
Sorry for the double PR. I've checked all other hooks.

rtivital:
Thanks!

PR #1311, opened 2022-04-23
[docs] use-focus-return: Update example for shouldReturnFocus to reflect description.

zhulien-ivanov:
Plus some minor code highlighting improvements.

### DIFF
--- a/docs/src/docs/hooks/use-focus-return.mdx
+++ b/docs/src/docs/hooks/use-focus-return.mdx
@@ -15,7 +15,7 @@ import { ModalDemos } from '@mantine/demos';
 
 ## Usage
 
-`use-focus-return` automatically manages focus returning to last focused element when given condition is met.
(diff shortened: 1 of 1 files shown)


docs/src/docs/hooks/use-focus-return.mdx
```diff
@@ -15,7 +15,7 @@ import { ModalDemos } from '@mantine/demos';
 
 ## Usage
 
-use-focus-return automatically manages focus returning to last focused element when given condition is met.
+`use-focus-return` automatically manages focus returning to last focused element when given condition is met.
```
rtivital, line 18: Hook name should not be wrapped with code, see other hooks

zhulien-ivanov: Yeah, you're right. My bad.

rtivital:
Thanks!

PR #1315, opened 2022-04-24
[@mantine/core] (Multi)Select/Autocomplete : add positionDependencies

achmurali:


src/mantine-core/src/components/Select/Select.tsx
```diff
@@ -87,6 +87,9 @@ export interface SelectSharedProps<Item, Value> {
 
   /** Whether to switch item order and keyboard navigation on dropdown position flip */
   switchDirectionOnFlip?: boolean;
+
+  /** useEffect dependencies to force update tooltip position */
```
rtivital, line 91: Invalid description

achmurali: Have followed this - https://github.com/mantinedev/mantine/blob/master/src/mantine-dates/src/components/DatePickerBase/DatePickerBase.tsx#L82, so should i update Datepicker's `positionDependencies` description?

rtivital: Yes

rtivital: It should be dropdown, not tooltip

rtivital:
Thanks!

Issue #1245, opened 2022-04-14
React 18 implicit children typing issue?


girvo:
### What package has an issue

@mantine/core

### Describe the bug

So the `@types/react` typing package for TypeScript, for React 18, has remove the implicit children for all component types, and this is causing me to run into some issues with our extensions of some Mantine components.

```typescript
import React, { ComponentType, Key, Fragment } from 'react'

// given a Props interface with this:
  header?: () => React.ReactNode
}

// snip
  return (
    <div className={classes.grid}>
      {props.header && (
        <MediaQuery largerThan="sm" styles={hide}>
          {props.header()}
        </MediaQuery>
      )}
```

When used with React 18.0.1, `@types/react` 18.0.1 and `@mantine/core` v4.1.3, this gives an error about React.ReactNode not being React.ReactNode!

At the `{props.header()}` line:
```
Type 'React.ReactNode' is not assignable to type 'import("/workspaces/cobalt-dev/node_modules/@types/react/index").ReactNode'.
  Type '{}' is not assignable to type 'ReactNode'.
  ```

I'm not entirely sure what is causing this, other than some kind of typing mismatch when trying to use React v18 and it's v18 @types package with Mantine?

### In which browser did the problem occur

All browsers

fl0wx:
@girvo yesterday i found this, which could be related: https://www.reddit.com/r/reactjs/comments/u2rbbx/library_authors_should_revise_typings_to_support/

rtivital:
Mantine does not use FC type anywhere, so most likely provided codemod will not help.

girvo:
@rtivital Yeah I initially assumed it was something to do with that as well initially (you can tell in my initial paragraph thats what I thought it was) but then checked the source and saw that the FC type isn't used anywhere, and yet the type error still exists. Super odd.

Is there anything I can do to help with tracking this down? It's unfortunately prevented us from migrating to v18 at work. Happy to do some more debugging or see if I can reproduce a tiny test case for it if need be, or post more of the source we have for it, but it's super weird. Basically thinks it's `React.ReactNode` type is the empty object, instead of the actual `ReactNode` definition, despite both my code and Mantine ostensibly using the same `@types/react` package

wes337:
> Mantine does not use FC type anywhere, so most likely provided codemod will not help.

Are you sure? I ran the codemod (https://github.com/eps1lon/types-react-codemod) and quite a few changes were made:

[...]

PR #1322, opened 2022-04-25
[@mantine/dates] AmPmInput: Fix onChange being called twice

### LINKED ISSUE #1256: TimeInput AM/PM React 18
### What package has an issue

@mantine/dates

### Describe the bug

TimeInput am/pm state is stuck on am (when format = '12').

- Issue is occurring on React 18

(Mac) Chrome, Firefox, Safari

_No response_

No

No

### Possible fix

_No response_

nrayburn-tech:
Fixes #1256 

When entering `a` or `p` into the AmPmInput, the onChange event is triggered twice.  This adds a call to `event.preventDefault`, so that onChange is called a single time.


rtivital:
Have you checked that with React 18? Storybook and docs are currently working with React 17 and there is not any way to check that in dev environment.

nrayburn-tech:
I manually changed the project to React 18, tested that the bug reproduced, made the code change, and tested that it didn’t reproduce. (Ignoring the type errors from the implicit children https://github.com/mantinedev/mantine/issues/1245.)
I used one of the TimeInput Storybooks to test. I didn’t check the docs.

I’ll look again to see if I can find any issues or a definite reason why there’s a difference between 17/18.

nrayburn-tech:
Updated so that the focus behaves the same as currently on React 17 without the change.  Also added a TimeInputRange demo, so that the focus can be tested in Storybook.

As far as I can tell both React 17 and React 18 both call the onChange function prop twice.  Once during keyDown and once during onChange.  The difference is that when React 17 does it, the value only changes once.  When React 18 does it, the value changes twice so it appears to not change at all to the user. (Goes from am to pm and then back to am.)

I don't think I'll be able to find out anything more specific as to why this happens differently now.

rtivital:
Alright, thanks for PR and testing!

PR #1329, opened 2022-04-25
[@mantine/modals] ConfirmModal: Add default labels

### LINKED ISSUE #836: Confirm Modal errors when labels not provided
### What package has an issue

@mantine/modals

### Describe the bug

`openConfirmModal` does not require labels in the props, but errors if `labels`, `cancelProps`, and `confirmProps` are not provided

Chrome

https://codesandbox.io/s/unruffled-boyd-fb1gk?file=/src/App.tsx

Yes

Yes

### Possible fix

Provide default values for labels, check labels exists before reading `labels.confirm` and `labels.cancel`, or make labels a required prop

mattrunyon:
Fixes #836 

Sets default confirm modal labels to empty string to avoid browser error (and no modal opening) if the labels are not. Empty string instead of English labels so it is not confusing for non-English development/usage

### DIFF
--- a/src/mantine-modals/src/ConfirmModal.tsx
+++ b/src/mantine-modals/src/ConfirmModal.tsx
@@ -20,7 +20,7 @@ export function ConfirmModal({
   id,
   cancelProps,
   confirmProps,
-  labels,
+  labels = { cancel: '', confirm: '' },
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #1339, opened 2022-04-26
[docs] Removed incorrect argument types for radialGradient

sunney-x:
`theme.fn.radialGradient` only takes an argument of type `string[]`, which wasn't represented correctly in the documentation.


rtivital:
Thanks!

PR #1342, opened 2022-04-28
[@mantine/dates] TimeInput: Call onChange prop function when clearing…

### LINKED ISSUE #1294: TimeInput onChange not called when cleared
### What package has an issue

@mantine/dates

### Describe the bug

When a TimeInput is cleared using the "clearable" prop, the onChange prop is not invoked.

https://github.com/mantinedev/mantine/blob/master/src/mantine-dates/src/components/TimeInput/TimeInput.tsx#L225

All

_No response_

Yes

Yes

### Possible fix

Since the `handleClear` function is not using `setDate`, it doesn't ever call `props.onChange`

The easiest solution would be to call `onChange?.(null)` inside `handleClear`, or refactor to use setDate

uf0h:
Fixes #1294


rtivital:
Thanks!

Issue #1333, opened 2022-04-26
Divider text color is not applied


stefanoborini:
### What package has an issue

@mantine/core

### Describe the bug

The following divider specification does not propagate the color to the label. 

```
<Divider label="hello" labelProps={{size: "xl", weight: 700, color: "blue"}} labelPosition="center"></Divider>
```

generates the following class:

```
.mantine-1emo7cc {
    -webkit-font-smoothing: antialiased;
    -moz-osx-font-smoothing: grayscale;
    font-family: -apple-system,BlinkMacSystemFont,Segoe UI,Roboto,Helvetica,Arial,sans-serif,Apple Color Emoji,Segoe UI Emoji;
    -webkit-tap-highlight-color: transparent;
    color: #495057;
    font-size: 20px;
    line-height: 1.55;
    -webkit-text-decoration: none;
    text-decoration: none;
    font-weight: 700;
    display: -webkit-box;
    display: -webkit-flex;
    display: -ms-flexbox;
    display: flex;
    -webkit-align-items: center;
    -webkit-box-align: center;
    -ms-flex-align: center;
    align-items: center;
    color: #868e96;
    margin-top: 2px;
}
```

Note that the weight is passed. The colors are a dark grey and a light gray, so blue is not passed.

### In which browser did the problem occur

Chrome latest

zhulien-ivanov:
The property is applied but it gets overriden. I can take a look and apply a fix tomorrow.

zhulien-ivanov:
Solution is merged into master. I guess this can be closed.

PR #1343, opened 2022-04-28
[@mantine-core] [Issue: #1333] Divider: Fix labelProp's color property getting overriden by Divider's color.

zhulien-ivanov:


zhulien-ivanov:
I've skipped writing a unit test to verify respecting labelProp's `color` as it will be either too specific and will break on styles/shade change or will be too broad. In short, it will be more of a nuisance than do good.

rtivital:
Thanks!

PR #1357, opened 2022-04-30
[@mantine/core] Alert: Fix a11y name without id

apoco:
Fix label association of alerts when they have no `id`


apoco:
:man_facepalming: annoyed I didn't catch this the first time around.

rtivital:
Thanks!

PR #1364, opened 2022-05-01
[@mantine/core] Autocomplete: Fix enter not submitting form

### LINKED ISSUE #1360: Autocomplete doesn't submit form when input doesn't match a data item
### What package has an issue

@mantine/core

### Describe the bug

I'm trying to build an autosuggestion search bar kind of like Google. Suggestions are provided based on what the user has currently typed, but they can also type in an arbitrary search string. My problem is that Autocomplete doesn't let the enter keydown event bubble up to submit my form when the user types something that doesn't match one of the autocomplete options.

In my linked CodeSandbox, type "Other" and hit "Enter" and notice how the onSubmit alert isn't triggered. It is triggered when you type "Cat" and hit Enter twice (once to select it and again to submit the form).

Brave v1.38.109

https://codesandbox.io/s/distracted-pascal-ut0qgh

No

Yes

### Possible fix

_No response_

canac:
Fixes #1360

Now pressing enter when there are no filtered items doesn't call event.preventDefault(), so a parent form will be submitted now. Try the Autocomplete -> Within form story to test the changes.

### DIFF
--- a/src/mantine-core/src/components/Autocomplete/Autocomplete.tsx
+++ b/src/mantine-core/src/components/Autocomplete/Autocomplete.tsx
@@ -139,11 +139,8 @@ export const Autocomplete = forwardRef<HTMLInputElement, AutocompleteProps>(
         }
 
         case 'Enter': {
-          if (dropdownOpened) {
-            event.preventDefault();
-          }
-
           if (filteredData[hovered] && dropdownOpened) {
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks! The fix will be released with 4.2.3 patch next Sunday.

PR #1370, opened 2022-05-02
[@mantine/core] Text: Change color's type to include dimmed value.

zhulien-ivanov:
Related to #1368.

Provides accurate intellisense for `dimmed` color value.

### DIFF
--- a/src/mantine-core/src/components/Text/Text.tsx
+++ b/src/mantine-core/src/components/Text/Text.tsx
@@ -6,9 +6,9 @@ import {
   PolymorphicComponentProps,
   PolymorphicRef,
   useMantineDefaultProps,
+  MantineColor,
 } from '@mantine/styles';
 import { Box } from '../Box';
-import { MantineTextColor } from './Text.types';
 import useStyles from './Text.styles';
 
 export interface SharedTextProps extends DefaultProps {
@@ -16,7 +16,7 @@ export interface SharedTextProps extends DefaultProps {
   size?: MantineSize;
 
   /** Text color from theme or dimmed variant */
-  color?: MantineTextColor;
+  color?: 'dimmed' | MantineColor;
 
   /** Sets font-weight css property */
   weight?: React.CSSProperties['fontWeight'];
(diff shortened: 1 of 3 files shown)


zhulien-ivanov:
`MantineTextColor` is extracted into `Text.types` as otherwise it causes a circular dependency when it resides in `Text` as it is used by both `Text` and `Text.styles` (which have to import each other). This is a bit inconsistent as we're not exporting the `MantineTextColor` from `Text` but I'm not sure how to solve this issue in a better way. Maybe a default export from `Text.types` and then import + re-export from `Text`?

src/mantine-core/src/components/Text/Text.tsx
```diff
@@ -3,20 +3,20 @@ import {
...
 import useStyles from './Text.styles';
 
 export interface SharedTextProps extends DefaultProps {
   /** Predefined font-size from theme.fontSizes */
   size?: MantineSize;
 
-  /** Text color from theme */
-  color?: MantineColor;
+  /** Text color from theme or dimmed variant */
+  color?: MantineTextColor;
```
rtivital, line 19: There is no need to extract this type, it should not be exported or used anywhere else, it should be just `color?: 'dimmed' | MantineColor`

zhulien-ivanov: It's good to have a separate type as you can use it to assign the text color prior to using it in `Text` (or `Text` based components like `Menu.Label` for example). If we use a simple union, we have code duplication which also makes it harder to maintain and spot errors. But I guess it can always be added later if you decide on it.
I'll refactor it to a union for now, then.

src/mantine-core/src/components/Text/Text.tsx
```diff
@@ -15,8 +15,8 @@ export interface SharedTextProps extends DefaultProps {
   /** Predefined font-size from theme.fontSizes */
   size?: MantineSize;
 
-  /** Text color from theme */
-  color?: MantineColor;
+  /** Text color from theme or dimmed variant */
+  color?: 'dimmed' | MantineColor;
```
timagixe, line 19: If 'dimmed' is supposed to use in some other components too, why not simply create a utility type like below
```typescript
type WithDimmed<T> = T | 'dimmed';
```

And after use it like:
```suggestion
  color?: WithDimmed<MantineColor>;
```

rtivital:
Thanks!

PR #1371, opened 2022-05-02
[@mantine/hooks] use-move: fix onScrubEnd on quick tap i.e without any scrubbing

achmurali:
`onScrubEnd` is being called before the `onChange` of parent component like `Slider`, hence a mismatch in the value being passed to `onScrubEnd`.


rtivital:
Thanks!

PR #1372, opened 2022-05-02
[@mantine-dropzone] Add support for React 18

bbar:
I struggled with docs a bit. I wasn't sure where to add the example since using an array is so simple and intuitive, it's nice having it right at the top so it's the first thing users see. At the same time, the two current examples flow so nicely into the mime types table that I didn't want to disrupt the flow. Anyway, I'm happy to make any changes.


Manokii (approved):

rtivital:
Thanks!

PR #1382, opened 2022-05-04
[@mantine/core] NumberInput: Allow user to pass inputMode

itsmoops:
The current logic to determine the NumberInput `inputMode` prop checks various combinations of the `step` and `precision` props to make that determination. I'm proposing that users should be allowed to supply their own `inputMode` prop in the event that the logic in `mantine/src/mantine-core/src/utils/get-input-mode` doesn't account for their needs.


rtivital:
Thanks!

PR #1383, opened 2022-05-04
[@mantine/form] use-form: Make functions referentially stable

### LINKED ISSUE #1196: Excessive re-rendering from useForm hook
### Discussed in https://github.com/mantinedev/mantine/discussions/1193

<div type='discussions-op-text'>

<sup>Originally posted by **rfdrake27** April  6, 2022</sup>
Hello, I'm noticing that updating any value within the form will trigger a re-render of the parent component where the useForm hook is located. Even after destructuring the returned object from useForm, I'm noticing that functions such as getInputProps or addListItem are also being changed. 

Are there plans to optimize performance of this hook? Such as wrapping functions in useCallback, or using refs?

Thank you Vitaly for the excellent library. </div>

mattrunyon:
Fixes #1196, #1193 

This makes all functions returned by `useForm` stable. I used the eslint react-hooks plugin for their exhaustive dependencies rule while converting these. It helped catch everywhere that the state needed to be switched to using the ref value within the `useCallback` uses

### DIFF
--- a/src/mantine-form/src/use-form.ts
+++ b/src/mantine-form/src/use-form.ts
@@ -1,9 +1,11 @@
-import { useState, useCallback, useRef, useEffect } from 'react';
+import { useCallback } from 'react';
 import { formList, isFormList, FormList } from './form-list/form-list';
 import { validateValues, validateFieldValue } from './validate-values/validate-values';
 import { filterErrors } from './filter-errors/filter-errors';
 import { getInputOnChange } from './get-input-on-change/get-input-on-change';
 import { getErrorPath } from './get-error-path/get-error-path';
+import { useStateRef } from './use-state-ref/use-state-ref';
+import { usePropRef } from './use-prop-ref/use-prop-ref';
 import type {
   FormErrors,
   FormRules,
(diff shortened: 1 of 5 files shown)


src/mantine-form/src/use-form.ts
```diff
@@ -63,46 +63,73 @@ export interface UseFormReturnType<T> {
...
 }
 
+/**
+ * Creates a ref that stays up to date with the state of the param
+ * Used to help with all the useCallback signatures in useForm
+ *
+ * @param value The value to clone in the ref
+ * @returns The ref which clones the value
+ */
+function useStateRef<T>(value: T): React.MutableRefObject<T> {
```
rtivital, line 73: It should be in a separate file

src/mantine-form/src/use-form.ts
```diff
@@ -63,46 +63,73 @@ export interface UseFormReturnType<T> {
   ) => GetInputProps<L>;
 }
 
+/**
+ * Creates a ref that stays up to date with the state of the param
+ * Used to help with all the useCallback signatures in useForm
+ *
+ * @param value The value to clone in the ref
```
rtivital, line 70: No need for documentation in comments, it's pretty clear what hook does

src/mantine-form/src/use-form.ts
```diff
@@ -63,46 +63,73 @@ export interface UseFormReturnType<T> {
...
+}
+
 export function useForm<T extends { [key: string]: any }>({
   initialValues,
   initialErrors,
   validate: rules,
   schema,
 }: UseFormInput<T>): UseFormReturnType<T> {
   const [errors, setErrors] = useState(filterErrors(initialErrors));
+  const errorsRef = useStateRef(errors);
```
rtivital, line 90: I think it is too wordy to create separate variables and calling hooks to get refs. Instead move this logic to `useRefState` hook, that will handle that, for example:

```tsx
const [errors, setErrors, errorsRef] = useRefState(filterErrors(initialErrors));
```

rtivital (commented):
I haven't checked how it works, but if tests are passing, then it is 👍

mattrunyon:
Made the changes you suggested (and also made another helper hook to track the state of a prop in a ref).

I also added a unit test for checking the functions are stable. Tested it failed when I removed `useCallback` from one of functions returned by the hook

rtivital:
Thanks!

rtivital:
I've merged it without checking how it works in real demos (stupid of me) and it turns out that it does not work at all – all demos in broken (values updates are random). I've reverted the changes, so if you want, you can open another PR for this.

PR #1387, opened 2022-05-04
use-list-state methods now return newState instead of void

tylim88:
As discussed in discord

returning newState from the method so we don't need to rely on useEffect to do something if the value changed

for example, after the user calls `handle.reorder`, the only way to update the new state to the database is via useEffect

```ts
useEffect(()=>{
  updateListInDatabase(list)
},[list])
```

There are a few drawbacks to this approach:
1. updateListInDatabase(or any other computation) will get called every time the state change, frequent API call is expensive, especially for someone that use cloud service like firebase firestore.
2. context is unknown. We don't know what triggers the change, for example, I want to do X with state after **append** and do Y with state after **sort**, this is impossible because the context is unknown

The easiest way to solve this is to return the new state from the methods, so we can do something like this

```ts
const reorderList=(indexes)=>{
  updateListInDatabase(handler.reorder(indexes))
}
```

with this:
1. API(or any other computation) will not get called every time a state changes (finer control)
2. context is known, we know what triggers the API call(informative)

### DIFF
--- a/docs/src/docs/hooks/use-list-state.mdx
+++ b/docs/src/docs/hooks/use-list-state.mdx
@@ -97,20 +97,20 @@ function useListState<T>(initialValue: T[] = []): [
   T[],
   {
     setState: Dispatch<SetStateAction<T[]>>;
-    append: (...items: T[]) => void;
-    prepend: (...items: T[]) => void;
-    insert: (index: number, ...items: T[]) => void;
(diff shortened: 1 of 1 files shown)


rtivital:
Please update documentation with new hook definition – https://mantine.dev/hooks/use-list-state/#definition

PR #1396, opened 2022-05-05
[@mantine/spotlight] Fix registerSpotlightActions and removeSpotlight…

KaiserBloo:
…Actions

Both of these functions were creating trigger events rather than their respective correct ones (registerActions & removeActions).

Tested on my own project and it fixed my issue with registerSpotlightActions seemingly not doing anything.


rtivital:
Thanks!

PR #1403, opened 2022-05-07
[@mantine/hooks] use-network: Include all properties and remove dummy object.

zhulien-ivanov:
Yeah, the support is truly terrible.

TypeScript types i.e. `NetworkInformation`, `ConnectionType` and such are incomplete and don't reflect the MDN docs as well so I didn't use them.

### DIFF
--- a/src/mantine-hooks/src/use-network/use-network.ts
+++ b/src/mantine-hooks/src/use-network/use-network.ts
@@ -2,22 +2,17 @@ import { useState, useEffect, useCallback } from 'react';
 import { useWindowEvent } from '../use-window-event/use-window-event';
 
 interface NetworkStatus {
-  downlink: number;
-  effectiveType: 'slow-2g' | '2g' | '3g' | '4g';
-  saveData: boolean;
-  rtt: number;
+  downlink?: number;
+  downlinkMax?: number;
+  effectiveType?: 'slow-2g' | '2g' | '3g' | '4g';
+  rtt?: number;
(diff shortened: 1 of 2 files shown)


src/mantine-hooks/src/use-network/use-network.ts
```diff
@@ -47,7 +44,7 @@ export function useNetwork() {
   useWindowEvent('offline', () => setStatus({ online: false, ...getConnection() }));
 
   useEffect(() => {
-    if (navigator.connection) {
+    if (navigator?.connection) {
```
rtivital, line 47: In which case navigator can be undefined?

zhulien-ivanov: I'm not sure what's the status in SSR. I've added it to be on the safe-side.
Also, there is `if (typeof navigator === 'undefined') {` in `getConnection` so I am keeping it consistent.

rtivital: useEffect hooks are never called during ssr, while `getConnection` function will be called since it is used as an initial value for useState

zhulien-ivanov: Got it. You want the check removed?

rtivital: Yep, seems like ut is not required

rtivital:
Thanks!

PR #1406, opened 2022-05-08
[@mantine/ssr] Update html-react-parser

### LINKED ISSUE #1381: Unmet peer dependencies
### What package has an issue

@mantine/core

### Describe the bug

It's actually @mantine/next but I can't choose that in the dropdown.

`pnpm add @mantine/next` gives the following error;
image

This looks to be due to the fact that @mantine/ssr is locked to html-react-parser@1.3.0
https://github.com/mantinedev/mantine/blob/master/src/mantine-ssr/package.json#L41

The latest version of html-react-parser supports react 18
https://github.com/remarkablemark/html-react-parser/blob/master/package.json#L79

N/A

_No response_

Yes

Yes

### Possible fix

I _think_ it's just a matter of upgrading the dependency in @mantine/ssr? It's only a patch release so there shouldn't be any breaking changes (checked the release notes, couldn't see any).

FizzBuzz791:
Update html-react-parser so that it supports react 18. 

Fixes #1381.

### DIFF
--- a/package.json
+++ b/package.json
@@ -132,7 +132,7 @@
     "clsx": "^1.1.1",
     "csstype": "3.0.9",
     "dayjs": "^1.10.5",
-    "html-react-parser": "1.3.0",
+    "html-react-parser": "1.4.12",
     "prism-react-renderer": "^1.2.1",
     "quill-mention": "^3.0.8",
     "react": "^17.0.2",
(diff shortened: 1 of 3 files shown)


rtivital:
Tests do not pass

rtivital:
Thanks!

PR #1408, opened 2022-05-08
[@mantine/labs] PinInput calls onChange twice

jerebtw:
Before this PR when you entered 4 numbers and remove one the onChange function is called twice and the second time with the old value so for example `1234` instead of `123` the real value that is display


rtivital:
Thanks!

PR #1423, opened 2022-05-10
[docs] Change the order of the ActionIcon variants

Jessuhh:
Changed the order of the ActionIcon variants so that they match with the order in the demo


rtivital:
Thanks!

PR #1437, opened 2022-05-11
Use `style` instead of `sx` to set thumb position

### LINKED ISSUE #1436: Slider component can have high CPU usage
### What package has an issue

@mantine/core

### Describe the bug

If the value of the slider changes often, the CPU is intensively used. This is because emotion needs to put new styles on the DOM because of the changing position. Probably, a fairly simple solution would be to use `style` instead of `sx` in the Thumb component to set the new position.

Any

https://codesandbox.io/s/friendly-hooks-vg11ph?file=/src/App.tsx

Yes

Yes

### Possible fix

Use `style` instead of `sx` in the `Thumb` component

ramiel:
This should solve high CPU usage problems.

NOTE: there is one test failing, but it's independent of my changes, it's failing on main branch as well.

Closes #1436

### DIFF
--- a/src/mantine-core/src/components/Slider/Thumb/Thumb.tsx
+++ b/src/mantine-core/src/components/Slider/Thumb/Thumb.tsx
@@ -1,4 +1,4 @@
-import React, { useState, forwardRef, useEffect } from 'react';
+import React, { useState, forwardRef } from 'react';
 import { DefaultProps, MantineNumberSize, MantineColor, Selectors } from '@mantine/styles';
 import { Box } from '../../Box';
 import { Transition, MantineTransition } from '../../Transition';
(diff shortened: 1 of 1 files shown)


rtivital:
Does not work for RTL

https://user-images.githubusercontent.com/10353856/167854535-f80129f3-c111-4ac9-abf4-87db94ec73fb.mov

ramiel:
I'll keep working on it later today

ramiel:
Latest commit solves the problem and now it works on RTL as well. Though, I don't know if it's the right way of doing it and if this is the right approach to the problem. I don't know the system well enough. Maybe we can discuss it further?

rtivital:
Thanks!

Issue #1394, opened 2022-05-05
Array form validation is not reactive


macist-m:
### What package has an issue

@mantine/form

### Describe the bug

Hello, when creating forms with "use-form lists" validation of the inputs are not reactive, meaning if there is an error on empty input box, filling it out is not clearing out the error message, yet if I submit the form again it clears out....

This happens with 3rd party lib like yup or with plain validation, it does not matter in my case...

This is not the case with regular forms with no array of inputs.

Hopefully I was able to explain it :)

Thanks

### In which browser did the problem occur

chrome/safari

### Do you know how to fix the issue

No

rtivital:
I cannot reproduce this, can you send a sandbox?

macist-m:
> I cannot reproduce this, can you send a sandbox?

Sure -> https://codesandbox.io/s/heuristic-rui-c6pxm5

rtivital:
So the issue is that when field changes the error is not cleared, right?

macist-m:
> So the issue is that when field changes the error is not cleared, right?

Yup, exactly...

OlegGulevskyy:
Hello,
can I take this one? I am looking for a place to dip my toes in open source so that would be my first attempt :)

rtivital:
sure

rtivital:
Fixed in 4.2.5, thanks @OlegGulevskyy

macist-m:
@OlegGulevskyy Thanks a lot...

PR #1438, opened 2022-05-11
[@mantine/form] `use-form`: Use correct error paths to clear single input errors

OlegGulevskyy:
This PR addresses issue raised in #1394 
The reason why input errors are not clearing out is because errors are stored in the format of `form name + input field index + field name`. Which results in the following errors state, if we take the example raised in the issue (given 2 same inputs are created).
```
{
  employees.0.name: "Name should have at least 2 letters"
  employees.1.name: "Name should have at least 2 letters"
}
```
Whenever we pass field to the `clearFieldError` utility, we always pass just the `form name`, thus receiving `employees`, instead of full error path.

This PR uses full error path for each individual input.

### DIFF
--- a/src/mantine-form/src/use-form.ts
+++ b/src/mantine-form/src/use-form.ts
@@ -83,21 +83,26 @@ export function useForm<T extends { [key: string]: any }>({
       return clone;
     });
 
-  const setFieldValue = <K extends keyof T, V extends T[K]>(field: K, value: V) => {
+  const setFieldValue = <K extends keyof T, V extends T[K]>(
+    field: K,
+    value: V,
+    errorPath?: string
+  ) => {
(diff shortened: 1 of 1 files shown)


src/mantine-form/src/use-form.ts
```diff
@@ -83,21 +83,26 @@ export function useForm<T extends { [key: string]: any }>({
       return clone;
     });
 
-  const setFieldValue = <K extends keyof T, V extends T[K]>(field: K, value: V) => {
+  const setFieldValue = <K extends keyof T, V extends T[K]>(
+    field: K,
+    value: V,
+    errorPath?: string
```
rtivital, line 89: Why errorPath is optional?

rtivital:
Thanks!

PR #1445, opened 2022-05-12
[@mantine/core] NumberInput: Allows the "-" symbol in an empty NumberInput

BWsix:
Fixes the bug in NumberInput raised in the [discord bugs channel](https://discord.com/channels/854810300876062770/972850721244471436/973654286154203156)

> Is NumberInput not allowing the "-" symbol in an empty NumberInput expected behavior? Currently, you are unable to type "-" in an empty NumberInput unless you first put a number then go back and place the "-" before it

Changes:
- You can now type `-` in an empty NumberInput.
- If `-` is the only thing in a NumberInput, the value will remain `undefined`, and the NumberInput will be cleared when the `onBlur` event fires.


coolRoger:
@rtivital great job, and when are you going to release this bug fixed to npm?

rtivital:
Patches are usually released every Sunday

PR #1447, opened 2022-05-12
[@mantine/styles] `sx`: Add `undefined` for array

AzzouQ:
With strictMode, TS will give an error following this use case: https://mantine.dev/theming/sx/#sx-array

This PR add `undefined` to `sx` array's definition.


AzzouQ:
Forked from dev instead of master on #1446, fixed it, sorry

rtivital:
Thanks!

PR #1448, opened 2022-05-12
[@mantine/core] Box: Remove unnecessary types

AzzouQ:
Do some clean up by removing unnecessary types from Box component.

This PR is just clean up code.

### DIFF
--- a/src/mantine-core/src/components/Box/Box.tsx
+++ b/src/mantine-core/src/components/Box/Box.tsx
@@ -5,11 +5,9 @@ import {
   PolymorphicRef,
   extractSystemStyles,
 } from '@mantine/styles';
-import { useSx, BoxSx } from './use-sx/use-sx';
+import { useSx } from './use-sx/use-sx';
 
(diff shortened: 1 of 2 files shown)


src/mantine-core/src/components/Box/Box.tsx
```diff
@@ -5,11 +5,9 @@ import {
   PolymorphicRef,
   extractSystemStyles,
 } from '@mantine/styles';
-import { useSx, BoxSx } from './use-sx/use-sx';
+import { useSx } from './use-sx/use-sx';
 
-interface _BoxProps extends Omit<DefaultProps, 'sx'> {
-  sx?: BoxSx;
-}
+interface _BoxProps extends DefaultProps {}
```
AzzouQ, line 10: Be aware that using an empty object will raise [@typescript-eslint/no-empty-interface](https://typescript-eslint.io/rules/no-empty-interface). Using directly DefaultProps instead of _BoxProps could fix it, but ignoring that warning is fine and makes more sense to keep consistency with others component.

rtivital:
Thanks!

PR #1449, opened 2022-05-12
Allow to layout radios evenly with the grow prop

lukemt:
Not sure if this works, but this is what i would like to do.


rtivital:
That's not a good idea, it will produce more bugs, we will provide a better of radio group customizations in 5.0

lukemt:
Ok, that's fine

PR #1456, opened 2022-05-13
[@mantine/nprogress] withinPortal default to false

jerebtw:


rtivital:
nprogress, modal and drawer should remain in Portal by default –  this change is only for dropdowns to fix click outside issues

jerebtw:
Okay

PR #1461, opened 2022-05-14
[docs] Use Discord's new blurple for light mode

ElijahPepe:
#7289da is used on Mantine's documentation page for Discord's background color; the actual color is #5865f2 which is used on the dark mode version of the page.


rtivital:
Thanks!

PR #1462, opened 2022-05-14
Update FullScreenDropzone.tsx

dexter4life:
Notice that children is having a different signature in this case a function that returns React.ReactNode but in React.ComponentPropsWithoutRef<'div'>... is also having a field with that same name, this will result to a conflict so to avoid this, the "children" must not be imported from the upper parent.


dexter4life:
So I am requesting to review this and see what I mean. If using other typescript less than 4.6.3 you might not notice this as an error, but for 4.6.3 is an error or maybe depending on the eslint configuration in your workspace. Let me know if my request is not valid. Thanks

rtivital:
React 18 types errors will be fixed with 5.0 release

PR #1463, opened 2022-05-14
Update global-types.d.ts

dexter4life:
To avoid this Error "Type '{ children: Element; }' has no properties in common with type 'IntrinsicAttributes'.ts" providing a prop { children: React.ReactNode }


rtivital:
Where did this error occur? Component does not accept child

rtivital:
React 18 errors will be fixed with 5.0 release

Issue #1443, opened 2022-05-12
TimeInput produces Date with non-zero milliseconds


Stouffi:
### What package has an issue

@mantine/dates

### Describe the bug

TimeInput onChange will produce Date with non-zero milliseconds.
e.g. for an input `12:00` in an environment with an UTC offset to -120 minutes date.toIsoString() returns `"2022-05-12T10:00:00.532Z"` instead of expected `"2022-05-12T10:00:00.000Z"`.

### In which browser did the problem occur

Any

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/young-framework-mgi9yr?file=/src/App.tsx

zhulien-ivanov:
Problem seems to be here but I haven't tested: https://github.com/mantinedev/mantine/blob/158c9cb4033c066db3a0304e242270d60ba8bcbd/src/mantine-dates/src/components/TimeInputBase/get-date/get-date.ts#L26-L31.
I can test it and supply a fix tomorrow, @rtivital.

PR #1464, opened 2022-05-14
[@mantine/dates] TimeInput: Clear milliseconds on input value change.

zhulien-ivanov:
Fix for #1443.

Outside change through the `onChange` handler in controlled variant still preserves milliseconds.


rtivital:
Thanks!

PR #1465, opened 2022-05-15
[@mantine/nprogress] Add autoReset prop

jerebtw:
This PR makes resetting progress easier and more stable.
And also fixes a bug when withinPortal is false where the progress is centered and not on the left


rtivital:
Thanks!

PR #1485, opened 2022-05-19
Fix aria-owns attribute

isaul32:
Fixes [aria-valid-attr-value](https://dequeuniversity.com/rules/axe/4.4/aria-valid-attr-value) issue I found with [@axe-core/react](https://www.npmjs.com/package/@axe-core/react).


rtivital:
Thanks!

PR #1491, opened 2022-05-21
[@mantine/nprogress] Fix react 18 support

jerebtw:


rtivital:
@jerebtw can this be merged?

jerebtw:
@rtivital Yes 😄

rtivital:
Thanks!

carlesba:
Is this taking into consideration dependencies that have react as peer dependency?

I'm still getting this warning when using the last mantine-core version:

Screenshot 2022-06-03 at 13 14 30

This should could be solved upgrading the dependencies. Do you know if there's any issue that prevents the upgrade?

rtivital:
This change was never published on npm, react 18 support will be added with version 5.0

carlesba:
Thanks for the info!

rtivital:
It will be released when it is ready, we do not share public roadmap

PR #1494, opened 2022-05-21
[@mantine/core] MultiSelect/Select: create item mutation support

achmurali:


src/mantine-core/src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -315,6 +315,15 @@ export const MultiSelect = forwardRef<HTMLInputElement, MultiSelectProps>(
 
     const handleItemSelect = (item: SelectItem) => {
       clearSearchOnChange && handleSearchChange('');
+
+      if (item.creatable) {
+        if (typeof onCreate === 'function') {
+          const newItem = onCreate(item.value);
+          // eslint-disable-next-line no-param-reassign
+          if (newItem) item = { ...newItem };
```
rtivital, line 323: Please do not disable eslint rules, it make code hard to follow

achmurali:
user from now on has to listen to `onCreate` and append the newly created item to the `data` array.

rtivital:
I've implemented it differently in 5.0 (appears to be the best way to fix it was with a breaking change), thanks for your work anyway!

Implementation – https://github.com/mantinedev/mantine/blob/v5/src/mantine-core/src/components/MultiSelect/MultiSelect.story.tsx#L64:L89

Issue #782, opened 2022-01-31
[@mantine/hooks] Interval Hook start() creates unbounded intervals


melloware:
### What package has an issue

@mantine/hooks

### Describe the bug

If you call `interval.start()` X times it starts X timers but calling stop can only stop the latest one.

### In which browser did the problem occur

ALL

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/sharp-matan-zx040?file=/src/App.tsx

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

PR being submitted with...

```typescript
  const start = () => {
    if (!active) {
       setActive(true);
       intervalRef.current = window.setInterval(fn, interval);
    }
  };
```

melloware:
PR was merged

prograpp:
I think this does not completly solve the issue. When calling the `start()` function multiple times in a short time, the `active` state won't be changed in-between due to reacts set state caching.

If I am correct, the following would solve that issue:

```typescript
const start = () => {
    setActive(old => {
        if (!old) {
            intervalRef.current = window.setInterval(fn, interval);
        }
        return true;
    });
};
```

melloware:
I agree you want to submit a PR?

prograpp:
Yes, did so

melloware:
@prograpp I still found a bug with this where I am seeing it hold two instances of my interval timer that was not happening before your fix.

I am going to submit another fix.

PR #1498, opened 2022-05-22
[@mantine/hooks] use-interval Fix mantinedev#782: No multiple interva…

prograpp:
…l starts if called several times

See issue #782 for the reasons of this pull request.


rtivital:
Thanks!

PR #1523, opened 2022-05-27
[docs] Fixed typo in cards page

WhyAsh5114:


rtivital:
The fix is incorrect

WhyAsh5114:
`It component is a first child in Card then it has negative top, left and right margins?`
I thought it should be It's?

WhyAsh5114:
oh my bad, it should be if, ill make the changes

rtivital:
Add more commits in this

WhyAsh5114:
done, I hope its correct now

PR #1526, opened 2022-05-27
[@mantine/form] - Export FormList type for explicit TypeScript users

iDavidB:
Exports FormList type so TS users can use it in their explicit form values definition without breaking `form.getListInputProps`.

Example of the issue:
https://codesandbox.io/s/eloquent-lake-1cb038?file=/src/Registeration.tsx


rtivital:
Thanks!

PR #1543, opened 2022-05-30
Add mime type .kml

na0495:
Mime type for .kml files


rtivital:
We cannot add all existing mime types, library contains only the most used, and I've never seen kml files before

PR #1552, opened 2022-06-02
added multiples mimes types

JigolKa:
Added mimes types


rtivital:
Thanks for PR, we do not need all those extra type, package exports only the most used mime types.

PR #1556, opened 2022-06-02
[@mantine/hooks] useTimeout: init

achmurali:


docs/src/docs/hooks/use-timeout.mdx
```diff
@@ -0,0 +1,50 @@
+---
+group: 'mantine-hooks'
+package: '@mantine/hooks'
+category: 'state'
+title: 'use-timeout'
+order: 1
+slug: /hooks/use-timeout/
+description: 'Wrapper around window.setTimeout'
```
rtivital, line 8: "Calls function at given timeout"

docs/src/docs/hooks/use-timeout.mdx
```diff
@@ -0,0 +1,50 @@
...
+
+```tsx
+const { start, clear } = useTimeout(fn, delay, { autoInvoke: true });
+```
+
+Arguments:
+
+- `fn` – function that will be called after the timer expires
+- `interval` – amount of milliseconds the timer should wait before the specified function is executed.
+- `options: { autoInvoke }` - If the timer should be invoked on component mount, defaults to true.
```
rtivital, line 30: autoInvoke should be false by default

rtivital:
Thanks!

Issue #1582, opened 2022-06-07
Switch on and off label not working correctly


tahseenio:
### What package has an issue

@mantine/core

### Describe the bug

When the Switch component is toggled on and off, the onLabel and offLabel text are both just staying on the right and not showing on the left when on.

### In which browser did the problem occur

Chrome

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/shy-surf-k59rsl?file=/src/App.js

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

No

tahseenio:
@rtivital This can be closed now? Thanks for the fix @SmitharyZach

PR #1588, opened 2022-06-08
[@mantine/core] Switch: fix on / off label

SmitharyZach:
Fix issue: https://github.com/mantinedev/mantine/issues/1582

The on label for the switch had incorrect styling resulting in the label not being visible. To fix this I just copied the styles from the off label which was showing correctly. I added some stories as well

image

image


rtivital:
Thanks!

PR #1589, opened 2022-06-08
[docs] Avoid double render from ColorSchemeProvider

justjavac:
fix 38a4067779951ef7bed72750167a806d5d8106f2


rtivital:
This change will break the docs website

PR #1599, opened 2022-06-10
[@mantine/dates] TimeInput: Add props for "am"/"pm" labels

### LINKED ISSUE #1496: am/pm cannot be localized for TimeInput
### What package has an issue

@mantine/dates

### Describe the bug

The `TimeInput` component is hard/coded to use am/pm for 12-hour clocks, and these values are not configurable for different locales.

all

_No response_

Yes

Yes

### Possible fix

At the very least it would be necessary to allow providing these values as props to the input so the localization can be provided from any external source. The more ideal solution would likely be a `TimePicker` component (similar to `DatePicker`) that uses dayjs locales to specify the clock type and the correct am/pm suffix.

Martin005:
Fixes #1496

Added props for `TimeInput` component which allow user to pass custom "am"/"pm" labels. As dayjs doesn't provide these translations, I turned to this solution.

### DIFF
--- a/src/mantine-core/src/components/Drawer/Drawer.tsx
+++ b/src/mantine-core/src/components/Drawer/Drawer.tsx
@@ -190,7 +190,7 @@ export function Drawer(props: DrawerProps) {
     return undefined;
   }, [trapFocus]);
 
-  useFocusReturn({ opened, transitionDuration: 0 });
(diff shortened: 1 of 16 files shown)


rtivital:
Thanks!

Issue #1600, opened 2022-06-10
Chrome browser is allowed to automatically suggest search params for the Autocomplete


victorlrz:
### What package has an issue

@mantine/core

### Describe the bug

![image](https://user-images.githubusercontent.com/60266782/173084059-4ff4d991-45e6-48e6-9a29-c98f79d4f56d.png)

When I use autocomplete on Chrome (V102.0.5005.61), after several searches, the browser suggests my old search suggestions which makes using the component really difficult. Threads already exist on the subject like the one https://stackoverflow.com/questions/12374442/chrome-ignores-autocomplete-off but I have the impression that the attribute is now no longer ignored and that they have pushed a fix on their side as shown in the picture below which prevents the browser suggestions.

![image](https://user-images.githubusercontent.com/60266782/173084232-a621373c-e19c-4b60-8156-63cdaf2ac099.png)

### In which browser did the problem occur

Chrome

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

Set the autoComplete attribute value to "off" : (Commit 1)
- So the browser is not permitted to automatically enter or select a value for this field. 

Set the type attribute to "search" : (Commit 2) : 
- Since we enter search strings on this input I think we could update it's type to "[Search](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Input)"

PR #1601, opened 2022-06-10
[@mantine/core] Autocomplete: Update input autoComplete and type attributes

victorlrz:
Fixes https://github.com/mantinedev/mantine/issues/1600

Added autocomplete="off" instead of "nope" and modified input type to "search" since it's a search field


rtivital:
Thanks!

PR #1621, opened 2022-06-13
[@mantine/core]: Changing the Group types to allow 'row-reverse' and 'column-reverse'

Abys5:
Not much need for an explanation it is a small change.

Stop the Type checking issues when using 'row-reverse' and 'column-reverse', the visual aspect work just fine so only needed to amend the types.


rtivital:
Group direction will be removed in 5.0

PR #1642, opened 2022-06-16
[docs] Dropzone: strict mode adjustment

MikevPeeren:
Adding null to useRef so that TypeScript does not throw an error when using this example inside a TS file.


rtivital:
Thanks!

PR #1652, opened 2022-06-19
[@mantine/core] NumberInput: Fix controls when min/max is set

medihack:
Fixes an issue that a control button gets disabled when a set min or max is reached and no mouse down event is triggered then. There are other ways to fix this (like returning the next value from increment/decrement and comparing to min/max), but this seems to me the most "React" way.


rtivital:
Thanks for PR and investigation, I've implemented it with pointer events instead – it is more clear – https://github.com/mantinedev/mantine/commit/7537adaf9b39abab74c828248e3be516a999cbed

PR #1682, opened 2022-06-25
[NumberInput] Fix late focus when using controls

medihack:
NumberInput does not seem to follow the correct order of the onChange events when using the controls. The blur of other inputs is triggered too late as the focus event is triggered after onChange. This could be resolved by triggering the focus earlier.


rtivital:
Please do not create PRs towards `v5` branch, use `master` branch as target for changes that should be published as a patch.

medihack:
"@mantine/dropzone/Dropzone › exposes static components" seems to fail also without my change in v5 branch.

medihack:
> Please do not create PRs towards `v5` branch, use `master` branch as target for changes that should be published as a patch.

Sure, I create a new PR then (and I guess this will then be merged in V5, too).

PR #1683, opened 2022-06-25
[@mantine/core] NumberInput: Fix late focus when using controls

medihack:
NumberInput does not seem to follow the correct order of the onChange events when using the controls. The blur of other inputs is triggered too late as the focus event is triggered after onChange. This could be resolved by triggering the focus earlier.


rtivital:
Thanks!

Issue #1686, opened 2022-06-26
Transition duration not correctly applied


Maetes:
### What package has an issue

@mantine/core

### Describe the bug

This is a react 18 issue. I think the problem is that the Transition component within the useTransition.ts is using the useDidUpdate hook. Within React 18 the strict mode now executes useeffect twice so the hook is more or less useless? Since because of the second loop it is deemed mounted but is not actually.

This cause several problems like the one i faced. The transitions are not correctly applied to the dropdown of the Select component. On top of that, there is a side effect issue within the useTransition.ts causing a duration missmatch between the variables _duration and transitionDuration.

I found a workaround to fix the sideeffect issue. Nevertheless I think the useDidUpdate hook is also worth to rethink. I'll create a pull request updating the useTransitions.ts to apply the transitions correctly in dev mode under react 18.

If I still miss something I'm happy to discuss

### In which browser did the problem occur

any

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

rtivital:
Fixed in 5.0

PR #1687, opened 2022-06-26
[@mantine/core] Fix useTransition.ts not correctly emit duration

Maetes:
Hi, this is my suggestion to fix the transition duration (possible temporary until the hook is also fixed) application in dev mode. It more or less is just a deletion of the useState which causes a side effect in the useDidUpdate hook where the transition duration is not correctly returned (probably cause the hook fired twice in the new React 18 strict mode). Please refer to the issue I created for a description of the faced error. #1686 

Happy to discuss

### DIFF
--- a/src/mantine-core/src/components/Transition/use-transition.ts
+++ b/src/mantine-core/src/components/Transition/use-transition.ts
@@ -34,7 +34,7 @@ export function useTransition({
   const theme = useMantineTheme();
   const reduceMotion = useReducedMotion();
   const [transitionStatus, setStatus] = useState<TransitionStatus>(mounted ? 'entered' : 'exited');
-  const [transitionDuration, setTransitionDuration] = useState(reduceMotion ? 0 : duration);
+  let transitionDuration = reduceMotion ? 0 : duration;
   const timeoutRef = useRef<number>(-1);
 
   const handleStateChange = (shouldMount: boolean) => {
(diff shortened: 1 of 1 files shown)


rtivital:
Does it work fine without strict mode?

Maetes:
Yes it does

rtivital:
I do not see any reason to fix issues with strict mode – you can turn that off and not to have any issues

Maetes:
ohh sorry I misunderstood your question. I though you mean whether the possible solutions works with strict mode off...
The issue persist regardless whether i turn of the strict mode or not. The only thing is that when switching off strict mode is, that the first dropdown is correctly rendered with transition. All after are redered without transition, so the error persist.

As I tried to explain in the referenced issue: I think the whole error comes from the hook use-did-update, because the React 18 fires 2 times the useeffect, hence it is not ensured anymore that on the first render the useeffect within use-did-update is not executed and that cause the line 
`46: const _duration = reduceMotion ? 0 : shouldMount ? duration : exitDuration;` 
`47: setTransitionDuration(_duration);`

within useTransition.ts to generate a deviant state, which result in the error in the application of the transition styles.

The problem is I acutally struggle with fixing useeffect within the hook. So actually the workaround in this request works and is good to go for now i think.

rtivital:
Thanks!

PR #1688, opened 2022-06-26
[@mantine/ds] Fix wrong dev dependency version

medihack:
`yarn test` fails because of some wrong dev dependency. This fixes it.


medihack:
Seems to be solved

PR #1689, opened 2022-06-26
[@mantine/core] NumberInput: Only fire onChange when value really cha…

medihack:
I am not completely sure if it would be better to compare against `finalValue`, but maybe it doesn't matter.


medihack:
Should I add some documentation to the `NumberInput.mdx` as well or is the documentation in `Component props` enough? If yes, then to which section? [Min, max and step](https://v5.mantine.dev/core/number-input/#min-max-and-step)? Should I add a test case, or is it too trivial?

rtivital:
Documentation in component props is enough

medihack:
Can I do something to improve this PR or just haven't had the time to look into it yet?

rtivital:
Nope, all fine, I usually merge PRs on Sundays, first for patch, then everything else.

medihack:
Cool, good to know. Thanks!

rtivital:
Thanks!

PR #1714, opened 2022-07-02
[@mantine/next] Improve next link

u3u:
## Improve next link

- The `href` prop should not be passed to the `a` element and should be handled automatically by `next/link` to fix the [`basePath`](https://nextjs.org/docs/api-reference/next.config.js/basepath#links)
- Respect `next/link` all props
- Add `displayName` as `@mantine/next/NextLink`
- Upgrade `next` dependence


rtivital:
Thanks!

PR #1725, opened 2022-07-03
[@mantine/core] NumberInput: Fixes repeating keys when step on hold is used

medihack:
As mentioned on [Discord](https://discord.com/channels/854810300876062770/972850721244471436/993098144067899472). Fixes that NumberField does not respect repeating keys (by holding them long) when "stop on hold" is used.


rtivital:
Thanks!

PR #1733, opened 2022-07-04
[@mantine/styles] fixed an issue with infinite cache recreation

Andarist:


src/mantine-styles/src/tss/use-emotion-cache.ts
```diff
@@ -12,7 +12,10 @@ export const { getCache } = (() => {
 
   function _getCache(options?: Options) {
     if (cache === undefined || _key !== options?.key) {
-      _key = options?.key || 'mantine';
+      if (process.env.NODE_ENV === 'development' && options && !options.key) {
+        console.error('`key` is required for `emotionOptions`');
+      }
+      _key = options?.key;
```
Andarist, line 18: I was investigating an issue reported to Emotion. I had no idea what are your actual requirements for the Emotion options, I've assumed that they would be merged with default so I didn't include the `key`. This resulted in a lot of styled being generated on a site - because styles were never flushed but each new cache was creating new ones (duplicates).

The new cache was always created because I never gave one as part of `options` and you were defaulting your closure variable to `'mantine'` - so my `undefined` never could match it and thus the "cycle" continued.

rtivital:
Thanks for explanation

PR #1750, opened 2022-07-07
[@mantine/hooks] use-hotkeys: Fix contenteditable event handling

### LINKED ISSUE #1749: Text input within RichTextEditor triggers global hotkeys
### What package has an issue

@mantine/rte

## Issue: Characters typed within RichTextEditor might trigger global hotkeys.

[CodeSandbox](https://codesandbox.io/s/happy-wu-qxy6vo?file=/src/App.tsx)

### Use case
In my app I have a global hotkey `'m'` to show or hide a navigation menu.
But whenever I use the `RichTextEditor`, typing the letter `m` triggers the hotkey.

### Expected Behavior
My expectation is that `RichTextEditor` would behave similarly to typical text input fields like `<input>` or mantine's `<TextInput>` and `<Textarea>`:
- all keyboard events are captured by the text input field
- `event.stopPropagation()` is called to prevent bubbling up the DOM

Chrome, safari, firefox

https://codesandbox.io/s/happy-wu-qxy6vo?file=/src/App.tsx

Yes

Yes

### Fix/Approach
The issue is that Quill's rich text editor is not a typical html input-style element, it's a `<div>` with the `contenteditable="true"` attribute set.

The conditional in [`mantine/use-hotkeys.ts`](https://github.com/mantinedev/mantine/blob/master/src/mantine-hooks/src/use-hotkeys/use-hotkeys.ts#L10) is not testing for this, so one approach is to simply update the conditional

```typescript
function shouldFireEvent(event: KeyboardEvent) {
  if (event.target instanceof HTMLElement) {
    return (
      !event.target.isContentEditable &&  // <-- quill uses editable div
      !["INPUT", "TEXTAREA", "SELECT"].includes(event.target.tagName)
    );
  }
  return true;
}
```

Reference: [MDN docs for `HTMLElement.isContentEditable`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/isContentEditable)

marcelchastain:
Fixes #1749

### Testing
Tested manually by adding a global hotkey to one of the stories in `RichTextEditor.story.tsx`.

```typescript
// RichTextEditor.story.tsx
export function Usage() {
  const [value, onChange] = useState(html);
  useHotkeys([['shift + b', () => console.log('Triggered')]]);
  return (
    <div style={{ padding: 40, maxWidth: 800, margin: 'auto' }}>
      <RichTextEditor
        value={value}
        onChange={onChange}
        onImageUpload={handleImageUpload}
        stickyOffset={0}
      />
    </div>
  );
}
```

_Note: I'm happy to write unit/integration tests for this if needed!_

### Results
- _Without_ this fix, the hotkey fired while typing within the `RichTextEditor`
- _With_ the fix, the hotkey did not fire, indicating that the bug was fixed
- In addition, all unit tests passed

### DIFF
--- a/src/mantine-hooks/src/use-hotkeys/use-hotkeys.ts
+++ b/src/mantine-hooks/src/use-hotkeys/use-hotkeys.ts
@@ -7,7 +7,10 @@ export type HotkeyItem = [string, (event: KeyboardEvent) => void];
 
 function shouldFireEvent(event: KeyboardEvent) {
   if (event.target instanceof HTMLElement) {
-    return !['INPUT', 'TEXTAREA', 'SELECT'].includes(event.target.tagName);
+    return (
+      !event.target.isContentEditable &&
+      !['INPUT', 'TEXTAREA', 'SELECT'].includes(event.target.tagName)
+    );
   }
   return true;
 }


rtivital:
Thanks!

PR #1755, opened 2022-07-09
Changed dependency array for use-debounced-value

Prasiddha22:
[use-debounced-value] - Added dependency array


rtivital:
Thanks for reporting the issue, it is already fixed in v5.

PR #1756, opened 2022-07-09
Feature/hooks/use callback ref

Prasiddha22:
[use-callback-ref]: Added new hook useCallbackRef. This helps to create a ref for the incoming props whose references change every time the app re-renders. This can be achieved by useRef but useRef takes the initial value and it does not change. But this hook helps to create a ref which values changes on prop value changes.


rtivital:
Hi @Prasiddha22, I do not really see much use case for this hook, so I cannot merge this PR. Please see contributing guide – https://mantine.dev/pages/contributing/ – it is required to discuss features with maintainers before submitting a PR to avoid working on PRs that won't be merged.

PR #1762, opened 2022-07-11
[docs] `setCookies` being deprecated

mrfambo:
`setCookies` method is being deprecated in favour of `setCookie` as discussed in PR https://github.com/andreizanik/cookies-next/pull/21

It is giving console warnings, we should better update our code with `setCookie`. It works as expected.


rtivital:
Thanks for reporting, the issue will be resolved with 5.0 release.

PR #1776, opened 2022-07-12
[@mantine/core/select] Select: passed name props to component

sevgeek:
Explicit transfer of the `props.name` property to the `Input` component, for correct processing of the `onBlur` event.

https://github.com/mantinedev/mantine/discussions/1767


rtivital:
Select already has a hidden input that handles uncontrolled component value. Adding name attribute to the actual input will make uncontrolled forms to receive incorrect value.

PR #1796, opened 2022-07-15
[@mantine/core] Timeline: Allow manually overriding of active state

Zegnat:
I was surprised that I was able to set `active={true}` (or just `active`) on a `Timeline.Item` to turn on the active presentation of an item, but if it was covered by `Timeline`’s `active` I was unable to set `active={false}` to turn off the active presentation.

Turns out falsy values for `active` were being ignored and always turned the element over to `Timeline`’s control. I propose to only let `Timeline` handle active state in case of `undefined`.

I am not too happy with the test, but it seems to confirm my fix. Not sure how deep Mantine usually goes with tests. Let me know if there is anything there you would want me to tweak!

Thanks for the awesome UI project ❤️

### DIFF
--- a/src/mantine-core/src/components/Timeline/Timeline.test.tsx
+++ b/src/mantine-core/src/components/Timeline/Timeline.test.tsx
@@ -14,6 +14,17 @@ const defaultProps = {
   ],
 };
 
+const manualActiveStates = {
+  children: [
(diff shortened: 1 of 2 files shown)


rtivital:
Hi @Zegnat, sorry for long response, can you please resolve the conflicts?

PR #1803, opened 2022-07-17
[@mantine/styles]: Fix radius function type, to be optional

BenJeau:
Made the `size` parameter optional, since the function already defaults to the `defaultRadius` from line 9-12.


rtivital:
Thanks for a PR, unfortunately that will not fix the type issue, I'll make changes on my side to make it work, `theme.fn.radius` argument will be optional.

Issue #1784, opened 2022-07-13
Error rendering Calendar when first month displays only disabled dates


Stouffi:
### What package has an issue

@mantine/dates

### Describe the bug

Note that downgrading to 4.2.11 makes the error not reproducible.

PR #1704 seems to have introduced the bug unfortunately.

### In which browser did the problem occur

Any

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/calendar-with-first-month-excluded-x9c9bu?file=/src/App.tsx

### Do you know how to fix the issue

Yes

### Possible fix

A quick fix could be to revert #1704

I noticed also that the code base is not using the strict compilation mode of TypeScript. Such issue would have been caught by the compiler due to the implied noImplicitAny parameter.

khobiziilyes:
So far, I discovered that the error happens when all the days in the month component are excluded, the commit you mentioned is not the reason the bug happened, the real reason is that [this line](https://github.com/mantinedev/mantine/blob/master/src/mantine-dates/src/components/Month/Month.tsx#L189) returns undefined (since no day is returned) and that is passed to another function [(here)](https://github.com/mantinedev/mantine/blob/master/src/mantine-dates/src/components/Month/Month.tsx#L274) ...

rtivital:
Fixed in 5.0

PR #1813, opened 2022-07-18
[@mantine/dates] Calendar: Fix bug when all days are excluded:

khobiziilyes:
Bare minimum to re-produce:

```
import React, { useState } from 'react';
import { Calendar } from '@mantine/dates';

function Demo() {
  const [value, setValue] = useState(null);
  
  return (<Calendar
    value={value}
    onChange={setValue}
    excludeDate={() => true}
  />);
}
```

The bug was caused because of a loop to find the first non-excluded date and then used in another function call.
The loop couldn't find a proper day so it was returning undefined, which caused the bug.

Fix for: #1784


rtivital:
Thanks!

Issue #1795, opened 2022-07-15
TimeInput field disabled state background is different color


hammo92:
### What package has an issue

@mantine/dates

### Describe the bug

When set to disabled the background of the TimeInput component is a different color to other disabled input components.

### In which browser did the problem occur

Firefox, Chrome

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/lively-flower-7skwwb?file=/src/App.js

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

Opacity for TimeInput is being set on Wrapper and Component so opacity is half the expected 0.6 value. Remove opacity from the TimeInput wrapper

khobiziilyes:
Did you fill a PR or should I do it instead?

hammo92:
Apologies, I've been away. I'll fill a PR

rtivital:
Fixed in 5.0

PR #1835, opened 2022-07-22
[@mantine/dates]: TimeInput: Fix opacity applied twice when disabled

hammo92:
Disabled time input had opacity set on both the component and wrapper, causing the final opacity to be incorrect.

fix for [#1795](https://github.com/mantinedev/mantine/issues/1795)


rtivital:
Thanks!

Issue #907, opened 2022-02-26
Select component is clearable when disabled


wes337:
### What package has an issue

@mantine/core

### Describe the bug

When the Select component has `clearable` and is `disabled`, you can still clear the input.

### In which browser did the problem occur

Chrome

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/wonderful-ishizaka-pz11wz?file=/src/App.js

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

The button to clear the input should also be disabled if the Select component is disabled.

rtivital:
Feel, free to create a PR, note that this issue most likely is also in MultiSelect component

wes337:
Sure. Checking the MultiSelect component and it doesn't even seem to render the clear button if disabled:

![image](https://user-images.githubusercontent.com/9464690/155848508-63079bbf-4df8-45aa-9e34-7274da493c67.png)

Seems like a separate but related issue. I can fix it in the same PR though.

PR #1843, opened 2022-07-24
[@mantine/core] Select: add missing 'disabled' prop to SelectRightSection

antoinesion:
Will now pass `disabled` prop to `SelectRightSection` component preventing a bug when `clearable` and `disabled` are `true`, the clear icon is visible and works when it should not.

This correctly fix https://github.com/mantinedev/mantine/issues/907.

However I agree with @wes337, I think the clear icon should still be visible but simply disabled when the `Select` component is, like the chevrons that are always visible. (cf. https://github.com/mantinedev/mantine/pull/908#discussion_r815323272)


antoinesion:
Or maybe we should leave the chevrons displayed in this case instead of not displaying anything at all, what do you think?

rtivital:
Thanks!

Issue #1844, opened 2022-07-24
DatePicker component seems to behave incorrectly when using localization


esistgut:
### What package has an issue

@mantine/dates

### Describe the bug

I'm having a little bit of trouble with the `DatePicker` component when using localization: if you select a date for a split second the first character of the month is uppercase and then it switches back to lowercase (it is very fast). Furthermore it doesn't seem to use any format of the locale I selected (https://github.com/iamkun/dayjs/blob/dev/src/locale/it.js) but it is translating month names correctly.

### In which browser did the problem occur

Firefox 102.0.1

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/goofy-browser-ducwiy?file=/src/App.tsx

### Possible fix

I worked around both problems this way https://codesandbox.io/s/ancient-glade-0ezu2h?file=/src/App.tsx
I dynamically load dayjs locales so I can read the formats and I apply them using `dateFormat` on the `MantineProvider`. I'm not even sure if `dateFormat` is a documented property as I got it from source code and couldn't find it on the documentation.

ercgrat:
Hi @esistgut, I opened a PR that fixes the capitalization glitch you noticed.

Otherwise, there is also an `inputFormat` prop on the `DatePicker` component that you can used. It also happens not to be documented. The locale-specific formats will only be recognized if you extend dayjs with the `localizedFormat` plugin. I made a CodePen that shows this: https://codesandbox.io/s/eager-shirley-7srfg2?file=/src/App.tsx

@rtivital would you mind if I open a PR adding documentation for the `inputFormat` prop with this context?

esistgut:
@ercgrat thank you for the fast answer. Using `dayjs.extend(localizedFormat);` and passing "LL" to `dateFormat` on MantineProvider or `inputFormat` on DatePicker itself sure seems a more elegant way to do what I was doing.
Still everything seems to be ignore when enabling `allowFreeInput` as it always seems to revert to the "MM/DD/YYYY" format. I could pass a `dateParser` custom function but I have a feeling that this is supposed to work with dayjs automatically.
Am I missing something?

ercgrat:
That's a great question, yes you are missing ANOTHER dayjs plugin `customParseFormat`... then it will work only if you type the format exactly to match.

Another code sandbox: https://codesandbox.io/s/nice-tess-6dhgti?file=/src/App.tsx

More docs: https://day.js.org/docs/en/parse/string-format

[...]

PR #1845, opened 2022-07-24
[@mantine/dates] Fix capitalization of date input state

ercgrat:
Fixes the capitalization issue in #1844 - just one place appeared to be missing the `upperFirst` call.


ercgrat:
@rtivital hope you can help me be a better contributor. I was seeing many tests fail running the test script (`yarn test`), even just on `DatePicker.tsx`, with or without my changes. Is there some sort of scaffolding I need to set up to get the tests to pass? For example, the `sets label on DatePickerBase based on inputFormat prop` fails because the default props aren't supplied correctly. The test doesn't include the MantineProvider, so it makes sense to me why it's failing but I assume you are doing some other step to supply it?

rtivital:
Thanks!

PR #1849, opened 2022-07-25
Remove Appshell demo

PringlePot:
This probably isnt the best way to update it 😂


rtivital:
Not a good idea

PringlePot:
Still not fixed tho

Issue #1717, opened 2022-07-02
Invalid background-image property in Button


7iomka:
### What package has an issue

@mantine/core

### Describe the bug

In generated styles for button I see wrong value for background-image
```css
background-color: #228be6;
background-image: #228be6;
```

### In which browser did the problem occur

Chrome

### Possible fix

set background-image: none, or not set this prop at all.

rtivital:
Thanks for reporting

rtivital:
I do not see background-image styles in devtools 
<img width="617" alt="Снимок экрана 2022-07-02 в 19 57 35" src="https://user-images.githubusercontent.com/10353856/177009547-0e36bfcf-1e01-43b7-8284-9785ef12c291.png">

7iomka:
@rtivital is for variant="outline"

<img width="1920" alt="image" src="https://user-images.githubusercontent.com/18473137/177010794-7a4e08f2-24b9-40f1-8ee1-7a2db1ff1dec.png">

7iomka:
Also I would think about consecrating the issue with the ability for the user to control `autoprefixer` settings, because, for example `-webkit-text-decoration` no more needed long time ago.
As a good practice is to consider user `browserslist` key in your `package.json`, how do this next.js
https://nextjs.org/docs/advanced-features/customizing-postcss-config#customizing-target-browsers

rtivital:
I cannot find background-image styles on outline button, can you create a sandbox that reproduces the issue?
<img width="623" alt="Снимок экрана 2022-07-02 в 22 20 51" src="https://user-images.githubusercontent.com/10353856/177013799-0ba7e69b-fdd7-47d1-b112-16cc1b9e294c.png">

7iomka:
@rtivital 
https://codesandbox.io/s/happy-breeze-jl6kcv?file=/src/App.tsx

rtivital:
Fixed in 5.0.2

rtivital:
Will be fixed in 5.0.3, PR is merged

PR #1850, opened 2022-07-25
[@mantine/core] fix: Invalid background-image property in Button

lawff:
#1717 , when a variant is not 'gradient', maybe don't need the 'background-image', which now returns invalid value 'transparent'.
image


rtivital:
Thanks!

PR #1873, opened 2022-07-26
[@mantine/demos]: Corrected Typo error on Docs of Form Validation

Eliasmgprado:
Typo error on initialValues of useForm() function. Both should be the same value.


Eliasmgprado:
sorry, i realized its correct

PR #1885, opened 2022-07-27
improve types

brunocrosier:
- create a named type for `Record<string, ThemeComponent>`
- remove unnecessary code (`fn` was being `Omit`ted twice)
- refactor to use `Partial<Pick<>`

### DIFF
--- a/src/mantine-styles/src/theme/types/MantineTheme.ts
+++ b/src/mantine-styles/src/theme/types/MantineTheme.ts
@@ -9,6 +9,7 @@ import type { CSSObject } from '../../tss';
 
 export type LoaderType = 'bars' | 'oval' | 'dots';
 export type MantineThemeOther = Record<string, any>;
+export type MantineThemeComponents = Record<string, ThemeComponent>;
 
 export interface HeadingStyle {
   fontSize: CSSProperties['fontSize'];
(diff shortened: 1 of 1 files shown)


src/mantine-styles/src/theme/types/MantineTheme.ts
```diff
@@ -98,9 +99,6 @@ interface ThemeComponent {
 }
 
 export type MantineThemeBase = Omit<MantineTheme, 'fn'>;
-export type MantineThemeOverride = DeepPartial<
-  Omit<MantineThemeBase, 'fn' | 'other' | 'components'>
```
brunocrosier, line 102: we don't need to omit `fn` as we're already doing that in `MantineThemeBase`

src/mantine-styles/src/theme/types/MantineTheme.ts
```diff
@@ -98,9 +99,6 @@ interface ThemeComponent {
...
-> & {
-  other?: MantineThemeOther;
-  components?: Record<string, ThemeComponent>;
-};
+
+export type MantineThemeOverride = DeepPartial<Omit<MantineThemeBase, 'other' | 'components'>> &
+  Partial<Pick<MantineThemeBase, 'other' | 'components'>>;
```
brunocrosier, line 104: better and safer to `Pick` directly from `MantineThemeBase`, otherwise the types could fall out of sync

rtivital:
Thanks!

PR #1891, opened 2022-07-27
[@mantine/prism] no translate

ryuhangyeong:
## How

#1872 

- In the code area of ​​the official documentation, I think it would be better for the translator not to translate.

## Work History

- add `translate="no"` in `Prism` and `PrismTabs`

  image

  In my opinion the tab area should also not be translated. Let me know if you have any other comments!

## Etc

Thanks for the hard work of the mantine development team!


rtivital:
Thanks!

PR #1892, opened 2022-07-27
Avoid manual type annotation for useToggle

yudai-nkt:
Fix #1876.

With this PR, users don't need a type parameter for `useToggle` anymore.

```ts
import { useToggle } from "@mantine/hooks";

const [value] = useToggle(["light", "dark"] as const);  // `value` is inferred as "light" | "dark"
```


rtivital:
Thanks!

PR #1893, opened 2022-07-27
[@mantine/hooks] use-mouse: Add resetOnExit option

Siris01:
This PR adds a `resetOnExit` option (false by default for backwards compatibility) to the `useMouse()` hook. When true, the values of x & y will be 0 when the mouse leaves the specified element

### DIFF
--- a/src/mantine-hooks/src/use-mouse/use-mouse.ts
+++ b/src/mantine-hooks/src/use-mouse/use-mouse.ts
@@ -1,7 +1,9 @@
 import { useEffect, useRef, useState } from 'react';
 import type { MouseEvent } from 'react';
 
-export function useMouse<T extends HTMLElement = any>(resetOnExit: boolean = false) {
+export function useMouse<T extends HTMLElement = any>(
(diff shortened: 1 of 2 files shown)


src/mantine-hooks/src/use-mouse/use-mouse.ts
```diff
@@ -1,7 +1,7 @@
 import { useEffect, useRef, useState } from 'react';
 import type { MouseEvent } from 'react';
 
-export function useMouse<T extends HTMLElement = any>() {
+export function useMouse<T extends HTMLElement = any>(resetOnExit: boolean = false) {
```
rtivital, line 4: options should be an object: `options: { resetOnExit?: boolean } = { resetOnExit: false }`

Siris01: Done

rtivital:
Please also update documentation – https://mantine.dev/hooks/use-mouse/#definition

rtivital:
Thanks!

PR #1894, opened 2022-07-27
[@mantine/demos] Stepper - Corrected example of using styles API

Eliasmgprado:
Replaced stepProgress and stepCompleted styles by data-progress and data-completed props.


rtivital:
Thanks!

PR #1910, opened 2022-07-28
[docs & core] refactor some conditional logics

SayefReyadh:
Refactored some conditional logics


rtivital:
I'd rather not do it as these changes are not required, and they may lead to unexpected errors

PR #1923, opened 2022-07-29
[docs] Fix useToggle example

hu-zhi-hang:


rtivital:
`use-toggle` hook documentation is correct, the PR is invalid

PR #1930, opened 2022-07-30
[@mantine/core] fixed typo `content` attribute to Notification and Switch

ryuhangyeong:
## Work History

- fixed typo
  It has been modified for consistency with the value of the existing content property.


rtivital:
Thanks!

PR #1931, opened 2022-07-30
[@mantine/core] Divider: border style shorthand property

ryuhangyeong:
## Work History

- `border` style shorthand property


rtivital:
I prefer the previous variant – it is much easier to understand

PR #1933, opened 2022-07-30
[@mantine/core] Divider: Add `role="separator"` attribute

ryuhangyeong:
## Work History

- add `role="separator"`


rtivital:
Thanks!

PR #1935, opened 2022-07-30
fix NumberInput on Pinyin input method

huijiewei:
NumberInput should only number


rtivital:
@huijiewei please fix broken tests

huijiewei:
Prettier format

rtivital:
Thanks!

Issue #1883, opened 2022-07-27
ActionIcon disabled state while in "transparent" variant still shows background


eliranbenishai:
### What package has an issue

@mantine/core

### Describe the bug

ActionIcon disabled state while in "transparent" variant still shows the predetermined background color. Also styling the disabled state with `classNames` (which is vastly preferrable to me since I'm using Tailwind) became confusingly difficult since v5. The `emotion` class takes precedence, keeping the background in its predetermined color.

### In which browser did the problem occur

Chrome

### Do you know how to fix the issue

Yes

### Possible fix

Reinstate the `disabled` option in `classNames`? Or maybe make the `transparent` variant truly transparent?

rtivital:
Fixed in 5.0.3

PR #1936, opened 2022-07-30
[@mantine/core] ActionIcon: Transparent bg for disabled State - Skeleton: Safari: Fixed border Radius

pmalliotas:
Transparent variant now has transparent background color and border color for disabled state (#1883)


pmalliotas:
Added a fix for issue #1937. Safari Skeleton now has a border radius when being animated

rtivital:
Hi @pmalliotas thanks for PR and investigation! Unfortunately, it is not really a good way to add styles to transparent ActionIcon, I've already fixed the issue on my side, so closing this PR.

PR #1939, opened 2022-07-30
[@mantine/core] AppShell/Header: Add withBorder Prop

tonydawhale:
Added an optional `withBorder` prop to the appshell header component. By default, it is set to true.


GLeBrocq:
looks good to me

rtivital:
This should be added to HorizontalSection as well

tonydawhale:
@rtivital I based it on whether or not the styles had the `borderPosition` value. HorizontalSection does not have that. Unless it is a default setting, I do not know if that is possible.

lelboatz:
looks good to me

rtivital:
Thanks!

PR #1950, opened 2022-08-01
[@mantine/rte] Change the display property of the mention to `inline-block`

ryuhangyeong:
## Package

Rich text editor

## Example Docs

https://mantine.dev/others/rte/#mentions

## Description

- Mention has width and height properties, but the actual size is not applied because it is an inline property.

image

- If the Mention has inline-block properties, width, height, margin, padding can be set, so I think it is very useful.
- Even for inline block properties, the width (65px) and height (24px) currently specified are too small, so I thought it would be better to delete them.

## Suggest

image

- If there are several lines of mentions, it would be better to see if there is a margin-bottom. What do you think?

image

If you agree with this, I'll add a commit.


rtivital:
Thanks!

Issue #1390, opened 2022-05-05
npm installs dependencies that have incompatible versions (FOUC Next.js, Remix)


codart1:
### What package has an issue

@mantine/core

### Describe the bug

I just follow the official guide to integrating Remix with Mantine and see this flash of unstyled content issue:

https://user-images.githubusercontent.com/1023519/166897801-c658f2cf-1184-42c4-80d5-866bd52ebe9e.mp4

 

### In which browser did the problem occur

Chrome

### If possible, please include a link to a codesandbox with the reproduced problem

https://github.com/codart1/mantine-remix-fouc

codart1:
@rtivital Thanks, but how can a package manager be the root cause of this issue?

rtivital:
I do not know, it is just how npm started resolving dependencies recently, no idea how to fix it

agergo:
Could the issue be somehow related to the regression in Next.js 12.1.6: https://github.com/vercel/next.js/issues/36675 ? The FOUC disappeared for me after downgrading to 12.1.5 (I'm using `pnpm` btw)

rtivital:
Next.js issue – https://github.com/vercel/next.js/issues/36675

rtivital:
@thisisandreww do you use npm or pnpm?

thisisandreww:
I’m using npm

codart1:
NextJS issue? But I'm using Remix.

tombohub:
I am using yarn with Nextjs 12.1.6 and having FOUC.

The FOUC doesn't appear for MUI and Chakra.

debel27:
Isn't the FOUC caused by the injection of `<style>` elements at runtime ? 
Consider the following render sequence

- the app gets server-side rendered. The rendered HTML doesn't contain `<style>` tags
- the browser displays the (unstyled) app
- the app gets hydrated by client-side code. Mantine components' effects run and inject `<style>` tags
- styles get updated on screen, causing the FOUC to happen

@rtivital can you validate this ?
If so, is there a pre-compiled CSS stylesheet we can use to prevent FOUC from happening ?

Unfortunately, using the (undocumented :no_mouth:) `injectStylesIntoStaticMarkup` function server-side doesn't cut it for folks like me who stream the HTML as it gets rendered.

rtivital:
Should be fixed in 5.0, `@emotion/react` and `@emotion/server` packages are now required to be installed as peer dependencies

PR #1967, opened 2022-08-02
[docs] Add pnpm to package installation component

NyCodeGHG:
This PR adds support for the [pnpm](https://pnpm.io/) package manager in the package installation instructions component in the documentation.

### DIFF
--- a/docs/src/components/PackagesInstallation/PackagesInstallation.tsx
+++ b/docs/src/components/PackagesInstallation/PackagesInstallation.tsx
@@ -1,13 +1,19 @@
 import React, { useState } from 'react';
 import { Table, Checkbox, Code, Text, Box } from '@mantine/core';
 import { Prism } from '@mantine/prism';
-import { NpmIcon, YarnIcon } from '@mantine/ds';
+import { NpmIcon, YarnIcon, PnpmIcon } from '@mantine/ds';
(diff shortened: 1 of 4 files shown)


rtivital:
Hi @NyCodeGHG, we include `yarn` and `npm` in installation instructions because these are the most popular Node.js package managers. And we sometimes have issues with packages installation (see #1390 as reference). We are not planning to maintain `pnpm` as it is not very popular, and thus installation instructions should not be listed on mantine.dev website.

NyCodeGHG:
pnpm is becoming very popular. It's supported by Node's [corepack](https://github.com/nodejs/corepack). Also a few known projects using it are [Vue](https://github.com/vuejs/core/blob/main/pnpm-lock.yaml), [postcss](https://github.com/postcss/postcss/blob/main/pnpm-lock.yaml), [SvelteKit](https://github.com/sveltejs/kit/blob/master/pnpm-lock.yaml), [Next.js](https://github.com/vercel/next.js/blob/canary/pnpm-lock.yaml) and [Vite](https://github.com/vitejs/vite/blob/main/pnpm-lock.yaml). These are pretty popular projects and I think it's worth supporting pnpm.
Next.js also has [support for pnpm](https://nextjs.org/docs/api-reference/create-next-app#options). So some of the mantine users (like me) use pnpm.

Issue #1968, opened 2022-08-02
High CPU usage, but only when using Safari


waltershewmake:
### What package has an issue

@mantine/core

### Describe the bug

Note: I selected @mantine/core because I'm not sure which package is causing the issue yet. 

When loading a page built with Mantine in Safari (on both Mac and iPhone), the page very quickly becomes unusable due to spikes in CPU usage. The page freezes, stutters, interactions are slow, newly scrolled sections of the page don't render, etc. When checking Activity Monitor on a Mac, the site frequently spikes up to 100% CPU, even with no interaction on page load.

![Screen Shot 2022-08-01 at 8 42 08 PM](https://user-images.githubusercontent.com/21002090/182436953-569431a0-f3eb-42ea-a1aa-560b3a3816f8.png)

These results are not only found on websites created with Mantine 5 but even on https://mantine.dev. Luckily that means it's easy to reproduce.

Steps to reproduce:
- Using Safari, open https://mantine.dev/
- Interact with the page

I am opening this issue but I intend to look into this myself as well. I'm going to begin by creating a NextJS app with every component rendered, and slowly–by process of elimination–find the package causing the issue.

### What version of @mantine/hooks page do you have in package.json?

5.0.2

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

I think it's highly likely that some component is forced into a continuous re-render cycle causing it to build up tons of CPU usage, but I have no way of knowing right now.

waltershewmake:
I think I may have tracked the issue down to the NumberInput component.

waltershewmake:
#### Works
```jsx
<div>
    {new Array(100).fill(0).map((_, i) => (
        <TextInput key={i}/>
    ))}
</div>
```
#### Works
```jsx
<div>
    {new Array(100).fill(0).map((_, i) => (
        <Checkbox key={i}/>
    ))}
</div>
```
#### Causes the extreme lag we talked about
```jsx
<div>
    {new Array(100).fill(0).map((_, i) => (
        <NumberInput key={i}/>
    ))}
</div>
```

The code creates an array of 100 inputs. That number is just an example. I also tried 50, 20, 5, 3, and 2. All of which produced the same result.

The number input works just fine with only 1 instance though.

rtivital:
The issue is resolved, thanks for investigation!

rtivital:
Fixed in 5.0.3

ndmt:
[...]

PR #1908, opened 2022-07-28
[@mantine/core] NumberInput: Preserve cursor selection on format/parse

armoniaechaos:
Related issue: https://github.com/mantinedev/mantine/issues/1905

Before:
https://user-images.githubusercontent.com/22859283/181459033-21669098-e90c-4173-9e60-677fbdc75a23.mov

After:
https://user-images.githubusercontent.com/22859283/181459057-3de74e75-c259-4f90-9f2c-69990f35af0d.mov

Issue #1905, opened 2022-07-28
NumberInput parser/formatter doesn't work well with backspace.


arel:
### What package has an issue

@mantine/core

### Describe the bug

When using the parser/formatter feature of NumberInput, the backspace key doesn't work correctly when editing a value.

This is reproducible even on the documentation site.

For instance, from the example given here: https://mantine.dev/core/number-input/#parser-and-formatter, type in the number: $1,234. Then move the cursor between the 2 and the 3 and try to delete the first two digits. After deleting one character, the cursor moves to the end of the number.

### In which browser did the problem occur

Chrome

### If possible, please include a link to a codesandbox with the reproduced problem

https://mantine.dev/core/number-input/#parser-and-formatter

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

One possible solution may be to not to format the string when deleting and always format on blur.

rtivital:
Fixed in 5.0.2

PR #1977, opened 2022-08-03
[@mantine/core] NumberInput: Preserve cursor selection on format/parse

armoniaechaos:
Related issue: https://github.com/mantinedev/mantine/issues/1905 and https://github.com/mantinedev/mantine/issues/1968

Before:
https://user-images.githubusercontent.com/22859283/181459033-21669098-e90c-4173-9e60-677fbdc75a23.mov

After:
https://user-images.githubusercontent.com/22859283/181459057-3de74e75-c259-4f90-9f2c-69990f35af0d.mov

Schermata 2022-08-03 alle 10 28 08

https://github.com/mantinedev/mantine/pull/1908#issuecomment-1203107099
I solved the performance problem on Safari by using a useMemo for the input value so that it doesn't change with each render. If anyone can give me a hand to verify that it is really fixed, you would help me out :)

### DIFF
--- a/src/mantine-core/src/NumberInput/NumberInput.tsx
+++ b/src/mantine-core/src/NumberInput/NumberInput.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, useRef, forwardRef } from 'react';
+import React, { useState, useEffect, useRef, forwardRef, useMemo } from 'react';
 import { useMergedRef, assignRef, useOs, clamp } from '@mantine/hooks';
 import { DefaultProps, Selectors, useComponentDefaultProps } from '@mantine/styles';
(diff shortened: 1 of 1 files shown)


mmahalwy:
oh yeah, need this 🙏

rtivital:
@armoniaechaos it does not work as expected when there are extra symbols (input some numbers in sequence in formatter demo for reference):

https://user-images.githubusercontent.com/10353856/183123826-fd509965-f634-409a-9946-226cc19db190.mov

rtivital:
Related issue – #2024

rtivital:
Alright, I do not think that there is a good solution for it, @armoniaechaos if you find a way to fix it, feel free to reopen the PR

PR #1981, opened 2022-08-03
[core] cancel same in-progress workflow

ryuhangyeong:
## Description

If you're doing multiple commits to the same PR, I thought it would be efficient to stop working on the CI of the previous commit.

For the configuration file, refer to vite.js below

https://github.com/vitejs/vite/blob/main/.github/workflows/ci.yml#L23


rtivital:
Thanks!

Issue #1990, opened 2022-08-03
Vertical Divider not showing unless sx modified


hunterwilhelm:
### What package has an issue

@mantine/core

### Describe the bug

I was using the `<Divider orientation="vertical"/>` element and noticed that I had to customize the `sx` property to get it working. Here is how I have it working right now `<Divider sx={{ height: "auto" }} orientation="vertical" />`

For some reason, the `height: 100%` property is causing it not to show up.

### What version of @mantine/hooks page do you have in package.json?

"5.0.2

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/mantine-report-divider-vertical-bug-3yvpwz?file=/src/App.tsx:493-551

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

Changing the `height: 100%` to `height: auto` in the styles of the vertical divider.

rtivital:
Fixed in 5.0.3

PR #1991, opened 2022-08-03
Fix display issue with Vertical Divider #1990

hunterwilhelm:


rtivital:
Please also update demo – https://github.com/mantinedev/mantine/blob/master/src/mantine-demos/src/demos/core/Divider/Divider.demo.orientation.tsx#L29

rtivital:
Thanks!

Issue #1994, opened 2022-08-03
Docs - useform nested list - validation examples backwards


tmyracle:
### Link to the page where something nasty is located

https://mantine.dev/form/nested/#list-values-validation

### Exact quote of what is wrong

the console log at the bottom has the validation errors swapped, the user[0] should have the age validation error

```
import { useForm } from '@mantine/form';

const form = useForm({
  initialValues: {
    users: [
      { name: 'John', age: 12 },
      { name: '', age: 22 },
    ],
  },

  validate: {
    users: {
      name: (value) => (value.length < 2 ? 'Name should have at least 2 letters' : null),
      age: (value) => (value < 18 ? 'User must be 18 or older' : null),
    },
  },
});

// Validate list item field
form.validate('users.1.name');

// Or with all other fields
form.validate();
console.log(form.errors);
// {
//  'users.0.name': 'Name should have at least 2 letters',
//  'users.1.age': 'User must be 18 or older'
// }
```

### Are you willing to create pull request with the fix?

Yes

PR #1995, opened 2022-08-03
[docs] Fix useform list validation example

tmyracle:
Fixes [#1994](https://github.com/mantinedev/mantine/issues/1994) - validation example for lists in useform has error messages swapped


rtivital:
Thanks!

Issue #1986, opened 2022-08-03
Clicking a `MultiSelect` item inside a `Popever.Dropdown` closes the dropdown


yudai-nkt:
### What package has an issue

@mantine/core

### Describe the bug

When I deselect an item of a `MultiSelect` component inside a `Popover.Dropdown` component, I expect the dropdown is still open. However, it is actually closed after the click.

https://user-images.githubusercontent.com/13291527/182641226-72a38aec-93c5-4adb-907f-3c3021c1e2b4.mov

I assumed this to be a bug because other clickable components (e.g., `Checkbox` and `Switch`) do not cause this behavior, but feel free to transfer to Discussion if this is just my mistake (and please let me know how to fix).

### What version of @mantine/hooks page do you have in package.json?

5.0.2

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/mantine-issue1986-repro-ip53ql

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

ercgrat:
You can work around it by setting `closeOnClickOutside={false}`. The root issue happens in the `useClickOutside` hook. Because the element you clicked is removed from the DOM, it is no longer a descendant of the popover and gets interpreted as a "click outside". I think this is a really challening edge case.

Suppose the hook ignores clicks on elements that have been removed from the DOM. In that case, you may have actually clicked outside, but because what you clicked was deleted or dismissed, we don't close the popover and require two clicks. This doesn't sound the worst to me, but I think it deserves consideration before changing. I created PR #2003 as a suggestion.

yudai-nkt:
Thanks for the investigation, @ercgrat! Now I understand why this happens. Your proposal looks reasonable to me, so I'll wait for the maintainers' opinion.

rtivital:
Fixed in 5.1.0, thanks @ercgrat!
https://codesandbox.io/s/mantine-issue1986-repro-forked-0qwu7v?file=/src/App.tsx

PR #2003, opened 2022-08-04
[@mantine/hooks] use-click-outside: ignore clicks on elements removed from DOM

ercgrat:
This PR makes `useClickOutside` less permissive. If an element is not inside the DOM (it was removed), assume that the deleted element could have been inside the relevant `nodes` and do NOT fire the click outside event.

This is a proposed solution to #1986.


rtivital:
Thanks!

Issue #2012, opened 2022-08-05
@mantine/next and core @emotion peer deps


abadire:
### What package has an issue

@mantine/core

### Describe the bug

Hey! There is a warning in Yarn 2+ regarding @emotion peer dependency:
![image](https://user-images.githubusercontent.com/49831355/183065335-ae7fbe41-b26a-41bf-b6d0-31eff5c9a6ac.png)

### What version of @mantine/hooks page do you have in package.json?

5.0.2

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

Add @emotion peer dependencies to @mantine/core and to @mantine/next ([reference](https://dev.to/arcanis/implicit-transitive-peer-dependencies-ed0))

rtivital:
It is required to install these dependencies as described in getting started guide, `@manine/core` and `@mantine/next` do not depend on these dependencies directly – they use them with `@mantine/styles` and `@mantine/ssr` packages. I do not think that we should add these packages as peer dependencies.

PR #2014, opened 2022-08-05
[@mantine/core] deps: add missing @emotion peer dependency (#2012)

abadire:


rtivital:
It is required to install these dependencies as described in getting started guide, `@manine/core` and `@mantine/next` do not depend on these dependencies directly – they use them with `@mantine/styles` and `@mantine/ssr` packages. I do not think that we should add these packages as peer dependencies.

Issue #2026, opened 2022-08-06
Running form.setFieldValue inside a useEffect causes Maximum update depth exceeded error


tahmeednabi:
### What package has an issue

@mantine/form

### Describe the bug

In the new version of mantine, because of the way form.setFieldValue is written, every value in form.values is being cloned, resulting in a new reference detected by useEffect (I think). As such, using form values in useEffect dependency array results in maximum update depth error, e.g.

```javascript
useEffect(() => {
  form.setFieldValue('y', 3);
}, [form.values.x]);
```

Although the useEffect function is updating a different value, because `klona` is used to update the object, all the values are being updated, so useEffect runs again (hence the update depth error)

### What version of @mantine/hooks page do you have in package.json?

5.02

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

I believe the error lies in the cloning of the values object in the `setPath` function in `setFieldValue`:

**Original**

```javascript
export function setPath<T>(path: unknown, value: unknown, values: T) {
  const splittedPath = getSplittedPath(path);

  if (splittedPath.length === 0) {
    return values;
  }

  const cloned = klona(values); <<< Cloning causes new references for each field value

  if (splittedPath.length === 1) {
    cloned[splittedPath[0]] = value;
    return cloned;
  }

  let val = cloned[splittedPath[0]];

  for (let i = 1; i < splittedPath.length - 1; i += 1) {
    if (val === undefined) {
      return cloned;
    }

    val = val[splittedPath[i]];
  }

  val[splittedPath[splittedPath.length - 1]] = value;

  return cloned;
}
```

**Proposed Fix**
```javascript
export function setPath<T>(path: unknown, value: unknown, values: T) {
  const splittedPath = getSplittedPath(path);

  if (splittedPath.length === 0) {
    return values;
  }

  if (splittedPath.length === 1) {
    values[splittedPath[0]] = value;
    return values;
  }

  let val = values[splittedPath[0]];

  for (let i = 1; i < splittedPath.length - 1; i += 1) {
    if (val === undefined) {
      return values;
    }

    val = val[splittedPath[i]];
  }

  val[splittedPath[splittedPath.length - 1]] = value;

  return values;
}
```

I don't see a reason why we need to clone the values? I've implemented this fix in my own repo and its working fine so far. Please let me know if this is possible, and if not, what else could be done about this issue.

rtivital:
[...]

PR #2027, opened 2022-08-06
Removed cloning from setPath function

tahmeednabi:
Removed cloning from setPath function to prevent Maximum update depth issue when values are used as dependencies in useEffect, and the useEffect body has a form.setFieldValue function within it.

#2026


ryuhangyeong:
Please check github actions~

PR #2029, opened 2022-08-06
[@mantine/core] add data-position attribute to popover dropdown

jvdsande:
As discussed on Discord, simply expose the `ctx.placement` from Floating UI as a `data-position` attribute on `Popover.Dropdown`


rtivital:
Thanks!

PR #2030, opened 2022-08-06
[@mantine/core] add all properties to polymorphic component types

jvdsande:
This PR adds missing Typescript properties on polymorphic components: `displayName`, `defaultProps`, `propTypes` and `contextTypes`.

It uses a future-proof select on `FunctionComponent` to include any property that could be added by the React team in the future

### DIFF
--- a/src/mantine-utils/src/create-polymorphic-component/create-polymorphic-component.ts
+++ b/src/mantine-utils/src/create-polymorphic-component/create-polymorphic-component.ts
@@ -35,7 +35,9 @@ export function createPolymorphicComponent<
     props: ComponentProps<C>
   ) => React.ReactElement;
 
-  type PolymorphicComponent = _PolymorphicComponent & StaticComponents;
+  type ComponentProperties = Omit<React.FunctionComponent<ComponentProps<any>>, never>;
+
+  type PolymorphicComponent = _PolymorphicComponent & ComponentProperties & StaticComponents;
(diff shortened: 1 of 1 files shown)


src/mantine-utils/src/create-polymorphic-component/create-polymorphic-component.ts
```diff
@@ -35,7 +35,11 @@ export function createPolymorphicComponent<
     props: ComponentProps<C>
   ) => React.ReactElement;
 
-  type PolymorphicComponent = _PolymorphicComponent & StaticComponents;
+  type ComponentProperties = Omit<FunctionComponent<ComponentProps<any>>, never>;
```
jvdsande, line 38: The `Omit<T, never>` here basically removes the function signature of `FunctionComponent`, to only keep the static properties

jvdsande: (changed the line to use `React.FunctionComponent` instead of importing FunctionComponent separately)

rtivital:
@jvdsande Have you fixed prettier issue? There were some changes to github workflow that prevented it from running twice

jvdsande:
It should be fixed now, sorry about that, my IDE did not report it and I forgot to check the workflow result

rtivital:
Thanks!

PR #2035, opened 2022-08-07
[@mantine/core] Refactor: rename isValidSizeValue, getSizeValue parameter

ryuhangyeong:
## Suggest

In addition to `margin`, also receive `padding` values, so I think need to change the name!


rtivital:
Thanks!

PR #2046, opened 2022-08-08
[@mantine/core] Replace fromEntries with native function in createStyles function

ryuhangyeong:
## Description

- I think you can replace that function with a native function. That makes it more concise.

## Docs

- [Object.fromEntries()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries)

## Etc

- Mantine Support

image

- `fromEntries()`

image


rtivital:
Thanks!

PR #2053, opened 2022-08-09
[@mantine/core] Group: extend props with BoxProps

zeropaper:
I was looking to create a (HTML) `<header>` whit the styling of a `Group`.  
Since the (`others`) props are spread and the `Box` component has a `component` prop that would accept a `header` value.

Not sure if some omitting should be done (`Omit<BoxProps, '...'>`) tho.

### DIFF
--- a/src/mantine-core/src/Group/Group.tsx
+++ b/src/mantine-core/src/Group/Group.tsx
@@ -1,10 +1,10 @@
 import React, { forwardRef } from 'react';
 import { DefaultProps, MantineNumberSize, useComponentDefaultProps } from '@mantine/styles';
-import { Box } from '../Box';
+import { Box, BoxProps } from '../Box';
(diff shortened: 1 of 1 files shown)


rtivital:
Group already has all BoxProps – it extends DefaultProps which is basically the same thing in this context.

zeropaper:
I may miss something but... I'm getting a TS error when trying to use something like

```ts
<Group component="header">
  {/* ... */}
</Group>
```

https://codesandbox.io/s/pedantic-goldstine-bh135s?file=/src/App.tsx

rtivital:
Group is not supposed to be used as polymorphic component – it manages layout, not functionality, you should wrap it with `header` element instead:
```tsx
<header><Group /></header>
```

zeropaper:
I guess I'll rather style the header element with some "good'ol CSS" than
nesting elements and cluttering the DOM to achieve that.

Thanks anyways.

PR #2050, opened 2022-08-08
[@mantine/spotlight] Add closeOnTrigger prop to SpotlightAction (#2040)

Meierschlumpf:
Pull request for issue #2040

PR #2066, opened 2022-08-10
[@mantine/spotlight] Fix closeOnTrigger with 'Enter'

morvud:
#2050 added closeOnTrigger to actions but it only works when clicking, this PR adds the check to 'Enter' handler.


rtivital:
Thanks!

PR #2094, opened 2022-08-12
Fix #2093: [@mantine/hooks] Interval hook ensure only 1 running timer

### LINKED ISSUE #2093: Interval hook can create two instances of the timer
### What package has an issue

@mantine/hooks

### Describe the bug

In some cases the Interval hook is firing the timer twice depending on how the component using it is being created.

Related to #782 and specifically this fix: https://github.com/mantinedev/mantine/pull/1498

cc @prograpp

5.1.1

_No response_

Yes

Yes

### Possible fix

the `intervalRef` must be nulled out PR on its way

melloware:
Fix #2093: Interval hook ensure only 1 running timer

This ensures even when creating and destroying components in React 18 it ensures that only 1 timer can be running at a time.

### DIFF
--- a/src/mantine-hooks/src/use-interval/use-interval.ts
+++ b/src/mantine-hooks/src/use-interval/use-interval.ts
@@ -11,7 +11,7 @@ export function useInterval(fn: () => void, interval: number) {
 
   const start = () => {
     setActive((old) => {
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #2097, opened 2022-08-13
[@mantine/hooks] use-id: replace with `useIsomorphicEffect` hooks

ryuhangyeong:
## Suggest

I thought I could replace that syntax with the already existing Hooks.

## Question

1. If you look at the use-isomorphic-effect hook, you are checking the document, not the window. Is it intentional?

```ts
// hooks/use-isomorphic-effect.ts
export const useIsomorphicEffect = typeof document !== 'undefined' ? useLayoutEffect : useEffect;
```


rtivital:
Thanks!

PR #2116, opened 2022-08-14
[@mantine/core] fixed issue when using collapse with reduced motion

jvdsande:
Fixes error reported on Discord where the Collapse component would sometimes make the page crash. "Sometimes" narrowed down to "people using reduced motion accessibility setting".


rtivital:
Thanks!

slk333:
good catch and thanks @jvdsande

PR #2125, opened 2022-08-14
[docs] Use NavLink instead of UnstyledButton in examples

ShahriarKh:
#2124


rtivital:
I'd rather keep it as it.

PR #2130, opened 2022-08-15
[@mantine/styles] Type color variants to get better autocomplete

### LINKED ISSUE #2129: Color props doens't autocomplete variants
### What package has an issue

@mantine/core

### Describe the bug

It would be great if this also autocompleted i.e. `red.3`

image

5.1.6

Any mantine project

Yes

Yes

### Possible fix

Linked PR

Zeko369:
Closes #2129


rtivital:
Hi @Zeko369, I do not think that we should include all these values in component types as:
- it may impact performance
- it will create a huge mess in IDE autocomplete – it is much easier to view all possible colors and only then append color index
- it is already handled with `(string & {})` type
- It may break custom colors type override – https://mantine.dev/theming/colors/#add-custom-colors-types

PR #2143, opened 2022-08-16
[@mantine/core] Grid: Add column prop

mdodell:
Adds an `order` prop to the `Grid.Col` component so that columns can be re-ordered/be ordered based on responsiveness.

### DIFF
--- a/src/mantine-demos/src/demos/core/Grid/Grid.demo.order.tsx
+++ b/src/mantine-demos/src/demos/core/Grid/Grid.demo.order.tsx
@@ -1,0 +1,37 @@
+import React from 'react';
+import { Grid } from '@mantine/core';
+import { ColWrapper as Col } from './_col-wrapper';
+
+const code = `
+import { Grid } from '@mantine/core';
+
+function Demo() {
+  return (
+    <Grid>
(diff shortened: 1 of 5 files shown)


src/mantine-demos/src/demos/core/Grid/Grid.demo.order.tsx
```diff
@@ -0,0 +1,37 @@
...
+import { Grid } from '@mantine/core';
+import { ColWrapper as Col } from './_col-wrapper';
+
+const code = `
+import { Grid } from '@mantine/core';
+
+function Demo() {
+  return (
+    <Grid>
+      <Col span={3} order={2}>2</Col>
```
rtivital, line 11: Please add responsive props (let's say `orderSm` and `orderLg` to the demo)

mdodell: Sounds good - should I also use the [`useMediaQuery`](https://mantine.dev/hooks/use-media-query/) hook here to relabel the children based on their new order?

rtivital:
Thanks!

PR #2146, opened 2022-08-17
[@mantine/core] ListItem:  Span limits the full width capability

glomyst:
By Removing the span we can now pass a div or any other JSX Element which can take the full width of the container the list is in.
Eg. I can put Text on the left side and have a button floated / flexed to the right side.


rtivital:
Have you checked if that works correctly in Firefox?

Manas-Nagelia:
This actually did more harm than good. If you see this image below, it overflows the List.Item tag. I worked around this by adding <span> tags inside the List.Item tag. 
Screenshot 2022-11-01 at 7 58 35 PM

PR #2157, opened 2022-08-18
[@mantine/styles] Improved extractSystemStyles function

ryuhangyeong:
## Suggest

If you use filterProps in the extractSystemStyles function, you can write more concise code.


rtivital:
Thanks!

PR #2162, opened 2022-08-18
[@mantine/core] Group: use css selectors to style children

drj17:
Replaces the logic in `<Group />` that passes `className` as prop with a direct child selector.   This removes the requirement for Group children to accept `className` as a prop.


rtivital:
Thanks!

danielemesh:
@drj17 this makes it impossible to override a `Group`'s direct child `flex-grow` rule without using `!important` or another css selector that is more "powerful" than `> *` 😢 
and IMHO it's a breaking change that needs to be documented...

PR #2163, opened 2022-08-18
[@mantine/core] Grid: Fix offset and order responsive props

mdodell:
Fixes issue with both `offset` and `order` props for `Grid.Col` that requires each to have a responsive prop (`xs`, `sm`, `md`, `lg`, `xl`) for their respective responsive props (`offsetXs`, `orderXs`, etc). to work.

Also updated the docs for `order` to properly use `Grid.Col` rather than the `Col` shorthand.

### DIFF
--- a/src/mantine-core/src/Grid/Col/Col.styles.ts
+++ b/src/mantine-core/src/Grid/Col/Col.styles.ts
@@ -50,15 +50,17 @@ function getBreakpointsStyles({
   columns: number;
 }) {
   return MANTINE_SIZES.reduce((acc, size) => {
-    if (typeof sizes[size] === 'number') {
-      acc[`@media (min-width: ${theme.breakpoints[size] + 1}px)`] = {
-        order: orders[size],
-        flexBasis: getColumnWidth(sizes[size], columns),
-        flexShrink: 0,
-        maxWidth: grow ? 'unset' : getColumnWidth(sizes[size], columns),
-        marginLeft: getColumnOffset(offsets[size], columns),
-      };
-    }
+    acc[`@media (min-width: ${theme.breakpoints[size] + 1}px)`] = {
+      order: orders[size],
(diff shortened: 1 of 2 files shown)


src/mantine-core/src/Grid/Col/Col.styles.ts
```diff
@@ -50,15 +50,17 @@ function getBreakpointsStyles({
...
-    }
+    acc[`@media (min-width: ${theme.breakpoints[size] + 1}px)`] = {
+      order: orders[size],
+      flexBasis: typeof sizes[size] === 'number' ? getColumnWidth(sizes[size], columns) : 'auto',
+      flexShrink: 0,
+      maxWidth: grow
+        ? 'unset'
+        : typeof sizes[size] === 'number'
+        ? getColumnWidth(sizes[size], columns)
+        : 'none',
```
mdodell, line 61: Not super in love with the double `ternary` so open to feedback here. If the `sizes[size]` isn't defined, we just use the initial value for that CSS property (`none` in the case of `max-width` and `auto` in the case of `flexBasis`).

rtivital:
Thanks!

rtivital:
@mdodell this PR broke all Grid responsive styles, I've reverted it

mdodell:
> @mdodell this PR broke all Grid responsive styles, I've reverted it

Interesting. I'll submit a follow-up PR, I think that the check for `maxWidth` and `flexBasis` was breaking it.

rtivital:
No need, I've already resolved the isuue

mdodell:
> No need, I've already resolved the isuue

Ah - perfect. My apologies here. Is it now resolved so that the `MantineSize` prop is no longer needed in order to check for order? I just fixed it locally before I saw this comment - it looks like `maxWidth` and `flexBasis` just both need to be set to `100%` if the `typeof sizes[size]` check is falsey.

PR #2164, opened 2022-08-19
Add support for dependabot

gaby:
* This PR adds support for dependabot to automatically keep dependencies up-to-date.

Source: https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuring-dependabot-version-updates#example-dependabotyml-file


rtivital:
Hi @gaby, this project does not need dependabot.

Issue #1715, opened 2022-07-02
The `@emotion/react` should be used as a peer dependency


u3u:
### What package has an issue

@mantine/core

### Describe the bug

https://github.com/emotion-js/emotion/discussions/2795#discussioncomment-3031404
https://github.com/emotion-js/emotion/issues/2343#issuecomment-817265816

According to the comments here `@emotion/react` should be used as a peer dependency instead of a direct dependency, which would lead to the problem of introducing `@emotion/react` repeatedly

Another related issue: https://github.com/emotion-js/emotion/issues/2803

### In which browser did the problem occur

None

### If possible, please include a link to a codesandbox with the reproduced problem

https://github.com/u3u/emotion-issue-2803

rtivital:
Seems reasonable, I'll see if I'll be able to make it work in 5.0 release.

rtivital:
Fixed in 5.0

PR #2174, opened 2022-08-19
[@mantine/core] - Add @emotion/react to package.json

RichardX366:
`@emotion/react` is currently a peer dependency for @mantine/styles, but when installing (tested with yarn), it will not be installed, causing mantine to crash on startup.


rtivital:
See #1715

PR #2179, opened 2022-08-20
[@mantine/core] Container: Change desc of fluid prop

### LINKED ISSUE #2175: Misleading description for Container fluid prop
### What package has an issue

@mantine/core

### Describe the bug

https://github.com/mantinedev/mantine/blob/8a2b20e7388d66fbb911c15f872ad1a545dd2e3b/src/mantine-core/src/Container/Container.tsx#L15

"always take 100% of width" makes me think that the Container has the css prop `width: 100%`, but in reality it has `max-width: 100%`. It took me 10 minutes and the devtools to find out that it follows the second behaviour. :(

^5.0

_No response_

Yes

Yes

### Possible fix

Change description to `If fluid is set to true, size prop is ignored and Container can expand to 100% of width` or something similar (maybe reference the `max-width` prop as well)

jvllmr:
Description of fluid prop was misleading because it implied that the prop would apply `width: 100%`, but it actually applies `max-width: 100%`

closes #2175

### DIFF
--- a/src/mantine-core/src/Container/Container.tsx
+++ b/src/mantine-core/src/Container/Container.tsx
@@ -12,7 +12,7 @@ export interface ContainerProps extends DefaultProps, React.ComponentPropsWithou
   /** Predefined container max-width or number for max-width in px */
   size?: MantineNumberSize;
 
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #2181, opened 2022-08-20
[@mantine/core] MultiSelect: prevent drop down from opening on item remove

achmurali:
Prevent dropdown from opening when item is removed when multiselect is in focus or not.


src/mantine-core/src/MultiSelect/MultiSelect.tsx
```diff
@@ -324,6 +324,10 @@ export const MultiSelect = forwardRef<HTMLInputElement, MultiSelectProps>((props
       valuesOverflow.current = true;
       setDropdownOpened(false);
     }
+
+    if (filteredData.length === 0) {
```
rtivital, line 328: Will this work correct with creatable prop?

achmurali: Yes vitaly, i have tested creatable multiselect and they are working fine

rtivital: Great, thanks!

PR #2182, opened 2022-08-20
[@mantine/core] Title: Fix `inline` prop being ignored

### LINKED ISSUE #2180: Title component does not respect inline=true
### What package has an issue

@mantine/core

### Describe the bug

A `<Title inline>` element does not set `line-height` to 1 because the property is defined twice.
grafik

5.2.0

https://codesandbox.io/s/optimistic-architecture-vj6t9f

Yes

Yes

### Possible fix

Amend [getLineHeight](https://github.com/mantinedev/mantine/blob/557d7728ccebfe2b15cf38ec2095efe83d81118c/src/mantine-core/src/Title/Title.styles.ts#L20) to return 1 if inline is set to true.

Narcha:
Fixes #2180

### DIFF
--- a/src/mantine-core/src/Title/Title.styles.ts
+++ b/src/mantine-core/src/Title/Title.styles.ts
@@ -20,7 +20,7 @@ function getFontSize(size: TitleSize, element: HeadingElement, theme: MantineThe
 
 function getLineHeight(size: TitleSize, element: HeadingElement, theme: MantineTheme) {
   if (typeof size !== 'undefined' && size in theme.headings.sizes) {
-    return theme.headings.sizes[element].lineHeight;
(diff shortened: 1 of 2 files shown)


Narcha:
To test this, I added an example to the storybook [here](https://github.com/mantinedev/mantine/blob/0326f9b75525a1acebaf69ae569c7b4b394a6fd3/src/mantine-core/src/Title/Title.story.tsx#L48-L62):
grafik
Should I push that as well?

Narcha:
Please also let me know if my solution doesn't fit your code style, there are many other ways to fix this.

rtivital:
Great job, thanks!

Narcha:
Thanks for making Mantine, it's a joy to work with :smile:

PR #2189, opened 2022-08-21
ActionIcon:  Fix mismatch between demo apparent code and real code

slk333:
Please see the screenshots and the videos in the discord thread: https://discord.com/channels/854810300876062770/1010668931842855064

This PR is more of a "Expose the problem to the maintainers" PR than anything, even though it does fix the mismatch

The demo real code was changing both the ActionIcon's size and the embedded Icon's size.
While this seems to be desirable, **it was not reflected in the apparent demo code**.

There are two way to "fix" this. One simple way, and one difficult way

1. (Easy way) **Make the real code match the apparent code**: do not change the embedded Icon's size when the slider moves.
2. (Hard way) **Make the apparent code match the real code**: continue to change the embedded Icon's size when the slider moves, but reflect that in the code template in real time as well.

This PR takes the first option, as it is the most straightforward, but I believe the second option might be better. **It would require extensive string manipulation though to change the template code and I want to leave that to the maintainer discretion**

Before:
Screenshot 2022-08-21 at 11 23 18

After: 
Screenshot 2022-08-21 at 11 20 31

### DIFF
--- a/src/mantine-demos/src/demos/core/ActionIcon/ActionIcon.demo.configurator.tsx
+++ b/src/mantine-demos/src/demos/core/ActionIcon/ActionIcon.demo.configurator.tsx
@@ -20,28 +20,28 @@ function Wrapper(props: ActionIconProps) {
   );
 }
 
-function computeChildIconSizeProp(props:string) {
-if (props.includes('size="xs"')) {
-return 'size={12}';
-}
-if (props.includes('size="sm"')) {
-return 'size={14}';
-}
-if (props.includes('size="md"')) {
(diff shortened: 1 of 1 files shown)


rtivital:
The demo should include icon size instead.

slk333:
> The demo should include icon size instead.

I changed the code to follow your suggestion, can you check the code?

https://user-images.githubusercontent.com/23578949/185793470-8edf0b16-3ace-48cf-809a-df79b0b91b42.mov

rtivital:
Thanks!

PR #2205, opened 2022-08-22
[@mantine/core] Notification: spread closeButtonProps after iconSize, color

emewjin:
[original discord thread](https://discord.com/channels/854810300876062770/887408088800448553/1011259560012566609)

Spread Notification's `closeButtonProps` after CloseButton's `iconSize` and `color` would be better experience to customize Notification's CloseButton. (override default prop `iconSize=16, color=gray`)

Like below code, for customizing, using  CloseButton's props will better than using `classNames` because it is more simple and It can use CloseButton's predefined props.
```tsx
// If using classNames to customize Notification's closeButton
const useStyles = createStyles(
  (theme) => ({
    closeButton: {
      color: theme.colors.gray[2],
      width: '44px',
      height: '44px',
    },
  })
);
const { classes } = useStyles();
<Notification classNames={{ ...classes }} />

// If using closeButtonProps to customize Notification's closeButton
<Notification closeButtonProps={{ color: 'gray.2', size: 'xl' }}  />
```

### DIFF
--- a/docs/src/docs/core/Notification.mdx
+++ b/docs/src/docs/core/Notification.mdx
@@ -32,6 +32,15 @@ Notification component has 3 variants:
 
 <Demo data={NotificationDemos.configurator} />
 
+## CloseButtonProps
+
+To customize CloseButton inside Notification, you can pass `CloseButtonProps`.
+
+```tsx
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #2214, opened 2022-08-23
[@mantine/hooks] use-session-storage: Implement hook.

zhulien-ivanov:
Implementation for: #2169.

Also fixes issue in `useLocalStorage` not supporting falsy values.


rtivital (changes requested):
We should not duplicate the same logic in two hooks, there should be one hook creator function `createStorage`, it should take single argument `type: 'localStorage' | 'sessionStorage'`. Also, documentation for use-session-storage is obsolete, there is not any difference with use-local-storage, so its docs should be placed in use-local-storage hook with mention that the hook works exactly the same way.

zhulien-ivanov:
Isn't it better if we leave the doc for `useSessionStorage` as a separate page and just present the interface with a reference to `useLocalStorage` and explain that it works the same? Not having a dedicated page for the hook is confusing and leaves the impression that the hook doesn't exist. You can't use the search to find it either.

rtivital:
I doubt that use-session-storage hook will be used frequently, so a separate page that will be exactly the same as use-local-storage hook is not necessary – it will just make maintenance more complicated (it will be required to update two pages when changes are made to shared base).

rtivital:
@zhulien-ivanov please resolve conflicts

zhulien-ivanov:
Done.

rtivital:
I'd better leave it as is and have two separate hooks, it will be easier to maintain.

PR #2222, opened 2022-08-24
[core]: Add volta support

MalikBagwala:
@rtivital

Volta is a node version manager just like **nvm** 
https://volta.sh/

By pinning the `node version` to `package.json`, volta can automatically pick up the node version to use when running `yarn install` making the process more user friendly.


rtivital:
what is volta and why do we need to support it?

PR #2223, opened 2022-08-24
[@mantine/hooks] use-local-storage: Fix for supporting falsy values.

zhulien-ivanov:
Fix until #2214 gets merged.

Currently, `useLocalStorage` returns default value in the `getInitialValueInEffect` `useEffect` if the stored value is falsy due to the `||` check. This can be verified even in the `SerializeBoolean` story. With the proposed change we support the storage of all falsy values including `null` (which should be completely valid).

### DIFF
--- a/src/mantine-hooks/src/use-local-storage/use-local-storage.ts
+++ b/src/mantine-hooks/src/use-local-storage/use-local-storage.ts
@@ -41,11 +41,20 @@ export function useLocalStorage<T = string>({
   deserialize = deserializeJSON,
   serialize = serializeJSON,
 }: UseLocalStorage<T>) {
-  const [value, setValue] = useState<T>(
-    typeof window !== 'undefined' && 'localStorage' in window && !getInitialValueInEffect
-      ? deserialize(window.localStorage.getItem(key) ?? undefined)
-      : ((defaultValue ?? '') as T)
+  const readLocalStorageValue = useCallback(
+    (skipStorage?: boolean): T => {
+      if (typeof window === 'undefined' || !('localStorage' in window) || skipStorage) {
+        return (defaultValue ?? '') as T;
+      }
+
+      const storageValue = window.localStorage.getItem(key);
+
+      return storageValue !== null ? deserialize(storageValue) : ((defaultValue ?? '') as T);
+    },
+    [key, defaultValue]
   );
+
+  const [value, setValue] = useState<T>(readLocalStorageValue(getInitialValueInEffect));
 
   const setLocalStorageValue = useCallback(
     (val: T | ((prevState: T) => T)) => {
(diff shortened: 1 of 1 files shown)


rtivital:
Are you 100% sure that it will not produce hydration mismatches? Have you tested that with Next.js?

zhulien-ivanov:
Currently, the hook is not working correctly. There is a mismatch between the value being set in `useState` and in `useEffect`.
` const [value, setValue] = useState<T>(
    typeof window !== 'undefined' && 'localStorage' in window && !getInitialValueInEffect
      ? deserialize(window.localStorage.getItem(key) ?? undefined)
      : ((defaultValue ?? '') as T)`

If `getInitialValueInEffect` is `false` and if the `key` doesn't exist in `localStorage`, `undefined` is assigned as initial value, not `defaultValue`. This happens because of this: `deserialize(window.localStorage.getItem(key) ?? undefined)`. `getItem` returns `null` so it becomes `deserialize(null ?? undefined)` which in turn becomes `deserialize(undefined)`. The default `deserialize` implementation tries to parse the value and if it fails, it returns the value itself. `JSON.parse(undefined)` throws so `undefined` is set as initial value. Then the hooks counts on the:

`useEffect(() => {
    if (defaultValue !== undefined && value === undefined) {
      setLocalStorageValue(defaultValue);
    }
  }, [defaultValue, value, setLocalStorageValue]);`

to assign the actual default value.

On the other hand, if you supply `getInitialValueInEffect` as `true`, `useState` will set initial value as `defaultValue` and then apply the:
`useEffect(() => {
    if (getInitialValueInEffect) {
      setValue(
        deserialize(window.localStorage.getItem(key) ?? undefined) || ((defaultValue ?? '') as T)
      );
    }
  }, []);`
which have a few problems and will definitely cause a mismatch. The most obvious case is storing a boolean of `false`.
`deserialize(window.localStorage.getItem(key) ?? undefined) || ((defaultValue ?? '') as T)` => `deserialize('false' ?? undefined) || ((defaultValue ?? '') as T)` => `deserialize('false') || ((defaultValue ?? '') as T)` => `false || ((defaultValue ?? '') as T)` which in the end returns default value instead of the actual boolean one. You can verify this in the story for the hook, it doesn't work properly on refresh with stored `false`. We have the same exact problem for all falsy values including `''`, `null`, etc..

I haven't tested it in Next.js as I just extracted the common logic in a reusable method and fixed the current issues. Like I stated in discord, I think the `getInitialValueInEffect` property is completely useless and just adds unnecessary complexity and an additional source of issues as you can just do:
1. useState -> if window and localStorage exists -> get value or default.
2. useEffect -> set value or default.

To verify this, I checked other implementations of hooks using local storage and all of them do exactly that. Example: https://usehooks-ts.com/react-hook/use-local-storage.

I'll make a setup using Next to test it out tomorrow.

rtivital:
Alright, I've checked and it seems to be fine, merging

PR #2230, opened 2022-08-25
[@mantine/hooks] use-color-scheme-value: Return value according to the system theme

AmelloAster:
I would like to add a more convenient Hook function for `useColorScheme`. It is `useColorSchemeValue` allows the user to enter two parameters `useColorSchemeValue('dark', ' light');` the first parameter will be returned when the current theme is light, and the second parameter will be returned when the current theme is dark, so that the user does not have to make frequent determinations `if theme === 'dark'`


rtivital:
Hi @AmelloAster, we already export use-color-scheme hook, there is no need for this one – it is specific to your application business logic.

PR #2259, opened 2022-08-28
[@mantine/core] Add name attribute to Radio.Group

nmay231:
The name prop should probably be required in a future major release because otherwise you get [odd behaviour](https://codesandbox.io/s/jolly-worker-yfclb5).


rtivital:
Thanks!

Issue #1753, opened 2022-07-08
`slide-left` transition looks off on modal with `size` set


emi2k01:
### What package has an issue

@mantine/core

### Describe the bug

If you set `size` to a big value (~900) in a `Modal`, the transition `slide-left` looks off when the modal is opened.

The `transitionDuration` is to make the problem more apparent but it still looks off without it.

### In which browser did the problem occur

Chrome, Firefox

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/mantine-playground-forked-ktf2rw?file=/src/App.tsx

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

Sorry, unfortunately I have no experience with mantine's codebase but since I'll start using this library for the months to come, I'd like to learn what I can. If someone has a clue to what's wrong, I'm willing to look into it.

emi2k01:
Using `withCloseButton={false}` seems to make it look OK. I updated the codesandbox to show both behaviors

emi2k01:
Using a smaller width (~600px) seems to make it look ok too, though I don't know why.

If I use 1000px, it looks wrong even though I'm trying it in a big enough viewport.

rtivital:
Fixed in 5.2.7 – https://codesandbox.io/s/mantine-playground-forked-ub6nd5?file=/src/App.tsx
Thanks @hotgen!

Issue #1982, opened 2022-08-03
Menu Component causing page to scroll (React strict mode only)


thivy:
### What package has an issue

@mantine/core

### Describe the bug

Hi Team, 

I found an issue  with Mantine 5.0.2 where the scroll position of the body changes if there are many Menu Components rendered on the page. Below are the examples of with and without the Menu component. 

Has anyone had this issue?  

With Menu component 
https://youtu.be/NFigCG_leGU 

Without Menu component
https://youtu.be/up2pdnbI-xQ

### What version of @mantine/hooks page do you have in package.json?

5.0.2

### Do you know how to fix the issue

No

rtivital:
Hi @thivy, please provide a codesandbox ([template](https://codesandbox.io/s/mantine-template-9ze89?file=/src/App.tsx:0-6)) with [minimal reproducible example](https://stackoverflow.com/help/minimal-reproducible-example). It is not possible to identify the issue without reproduction.

nostrapollo:
I experienced the same issue, unfortunately I can't seem to reproduce using the codesandbox template...after experimenting with it I found setting `transitionDuration={0}` on the `<Menu>` component stopped the scrolling issue in my project.

I suspect it has something to do with `data-autofocus` and the useFocusTrap hook because the Menu component is scrolled into view.

nacasha:
> I experienced the same issue, unfortunately I can't seem to reproduce using the codesandbox template...after experimenting with it I found setting `transitionDuration={0}` on the `<Menu>` component stopped the scrolling issue in my project.
> 
> I suspect it has something to do with `data-autofocus` and the useFocusTrap hook because the Menu component is scrolled into view.

thanks, for now I have to use this workaround to prevent scrolling to bottom when using Menu component

wes337:
Would adding a default prop `transitionDuration: 0` to the Menu component be a valid fix for this?

rtivital:
No, I still do not see the issue, Menu component does not scroll anything when used on mantine.dev docs.

keith-loww:
I have also experienced the same issue, and the transitionDuration={0} "fix" also worked for me.

nacasha:
> No, I still do not see the issue, Menu component does not scroll anything when used on mantine.dev docs.

Yeah, It's weird that on mantine docs it doesn't scroll, but the issue is exist to other people projects, and by adding `transitionDuration: 0` prop fix the issue

rtivital:
[...]

PR #2267, opened 2022-08-28
[@mantine/hooks] use-focus-trap: Fix animation bug for modal (#1753)

hotgen:


rtivital:
Hi @hotgen, is this related to one of the open issues?

hotgen:
@rtivital Hi, yes, it's fix for "slide-left transition looks off on modal with size set" #1753. The `focus` call shifts the scroll and thus breaks the animation.

wes337:
This does seem to fix the bug in the issue. But are there other side effects to this solution? What if the `useFocusTrap` hook focuses an element outside the viewport, and we want to scroll?

rtivital:
> What if the useFocusTrap hook focuses an element outside the viewport, and we want to scroll?

There should not be such case, focus trap should always be used on visible elements, any scrolling caused by use-focus-trap in unwanted. 

I was thinking that it will resolve the issue with Menu – #1982

rtivital:
Thanks!

Issue #1812, opened 2022-07-18
allowLevelChange not respected


rfdrake27:
### What package has an issue

@mantine/dates

### Describe the bug

When allowLevelChange is set to false, and the initialLevel is set to month, the level still changes to days when a month is selected.

```
function Demo() {
  return <Calendar allowLevelChange={false} initialLevel='month' />;
}
```

With this combination of props, I would expect the calendar to select the month immediately when the user clicks a month. 

### In which browser did the problem occur

Edge 103

khobiziilyes:
Already started working on this issue.

rfdrake27:
Yes. 
My use case is that the user is presented with the "month" picker. Once they click the month, the input will blur, and show the value as the first of that month. 

So as an example, the user clicks "July" on the calendar. It blurs. The value in the input is "July 1st, 2022" (or however the format prop is configured). 
Hypothetically, if the calendar were set to the following props:
`<Calendar allowLevelChange={false} initialLevel='year' />`

If the user selects "2022", I would expect that the calendar will select the date value to "Jan 1st, 2022".  

Presently those two examples are not possible with the calendar picker. Even with the allowLevelChange prop set to false, the calendar will continue changing levels down until the user is required to select a day.

rtivital:
DatePicker does not support this option – it is designed to pick dates, not months. We are not planning to add month picker any time soon.

Issue #547, opened 2021-12-20
[Feature Request] [@mantine/dates] Week select


jerebtw:
### Usage example, including component/hook API

Add a week select

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/week

rtivital:
Well, I guess we can also add MonthPicker, YearPicker and QuarterPicker components as well including range variants. Just curious for what are you using week select?

jerebtw:
For a timetable app

jerebtw:
@rtivital could you please add the `Help wanted` tag because i don't know much about the date package

rtivital:
Help is not wanted, it's a large feature, I'll work on it on my own when I have time for this

jerebtw:
Ohhh okay

LucasLemanowicz:
A more generic implementation of this could be to allow for "Defined Ranges" that the implementer defines. The interface could look something along the lines of:

```typescript
interface DefinedDateRange {
    label: string;
    dateRange: {
        start?: Date,
        end?: Date,
    }
}
```

and the `DateRangePicker` would take in a `staticRanges: DefinedDateRange[]` prop, similar to what `react-date-range` does: https://hypeserver.github.io/react-date-range/

This would be incredibly useful for me (if you need any additional signal on the value of this feature).

PR #2282, opened 2022-08-30
[@mantine/dates] Support YearPicker component

anhdo95:
Based on comments from these issues https://github.com/mantinedev/mantine/issues/1812, https://github.com/mantinedev/mantine/issues/547. I'd like to support YearPicker to meet people's requirements


anhdo95:
@rtivital could you please review and consider this support?

rtivital:
Hi @anhdo95, this component does not fit well with future updated coming in next major release which will introduce dates package revamp – introducing this component now will mean that we will need to make even more breaking changes to the package when 6.0 is released. So, I'd rather postpone the addition of this component until dates package revamp in 6.0 is finished.

PR #2297, opened 2022-08-31
[@mantine/core] Autocomplete: fix grouping in Autocomplete

### LINKED ISSUE #2278: Autocomplete data grouping does not work perfectly when data is updated
### What package has an issue

@mantine/core

### Describe the bug

image

I posted a screenshot showing the issue. Item2 has group Item3. But when I add two new items also with group Item3, they are group by themselves.

The actually data has no difference between the groups. This issue only resolves itself when I refresh the app completely. All three items will be with the same group. 

The issue only happens when I update the data behind Autocomplete.

5.2.3

_No response_

No

No

### Possible fix

_No response_

achmurali:
Fixes #2278


rtivital:
Thanks!

PR #2317, opened 2022-09-01
[@mantine/core] Grid: Add auto and content spans

leeweisberger:
Discussion https://github.com/mantinedev/mantine/discussions/2292

Add two additional options for grid columns spans:
- `auto` - A column with auto grows to fill additional space on the row. All columns with span `auto` will be the same size.
- `content` - The column fits its content. Using the same name as flex-basis for the same concept: https://developer.mozilla.org/en-US/docs/Web/CSS/flex-basis

Auto
image

Content
image

### DIFF
--- a/docs/src/docs/core/Grid.mdx
+++ b/docs/src/docs/core/Grid.mdx
@@ -31,7 +31,7 @@ Use xs, sm, md, lg, xl values to set spacing from `theme.spacing` or number to s
 
 ## Grow
 
-Set `grow` prop on Grid component to force last row take 100% of container width:
+Set `grow` prop on Grid component to force all rows to take 100% of container width:
 
 <Demo data={GridDemos.growConfigurator} />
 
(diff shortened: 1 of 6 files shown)


docs/src/docs/core/Grid.mdx
```diff
@@ -31,7 +31,7 @@ Use xs, sm, md, lg, xl values to set spacing from `theme.spacing` or number to s
 
 ## Grow
 
-Set `grow` prop on Grid component to force last row take 100% of container width:
+Set `grow` prop on Grid component to force all rows to take 100% of container width:
```
leeweisberger, line 34: Unrelated - this documentation is incorrect. `grow` affects all rows, not just the last.

src/mantine-core/src/Grid/Col/Col.styles.ts
```diff
@@ -30,8 +32,29 @@ interface ColStyles {
   orderXl: React.CSSProperties['order'];
 }
 
-const getColumnWidth = (colSpan: number, columns: number) =>
-  colSpan ? `${100 / (columns / colSpan)}%` : undefined;
+const getColumnFlexBasis = (colSpan: ColSpan, columns: number) => {
+  if (colSpan === 'content') {
+    return 'auto';
+  }
```
leeweisberger, line 38: Rather than using flex basis `content`, we're using flex basis `auto` together with a width of `auto` to support older browsers

image

PR #2322, opened 2022-09-02
[@mantine/hooks]use-text-selection: Fixed typo on useTextSelection document

Motonosuke:
from ...  `import { useClipboard } from '@mantine/hooks';`
to ...  `import { useTextSelection } from '@mantine/hooks';`


rtivital:
Thanks!

PR #2329, opened 2022-09-03
[@mantine/core] RingProgress: remove duplicate types

ryuhangyeong:
## Description

Refactored by removing duplicate types.


src/mantine-core/src/RingProgress/RingProgress.tsx
```diff
@@ -23,7 +24,7 @@ export interface RingProgressProps
   roundCaps?: boolean;
 
   /** Ring sections */
-  sections: { value: number; color: MantineColor; tooltip?: React.ReactNode }[];
+  sections: CurveData[];
```
rtivital, line 27: This change will remove type information from props table and users will not know what properties `sections` supports.

ryuhangyeong: sorry. I didn't consider what you said. I will close the issue.

PR #2335, opened 2022-09-04
[@mantine/carousel] Carousel: add options `skipSnaps` and `containScroll`

italodeandra:
Current workaround is to reinit embla instance.

```ts
const [embla, setEmbla] = useState<Embla | null>(null);
useEffect(() => {
  if (embla) {
    embla.reInit({ containScroll: "trimSnaps", skipSnaps: true });
  }
}, [embla]);
```

So to me it seems reasonable to add it directly to Mantine.

### DIFF
--- a/src/mantine-carousel/src/Carousel.tsx
+++ b/src/mantine-carousel/src/Carousel.tsx
@@ -104,6 +104,12 @@ export interface CarouselProps
 
   /** Previous control icon */
   previousControlIcon?: React.ReactNode;
+
+  /** Allow the carousel to skip scroll snaps if it's dragged vigorously. Note that this option will be ignored if the dragFree option is set to true, false by default */
+  skipSnaps?: boolean
+
+  /** Clear leading and trailing empty space that causes excessive scrolling. Use trimSnaps to only use snap points that trigger scrolling or keepSnaps to keep them. */
+  containScroll?: 'trimSnaps' | 'keepSnaps'
 }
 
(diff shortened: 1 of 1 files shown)


italodeandra:
Not sure what this error from the workflow means:

```
FAIL src/mantine-carousel/src/Carousel.test.tsx
  ● Console

    console.error
      Error: Uncaught [TypeError: Cannot read property 'left' of undefined]
```

I didn't really change anything on the Carousel itself so it would break on the test?

rtivital:
It may have broken something in embla integration, you can run tests locally with `npr jest src/mantine-carousel` command

italodeandra:
> It may have broken something in embla integration

But I just added the possibility to add two properties without changing the test, so if those properties aren't being used, the test shouldn't break. Weird. 

Does the test do something automatically that I don't have control of? I'll check it later.

italodeandra:
After some investigations, I discovered that this error happens when you try to load the option `containScroll: "trimSnaps"` but without any slides (empty). I'll fix it.

rtivital:
@italodeandra any progress on this?

italodeandra:
@rtivital 

Just pushed the fix for the property, by using the right default values. 

Loading an empty carousel with the `containScroll: "trimSnaps"` is a separate issue that I don't think should be fixed on the library level. What do you think?

rtivital:
It is managed by embla, should be fine.

rtivital:
Thanks!

PR #2336, opened 2022-09-04
[@mantine/carousel] CarouselSlide: remove margin from the last slide

italodeandra:
Removes this last margin:

image


rtivital:
Thanks!

PR #2351, opened 2022-09-04
[@mantine/core] Drawer: `withOverlay={false}` should let you click-through

italodeandra:


rtivital:
Hi @italodeandra, we do not need separate prop for that, if `withOverlay` is set to false, then it should allow click thought.

italodeandra:
Hey @rtivital,

I also thought that, but it only removes the overlay styles. To be able to click through, I still need to do this:

```tsx
<Drawer
    opened={opened}
    onClose={handleDrawerClose}
    title="Filme"
    padding="xl"
    size="xl"
    position="right"
    withOverlay={false}
    sx={{
      pointerEvents: "none",
      "& > *": {
        pointerEvents: "auto",
      },
    }}
  >
```

So perhaps I should remove the `clickThrough` from the Overlay and just keep `overlayClickThrough` on the Drawer?

italodeandra:
> Check if `withOverlay={false}` and add those styles on library level, there is no need for extra prop for that.

Done.

rtivital:
Thanks!

ashishsurya:
I dont thibk this works

heejj1206:
I don't think it's solved

italodeandra:
Hey bros, my changes were removed by another PR a couple years ago. I don't use the lib anymore so I think you should create a new issue.

PR #2377, opened 2022-09-07
[mantine/core] Switch : Add thumbIcon prop and onLabel, offLabel can now support ReactNode

DhrjArun:


rtivital:
@DhrjArun component should adapt to labels with different sizes, currently it becomes messy if label is too large
```tsx
<Switch label="Hello" onLabel="Another label" offLabel="Large label" value="1" m="xl" />
```
Снимок экрана 2022-09-07 в 14 46 15

DhrjArun:
@rtivital 

only width? or do you also want the height to increase based on the label size?

rtivital:
No, only width should be increased

DhrjArun:
Can I put onLabelWidth and offLabelWidth props? just like we have rightSectionWidth prop in the input component


Screenshot 2022-09-08 at 11 09 05
Screenshot 2022-09-08 at 11 10 06

DhrjArun:
@rtivital

forgot to mention

DhrjArun:
or Instead of onLabelWidth, offLabelWidth, we can have trackWidth prop.

rtivital:
@DhrjArun Is not it possible to implement it without these props?

DhrjArun:
@rtivital  I have added one prop, trackWidth. check it.

I think it can be done without trackWidth prop, but it would look like this:
Screenshot 2022-09-08 at 11 10 06

Screenshot 2022-09-08 at 11 09 05

zhulien-ivanov:
Using fixed width defeats the purpose of introducing this change. You can checkout the implementation of Switch in antd: https://ant.design/components/switch/. It should simply render the children and it should work dynamically without the need for a fixed width. Otherwise it causes more headaches than actually helps, tbh.

DhrjArun:
@zhulien-ivanov  It can be done. I did the same thing, like ant-ui previously. but what happens with it is if onLabel and offLabel is of different sizes then switch will have different size on different state.

And if we make it same size, then It will look like this: 
image

DhrjArun:
I will do it just like ant-ui then.

DhrjArun:
@rtivital I did onLabel and offLabel without any props or giving fixed value. Do you also want animation when state change?


https://user-images.githubusercontent.com/69076988/189277345-909790a5-7b88-4b74-8cc6-6d0a1c06d5fe.mov

rtivital:
Yep, animation will be super cool

DhrjArun:
@rtivital Is it fine?

https://user-images.githubusercontent.com/69076988/189284954-1a46ae72-42a3-47d9-b970-ee33653fb666.mov

rtivital:
Thanks!

PR #2380, opened 2022-09-07
[@mantine/core] Theme: add support other unit

auronsan1st:
Add support theme with other unit on MantineSizes.  
using reference to-px library for convert string to px

### DIFF
--- a/src/mantine-styles/src/theme/utils/to-px/to-px.ts
+++ b/src/mantine-styles/src/theme/utils/to-px/to-px.ts
@@ -1,0 +1,29 @@
+const PIXELS_PER_INCH = 96;
+
+const defaults = {
+  ch: 8,
+  ex: 7.15625,
+  em: 16,
+  rem: 16,
+  in: PIXELS_PER_INCH,
+  cm: PIXELS_PER_INCH / 2.54,
+  mm: PIXELS_PER_INCH / 25.4,
+  pt: PIXELS_PER_INCH / 72,
+  pc: PIXELS_PER_INCH / 6,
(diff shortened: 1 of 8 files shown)


src/mantine-styles/src/theme/utils/to-px/to-px.ts
```diff
@@ -0,0 +1,29 @@
...
+function parseUnit(str: string): [number, string] {
+  const out: [number, string] = [0, 'px'];
+  const currentString = String(str);
+  const num = parseFloat(currentString);
+  out[0] = num;
+  out[1] = currentString.match(/[\d.\-+]*\s*(.*)/)[1] || '';
+  return out;
+}
+
+export function toPx(str: string) {
```
rtivital, line 25: That's not really a great idea. I want to provide support for other units to add an option to switch what units are used in all components. For example, when the user adds all units as `rem` in theme, then all components will use `rem` units. It defeats the purpose of this new feature if we just convert these units.

Furthermore, conversion is not accurate as rem, em units are relative to parent/html tag. There are no `vh/vw` units.

I guess it is not fully possible to implement what I've described without breaking, but I would still appreciate a PR that will convert all operations with number theme tokens to `calc`.

auronsan: oh did you mean just need change one by one component styles the number theme tokens to calc?
ex : `marginBottom: theme.spacing.xl * 2,` to `marginBottom: `calc(${ theme.spacing.xl} * 2)` @rtivital

PR #2384, opened 2022-09-07
fix: allow select item value for other types

mmahalwy:


rtivital:
Hi @mmahalwy, Select does not support values apart from strings because it:
- is how native select element works
- is much easier to maintain
- can produce all sorts of issues if we decide to support non-primitive values

Select and other similar components are already pretty complex, and I would rather not waste a lot of time maintaining this feature.

PR #2386, opened 2022-09-08
[docs] use-list-state: Fix filter description

### LINKED ISSUE #2385: example filter usage in `use-list-state` is reversed
### Link to the page where something nasty is located

https://mantine.dev/hooks/use-list-state/

### Exact quote of what is wrong

```js
// remove objects that have a = 'new-prop'
const filter = () => handlers.filter((item) => item.a === 'new-prop');
// values -> [{ a: 8 }]
```
should be
```js
// filter objects that have a = 'new-prop'
const filter = () => handlers.filter((item) => item.a === 'new-prop');
// values -> [{ a: 'new-prop' }]
```
because implementation uses plain filter.
https://github.com/mantinedev/mantine/blob/e267c1dfa4356566ad70b47cde9aa008ec7ff991/src/mantine-hooks/src/use-list-state/use-list-state.ts#L86-L88

### Are you willing to create pull request with the fix?

Yes

scarf005:
- fixed #2385

### DIFF
--- a/docs/src/docs/hooks/use-list-state.mdx
+++ b/docs/src/docs/hooks/use-list-state.mdx
@@ -64,9 +64,9 @@ const setItem = () => handlers.setItem(0, { a: 8 });
 const setItemProp = () => handlers.setItemProp(1, 'a', 'new-prop');
 // values -> [{ a: 8 }, { a: 'new-prop' }]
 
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #2398, opened 2022-09-08
[docs] Include usage of Radio.Group name prop and fix Calendar example

nmay231:
Previously, if you copied the [example from the Calendar Styles API demo](http://localhost:7521/dates/calendar/#customize-with-styles-api), you were unable to switch the current month.

I noticed you updated to `useState<Date | null>`. Sneaky sneaky :grin:


rtivital:
Thanks!

PR #2420, opened 2022-09-11
[@mantine/rte] fix prop types

Jawell:
- Quill can proceed undefined values
- defaultValue is related to defaultValue of HTML input, this is not right
- Also, onChange can be undefined because field can be readOnly

### DIFF
--- a/src/mantine-rte/src/components/RichTextEditor/RichTextEditor.tsx
+++ b/src/mantine-rte/src/components/RichTextEditor/RichTextEditor.tsx
@@ -43,7 +43,7 @@ function defaultImageUpload(file: File): Promise<string> {
 
 export interface RichTextEditorProps
   extends DefaultProps<RichTextEditorStylesNames>,
-    Omit<React.ComponentPropsWithoutRef<'div'>, 'onChange'> {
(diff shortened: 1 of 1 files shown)


src/mantine-rte/src/components/RichTextEditor/RichTextEditor.tsx
```diff
@@ -45,10 +45,13 @@ export interface RichTextEditorProps
   extends DefaultProps<RichTextEditorStylesNames>,
     Omit<React.ComponentPropsWithoutRef<'div'>, 'onChange'> {
   /** HTML content, value not forced as quill works in uncontrolled mode */
-  value: string | Delta;
+  value?: string | Delta;
+
+  /** Initial value of input */
+  defaultValue?: string | Delta;
```
rtivital, line 51: This will not work because the component does not pass thought defaultValue to Editor component. Either set defaultValue or remove this prop.

Jawell: Done! Thank you, I pass it

rtivital:
Thanks!

PR #2433, opened 2022-09-13
Add onChangeEnd prop on ColorPicker component

DhrjArun:
**Prop**
> Just like Slider onChangeEnd prop
```onChangeEnd?: (value: string) => void;```

**Effected Components**
- ColorPicker
- ColorInput

### DIFF
--- a/src/mantine-core/src/ColorInput/ColorInput.tsx
+++ b/src/mantine-core/src/ColorInput/ColorInput.tsx
@@ -134,6 +134,9 @@ export const ColorInput = forwardRef<HTMLInputElement, ColorInputProps>(
     };
 
     useEffect(() => {
+      if (isColorValid(_value)) {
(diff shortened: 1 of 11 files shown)


rtivital:
@DhrjArun `onChangeEnd` should be called:
- when user selects color by clicking on color swatch
- with valid value when user enters text in ColorInput

DhrjArun:
@rtivital currently onChangeEnd is being called on
- arrow buttons
- saturation slider
- hue slider
- alpha slider

Do you want this behaviour?

rtivital:
@DhrjArun It actually does not matter where onChangeEnd is called, what's important is that it is called every time user finishes picking color, this includes clicking color swatch and changing color in input

DhrjArun:
@rtivital , let me know if anything is missing

rtivital:
Works great, thanks!

PR #2434, opened 2022-09-13
Add cache key in classNames prefix

Yomyer:
Right now if you define a different key when you create a custom EmotionCache, it only changes in the serialized class "newkey-j45j320" with this small feature it would also change the prefix of the classes.


rtivital:
Have you checked whether it works correctly with ssr?

rtivital:
Thanks!

PR #2440, opened 2022-09-14
[@mantine/core] Button: Button is disabled when inside fieldset disabled

### LINKED ISSUE #2428: Button not disabled when inside HTML <fieldset disabled>
### What package has an issue

@mantine/core

### Describe the bug

When making a React form, it's common to wrap the entire form inside a [<code>fieldset</code>](https://www.w3schools.com/tags/att_fieldset_disabled.asp) element, so you can quickly disable the entire form while it is submitting. 

The bug is that the Button component is not set as disabled when inside a fieldset element.

TextInput, Switch and Checkbox are working as expected, and do get disabled.

```tsx
import {
  Button,
  Checkbox,
  MantineProvider,
  Stack,
  Switch,
  TextInput,
  Text
} from "@mantine/core";
import { useToggle } from "@mantine/hooks";

  const [disabled, toggleDisabled] = useToggle();
  return (
    <MantineProvider withGlobalStyles withNormalizeCSS>
      <Stack p="md">
        <Text>{`Form is currently: [${
          disabled ? "disabled" : "enabled"
        }]`}</Text>
        <Button onClick={() => toggleDisabled()}>toggle disable state</Button>

        <fieldset disabled={disabled}>
          <Stack>
            <TextInput
              label="Text input"
              placeholder="is correctly disabled when inside fieldset"
            ></TextInput>
            <Switch defaultChecked label="Switch get disabled!"></Switch>
            <Checkbox label="Checkbox gets disabled!"></Checkbox>
            <Button
              variant="outline"
              onClick={() => alert("should be disabled")}
            >
              <span role="img" aria-label="bug">
                🐞
              </span>
              should get disabled but isn't
            </Button>
            <button type="button" onClick={() => alert("basic html button")}>
              basic html button gets disabled
            </button>
          </Stack>
        </fieldset>
      </Stack>
    </MantineProvider>
  );
}
```

I wonder if it's something weird happening when using the Box element as the root for UnstyledButton, as switch and checkbox components (which are working fine) have an extra <input> html element inside the box, but Unstyled button only returns a Box with component="button".

5.2.5

https://codesandbox.io/s/confident-ishizaka-dt9y3d?file=/src/App.tsx

No

No

### Possible fix

_No response_

josiahayres:
Fixes #2428 using adding a disabled style to button, as suggested by @UgurKacak.

I added simple storybook story, wrapping the states example in a fieldset disabled, all buttons should be disabled.

I have added a basic test as well, please provide feedback if you'd like to see more details on top.

### DIFF
--- a/src/mantine-core/src/Button/Button.styles.ts
+++ b/src/mantine-core/src/Button/Button.styles.ts
@@ -135,6 +135,15 @@ export default createStyles(
 
       '&:active': theme.activeStyles,
 
+      '&:disabled': {
+        borderColor: 'transparent',
+        backgroundColor: theme.colorScheme === 'dark' ? theme.colors.dark[4] : theme.colors.gray[2],
+        color: theme.colorScheme === 'dark' ? theme.colors.dark[6] : theme.colors.gray[5],
+        cursor: 'not-allowed',
+        backgroundImage: 'none',
+        pointerEvents: 'none',
+      },
+
       '&[data-disabled]': {
         borderColor: 'transparent',
         backgroundColor: theme.colorScheme === 'dark' ? theme.colors.dark[4] : theme.colors.gray[2],
--- a/src/mantine-core/src/Button/Button.styles.ts
+++ b/src/mantine-core/src/Button/Button.styles.ts
@@ -135,21 +135,13 @@ export default createStyles(
 
       '&:active': theme.activeStyles,
 
-      '&:disabled': {
+      '&:disabled, &[data-disabled]': {
         borderColor: 'transparent',
         backgroundColor: theme.colorScheme === 'dark' ? theme.colors.dark[4] : theme.colors.gray[2],
         color: theme.colorScheme === 'dark' ? theme.colors.dark[6] : theme.colors.gray[5],
         cursor: 'not-allowed',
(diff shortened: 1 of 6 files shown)


src/mantine-core/src/Button/Button.styles.ts
```diff
@@ -135,6 +135,15 @@ export default createStyles(
 
       '&:active': theme.activeStyles,
 
+      '&:disabled': {
```
rtivital, line 138: Should be `&:disabled, &[data-disabled]` to avoid styles duplication

rtivital (commented):
Please also add the same changes to ActionIcon component

src/mantine-core/src/Button/Button.styles.ts
josiahayres: @rtivital sure I will add this to ActionIcon too. 

Should I be adding this as a new style (as in this screenshot) or simply updating the existing style (the one marked 2 in screenshot): 
image

rtivital: Update the existing one

rtivital:
Thanks!

PR #2462, opened 2022-09-16
SegmentedControl value: null

alexbakers:
Hello, friends.

image

Ability to set null so that none of the options is selected.

`value: "Hello"`

image

`value: undefined`

image

`value: "none"`

image

`value: null`

image


rtivital:
SegmentedControl works as expected, null value should not be supported

alexbakers:
> SegmentedControl works as expected, null value should not be supported

How to make that no option is selected without using null?

image

rtivital:
Segmented control does not support not having any option selected

alexbakers:
> Segmented control does not support not having any option selected

Support (`value = null`)

image

image

PR #2463, opened 2022-09-16
[@mantine/modals] ConfirmModal: Adds pending confirm function

Angusross111:
This function allows me to send off an external request on clicking the confirm button and waits until it completes before closing, while also setting the loading state on the confirm button untill it completes.
Happy to change up the name to something more descriptive.


rtivital:
Hi @Angusross111, thanks for a PR, but we do not need this feature as a part of the library

Issue #2436, opened 2022-09-13
closeModal not working with openModal


Eliasmgprado:
### What package has an issue

@mantine/modals

### Describe the bug

Want to open mutiple modals with openModal() function and close the modals with closeModal() function, using the modalId provided to the openModal() function. However, closeModal() is closing all opened modals, insted of the modal with the provided modalId.

### What version of @mantine/hooks page do you have in package.json?

5.3.0

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/embed/thirsty-archimedes-bydsjh?fontsize=14&hidenavigation=1&theme=dark

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

No

Eliasmgprado:
The example uses openContextModal instead of openModal, the problem is with openModal function. As showed in the code sandbox, when multiple modal are opened with openModal function, you can`t close the modals with closeModal() function, providing the modalId, passes to the props of openModal(). The closeModal() function close all the opened modals, and not just the one of the provided id.

UgurKacak:
Hello,
The top right close button is working as expected. However, closeModal("butonId") method call inside <Button> onClick, not working as expected. Because while this call happened, ModalProvider's closeModal function faced an empty state instead of the state that holds two opened modals. I think the problem is here.

rtivital:
Fixed in 5.3.3

PR #2464, opened 2022-09-16
[@mantine/utils] fix: closeModal not working with openModal

lawff:
#2436, when creating the closeModal function, the state is empty. But openModal will change the state modals, so I think we should update the handlers. Maybe there are other problems with this, I think we can update the handlers to get the latest state in all cases. Not sure of the best practice for this. Please give feedback or give improvement advice.


rtivital:
Thanks!

Issue #2467, opened 2022-09-16
`Skeleton` ignores `classNames`


108EAA0A:
### What package has an issue

@mantine/core

### Describe the bug

Passing `classNames` to `Skeleton` is ignored.
As a result, `classnames="[object Object]"` is output to the DOM.
![SS 2022-09-16 15 00 17](https://user-images.githubusercontent.com/23615952/190567825-1b3dfbe8-1a36-4259-bc2c-04d1c6fe92cc.png)

Perhaps it's because you're not explicitly receiving and processing `classNames` props in `Skeleton`.
https://github.com/mantinedev/mantine/blob/5cf81d288a7c47c115b2dd92efe9a6b0008fdb45/src/mantine-core/src/Skeleton/Skeleton.tsx#L36

### What version of @mantine/hooks page do you have in package.json?

5.3.2

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/mantine-example-ypvqjp?file=/src/App.tsx

rtivital:
It is expected behavior, Skeleton renders one node, you can add styles to it with sx or className props

108EAA0A:
I didn't know that, thanks!
But then, wouldn't it be better not to accept `classNames` ?
It is unkind that there is no error despite the unintentional rendering to the DOM.
I think the type definition needs to be improved.

rtivital:
I wasn't able to make it work, if you know how to implement that, feel free to send a PR to DefaultProps type

108EAA0A:
@rtivital
Since this is a Skeleton-specific issue, we have created a PR against SkeletonProps.
https://github.com/mantinedev/mantine/pull/2471

PR #2471, opened 2022-09-16
[@mantine/core] Skeleton: Omit classNames props

108EAA0A:
There are no type errors with respect to arguments that do nothing when received, so `classnames="[object Object]"` is unintentionally rendered in the DOM.
https://github.com/mantinedev/mantine/issues/2467#issuecomment-1249020944

Improvements have been made to omit classNames props from `SkeletonProps` and pass classNames props to cause a type error.


rtivital:
Hi @108EAA0A, it is not specific to Skeleton, this solution is not good

PR #2485, opened 2022-09-18
Controlled search in Select/MultiSelect

dipiash:
Added controlled/uncontrolled searchValue prop into Select/MultiSelect: (Original discussion: https://github.com/mantinedev/mantine/discussions/2480)


rtivital:
Thanks!

PR #2509, opened 2022-09-20
[@mantine/core] Switch: add data-disabled + change color when disabled on label 

### LINKED ISSUE #2507: Disabled Mantine labels are inconsistent for radios and checkboxes.
### Link to the page where something nasty is located

https://codesandbox.io/s/wild-sunset-6zkf4u?file=/src/App.tsx

### Exact quote of what is wrong

Now when we disable the radio we have the label of the input changing color and the data-disabled is changed, which will also make styling the component easier. 
Capture d’écran 2022-09-20 à 20 04 59

I would like to have the same behaviour for the checkbox.

### Are you willing to create pull request with the fix?

Yes

omarsy:
fixes #2507


rtivital:
Thanks!

PR #2519, opened 2022-09-21
[@mantine/core] Table: add withBorder and withColumnBorders props

DhrjArun:


rtivital:
@DhrjArun does it also include `size` prop?

DhrjArun:
@rtivital You said `size` prop should change fontSize, and there already `fontSize: MantineNumberSize` prop.

rtivital:
Thanks!

Issue #1508, opened 2022-05-24
RichTextEditor image upload (Cannot read properties of null (reading 'index'))


timahuna:
### What package has an issue

@mantine/rte

### Describe the bug

1. Wrap react app in a Strict Mode component. 
2. Add a RichTextEditor component. 
3. Add an image via the embed image icon in the toolbar:
<img width="606" alt="image" src="https://user-images.githubusercontent.com/104216931/170070300-2a62d14a-3f1c-45cc-8d9c-8bc539cad3a7.png">
4. You should see the error `Cannot read properties of null (reading 'index')`

Drag/drop works, pasting works, just the toolbar is causing this error (when wrapped in `<StrictMode>`). 

### In which browser did the problem occur

Chrome

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/bold-field-ilkjez?file=/src/App.js

### Do you know how to fix the issue

No

### Possible fix

Looks like this.range is reset in the [ImageUploader](https://github.com/mantinedev/mantine/blob/master/src/mantine-rte/src/modules/image-uploader/ImageUploader.ts#L139) in strict mode.  

I could be wrong (and I haven't tested this) but I think we might want to set the range [here](https://github.com/mantinedev/mantine/blob/ea0b78ca7bca36c821f34ff34f47d161f82a1288/src/mantine-rte/src/modules/image-uploader/ImageUploader.ts#L139) via `this.range = this.quill.getSelection();`.

wes337:
Duplicate of or closely related to #1386

rtivital:
Fixed in 4.2.7

mvskvortsov:
I'm still have this error on almost all of toolbar button: lists, image... And "addRange(): The given range isn't in document." on other toolbar buttons.

wes337:
> I'm still have this error on almost all of toolbar button: lists, image... And "addRange(): The given range isn't in document." on other toolbar buttons.

I'm seeing it on lists too. But it is fixed with images, just make sure you're using at least version 4.2.7

rtivital:
The issue happens only in with react strict mode, disable strict mode to fix it. The issue cannot currently be resolved on library side.

lucasvilaca27:
### ###

lucasvilaca27:
****

Issue #1386, opened 2022-05-04
error: addRange(): The given range isn't in document.


VeoScript:
### What package has an issue

@mantine/rte

### Describe the bug

I found an error while I uploading some photo in RichTextEditor.

Error: **quill.js?99f4:3195 addRange(): The given range isn't in document.**

![image](https://user-images.githubusercontent.com/26340308/166686595-3ed15599-d4a0-4b2b-8e4b-406a05e04e9b.png)

![image](https://user-images.githubusercontent.com/26340308/166686803-05214318-c266-43bf-9b5d-834161f45b3b.png)

### In which browser did the problem occur

Chrome

### Possible fix

I think the error is from the latest version of Quill.

wes337:
This is a duplicate of #1288

rtivital:
The PR was merged, will be fixed with next patch, thanks @wes337!

Hartaithan:
> The PR was merged, will be fixed with next patch, thanks @wes337!

I have updated mantine to version 4.2.5, but I am still seeing this problem.
RichTextEditor file - [link](https://github.com/Hartaithan/next-posts/blob/main/components/Editor.tsx)
Code - [link](https://github.com/Hartaithan/next-posts)
Demo - [link](https://next-posts-hartaithan.vercel.app/) (Authorization is required to access the editor page)

Hartaithan:
@wes337 Any updates?

wes337:
I will take another look at this. It seems to be a bug from quilljs and there might be a workaround.

Would you be able to provide a code sandbox where this issue can be reproduced?

rtivital:
Fixed in 4.2.7

Hartaithan:
@rtivital @wes337 
Updated to version 4.2.7. Still seeing this problem. But now file input is not triggered multiple times. And the image is displayed blurred, even after successfully loading the image.
[...]

PR #2529, opened 2022-09-21
Rte/image uploader index error

N-Argyle:
There are at least two error reports regarding the same error when uploading images (for me, specifically pasting in images) into rte. 

`Error: Cannot read properties of null (reading 'index')`

They were marked closed but the issue persists for me, even when running mantine in storybook. 
Changing the settimeout period on the handlePaste function of ImageUploader.ts to anything other than 0 resolves the issue. I've set it to 1ms. All tests pass, all three image upload methods work as intended.

https://github.com/mantinedev/mantine/issues/1508
https://github.com/mantinedev/mantine/issues/1386


rtivital:
Thanks!

Issue #2359, opened 2022-09-05
Leave animation should be the progress getting complete instead of resetting to the start


italodeandra:
### What package has an issue

@mantine/nprogress

### Describe the bug

When the navigation progress completes, the animation of the progress should be to fill the entire width instead of going to 0.

Reference: https://ricostacruz.com/nprogress/

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

rtivital:
Feel free to send a PR with this feature

rtivital:
Help is still wanted on this, if someone wants to submit a PR feel free to ping me here.

pmalliotas:
Hi @rtivital. I have a fix ready for this issue. Should i submit a PR?

zulianrizki:
> Help is still wanted on this, if someone wants to submit a PR feel free to ping me here.

i tried to fix this issue. The idea is instead of calling `resetNavigationProgress` when the page load is done which will set the progress to 0 and reset the progress, now can use `finishNavigationProgress` to set the progress to 100, so progress will fill the entire width and then set the `autoReset` to `true` to automatically reset the progress bar.

rtivital:
Fixed in 5.5.0

PR #2545, opened 2022-09-24
[@mantine/nprogress] add finishNavigationProgress

zulianrizki:
tried to resolve this issue https://github.com/mantinedev/mantine/issues/2359
the idea is instead of calling `resetNavigationProgress` when the page load is done which will set the progress to 0 and reset the progress, now can use `finishNavigationProgress` to set the progress to 100, so progress will fill the entire width and then set the `autoReset` to `true` to automatically reset the progress bar.


rtivital:
Thanks!

PR #2546, opened 2022-09-24
Update docs to include import useParams

keypuncherlabs:
Update docs to include import useParams for example code to work properly.


rtivital:
Thanks!

PR #2548, opened 2022-09-25
Allow `auto` value for prop `slidesToScroll`

7iomka:
Allow `auto` value for prop `slidesToScroll` 
Docs:  https://www.embla-carousel.com/api/options/#slidestoscroll


rtivital:
Thanks!

PR #2554, opened 2022-09-26
[docs] Text inherit - to have consistency between example and code example

TheKinng96:
I am using mantine for my project and currently heavily reading docs. 
Found one example that got me confused and thought to have it fixed by having consistency between the example and the code example, so users can just copy and paste to their code base and start modifying. 

It is not a big deal but hopes it can make the documentation even better


src/mantine-demos/src/demos/core/Text/Text.demo.inherit.tsx
```diff
@@ -5,7 +5,13 @@ const code = `
 import { Text, Title } from '@mantine/core';
 
 function Demo() {
-  return <Title order={3}>Highlight <Text color="blue" inherit component="span">something</Text>in title</Title>;
+  return <Title order={3}>
```
rtivital, line 8: It is formatted incorrectly, should be `<Title order={3}>Title in which you want to<Text color="blue" inherit component="span">highlight</Text>something</Title>`

TheKinng96: let me update it

rtivital:
Thanks!

PR #2581, opened 2022-09-29
Breakpoints usage fix

### LINKED ISSUE #2550: Wrong mediaQueries calculations from breakpoints
### What package has an issue

@mantine/core

### Describe the bug

Hi.
I started using this library but just now I noticed that the math used to calculate mediaqueries is wrong
I have over 8 years in layout and most libraries like `bootstrap` use the following calculation model:

`N` - breakpoint shorthand (like `sm`, `md`, ...)

- MobileFirst approach
   -> minWidth N -> `@media (min-width: ${N}px)`
   -> maxWidth N -> `@media (max-width: ${N-1}px`
  
- DesktopFirst approach
   -> maxWidth N -> `@media (max-width: ${N}px)`
   -> minWidth N -> `@media (min-width: ${N+1}px`

   
Also, for this 2 approaches value for N in config will be different, because usually breakpoints are set up in anticipation of a change in device type or orientation.
For example, we know that
- minimal supported mobile device has width 360px,
- tablets starts from 768px (popular res. for tablets is 1024x768)
- desktops  starts from 1200px width

Config looks like this
```
const breakpoints = {
  xs: 360,
  md: 768,
  xl: 1200,
};
```

All modern libraries have long since passed as mobileFirst approach,
So, if I use `tailwindcss`, and configure it to use the same `breakpoints` like I used in mantine, currently i get unexpected result:

Tailwind get me correct and expected code
`md:text-black` -> `@media (min-width: 768px){ color: black }`

However, Mantine, for some unknown reason used mobileFirst approach, but `N` is always increased by 1.

Mantine
`<Grid.Col md={6}>1</Grid.Col>` -> `@media (min-width: 769px){ color: black }`

How you see, `769px` value is **unexpected**
This value does not cover the interval expected by the user (in fact, when accessing the site from a tablet with a width of 768 pixels, the user will still see the settings for the previous breakpoint)

This issue can be fixed inside all components that uses `breakpoints` config **_directly_** (like `Grid`), or **_indirectly_** (components like `Carousel` which has the ability to set different settings depending on the breakpoint)

I am facing this issue now when I am trying to hotfix [this issue](https://github.com/mantinedev/mantine/issues/2447) locally.

This issue needed to be fixed internally, not by the end-user, because even if I use the correct values for the breakpoints in my custom styles for example in the carousel, I will still get unexpected results, since my styles will be overwritten by the core styles, because, as in example from above, 769px breakpoint will take precedence.

The solution to this problem, I believe, will not affect user expectations in terms of something that breaks the current implementation, but for users like me using common configuration files for several libraries that work with breakpoints correctly, this is a critical problem that needs to be addressed, otherwise I can't make working customization through breakpoints at all.

latest

_No response_

_No response_

_No response_

### Possible fix

In case of Carousel
```ts
const property = 'maxWidth' in breakpoint ? 'max-width' : 'min-width';
const breakpointSize = theme.fn.size({
  size: property === 'max-width' ? breakpoint.maxWidth : breakpoint.minWidth,
  sizes: theme.breakpoints,
});

acc[`@media (${property}: ${breakpointSize + (property === 'max-width' ? 0 : 1)}px)`] = {}
```

String with mediaQuery needed to be changed to

```ts
`@media (${property}: ${breakpointSize - (property === 'max-width' ? 1 : 0)}px)`
```

In case of Grid Col
```
acc[`@media (min-width: ${theme.breakpoints[size] + 1}px)`] = {}
```

becomes

```
acc[`@media (min-width: ${theme.breakpoints[size]}px)`] = {
```

7iomka:
Closes #2550
Docs examples have not been affected, because used just for demo and are not critical in terms of actual breakpoints for the end user.

### DIFF
--- a/src/mantine-core/src/AppShell/AppShell.styles.ts
+++ b/src/mantine-core/src/AppShell/AppShell.styles.ts
@@ -29,11 +29,11 @@ function getPositionStyles(props: AppShellStylesParams, theme: MantineTheme): CS
     paddingLeft: `calc(var(--mantine-navbar-width, 0px) + ${padding}px)`,
     paddingRight: `calc(var(--mantine-aside-width, 0px) + ${padding}px)`,
 
-    [`@media (max-width: ${navbarOffset}px)`]: {
+    [`@media (max-width: ${navbarOffset - 1}px)`]: {
       paddingLeft: padding,
     },
 
-    [`@media (max-width: ${asideOffset}px)`]: {
+    [`@media (max-width: ${asideOffset - 1}px)`]: {
       paddingRight: padding,
     },
   };
--- a/src/mantine-core/src/AppShell/HorizontalSection/HorizontalSection.styles.ts
+++ b/src/mantine-core/src/AppShell/HorizontalSection/HorizontalSection.styles.ts
@@ -38,7 +38,7 @@ export default createStyles(
     const breakpoints =
       typeof width === 'object' && width !== null
         ? getSortedBreakpoints(width, theme).reduce((acc, [breakpoint, breakpointSize]) => {
-            acc[`@media (min-width: ${breakpoint + 1}px)`] = {
+            acc[`@media (min-width: ${breakpoint}px)`] = {
               width: breakpointSize,
               minWidth: breakpointSize,
             };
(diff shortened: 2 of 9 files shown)


rtivital:
Thanks!

PR #2598, opened 2022-10-01
[@mantine/core] Fix the position of `errorProps`

A-Marzoug:
`errorProps` were overridden by `sharedProps` as a result the error props had no effect.


rtivital:
Current behavior is correct: `classNames`, `styles`, `unstyled`, `size` and `__staticSelector` props should be always controlled by `Input.Wrapper` component.

PR #2602, opened 2022-10-02
[@mantine/core] MultiSelect: fix dropdown flicker and ondropdown handlers being called when dropdown isn't visible

### LINKED ISSUE #2503: MultiSelect - a regression makes the MultiSelect unusable in some circunstances
### What package has an issue

@mantine/core

### Describe the bug

This is a regression as the same symptoms mentioned in #2059 are obtained for the same props given to MultiSelect

This new issue may have been introduced by this commit https://github.com/mantinedev/mantine/commit/80176313c6a5c58aee2bcc6917e79c293e36af7d

This line https://github.com/mantinedev/mantine/blob/5.4.0/src/mantine-core/src/MultiSelect/MultiSelect.tsx#L333 makes the dropdown close immediately after focus if the filteredData array is empty. By default the defaultFilter returns false for any selected item, thus making the filteredData empty when all items are selected.

One workaround is to provide a filter that will return always true and to manage filtering outside of the component.

5.4.0

https://codesandbox.io/s/repro-multi-select-dropdown-5-4-0-h1jsnj

_No response_

_No response_

### Possible fix

_No response_

achmurali:
Fixes #2503

### DIFF
--- a/src/mantine-core/src/MultiSelect/MultiSelect.tsx
+++ b/src/mantine-core/src/MultiSelect/MultiSelect.tsx
@@ -226,7 +226,7 @@ export const MultiSelect = forwardRef<HTMLInputElement, MultiSelectProps>((props
   const wrapperRef = useRef<HTMLDivElement>();
   const itemsRefs = useRef<Record<string, HTMLDivElement>>({});
   const uuid = useId(id);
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #2605, opened 2022-10-02
[@mantine/core] Select: close dropdown on click

### LINKED ISSUE #2586: Inconsistent focus management with `MultiSelect` and `Select`
### What package has an issue

@mantine/core

## Problem

> To understand the below lines please refer to the attached Codesandbox

Clicking on a `<MultiSelect>` input/trigger toggles the open state for the dropdown. In simple words: If the dropdown is closed and i click on the input/trigger it opens and if i click again it closes.

This behaviour is different to `<Select>` where a click on input/trigger with an opened dropdown **does not** toggle the dropdown. This only happens if the `<Select>` is created with a `searchable={true}` prop. If `searchable={true}` this code part makes sure the dropdown is always open on input focus: https://github.com/mantinedev/mantine/blob/master/src/mantine-core/src/Select/Select.tsx#L506

## Expected

I would expect that both `<MultiSelect>` and `<Select>` behave the same if clicked independent of the value of `searchable`. I suggest to also toggle the dropdown for `<Select>` (adapt behaviour of `<MultiSelect>`)

5.1.1

https://codesandbox.io/s/happy-drake-qdc5j7?file=/src/App.tsx

Yes

Yes

### Possible fix

Replace `handleInputFocus` at https://github.com/mantinedev/mantine/blob/master/src/mantine-core/src/Select/Select.tsx#L506 with (not tested)
```typescript
const handleInputFocus = (event: React.FocusEvent<HTMLInputElement>) => {
  typeof onFocus === 'function' && onFocus(event);

  setDropdownOpened(!dropdownOpened);
  if(dropdownOpened) {
    scrollSelectedItemIntoView();
  }
};
```

achmurali:
Fixes #2586

### DIFF
--- a/src/mantine-core/src/MultiSelect/MultiSelect.tsx
+++ b/src/mantine-core/src/MultiSelect/MultiSelect.tsx
@@ -226,7 +226,7 @@ export const MultiSelect = forwardRef<HTMLInputElement, MultiSelectProps>((props
   const wrapperRef = useRef<HTMLDivElement>();
   const itemsRefs = useRef<Record<string, HTMLDivElement>>({});
   const uuid = useId(id);
-  const [dropdownOpened, _setDropdownOpened] = useState(initiallyOpened);
+  const [dropdownOpened, setDropdownOpened] = useState(initiallyOpened);
   const [hovered, setHovered] = useState(-1);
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #2528, opened 2022-09-21
Unordered list indent fix

devstardude:
Issue #2494 
=> Keep the indentation with multiline contents of Item in List

Change in @mantine/core package at List/List.styles.ts component
Fix:
list-style-position: outside;

Reference:
https://www.w3schools.com/cssref/pr_list-style-position.asp

PR #2606, opened 2022-10-02
[@mantine/core] Fix indentation on List (listStylePosition: 'inside')

### LINKED ISSUE #2494: Keep the indentation with multiline contents of Item in List
### What package has an issue

@mantine/core

### Describe the bug

When content in List.Item is wrapping, it starts on start of a new line without the same space as mark has.
image

I think, this is because you use `list-style-position: inside`.
image

latest

_No response_

_No response_

_No response_

### Possible fix

If this is a native list (no custom icon, etc..)
for `itemWrapper` use styles
```css
    display: inline-block;
    vertical-align: top;
    max-width: calc(100% - 1.5em);
```

or as for custom one use flex
```css
    display: inline-flex;
    flex-direction: column;
    max-width: calc(100% - 1.5em);
```

`1.5em` is a value of a dot width plus default space between it and content of item.

Currently, there is no other way to fix this, except to make a custom implementation for the default list types

7iomka:
Revert #2528
Closes #2494

### DIFF
--- a/src/mantine-core/src/List/List.styles.ts
+++ b/src/mantine-core/src/List/List.styles.ts
@@ -15,6 +15,6 @@ export default createStyles((theme, { withPadding, size, listStyleType }: ListSt
     lineHeight: theme.lineHeight,
     margin: 0,
     paddingLeft: withPadding ? theme.spacing.xl : 0,
-    listStylePosition: 'outside',
+    listStylePosition: 'inside',
   },
 }));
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

vrtischev:
@7iomka, I'll have to revert it, the PR broke the component – https://mantine.dev/core/list/#with-icons

7iomka:
> @7iomka, I'll have to revert it, unless you send a PR with a fix, the PR broke the component – https://mantine.dev/core/list/#with-icons

Nope, I'll fix now

7iomka:
@rtivital #2624

PR #2608, opened 2022-10-02
[@mantine/core] fixed hook issue in segmented-control

jvdsande:
Fix an edge-case in SegmentedControl with dynamic data, where going from 0 option to >0 option or the other way around would crash because of broken rule of hooks


rtivital:
Thanks!

PR #2611, opened 2022-10-03
Fix mouse click for closing modal

MoreBlood:
Modal should not close on right mouse click, on 4/5 buttons and so on

### DIFF
--- a/src/mantine-core/src/Modal/Modal.tsx
+++ b/src/mantine-core/src/Modal/Modal.tsx
@@ -263,7 +263,7 @@ export function Modal(props: ModalProps) {
                   className={classes.overlay}
                   sx={{ position: 'fixed' }}
                   zIndex={0}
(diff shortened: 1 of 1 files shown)


rtivital:
This will cause some components dropdowns to close modal, for example, MultiSelect dropdown.

MoreBlood:
@rtivital Honestly, I don’t really understand how this is connected, can you give an example? 
Now the name of the prop is misleading that click should be used instead of pressing any mouse button

rtivital:
When MultiSelect dropdown is rendered within portal, all clicks are registered as outside (because its actually outside). But this is unwanted behavior. onMouseDown event is used to solve that issue – you can view Select/MultiSelect code to learn how it is managed.

rtivital:
Fixed by https://github.com/mantinedev/mantine/pull/2669

PR #2613, opened 2022-10-03
@mantine/hooks] use-local-storage: Fix error when localStorage is null

kdrag0n:
By default, `window.localStorage` is set to null in Android WebViews unless the embedding app has explicitly enabled storage, leading to an uncaught exception in this hook. Fix it by checking for null and returning the default value.


kdrag0n:
Reformatted with Prettier, so it should pass now.

rtivital:
Thanks!

kdrag0n:
Thanks for the quick merge!

PR #2614, opened 2022-10-03
added URL-driven notification subsection

SomiDivian:
where a message listen to a URL param to get opened or closed

### DIFF
--- a/docs/src/docs/others/notifications.mdx
+++ b/docs/src/docs/others/notifications.mdx
@@ -258,3 +258,116 @@ interface NotificationProps {
   onOpen?(props: NotificationProps): void;
 }
 ```
+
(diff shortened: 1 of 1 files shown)


rtivital:
Hi @SomiDivian, this logic seems to be specific for your application, it should not be included in documentation. We also do not recommend showing/updating notifications in useEffect.

SomiDivian:
Hey @rtivital True, I just wanted to show how to know if a specific notification is active.
You can see this feature in Chakra `useToast` hook built-in.

I have a question though about useEffect, Is there a best-practice way of showing/updating notification without user interaction.
for example: when an error happened after submitting a form.

Thanks for maintaining this great repo.

Issue #2615, opened 2022-10-03
Dropzone on excel MIME type, Microsoft Excel 97-2003 Worksheet is the first option in the dropdown list of file browser


micro9000:
### What package has an issue

@mantine/dropzone

### Describe the bug

We're using Dropzone with MS_EXCEL_MIME_TYPE and the first option of the file browser is Microsoft Excel 97-2003 Worksheet, which is I think an old one, can we swap the position of xls and xlsx
I know we can use the MIME_TYPES constant and specify each types we need in an array but it's convenient to use the MS_EXCEL_MIME_TYPE

### What version of @mantine/hooks page do you have in package.json?

5.2.3

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

https://github.com/mantinedev/mantine/blob/master/src/mantine-dropzone/src/mime-types.ts

### Line 33: 
**from**
`export const MS_EXCEL_MIME_TYPE = [MIME_TYPES.xls, MIME_TYPES.xlsx];`

**to**
`export const MS_EXCEL_MIME_TYPE = [MIME_TYPES.xlsx, MIME_TYPES.xls];`

PR #2616, opened 2022-10-03
[@mantine/dropzone] MIME TYPE: Swap xlsx and xls for MS_EXCEL_MIME_TYPE – Change in @mantine/dropzone package at mime-types.ts file

micro9000:
Change in @mantine/dropzone package at mime-types.ts file


rtivital:
why?

micro9000:
Hi, sorry I forgot to include this on my description: https://github.com/mantinedev/mantine/issues/2615#issue-1394360844

micro9000:
There's no impact on the UI but the first option in the file browser will always Microsoft Excel 97-2003 which is an old one and most of the users now are using the new version (xlsx), so it's better to show all xlsx files by default instead of xls

I don't have the right example but I'm referring to, is the drop down menu of the the file browser.
image

micro9000:
Hey,

Issue #2660, opened 2022-10-07
Radio labelWrapper


dekisr:
This [commit](https://github.com/mantinedev/mantine/commit/0513a481428fac9b1b4cd4547df07b3393630fe4#diff-e124c5671eb583bcac043dcf043a4e3f25f9f1ffcb13a0998d3217d4c4ca3ba5) removed some properties that broke when using some react node as label.

- before:
![Captura de Tela 2022-10-07 às 14 10 40](https://user-images.githubusercontent.com/38201274/194617483-6c8e9431-898e-4739-b7be-0f24628b4f98.png)

- after:
![Captura de Tela 2022-10-07 às 14 10 46](https://user-images.githubusercontent.com/38201274/194617555-766debd0-85de-4fb7-8c60-b8f07ebf626a.png)

aiheon:
does this break edit field?

rtivital:
Fixed in 5.7.2

PR #2661, opened 2022-10-07
[@mantine/core] Radio: fix labelWrapper styles

dekisr:
re #2660

### DIFF
--- a/src/mantine-core/src/Radio/Radio.styles.ts
+++ b/src/mantine-core/src/Radio/Radio.styles.ts
@@ -118,6 +118,8 @@ export default createStyles(
 
       labelWrapper: {
         ...theme.fn.fontStyles(),
+        display: 'flex',
+        alignItems: 'flex-start',
         fontSize: theme.fontSizes[size] || theme.fontSizes.md,
         lineHeight: `${theme.fn.size({ sizes, size })}px`,
         color: theme.colorScheme === 'dark' ? theme.colors.dark[0] : theme.black,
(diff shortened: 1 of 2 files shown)


dekisr:
It's also affecting the `Switch`:
Captura de Tela 2022-10-07 às 18 38 48

src/mantine-core/src/Radio/Radio.styles.ts
```diff
@@ -118,6 +118,8 @@ export default createStyles(
 
       labelWrapper: {
         ...theme.fn.fontStyles(),
+        display: 'flex',
+        alignItems: 'flex-start',
```
Phosphorus-M, line 122: In my opinion, it would be "center", not flex-start, it's weird I guess.

src/mantine-core/src/Switch/Switch.styles.ts
```diff
@@ -178,6 +178,8 @@ export default createStyles(
 
       labelWrapper: {
         ...theme.fn.fontStyles(),
+        display: 'flex',
+        alignItems: 'flex-start',
```
Phosphorus-M, line 182: And the same there

Phosphorus-M:
I just leave some comments it does not depend on me.

src/mantine-core/src/Radio/Radio.styles.ts
dekisr: Yeah, makes sense. I just changed it to how it was before these properties were removed

rtivital:
Fixed in #2881

PR #2663, opened 2022-10-08
[@mantine/spotlight] Use 'off' autocomplete value for Spotlight input

joelzwarrington:
The 'off' autoComplete value is an expected attribute value: https://developer.mozilla.org/en-US/docs/Web/HTML/Attributes/autocomplete

Not all browsers may understand the current value of `chrome-please-just-do-not-show-it-thanks`.


zhulien-ivanov:
Chrome doesn't respect `off` value. That's the whole purpose of using a random string. It's the only way currently.

joelzwarrington:
Ahh I see, I wasn't aware of this issue. I'll close this PR in that case, I thought Chrome would respect that value, but doesn't seem to in certain cases.

PR #2668, opened 2022-10-09
[@mantine/dates] DatePickerBase: corrected inputRef type to HTMLInputElement

ikapta:
corrected inputRef type from HTMLButtonElement to HTMLInputElement


rtivital:
Thanks!

PR #2669, opened 2022-10-09
[@mantine/core] Modal overlay respects scrollbar width

### LINKED ISSUE #2371: Modal overlay is rendered above the scroll
### What package has an issue

@mantine/modals

### Describe the bug

image

You can scroll normally using the mouse scroll, but if you try clicking the scroll bar you actually click the overlay triggering the modal to close.

https://mantine.dev/others/modals/

Yes

Yes

armanatz:
Fixes #2371 

If a modal has overflow set to outside, the overlay covers the scrollbar. This causes the modal to close by default if the scrollbar is clicked.

This fix moves the overlay outside the modal-inner container, removes the mousedown event from the overlay and instead puts an onClick handler onto the modal-inner div.

Event propagation is also stopped when clicking the modal itself so that clicking inside the modal won't close the modal. This should take care of any clickable components or components like MultiSelect that render outside the modal to work correctly.

### DIFF
--- a/src/mantine-core/src/Modal/Modal.tsx
+++ b/src/mantine-core/src/Modal/Modal.tsx
@@ -214,56 +214,13 @@ export function Modal(props: ModalProps) {
         }}
       >
         {(transitionStyles) => (
-          <Box id={baseId} className={cx(classes.root, className)} {...others}>
-            <div
-              className={classes.inner}
-              onKeyDownCapture={(event) => {
-                const shouldTrigger =
-                  (event.target as any)?.getAttribute('data-mantine-stop-propagation') !== 'true';
-                shouldTrigger && event.key === 'Escape' && closeOnEscape && onClose();
-              }}
-              ref={focusTrapRef}
-            >
-              <Paper<'div'>
-                className={classes.modal}
-                shadow={shadow}
-                p={padding}
-                radius={radius}
-                role="dialog"
-                aria-labelledby={titleId}
-                aria-describedby={bodyId}
-                aria-modal
-                tabIndex={-1}
-                style={transitionStyles.modal}
-                unstyled={unstyled}
-              >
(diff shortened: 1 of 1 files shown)


rtivital:
Click outside events should be handled by `onMouseDown`, `onClick` event will not work in all cases, we had this exact logic with `onClick` before and had to move to `onMouseDown`.

armanatz:
> Click outside events should be handled by `onMouseDown`, `onClick` event will not work in all cases, we had this exact logic with `onClick` before and had to move to `onMouseDown`.

@rtivital May I know what was the exact reason for switching to `onMouseDown` events?

As I can tell based on #2611 the reason `onClick` shouldn't be used is for components that render outside the modal dismissing the modal when clicked on. So if that's the case, I have tested my change against that and don't seem to see that behavior happening.

You may refer to the video below:

[google_screen_recording_2022-10-09T16-17_06.146Z.webm](https://user-images.githubusercontent.com/3622268/194767798-bbede358-b335-437d-b3a9-4363899fe050.webm)

rtivital:
As far as I remember, it did not work with elements that were rendered within portal, for example `<Select withinPortal />`

armanatz:
> As far as I remember, it did not work with elements that were rendered within portal, for example `<Select withinPortal />`

So in the video I showed, I have `<MultiSelect withinPortal />` which I assume should be the same as `<Select withinPortal />` and you can see that there is no abnormal behavior occurring due to `onClick` handler. You may test it yourself as well and let me know if you find any problems and I will be more than happy to dig deeper :)

rtivital:
Alright, I've checked, it seems to be working fine, thanks for the fix and investigation

armanatz:
No worries. Glad to help!

rtivital:
Thanks!

PR #2670, opened 2022-10-09
[docs] import FileWithPath type in Dropzone for image previews example

adarshaacharya:
- fix[docs] : add FileWithPath type imports for  image previews example in Dropzone


rtivital:
Thanks!

PR #2672, opened 2022-10-09
[@mantine/nprogress] Fix Next.js usage example

joelzwarrington:
@mantine/nprogress doesn't have an exported member `finishNavigationProgress`, and from looking at previous examples, I believe it's supposed to be `completeNavigationProgress`


rtivital:
Thanks!

PR #2676, opened 2022-10-09
[@mantine/core] Badge: default cursor behaviour

joelzwarrington:
I believe the StatusBadge should by default inherit it's cursor from the parent element. One shouldn't have to override the styles when the Badge is within an UnstyledButton for example. I believe the inherit cursor attribute would make sense for more cases than the `default` one.

Before:

https://user-images.githubusercontent.com/80279872/194773960-a384ba56-d25a-4873-afd7-37390c8d264b.mov

After:

https://user-images.githubusercontent.com/80279872/194774032-ca514fd6-4f36-42b2-9844-abb9afadfd1d.mov

I will say it is possible to change the cursor property with `sx` or CSS in JS styling (useStyles/className for example), but just thought the default behaviour might be best with `inherit` over `default`:
```jsx
function MyBadge() {
  return <Badge sx={{ cursor: "inherit" }}>Text</Badge>
}
```


rtivital:
Thanks!

PR #2689, opened 2022-10-12
Fix native select placeholder

nmay231:
Fixes `placeholder` not being the default for the NativeSelect component.

The example in usage should show "Pick one" https://mantine.dev/core/native-select/#usage


.gitignore
```diff
@@ -6,6 +6,11 @@
 .vscode/
 .nova/
 
+# Version Managers
+.tool-versions
+.nvmrc
```
rtivital, line 11: Do not remove `.nvmrc` is required for `nvm use` command to work.

rtivital: Never heard of `.tool-version` before, but I use `.nvmrc` often – local development work well only with Node 14 or 16, while 18 produces errors.

nmay231: `.tool-versions` is for [asdf](https://github.com/asdf-vm/asdf). It's basically an all-in-one tool for programming language version managers. Personally, I use it to manage versions of node, rust, and python.

Since node 18 has issues in development, I'll revert the commit and just synchronize the two files.

nmay231: I probably should have just put that in a separate PR, but ah well.

rtivital:
Thanks!

PR #2704, opened 2022-10-14
[@mantine/hooks] use-idle: improves typing of events options

arimariojesus:
Improve typing of events option to useIdle hook. This change allows for a better development experience, verifying code-time typing and making it easier to find correct events.


rtivital:
Thanks!

PR #2716, opened 2022-10-17
[@mantine/core] Image: fix figcaption overflow with removed height from root style.

ikapta:
Remove height from `Image` root style.

Now storybook demo can see this error directly，I also test when container is less than fixed image height, placeholder also work fine.
image


rtivital:
@ikapta please resolve conflicts

rtivital:
Thanks!

PR #2718, opened 2022-10-17
[@mantine/core] Select: do not rely on searchable in readonly prop

antoinesion:
All non searchable `Select` component are marked as `readonly`. It causes issues especially for CSS selection.

Example when we want to apply special style to readonly Select like this :
```css
input[readonly] {
  cursor: "not-allowed"
}
```
=> this style applies to all non searchable `Select` that are still editable. I think it makes no sense. What do you think?

### Solution

Get rid of `!searchable` in `readOnly={!searchable || readOnly}`


rtivital:
This logic is required for non-searchable Select to work correctly.

PR #2721, opened 2022-10-17
[@mantine/hooks] use-timout: Add parameters for callback function

linardsblk:
Issue discussed here - https://discord.com/channels/854810300876062770/1029643292431892491/1029643292431892491

As of now, setTimeout callback function is called without any parameters. I found it to be useful when these parameters can be pased from start() function.

### DIFF
--- a/src/mantine-hooks/src/use-timeout/use-timeout.ts
+++ b/src/mantine-hooks/src/use-timeout/use-timeout.ts
@@ -1,16 +1,16 @@
 import { useRef, useEffect } from 'react';
 
 export function useTimeout(
-  fn: () => void,
+  callback: (callbackParams?: any) => void,
   delay: number,
   options: { autoInvoke: boolean } = { autoInvoke: false }
 ) {
   const timeoutRef = useRef<number>(null);
 
-  const start = () => {
+  const start = (callbackParams?: any) => {
     if (!timeoutRef.current) {
       timeoutRef.current = window.setTimeout(() => {
-        fn();
+        callback(callbackParams);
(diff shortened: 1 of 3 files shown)


src/mantine-hooks/src/use-timeout/use-timeout.ts
```diff
@@ -1,16 +1,16 @@
 import { useRef, useEffect } from 'react';
 
 export function useTimeout(
-  fn: () => void,
```
rtivital, line 4: Why fn was renamed to callback?

src/mantine-hooks/src/use-timeout/use-timeout.ts
```diff
@@ -1,16 +1,16 @@
...
+  callback: (callbackParams?: any) => void,
```
rtivital, line 4: What if you want to pass more than one argument? Change type to allow any amount of arguments.

rtivital (changes requested):

src/mantine-hooks/src/use-timeout/use-timeout.ts
linardsblk: It was called callback in docs so I thought I change it to be consistant. callback seemed more descriptive than fn to me.

src/mantine-hooks/src/use-timeout/use-timeout.ts
linardsblk: Pushed changes based on your feedback.

rtivital:
Thanks!

PR #2735, opened 2022-10-18
[@mantine/hooks] use-focus-trap: Fix aria-hidden

### LINKED ISSUE #2726: FocusTrap not removing aria-hidden attribute
### What package has an issue

@mantine/hooks

### Describe the bug

Just found a bug which appears to affect all components using FocusTrap - Modal, Drawer, etc..
If you turn the focus trap on, the root gets hidden using "aria-hidden"="true", which makes sense. But this attribute is never removed and even when the focus trap is removed, the content remains inaccessible.

^5.5.2

https://codesandbox.io/s/hopeful-black-jopb3u?file=/src/App.js

Yes

Yes

### Possible fix

I *believe* this is caused by https://github.com/mantinedev/mantine/blob/9128d03970df766429c926c000c29e8c225e6c33/src/mantine-hooks/src/use-focus-trap/create-aria-hider.ts#L32, as the aria hidden only gets removed if it is already null.

If that is the case, I'd be willing to fix it. I have never contributed to an open-source project of this scale before, so it might take some time to figure things out - hope that'd be okay. But given the fact that this file was last changed in april of 2021, I don't reckon it's a high priority fix

Keilo75:
This PR resolves #2726.

Turns out that `restoreAria.current();` was simply not called on unmount; this has now been changed.
Furthermore, I fixed an additional issue where aria-hidden attributes would only be applied on the first render, but would not update after that. This was caused by https://github.com/mantinedev/mantine/blob/f7da1459fba6284e23ff57f3e839e691bbb8e980/src/mantine-hooks/src/use-focus-trap/use-focus-trap.ts#L16 After the first render, `ref.current` would already hold the node and aria attributes would not be reapplied again.

Finally, I added a test to `FocusTrap.test.tsx` which makes sure aria-hidden is added and removed. I'll be the first to admit it is kinda ugly but I could not find a better version. Also tested it manually with storybook - components like `Modal` or `Drawer` now remove the attribute correctly.

### DIFF
--- a/src/mantine-core/src/FocusTrap/FocusTrap.test.tsx
+++ b/src/mantine-core/src/FocusTrap/FocusTrap.test.tsx
@@ -51,6 +51,29 @@ describe('@mantine/core/FocusTrap', () => {
     expect(document.body).toHaveFocus();
   });
 
+  it('manages aria-hidden attributes', () => {
+    const adjacentDiv = document.createElement('div');
+    adjacentDiv.setAttribute('data-testid', 'adjacent');
+    document.body.appendChild(adjacentDiv);
+
+    const { rerender } = render(
+      <FocusTrap active>
+        <div />
+      </FocusTrap>
+    );
+
+    const adjacent = screen.getByTestId('adjacent');
+    expect(adjacent).toHaveAttribute('aria-hidden', 'true');
+
+    rerender(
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #2752, opened 2022-10-20
[@mantine/core] fixed preact infinite loop in popover

jvdsande:
As discussed on discord, this adds memoization for `reference` and `floating` functions in popover


rtivital:
Thanks!

Issue #2762, opened 2022-10-21
Floating point value in playground of `Rating` component crashes the site.


PhantomKnight287:
### Link to the page where something nasty is located

https://mantine.dev/core/rating/

### Exact quote of what is wrong

Go to the [Rating Component](https://mantine.dev/core/rating), in the playground of the rating component change the integer into a floating point number(ex: 2.5), the site crashes.

The logs in the console
![image](https://user-images.githubusercontent.com/76196237/197138545-14aaceb1-2551-437b-9058-919789965e1a.png)

zhulien-ivanov:
I'm onto it.

rtivital:
Fixed

PR #2763, opened 2022-10-21
[@mantine/core] Rating: Fix count and fractions parameters to accept integers only.

zhulien-ivanov:
Quickfix for #2762. 

However, this component seems prone to other errors related to how value and fractions interact with each other and should be reviewed more extensively.


zhulien-ivanov:
@DhrjArun Please, take a look at tests.

rtivital:
Thanks!

PR #2766, opened 2022-10-21
[@mantine/core] add custom icons capability to transfer-list

jvdsande:
Adds custom icon support for TransferList controls, to allow people to use their own design system icons.

The props take a component with a `reversed` props to allow handling both lists correctly.


jvdsande:
Closing, superseded by #2769

PR #2768, opened 2022-10-21
[@mantine/core] Tooltip: Fixed nested tooltips

### LINKED ISSUE #2193: Nested tooltips don't function as intended
### What package has an issue

@mantine/core

### Describe the bug

Nested tooltips seem to inherit each-others properties leading to buggy behaviors or not showing at all.

See the linked [demo](https://codesandbox.io/s/optimistic-meadow-95x4i9) where we attempt to nest a clickable Tooltip inside of a hoverable Tooltip, and instead neither fires.

5.2.0

https://codesandbox.io/s/optimistic-meadow-95x4i9

No

Yes

### Possible fix

Unknown. It works with TippyJS, however they manage this.

We had a custom wrapper to also help with this, but that doesn't help in this case either unfortunately:

```js
function mergeRefs(refs) {
  return value => {
    refs.forEach(ref => {
      if (typeof ref === 'function') {
        ref(value);
      } else if (ref != null) {
        ref.current = value;
      }
    });
  };
}


const TooltipWrap = forwardRef(({ label, tooltipOpts = {}, children }, ref) => {
  const mergedRefs = mergeRefs([children.ref, ref]);
  return label ? (
    <Tooltip
      {...{
        label,
        // NOTE: Doesn't work even if we use ...props here
      }}
      {...tooltipOpts}
    >
      {React.cloneElement(children, { ref: mergedRefs })}
    </Tooltip>
  ) : (
    React.cloneElement(children, { ref: mergedRefs })
  );
});
```

nfbuckley:
Fixes #2193

Works by merging the existing @floating-ui events from a parent tooltip (if it exists) onto the tooltip reference.


rtivital:
Thanks!

PR #2769, opened 2022-10-21
[@mantine/core] add controlled search and per-list wording to transfer-list

jvdsande:
As discussed on Discord, a few improvements to the `TransferList` component:

- `transferIcon` and `transferAllIcon` for customising the icons. The props take a component with a reversed props to allow handling both lists correctly.
- `placeholder` prop that is used in place of `nothingFound` if query is empty. If `placeholder` is omitted, `nothingFound` is used in its place
- `searchPlaceholder`, `nothingFound` and `placeholder` can now take a tuple instead of a single value, to customize each panel independently. A single value can still be passed, in which case it's applied to both panels.
- `searchValues` and `onSearch` props allowing to control the searches

### DIFF
--- a/docs/src/docs/core/TransferList.mdx
+++ b/docs/src/docs/core/TransferList.mdx
@@ -44,6 +44,27 @@ Value should be a tuple of two arrays which contain values from data:
 
 <Demo data={TransferListDemos.initialSelection} />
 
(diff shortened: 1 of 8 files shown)


rtivital:
Thanks!

PR #2781, opened 2022-10-24
[@mantine/rte]: Respect `defaultRadius` from `MantineProvider`.

zhulien-ivanov:
Respect theme `defaultRadius` for:
1. Tooltip input
2. Tooltip button
3. Mention
4. Toolbar actions

Even though `Popover` component respects it by default, I skipped adding default radius to Quill's tooltip as there is no way to change it afterwards.


rtivital:
Thanks!

PR #2785, opened 2022-10-24
Button loader position center

odysseaspapadimas:
Added a center option for Button loaderPosition


rtivital:
Thanks! These changes will be available with 5.7 release.

PR #2796, opened 2022-10-25
[@mantine/core] Pagination: Respect `font-family` from theme.

zhulien-ivanov:
`font-weight` is removed as it didn't affect the boldness before but it does now.


rtivital:
Thanks!

PR #2807, opened 2022-10-26
Minor typo fix in CopyButton (No iframes section)

drenther:
Minor typo fix in CopyButton (No iframes section)


rtivital:
Thanks!

PR #2819, opened 2022-10-27
[@mantine/core] Add header/footer breakpoints

auronsan1st:


rtivital:
Does not work as expected, see this story http://localhost:7520/?path=/story/appshell-appshell--fixed-with-header-breakpoints

Content should be pushed down based on current height, same as with Navbar – http://localhost:7520/?path=/story/appshell-appshell--fixed-with-navbar-breakpoints

auronsan1st:
fixed. sorry missed the part

rtivital:
Thanks!

PR #2822, opened 2022-10-27
Update MultiSelect.mdx

noahsb:
According to:

https://github.com/mantinedev/mantine/blob/4784c078b8b611e3c8b240a4f28c324394577984/src/mantine-core/src/Select/Select.tsx#L21

Looks like the type of `value` is actually `string[] | undefined` rather than `string[] | null`


rtivital:
Thanks!

PR #2842, opened 2022-10-29
Add peer dependencies to @mantine/core and @mantine/next

MariaSolOs:
This PR adds a couple of peer dependencies:
- Both `@mantine/core` and `@mantine/next` depend on `@mantine/styles`, which has a peer dependency on `@emotion/react`.
- `@mantine/next` depends on `@mantine/ssr`, which has a peer dependency on `@emotion/server`.

### DIFF
--- a/src/mantine-core/package.json
+++ b/src/mantine-core/package.json
@@ -28,6 +28,7 @@
   ],
   "peerDependencies": {
     "@mantine/hooks": "5.6.3",
+    "@emotion/react": ">=11.9.0",
     "react": ">=16.8.0",
     "react-dom": ">=16.8.0"
   },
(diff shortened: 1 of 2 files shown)


rtivital:
There is no need to describe peer dependencies in each package, it will only make dependencies versions harder to maintain. `@mantine/styles` and `@mantine/ssr` depending on `@emotion/*` packages is enough.

MariaSolOs:
@rtivital Modern versions of `yarn` don't think that that's enough.
Here's the error message I obtain when installing the dependencies in a project using `yarn` 3.2.4:
image

MariaSolOs:
If you wish, I could provide an example project with the issue @rtivital. 
I'm not trying to make things any harder for the library though, I'm just trying out Mantine for the first time and loving it so far! This is why I'm trying to contribute to make things better for everyone, specially Yarn users :)

rtivital:
These changes can be breaking for some package managers, I'll change the dependencies with next major version

rtivital:
Let's keep it opened as a reminder

PR #2856, opened 2022-11-01
PasswordInput: Fix validations Example doc typo

drenther:
Updated the docs for [PasswordInput Validations example](https://mantine.dev/core/password-input/#invalid-state-and-error) from "Incorrect email" to "Incorrect password" to match the code example


rtivital:
Thanks!

PR #2857, opened 2022-11-02
Added <span> tag around {children} in ListItem.tsx

Manas-Nagelia:
#2146 makes a pull request that removes the `<span>` tag around the `{children}` in `ListItem.tsx` This causes an overflow to happen in the `ListItem`, as represented in the image below. In this pull request, I re-added the `<span>` tags.

Before (from MantineUI, it overflows in to 2 lines):

Screenshot 2022-11-01 at 7 58 35 PM

After (my own personal project, it stays on one line):

Screenshot 2022-11-01 at 8 20 40 PM


rtivital:
Thanks!

PR #2866, opened 2022-11-02
Fix useScrollIntoView ignoring parameter changes

rChaoz:
Currently, useScrollIntoView returns a memoized callback that always refers to the params passed during the first render of the calling component, so calling with dynamically computed params, e.g. useScrollIntoView({ offset: someVariable }) will memoize the offset passed during the first render. If the component re-renders, any new offset passed is ignored.

Aditionally, 'scrollableRef.current' is an invalid dependency, see [https://epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array/](https://epicreact.dev/why-you-shouldnt-put-refs-in-a-dependency-array/)


rtivital:
Thanks!

PR #2872, opened 2022-11-03
[@mantine/hooks] use-local-storage: return undefined instead of empty string

### LINKED ISSUE #2642: useLocalStorageHook set defaultValue to undefined but get empty string
### What package has an issue

@mantine/hooks

### Describe the bug

Hello, why choose Nullish coalescing operator to make defaultValue which is undefined or null to empty string in useLocalStorageHook?

It's a little weird if I expect an undefined/null defaultValue but get an empty string.

---

Also, is there a typo here in [use-local-storage](https://mantine.dev/hooks/use-local-storage/#definition)?

> If set to true, value will be update is useEffect after mount

If set to true, value will be update **in** useEffect after mount

latest

https://codesandbox.io/s/kind-star-v25pry?file=/src/App.tsx

Yes

Yes

### Possible fix

```
const readStorageValue = useCallback(
      (skipStorage?: boolean): T => {
        if (typeof window === 'undefined' || !(type in window) || skipStorage) {
          return defaultValue as T;
          // return (defaultValue ?? '') as T;
        }

        const storageValue = window[type].getItem(key);

        return storageValue !== null ? deserialize(storageValue) : (defaultValue as T);
        // return storageValue !== null ? deserialize(storageValue) : ((defaultValue ?? '') as T);
      },
      [key, defaultValue]
    );
```

Narcha:
Fixes #2642

### DIFF
--- a/src/mantine-hooks/src/use-local-storage/create-storage.ts
+++ b/src/mantine-hooks/src/use-local-storage/create-storage.ts
@@ -54,12 +54,12 @@ export function createStorage<T>(type: StorageType, hookName: string) {
           window[type] === null ||
           skipStorage
         ) {
-          return (defaultValue ?? '') as T;
+          return defaultValue as T;
         }
 
         const storageValue = window[type].getItem(key);
 
-        return storageValue !== null ? deserialize(storageValue) : ((defaultValue ?? '') as T);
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #2876, opened 2022-11-03
[@mantine/core] ScrollAreaProps missing children prop

zhGio:
Ran into an issue using `ScrollArea` complaining that `children` is an unknown prop (as shown on the screenshot). 
Hopefully I patched it in the right place :)

Manually editing it in my own local `ScrollArea.d.ts` made typescript happy.

image

### DIFF
--- a/src/mantine-core/src/ScrollArea/ScrollArea.tsx
+++ b/src/mantine-core/src/ScrollArea/ScrollArea.tsx
@@ -15,6 +15,9 @@ export type ScrollAreaStylesNames = Selectors<typeof useStyles>;
 export interface ScrollAreaProps
   extends DefaultProps<ScrollAreaStylesNames, ScrollAreaStylesParams>,
     React.ComponentPropsWithRef<'div'> {
+      
+  children?: React.ReactNode;
+
   /** Scrollbar size in px */
   scrollbarSize?: number;
 


zhGio:
Oh darn you're right... My project is using some weird badly typed preact version which doesn't want to load any `React.` types.. sorry for taking your time :(

ljrahn:
> My project is using some weird badly typed preact version which doesn't want to load any React. types

Can you elaborate on this? what was your issue? I am running into the same issue only for ScrollArea component

ogmzh:
> > My project is using some weird badly typed preact version which doesn't want to load any React. types
> 
> Can you elaborate on this? what was your issue? I am running into the same issue only for ScrollArea component

gah it's been quite a while ago but i do recall some incompatibility with mantine types relying on react and me not using react but preact with some weirdly set up typescript configuration.. can't really help you much there anymore as that project is done for me.

ljrahn:
> > > My project is using some weird badly typed preact version which doesn't want to load any React. types
> > 
> > 
> > Can you elaborate on this? what was your issue? I am running into the same issue only for ScrollArea component
> 
> gah it's been quite a while ago but i do recall some incompatibility with mantine types relying on react and me not using react but preact with some weirdly set up typescript configuration.. can't really help you much there anymore as that project is done for me.

ahh no worries. Yeah not entirely sure what the issue is, but it looks like mantine didnt really like the yarn workspaces. i just threw a nohoist in the root package.json, and it seemed to fix the issue. not gonna be using mantine outside of my client side library, so seems like a good enough fix for me:

`"nohoist": ["**/@mantine/**"]`

PR #2881, opened 2022-11-05
[@mantine/core] Adjust Inline input label while passing component as label

adarshaacharya:
When a component is passed as label then it appears on the bottom right of the component like  Checkbox, Switch or Radio component, added inline flex on InlineInput labelWrapper class to fix this issue


rtivital:
This will break error and description styles
Снимок экрана 2022-11-06 в 11 06 20

adarshaacharya:
@rtivital sry, added direction column to avoid this .

Sajarin-M:
There is the same issue with the `Radio` and `Switch` components.
Does it fix them also?

rtivital:
Thanks!

PR #2885, opened 2022-11-05
[core]: Fix building script is not work when use --project

qiugu:


rtivital:
There is no need for `npm run build` script to support `--project` flag. You can build packages by running the following commands:

- `npm run build:all` – build all packages
- `npm run build core` – build one package
- `npm run build hooks styles core` – build multiple packages

All packages already have associated tsconfig file which should not be redefined.

qiugu:
why set --project flag in build.ts

PR #2887, opened 2022-11-05
[@mantine/core] Add missing right prop to system props

adarshaacharya:
- add support for right to mantine system props


rtivital:
Thanks!

PR #2895, opened 2022-11-06
Add hexa format to ColorPicker

Keilo75:
This PR adds a new color format to both color picker and color input. Both #rrggbbaa and #rgba are supported, as defined in https://developer.mozilla.org/en-US/docs/Web/CSS/hex-color. 

(Didn't expect I'd have enough time to finish this today, but here we are)


rtivital:
Thanks!

PR #2896, opened 2022-11-06
Fix bug when dragging mouse outside

Keilo75:
As discussed in #2890 and in the Discord, this PR fixes a bug where the modal would close when you mouse down in the modal but drag the mouse outside before releasing it. I've added a story which shows this working, even when the modal content uses portals.

The fix also doesn't cause a regression in #2669.


hamitaksln:
Changes look nice. I hope this issue will be resolved.

jvdsande:
Ha indeed, I missed this one while skimming PRs earlier.

This implementation indeed covers the case where initial mousedown happens in a Portal 👌

Changes look good for me!

rtivital:
Thanks!

PR #2907, opened 2022-11-08
[@mantine/demos] Update Text demos according to props description

anton-rodin:
Fix Text's props according to docs


rtivital:
The demo is correct, the props you've changed to will be removed in 6.0.

anton-rodin:
@rtivital when you use props from this example with typescript, they highlighted as en error, it's confusing

rtivital:
You need to update to the latest version, these props were introduced in 5.7.0

PR #2912, opened 2022-11-08
[@mantine/hooks] use-toggle: Add handling for more than two values in an array

yevhverb:


src/mantine-hooks/src/utils/is-function/is-function.ts
```diff
@@ -0,0 +1,3 @@
+export function isFunction(x: unknown): x is Function {
```
rtivital, line 1: Please remove this

yevhverb: In such a case, I can use `action instanceof Function` inside the hook (but this is a workaround), since there is a known issue with checking such type with `typeof` https://github.com/microsoft/TypeScript/issues/37663

If it is okay I will remove this

rtivital: typeof should work just fine

yevhverb: Unfortunately no
ts

rtivital: okay, then use instance of

rtivital:
Thanks!

PR #2913, opened 2022-11-08
[@mantine/core] Radio: Respect `Radio`'s `size` parameter over context.

zhulien-ivanov:
Allows overriding the size of a specific radio button when used in `Radio.Group`.
Fixes configurator demo in the `Radio` page: https://mantine.dev/core/radio/#usage.


rtivital:
Thanks!

PR #2921, opened 2022-11-09
Fixed performance in NavLink

jacob-j:
We have noticed that NavLink can cause some performance issues since its always rendering a Collapse. Using the already defined "withChildren", this is avoided.
What do you think?


rtivital:
@jacob-j, tests are failing

PR #2942, opened 2022-11-12
[@mantine/core] Add `right` system prop to extracted system styles and remove from docs.

zhulien-ivanov:
Fix for #2887.


rtivital:
Thanks!

PR #2978, opened 2022-11-15
[@mantine/core] Autocomplete: support selecting/creating item that's not in the list

sickelap:
Created demo section for new feature. Not sure about using console.log though.
Also feel free to suggest better use of selecting/finding input element in the test.


rtivital:
Autocomplete component provides **suggestions**, you can think of it like a Google search field. There is no need to create new values for suggestions because the user can already type in any value.

rtivital:
Discussion on Discord – https://discord.com/channels/854810300876062770/887408088800448553/1042210254001274890

PR #2981, opened 2022-11-16
Feature: Drawers Manager

josiahayres:
Drawers Manager
This is a clone of the existing Modals manager - but with the drawer component instead of Modals.

The idea being to give developers the same flexibility to show drawers, that we have with the modal system.


rtivital:
Hi @josiahayres, thanks for showing interest in contributing to Mantine. Unfortunately, this feature is not planned to be added, usually it is a good idea to discuss changes with maintainers before jumping into the coding (like described in the contributing guideline – https://mantine.dev/pages/contributing/#contributing-workflow) to avoid working on PR that will not get merged.

josiahayres:
Ah my apologies I missed that - makes sense. Will take that onboard for the next idea! 
Happy for this to be closed and to maintain this feature in my codebase instead.

PR #2984, opened 2022-11-17
doc: ScrollArea - type=auto description correction

jgravois:
disclaimer: i'm only ~90% sure what i'm proposing matches the actual intention/behavior.

i can't detect the difference between `auto` and `always` based on the current description. but i'm also having a hard time based on their behavior.


rtivital:
Thanks!

PR #2998, opened 2022-11-18
[core] RangeSlider maxRange Property

ultracodez:
This PR adds a feature that I think is very useful, particularly in a use case such as a video-length cropping application. This PR introduces the sibling feature to `minRange`, which allows you to not only customize the minimum distance between the two thumbs of the RangeSlider but also allows you to limit the distance possible between the thumbs of RangeSlider via the `maxRange` property.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 .idea/
 .vscode/
 .nova/
+.gitpod.yml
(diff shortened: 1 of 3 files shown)


rtivital:
Thanks!

PR #3004, opened 2022-11-19
[@mantine/tiptap] Change label to Strikethrough in tiptap text editor

adarshaacharya:
- change label from strike though to strikethrough in mantine tip tap editor


rtivital:
Thanks!

PR #3006, opened 2022-11-19
[@mantine/core] MultiSelect: remove unneccessary wrapper ref

MohammadHH:
It seems that wrapperRef in Multiselect component is defined and attached to a dom element, but it's not actually being used


rtivital:
Thanks!

Issue #3003, opened 2022-11-19
Dropzone: Rejected and accepted files on the same callback function (Proposed fix)


nicolassanmar:
## Context
Consider the following scenario, where a user uploads files to a dropzone and we want to keep track of the uploaded files:
``` typescript

const MyDropzoneComponent = () => {
  const [accepted, setAccepted] = useState([]);
  const [rejected, setRejected] = useState([]);
  
  return (
  <Dropzone onDrop={setAccepted} onReject={setRejected} accept={PDF_MIME_TYPE} />
  )
}
```
## Problematic scenario
1. User tries to upload a single '.png' file. It will get rejected, setting the `rejected` variable by `onReject`
2. User uploads only '.pdf' files. This will only call `onDrop` since user did only upload valid files.

## Problem
The final state of `accepted` and `rejected` do not reflect the last user drop

## Solution
Add a new prop to the `Dropzone` called something like `onDropAny` that gets both accepted and rejected files. `react-dropzone` already provides this functionality so it should be easy to implement. 
(it is called `onDrop` on the `useDropzone` hook)

This would make it possible to update the state of both `accepted` and `rejected` on the same function, guaranteeing to be able to set the state of the last user interaction

I am currently trying this on a fork!
This PR shows a way that this could be implemented:
https://github.com/mantinedev/mantine/pull/3010

rtivital:
Added in 5.8.3

PR #3010, opened 2022-11-19
[@mantine/dropzone] Add onDropAny prop

nicolassanmar:
Add a new prop to the `<Dropzone />` component that receives a function that is called on any file drop. 
This callback function will receive two arguments, the accepted files, and the rejected ones.

This solves the following issue: https://github.com/mantinedev/mantine/issues/3003


rtivital:
Thanks!

Issue #2847, opened 2022-10-31
[@mantine/core] Changing NumberInput doesn't change precision unless triggering onBlur 


joelzwarrington:
### What package has an issue

@mantine/core

### Describe the bug

Changing the precision of a `<NumberInput />` programatically (say from 0 to 4) doesn't change the precision of the value unless the user triggers `onBlur`.

https://user-images.githubusercontent.com/80279872/198915063-cbaed5af-50ec-4ef7-8c5e-681aa0a3888c.mov

### What version of @mantine/hooks page do you have in package.json?

5.6.3

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/focused-proskuriakova-txino2?file=/src/App.tsx

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

rtivital:
Fixed in 5.8.3

PR #3011, opened 2022-11-19
[@mantine/core] NumberInput: Fix precision update

auronsan1st:
fixes Issue: https://github.com/mantinedev/mantine/issues/2847


rtivital:
Thanks!

PR #3019, opened 2022-11-20
Custom avatar image component

ultracodez:
This was first mentioned in #719.

This is an excellent feature, even if not a high priority. It would also allow interoperability with other image components if the library user would so wish to use those.

This is accomplished by adding an additional property to the `Avatar` component:
```
  /** Custom image component */
  imageComponent?: React.FunctionComponent<AvatarCustomComponentProps>;
```

I added a new file `AvatarCustomComponentProps.tsx` in the `Avatar` directory to contain the TypeScript `interface` responsible for making sure the passed React Component has the required props:
```
interface AvatarCustomComponentProps {
  className: string;
  src: string;
  alt: string;
  onError: any;
}

export type { AvatarCustomComponentProps };
```
Finally, the actual custom component is consumed by a simple inline if statement:
```
imageComponent ? (
        <ImageComponent
          {...imageProps}
          className={classes.image}
          src={src}
          alt={alt}
          onError={() => setError(true)}
        />
      ) : (
        <img
          {...imageProps}
          className={classes.image}
          src={src}
          alt={alt}
          onError={() => setError(true)}
        />
      )
```
It is worth noting that the prop is renamed inline to an uppercase letter for React:
```
const ImageComponent = imageComponent;
```
I'd like to see if the same type of thing could be done for the Mantine `Image` component so that the styles could be applied to a `next/image` `Image` if it were passed as a parameter.

Thanks for your time!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -5,6 +5,7 @@
 .idea/
 .vscode/
 .nova/
+.gitpod.yml
 
 # node
 node_modules/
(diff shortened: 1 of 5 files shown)


ultracodez:
Closed because there is a commit that I didn't want in here.

Issue #3001, opened 2022-11-18
form.insertListItem does not make the form dirty


gdostie:
### What package has an issue

@mantine/form

### Describe the bug

When calling `insertListItem` on a form instance, it currently doesn't make the form dirty, but it should. The same is true for `reorderListItem`.

### What version of @mantine/hooks page do you have in package.json?

5.7.1

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/blazing-meadow-t6hvtq?file=/src/App.tsx

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

Issue #2937, opened 2022-11-10
[@mantine/form] setFieldValue doesn't trickle down on nested fields


gdostie:
### What package has an issue

@mantine/form

### Describe the bug

When a form has nested fields, calling `setFieldValue` on the nested fields parent does not clear the dirty state or the error state of its nested fields. This is problematic because you can end-up in confusing states like a form for which `form.values === initialValues` but `form.isDirty()` returns `true`.

See attached sandbox for a live example of the problem.

### What version of @mantine/hooks page do you have in package.json?

5.7.1

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/peaceful-browser-3kgjf4?file=/src/App.tsx

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

minosss:
why not use list handlers? https://mantine.dev/form/nested/#list-handlers

gdostie:
> why not use list handlers? https://mantine.dev/form/nested/#list-handlers

List handlers would indeed work in that case, but my issue is more that `setFieldValue` can lead to bugs, not so much about how to implement this behaviour so that it works properly.

auronsan1st:
@gdostie 
I try review it seems it's the problem between switch using `users` and `users.2` 
and implementation step 4 is mistake should be `setFieldValue("users.2","josh")` and when do step 1 only + step 4 the dirty will gone.

step 3 bug occur caused error path is `users.2` but setFieldValue to `users` as will little hard to remove error on users.2 if setField the parent. caused we will need do recursively and will make longer computing.

@rtivital I confused should I try make fixes by checked children error when parent value changed by setFieldValue?

rtivital:
@auronsan1st There is already some logic that clears errors for lists, maybe it can be applied for this case? https://github.com/mantinedev/mantine/blob/dev/src/mantine-form/src/clear-list-state/clear-list-state.ts

auronsan1st:
yeah error solved by added it it on inside _setValues(). @rtivital

gdostie:
If its too hard / bug prone to get it working properly, here are a few ideas that could help mitigate this bug IMHO:
1. Adding a notice on the documentation site to acknowledge this limitation by saying setFieldValue shouldn't be used for setting the value of parent fields _(IMPORTANT)_
[...]

PR #3025, opened 2022-11-21
[@mantine/form] useForm: fix isDirty for list fields

gdostie:
By computing the isDirty only on demand using deep comparison of the `values` state, it removes many bugs related to `isDirty` and nested fields. Manually managing a dirty state parallel to the values state was very much bug prone, considering all the possible updates to form values that can occur.

~Disclaimer: by doing this we loose 2 functionalities `initialDirty` and `setDirty`. IMHO the pros outweighs the cons, but I am curious to see what the community thinks.~ Fixed ✅ 

Fixes:  https://github.com/mantinedev/mantine/issues/3001 and https://github.com/mantinedev/mantine/issues/2937

### DIFF
--- a/src/mantine-form/src/use-form.ts
+++ b/src/mantine-form/src/use-form.ts
@@ -28,6 +28,7 @@ import {
   ResetDirty,
   IsValid,
   _TransformValues,
+  ClearFieldDirty,
 } from './types';
 
 export function useForm<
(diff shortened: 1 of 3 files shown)


gdostie:
Converting to draft, will try to retain `initialDirty` and `setDirty`

gdostie:
I managed to retain `setDirty` and `initialDirty` functionalities. 🎉 

I renamed them in the code as "manual overrides" to the standard dirty behaviour.

Reopening for review.

src/mantine-form/src/use-form.ts
```diff
@@ -45,7 +46,7 @@ export function useForm<
   validate: rules,
 }: UseFormInput<Values, TransformValues> = {}): UseFormReturnType<Values, TransformValues> {
   const [touched, setTouched] = useState(initialTouched);
-  const [dirty, setDirty] = useState(initialDirty);
+  const [manualDirtyOverride, setDirtyManualOverride] = useState(initialDirty);
```
rtivital, line 49: Is there a reason to rename these variables? I think dirty and setDirty are less noisy

gdostie: The goal was to make it clearer that those are overrides of the normal `initial not equal current` behaviour. I can revert if you feel keeping the original naming this makes sense 👍

rtivital:
Thanks!

PR #3027, opened 2022-11-22
[@mantine/core] Drawer: Add id for aria- attributes

noahsb:
As per [discussion](https://github.com/mantinedev/mantine/discussions/3016), added `id` to `Drawer` to allow `aria-` attributes on title and body of `Drawer`. 

Unlike `Modal`, I did not apply the `baseId` directly, but use it only for generating `titleId` and `bodyId`. This now passes an axe DevTools scan 

Added a test to check that passing `id` gets sent to title and body of `Drawer`

### DIFF
--- a/src/mantine-core/src/Drawer/Drawer.test.tsx
+++ b/src/mantine-core/src/Drawer/Drawer.test.tsx
@@ -36,6 +36,12 @@ describe('@mantine/core/Drawer', () => {
     expect(screen.getByText('test-title')).toBeInTheDocument();
   });
 
+  it('uses the provided id prop on the title and body', () => {
+    const { container } = render(<Drawer {...defaultProps} id="my-drawer" />);
+    expect(container.querySelectorAll('#my-drawer-title')).toHaveLength(1);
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

brykov:
@noahsb so now you have introduced a new level of DOM hierarchy in Drawer that has `id` but no class name and is not accessible via [styles API](https://mantine.dev/core/drawer/?t=styles-api). It literally breaks my UI. And there is no non-hacky way to fix it. Can you please consider adding styles API to this [new container](https://github.com/mantinedev/mantine/commit/51a0299a994373dd7b8a08aaa5afa260dd72ca04#diff-199cf7413f71cc441275289e31bff96b9a025ec6a291ce5b8033a1e67162a5efR268)

noahsb:
Could you please move this to an issue and provide a working example, @brykov? I see value in adding an additional styles API to the new container but this discussion shouldn't be inside a merged PR without an example.

brykov:
> Edit: A quick search of the current issues shows #3056 and the relevant open PR for it #3061...

Yes, that sounds about right. It says it's already been released. I'll check.

PR #3044, opened 2022-11-23
[@mantine/hooks] use-hot-keys: Add option to hotkey handlers to disable event.preventDefault always running

dbrxnds:


rtivital:
Thanks!

PR #3053, opened 2022-11-24
[@mantine/dropzone] Add the getFilesFromEvent and validator props

mpinedami:
## Description

Add getFilesFromEvent and validator props to the Dropzone component, both are methods that React Dropzone supports

## getFilesFromEvent prop

The Dropzone component a getFilesFromEvent prop that enhances the handling of dropped file system objects and allows more flexible use of them e.g. passing a function that accepts drop event of a folder and resolves it to an array of files adds plug-in functionality of folders drag-and-drop.

Though, note that the provided getFilesFromEvent fn must return a Promise with a list of File objects (or DataTransferItem of {kind: 'file'} when data cannot be accessed). Otherwise, the results will be discarded and none of the drag events callbacks will be invoked.

In case you need to extend the File with some additional properties, you should use [Object.defineProperty()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/defineProperty) so that the result will still pass through our filter:

## validator prop

By providing validator prop you can specify custom validation for files.

The value must be a function that accepts File object and returns null if file should be accepted or error object/array of error objects if file should be rejected.

### DIFF
--- a/src/mantine-dropzone/src/Dropzone.tsx
+++ b/src/mantine-dropzone/src/Dropzone.tsx
@@ -1,27 +1,27 @@
+import React from 'react';
 import {
-  Box,
+  useDropzone,
+  FileRejection,
+  Accept,
+  FileWithPath,
+  DropEvent,
+  FileError,
+} from 'react-dropzone';
+import { fromEvent } from 'file-selector';
+import {
   DefaultProps,
-  LoadingOverlay,
-  MantineNumberSize,
   Selectors,
-  useComponentDefaultProps
+  MantineNumberSize,
+  LoadingOverlay,
+  Box,
+  useComponentDefaultProps,
 } from '@mantine/core';
 import { assignRef } from '@mantine/hooks';
 import { ForwardRefWithStaticComponents } from '@mantine/utils';
-import { fromEvent } from 'file-selector';
-import React from 'react';
-import {
-  Accept,
-  DropEvent,
-  FileError,
-  FileRejection,
-  FileWithPath,
-  useDropzone
-} from 'react-dropzone';
 import { DropzoneProvider } from './Dropzone.context';
-import useStyles from './Dropzone.styles';
-import type { DropzoneFullScreenType } from './DropzoneFullScreen';
 import { DropzoneAccept, DropzoneIdle, DropzoneReject } from './DropzoneStatus';
+import type { DropzoneFullScreenType } from './DropzoneFullScreen';
+import useStyles from './Dropzone.styles';
 
 export type DropzoneStylesNames = Selectors<typeof useStyles>;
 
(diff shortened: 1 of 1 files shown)


src/mantine-dropzone/src/Dropzone.tsx
```diff
@@ -110,6 +124,8 @@ export const defaultProps: Partial<DropzoneProps> = {
   dragEventsBubbling: true,
   activateOnKeyboard: true,
   useFsAccessApi: true,
+  getFilesFromEvent: fromEvent,
```
rtivital, line 127: There is no need to specify default prop for this

src/mantine-dropzone/src/Dropzone.tsx
```diff
@@ -1,19 +1,27 @@
...
-  Box,
-  useComponentDefaultProps,
+  MantineNumberSize,
+  Selectors,
+  useComponentDefaultProps
 } from '@mantine/core';
 import { assignRef } from '@mantine/hooks';
 import { ForwardRefWithStaticComponents } from '@mantine/utils';
+import { fromEvent } from 'file-selector';
+import React from 'react';
```
rtivital, line 12: Please do not change imports order

rtivital (changes requested):

src/mantine-dropzone/src/Dropzone.tsx
mpinedami: I must set a default value to these props otherwise they are undefined values ​​and the behavior will not be as expected.

Although I could do the following in the useDropzone params:

  ```
const { getRootProps, ...otherValues } = useDropzone({
    ...otherParams
    getFilesFromEvent: getFilesFromEvent || fromEvent,
    validator: validator || null,
  });
```

Comment (rtivital) [reply src/mantine-dropzone/src/Dropzone.tsx]: useDropzone hook has its own default values. There is no need to provide default value from our side. Another unwanted side effect is that `@mantine/dropzone` package will have to depend on `file-selector` to make it work correctly with yarn v3.

File: src/mantine-dropzone/src/Dropzone.tsx [hunk 3]
```diff
@@ -1,5 +1,13 @@
...
-import { useDropzone, FileRejection, Accept, FileWithPath } from 'react-dropzone';
+import {
+  useDropzone,
+  FileRejection,
+  Accept,
+  FileWithPath,
+  DropEvent,
+  FileError,
+} from 'react-dropzone';
+import { fromEvent } from 'file-selector';
```
rtivital, line 10: Please remove this import

rtivital:
Thanks!

porkopek:
After upgrading to 5.9.1, the `Dropzone` component throws this exception

`getFilesFromEvent is not a function`

so it seems that providing a value to the getFilesFromEvent is mandatory.
Before this, dropzone worked without this prop

porkopek:
As a workaround, I made it work with
```js
<Dropzone

   getFilesFromEvent={e => {
          return Promise.resolve([...(e.target as EventTarget & HTMLInputElement)?.files]);
        }}
  ...
/>
```

Issue #3057, opened 2022-11-25
Carousel number of indicators not updating as per state value change


Adel-ak:
### What package has an issue

@mantine/carousel

### Describe the bug

The number of carousel indicators is not updating as per the state value change

if the prop slides to scroll is  `slidesToScroll={mobile ? 2 : 3}` and you start off on mobile size and move up the indicators much change accordingly to the number of child slides

Visual Example:

https://user-images.githubusercontent.com/47484379/203966501-f7c6f9eb-e861-44d9-ace6-cf1c6ff2fe19.mp4

### What version of @mantine/hooks page do you have in package.json?

5.8.3

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/proud-cookies-mg6d1y?file=/src/App.tsx:1228-1236

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

Including `slidesToScroll` to the dependency list [here](https://github.com/mantinedev/mantine/blob/9febdd1a1fb82950ad7cf68c2c4509bdad573b0c/src/mantine-carousel/src/Carousel.tsx#L268) and [here](https://github.com/mantinedev/mantine/blob/9febdd1a1fb82950ad7cf68c2c4509bdad573b0c/src/mantine-carousel/src/Carousel.tsx#L258) will fix it

PR #3058, opened 2022-11-25
[@mantine/carousel] Carousel: fixed carousel indicator not changing when slidesToScroll value is changed

Adel-ak:
Included a fix to make sure the indicator count is updated based on the value of slidesToScroll

Fixes: #3057


rtivital:
Thanks!

Issue #3060, opened 2022-11-25
Last item in stepper disappears when max-height is set on the steps div.


arvind0598:
### What package has an issue

@mantine/core

### Describe the bug

My goal is to have a vertical stepper with the content on the right, and to have both sections be independently scrollable.
To do this, I've added `max-height` and `overflow: auto` to the steps div.

Expected:
It works properly!

Actual:
The last element keeps getting hidden - checked for many values of max height and steps.

### What version of @mantine/hooks page do you have in package.json?

5.8.2

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/gifted-satoshi-6pkv0p

### Do you know how to fix the issue

Possibly related to max height and some display: flex oddities.

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

reza-sadeghzadeh:
Each step is rendered as a button element, and the last step (last button) has a property called `min-height` and is set to `auto`, which causes this issue. this property is added via the` :last-of-type` pseudo-class that I cannot find the point of it!

rtivital:
Fixed in 5.8.4

PR #3062, opened 2022-11-25
[@mantine/core] Stepper: Fix vertical last step

auronsan1st:
fixes issue https://github.com/mantinedev/mantine/issues/3060

minHeight:"auto" result 0px on overflow-y 
by look I think this intended for hide the separator. 
so by display none the last item separator works good.


rtivital:
Thanks!

PR #3065, opened 2022-11-26
[@mantine/core] always try centering arrow on target for popover

jvdsande:
This PR tries to fix some outstanding issues introduced with the switch to Floating UI:

- Position of the arrow is now correctly recomputed when the dropdown shifts to stay in viewport even when placement is `-end` or `-start``
- To fix the point above, a change was introduced: `-end` and `-start` placement now only impact the dropdown, not the arrow. The arrow will always try to align itself on the center of the target.
- `arrowOffset` now acts as a min distance from dropdown border, rather than a fixed position in `-end` and `-start` placement.
- When placement/position is dynamic, scroll/window resize would reset everything to the initial values on first mount. This is now fixed by remounting the dropdown when the placement/position changes
- The arrow was not correctly positioned and the border was wrong in RTL mode. This is fixed by actually removing any RTL-specific styling, as Floating UI already handles thi internally.


src/mantine-core/src/Floating/FloatingArrow/FloatingArrow.tsx
```diff
@@ -1,36 +1,19 @@
 import React, { forwardRef } from 'react';
-import { useMantineTheme } from '@mantine/styles';
 import { getArrowPositionStyles } from './get-arrow-position-styles';
 import { FloatingPosition } from '../types';
 
 interface FloatingArrowProps extends React.ComponentPropsWithoutRef<'div'> {
   withBorder: boolean;
   position: FloatingPosition;
   arrowSize: number;
-  arrowOffset: number;
```
jvdsande, line 10: We don't need the arrow offset anymore in this component, as the offset is now used as a configuration for the `arrow` middleware of floating UI directly.

src/mantine-core/src/Floating/FloatingArrow/FloatingArrow.tsx
```diff
@@ -1,36 +1,19 @@
...
-      arrowOffset,
-      arrowRadius,
-      visible,
-      arrowX,
-      arrowY,
-      ...others
-    },
-    ref
-  ) => {
-    const theme = useMantineTheme();
```
jvdsande, line 32: We don't need the theme at all anymore since the RTL direction is already handled by Floating UI

src/mantine-core/src/Floating/FloatingArrow/get-arrow-position-styles.ts
```diff
@@ -1,47 +1,6 @@
...
-  if (placement === 'end') {
-    return { [dir === 'ltr' ? 'right' : 'left']: arrowOffset };
-  }
-
-  if (placement === 'start') {
-    return { [dir === 'ltr' ? 'left' : 'right']: arrowOffset };
-  }
-
-  return {};
-}
```
jvdsande, line 43: Those two functions are not needed anymore, as position is correctly handled by Floating UI, including RTL. So we can directly set `left` for vertical and `top` for horizontal sides

src/mantine-core/src/Floating/FloatingArrow/get-arrow-position-styles.ts
```diff
@@ -111,20 +66,20 @@ export function getArrowPositionStyles({
...
   }
 
   if (side === 'bottom') {
     return {
       ...baseStyles,
-      ...verticalSide(placement, arrowX, arrowOffset, dir),
+      left: arrowX,
       top: arrowPosition,
       borderBottom: 0,
-      [dir === 'ltr' ? 'borderRight' : 'borderLeft']: 0,
```
jvdsande, line 127: This check was actually messing things up for RTL, simply removing it fixes things

PR #3067, opened 2022-11-26
[docs] Update Dropzone docs for Enable child pointer event

coadtan:
I raised this question on Discord as it's not documented. This PR document Enables child pointer events with an example for Dropzone.


rtivital:
Thanks!

PR #3068, opened 2022-11-27
[@mantine/tiptap]: add forward ref to mantine tip tap editor

adarshaacharya:
- forward ref to avoid ref warning in mantine tip tap editor

### DIFF
--- a/src/mantine-tiptap/src/RichTextEditor.tsx
+++ b/src/mantine-tiptap/src/RichTextEditor.tsx
@@ -1,5 +1,5 @@
 /* eslint-disable react/no-unused-prop-types */
-import React, { useMemo } from 'react';
+import React, { useMemo, forwardRef } from 'react';
 import {
   Box,
   useComponentDefaultProps,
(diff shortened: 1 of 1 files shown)


src/mantine-tiptap/src/RichTextEditor.tsx
```diff
@@ -50,7 +50,7 @@ const defaultProps: Partial<RichTextEditorProps> = {
   withTypographyStyles: true,
 };
 
-export function RichTextEditor(props: RichTextEditorProps) {
+export const RichTextEditor = forwardRef<HTMLInputElement, RichTextEditorProps>((props, ref) => {
```
rtivital, line 53: ref type is incorrect, it is div element, not input

adarshaacharya: corrected, thanks

rtivital:
Thanks!

src/mantine-tiptap/src/RichTextEditor.tsx
AabhasDhaubanja: thanks adarsha

PR #3070, opened 2022-11-28
[@mantine/core] Select: fixed the display of values in the dropdown i…

cheechoo28:
Issue example: https://codesandbox.io/s/serverless-framework-kl3vu3?file=/src/App.tsx
Issue description:
When I added a prop limit, after selecting the value, the dropdown is closed and if it is opened again, all values are displayed ignoring the limit.


cheechoo28:
@rtivital That's what I intended. The idea is that two things are always shown in the dropdown . If nothing is selected, the first 3 are shown. If you have selected something, closed the dropdown and reopened, the first selected and subsequent ones are shown.

rtivital:
Nope, this is not the correct logic that will create a lot of confusion

cheechoo28:
@rtivital Well, I'll think about this problem again. I think the way it works now is also a confusion. I set the limit , but after I selected the value and reopened the dropdown , all the items are shown again

cheechoo28:
@rtivital should now work correctly

rtivital:
Thanks!

PR #3080, opened 2022-11-29
[@mantine/form] validate-field-value: parent path validation

dmgauthier:
### Issue summary

This is the current behavior
```
// 'parent.path.precise' is in error.
form.isValid() // false
form.isValid('parent.path') // true  
form.isValid('parent.path.precise') // false
```
This is the wanted behavior
```
// 'parent.path.precise' is in error.
form.isValid() // false
form.isValid('parent.path') // false  
form.isValid('parent.path.precise') // false
```

### Fix summary

**Current**
`validateFieldValue` receives the path used in `form.isValid('path.to.check')` and verifies if it match an exact key.

**Fixed**
`validateFieldValue` parses the path used in `form.isValid('path.to.check')` with `Array.prototype.split` and verifies each path indexes for a match with keys in error.

### DIFF
--- a/src/mantine-form/src/validate/validate-field-value.test.ts
+++ b/src/mantine-form/src/validate/validate-field-value.test.ts
@@ -29,6 +29,16 @@ describe('@mantine/form/validate-field-value', () => {
     ).toStrictEqual({ hasError: true, error: 'error-b' });
   });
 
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

Issue #2965, opened 2022-11-15
spotlight doesn't scroll when using ArrowDown/ArrowUp


ika18:
### What package has an issue

@mantine/spotlight

### Describe the bug

Spotlight content overflows, and spotlight doesn't scroll when using ArrowDown/ArrowUp.

### What version of @mantine/hooks page do you have in package.json?

5.7.2

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/naughty-wiles-1g7qm5

### Do you know how to fix the issue

No

### Possible fix

Maybe we can add a hook to check hovered Action whether need to scroll into view in ActionList component.

colinricardo:
+1 on this one

ChristophWahnsinn:
+1

chhpt:
Is there any updates on the progress here?

rtivital:
This issue is closed for one of these reasons:

- It was fixed in 7.0
- It is no longer reproducible in 7.0
- It is no longer applicable in 7.0
- It does not have a reproduction

If you think that this issue was closed by mistake and it is still an issue in version 7.0, please reopen it.

PR #3081, opened 2022-11-29
[@mantine/spotlight] Fixing Paper scroll position when navigating using arrow keys

hassanzohdy:
Fixing scroll repositioning when using arrows to navigate between actions if Paper has a scroll

issue #2965


rtivital:
It does not work as expected, seems like arrows now move selection to random position

https://user-images.githubusercontent.com/10353856/205488872-3667d9ce-5865-43ca-8d20-2fb559f1ba57.mov

hassanzohdy:
@rtivital Can you provide an example when this behavior happens, i tried it over storybook examples, and made now a slight update.

PR #3085, opened 2022-11-30
Pr/media query initialvalue

jsheely:
There is a a rendering conflict flicker on client side rendered websites with `useColorScheme` due to the default value not being set on initial render.

You'll get a light -> dark render when browser is set to dark mode

Added storybook and unit test to prove situation


rtivital:
Thanks!

PR #3087, opened 2022-11-30
Update Drawer.mdx

noahsb:
Update docs now that #3027 is merged which supports aria- attribute support by default (now same as modal)


rtivital:
Thanks!

PR #3097, opened 2022-12-02
[@mantine/hooks] Fixing `use-hash` extra rerendering issue

hunterwilhelm:
Two bugs fixed:
* If you `setHash` to something that doesn't start with "#" it would be reflect that temporarily in `hash`
* `setHashValue` was being called multiple times (one in the listener and the set hash) resulting in the multiple renders

Result:
* `setHash` only rerenders the page once.
* `hash` consistently starts with '#'

### DIFF
--- a/src/mantine-hooks/src/use-hash/use-hash.ts
+++ b/src/mantine-hooks/src/use-hash/use-hash.ts
@@ -1,21 +1,23 @@
-import { useState, useEffect } from 'react';
+import { useState } from 'react';
 import { useWindowEvent } from '../use-window-event/use-window-event';
 
 export function useHash() {
-  const [hash, setHashValue] = useState<string>('');
+  const [hash, setHashValue] = useState<string>(() => window.location.hash);
 
   const setHash = (value: string) => {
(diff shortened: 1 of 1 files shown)


src/mantine-hooks/src/use-hash/use-hash.ts
```diff
@@ -1,21 +1,23 @@
-import { useState, useEffect } from 'react';
+import { useState } from 'react';
 import { useWindowEvent } from '../use-window-event/use-window-event';
 
 export function useHash() {
-  const [hash, setHashValue] = useState<string>('');
+  const [hash, setHashValue] = useState<string>(() => window.location.hash);
```
rtivital, line 5: This will fail during ssr, window object does not exist.

rtivital (changes requested):

rtivital:
Please fix failed tests

rtivital:
Please fix failed tests

hunterwilhelm:
Thanks! Sorry about that, I'm a little new to this.

rtivital:
Thanks!

PR #3098, opened 2022-12-02
[@mantine/core] Image: Fix image component update bug with placeholder

sctape:
If you render an Image component with a valid src, then update the src prop to null, the placeholder does not render. With this fix, the placeholder will be properly rendered when the component updates.


sctape:
Some how my eslint plugin was off. I've fixed the linting errors.

rtivital:
Thanks!

PR #3100, opened 2022-12-02
[@mantine/core] add a prop to choose how to position floating arrow

jvdsande:
This MR adds a `arrowPosition?: 'side' | 'center'` prop to Tooltip and Popover (and components inheriting from Popover).

`'side'` is the previous behavior, where arrow sticks to the side of the tooltip in `end` and `start` position, but it not correctly placed when the dropdown is shifted to stay in the viewport.

`'center'` is the newer behavior, where arrow tries to center itself to the target, but ends up aligned to the "wrong" side if the tooltip is smaller than the target.


rtivital:
Thanks!

PR #3101, opened 2022-12-02
[@mantine/form] fix type inference of validation for union types

jvdsande:
Union types were incorrectly distributed when computing validation functions. This was not visible for the most classic unions, those including `undefined` and `null`, because Mantine's codebase is not in strict mode.

In strict mode, if the type of a value includes `null` or `undefined` in a union, it will get incorrectly distributed, as `undefined` and `null` will be inferred independently.

To avoid that, I wrap the `Value` inferred type into `NonNullable` before checking it against Array and Record cases. This prevents early distribution of the `Value` inferred type since the check is not directly against it. 

It is safe since `NonNullable` will get rid of `undefined` and `null`, which would be "false" in those checks anyway (not an Array, not a Record), so the comparison is still correct.

### DIFF
--- a/src/mantine-form/src/types.ts
+++ b/src/mantine-form/src/types.ts
@@ -21,7 +21,7 @@ export interface ReorderPayload {
 
 type Rule<Value, Values> = (value: Value, values: Values, path: string) => React.ReactNode;
 
-type FormRule<Value, Values> = Value extends Array<infer ListValue>
+type FormRule<Value, Values> = NonNullable<Value> extends Array<infer ListValue>
   ?
       | Partial<{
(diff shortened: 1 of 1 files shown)


auronsan1st:
just note
I don't know why my vscode not complained. typescript: 4.8.2
Screen Shot 2022-12-03 at 23 43 47

jvdsande:
@auronsan1st if you are not using strict mode for Typescript, then union with undefined or null work correctly already.

Unions with other types such as `string | number` would fail.

And with TS in strict mode, all unions fail, including with null and undefined

rtivital:
Thanks!

PR #3102, opened 2022-12-02
[@mantine/core] (Select, MultiSelect, Autocomplete) - Add hoverOnChange/hoverOnSearchChange prop

SrBrahma:
Implements https://github.com/mantinedev/mantine/discussions/3092 (first item):

> It's not currently possible in [MultiSelect](https://github.com/mantinedev/mantine/blob/master/src/mantine-core/src/MultiSelect/MultiSelect.tsx) to type and press enter to select the first result; it's required to hover a result with the cursor.
> 
> I suggest a new property such as hoverOnType to automatically select the first result, so the user can press Enter to select it.

Instead of my suggested `hoverOnType`, I used `hoverOnSearchChange` to follow the pattern of the `onSearchChange`. In `Autocomplete`, as it doesn't have this property, I used `hoverOnChange` to follow its pattern of `onChange`.

Video of usage:

[](https://www.loom.com/share/39bc718a7630432197770af98f2e9977 "")
*https://www.loom.com/share/39bc718a7630432197770af98f2e9977*

(I tried to generate the docs for the new props but I wasn't being able to due to some weird errors. Not sure what went wrong but it seems that you run the docgen when releasing a new version?)


rtivital:
Thanks!

Issue #3049, opened 2022-11-24
`Table` props like `striped`, `withBorder` or `withColumnBorders` don't work well with `td` `rowSpan` or `colSpan` attributes


mcmxcdev:
### What package has an issue

@mantine/core

### Describe the bug

The code for striping table rows taken from the DOM looks like this:

```
.mantine-1ba39vj[data-striped] tbody tr:nth-of-type(odd)
```

Which leads to wrong styling when the table uses e.g. `rowSpan` props.

***

Truncated code:

```jsx
<Table striped withBorder withColumnBorders>
  <thead>
    <tr>
      <th>Health Plan</th>
      <th>Highmark</th>
    </tr>
  </thead>
  <tbody>
    <tr>
      <td rowSpan={2}>Premium</td>
      <td>Total Premium</td>
      <td>{formatAmount(plan.premium)}</td>
    </tr>
    <tr>
      <td>After allowance</td>
      <td>TODO</td>
    </tr>

    <tr>
      <td rowSpan={4}>Deductible</td>
      <td>Individual: Medical</td>
      <td>{plan.individualMedicalDeductible}</td>
    </tr>
    <tr>
      <td>Individual: Drug</td>
      <td>{plan.individualDrugDeductible}</td>
    </tr>
    <tr>
      <td>Family: Medical</td>
      <td>{plan.familyMedicalDeductible}</td>
    </tr>
    <tr>
      <td>Family: Drug</td>
      <td>{plan.familyDrugDeductible}</td>
    </tr>
  </tbody>
</Table>
```

Visualized:

![Screenshot from 2022-11-24 13-48-39](https://user-images.githubusercontent.com/16797721/203790390-c6524d3c-7c44-4566-947b-a4276e12dcea.png)

### What version of @mantine/hooks page do you have in package.json?

5.8.2

rtivital:
Fixed in 5.9.1

mcmxcdev:
I can confirm that the missing borders were indeed fixed in 5.9.1, but the background colors are still not alternating correctly.

The `td` with `rowSpan` is sometimes gray and other times white, with no clear pattern. You can see this on the left-most column in this picture: 
![Screenshot from 2022-12-05 08-56-36](https://user-images.githubusercontent.com/16797721/205583730-2bbcf33e-5c0c-40b1-994e-5f6ce872d481.png)

rtivital:
[...]

PR #3106, opened 2022-12-03
[@mantine/core] Table: Fix wrong styling with borders and rowspan

seeratawan01:
# Issue: #3049

## Description
Fixed the style issues with table component. There are multiple issues with borders when we use the table component with different props such as striped, withBorder or withColumnBorders.

## What package has changed?
@mantine/core

### Fixes # (issues)

[x] When we use withColumnBorders and rowSpan, the border left is not visible of the cells which are now in front of the rowSpan cell. (This is fixed by using borderRight instead of borderLeft for the cells)

[x] When we use rowSpan and rowspaned cell reaches the last row, the unexpected border bottom is added to the last row. (This is fixed by using borderTop for the rows instead of borderBottom)

[ ] When we use striped, striping works well but when the rowspan is used along, the user might expect the striping to be applied to the cells in front of the rowspan cell regarded as one row. (This is not fixed)

## Screenshots
normal without any props
image

props: withBorder, withColumnBorders and striped
image

props: withColumnBorders and striped
image

### DIFF
--- a/src/mantine-core/src/Table/Table.styles.ts
+++ b/src/mantine-core/src/Table/Table.styles.ts
@@ -79,9 +79,8 @@ export default createStyles(
           '&:last-of-type': {
             borderRight: 'none',
             borderLeft: withColumnBorders ? border : 'none',
-          }
+          },
         },
-
 
         '&[data-striped] tbody tr:nth-of-type(odd)': {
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #3108, opened 2022-12-04
types: make createStyle go to original definition

imranbarbhuiya:
By making this change, a user can click on the className to go to the original definition. Like clicking on `classes.wrapper`, it'll go to the object where `wrapper` was defined. Also renaming `wrapper` will rename it from all the places

### DIFF
--- a/src/mantine-styles/src/tss/create-styles.ts
+++ b/src/mantine-styles/src/tss/create-styles.ts
@@ -49,14 +49,14 @@ function getStyles<Key extends string>(
   return extractStyles(styles);
 }
 
-export function createStyles<Key extends string = string, Params = void>(
+export function createStyles<
+  Key extends string = string,
+  Params = void,
+  Input extends Record<string, CSSObject> = Record<Key, CSSObject>
+>(
   input:
-    | ((
-        theme: MantineTheme,
-        params: Params,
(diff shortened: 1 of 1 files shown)


rtivital:
@imranbarbhuiya please fix failing tests

imranbarbhuiya:
I wasn't at home so did it from phone 😂 I'll fix the formatting once I open my laptop.

rtivital:
@imranbarbhuiya please fix failing tests

imranbarbhuiya:
Hi, I think my ide was merging this prettier config with the default one. I formatted from cli now can u run the workflow again please?

rtivital:
It does not work, example with `Accordion.Panel` component after and before your changes
Снимок экрана 2022-12-05 в 20 17 26
Снимок экрана 2022-12-05 в 20 17 48

imranbarbhuiya:
It might be due to the last change I did which I didn't tested. I'll test everything and will inform you.

imranbarbhuiya:
The `extends` was causing the issue. Now to fix it, I had to use the `generic` and an `assertion` too. But if u don't like the assertion, lmk I'll try to make it work with a simpler solution without breaking existing typings

rtivital:
It works now, thanks!

PR #3140, opened 2022-12-07
[@mantine/core] Image: Edit error setting when src updates

Dinishka:
Currently the Image component's placeholder is only visible in the first time when the src is `null`. If a src is set and then changed to `null` the placeholder wouldn't show because anytime the src changes the error state becomes `false`.

So, I changed it to only set the error state to `false` whenever there's a src. Otherwise it will be set to `true`.


rtivital:
It is already fixed in #3098

Dinishka:
Sorry! My bad. I'll close the pull request.

PR #3149, opened 2022-12-08
[@mantine/core] RingProgress: Add rootColor property

Meierschlumpf:
Pr for discussion #3147
Adds the `rootColor` property to the RingProgress which can be used to define the color of the root curve.


src/mantine-core/src/RingProgress/RingProgress.tsx
```diff
@@ -30,6 +30,9 @@ export interface RingProgressProps
 
   /** Ring sections */
   sections: RingProgressSection[];
+
+  /** Root color */
```
rtivital, line 34: Should be "Color of the root section, key of theme.colors or CSS color value"

rtivital:
Thanks!

PR #3150, opened 2022-12-08
[docs] Add pnpm installation command

Sajarin-M:
**Mantine** docs are awesome! 👏
It would be nicer if we add a pnpm installation command just like npm and yarn.
Added a pnpm icon to the `@mantine/ds` package. is it ok ?


rtivital:
Well, one more PR for pnpm, see https://github.com/mantinedev/mantine/pulls?q=is%3Apr+pnpm

PR #3151, opened 2022-12-08
Update getLineClamp in Text.styles.ts to support 1 liner

paulm17:
This PR updates the getLineClamp function to behave like how Tailwind or Bootstrap works.

### Something to consider?
There is also a case for changing the getLineClamp function to truncate as it's more descriptive.  Also perhaps something like:

```
<Text truncate>Foo foo foo foo</Text> 
```

> Foo foo foo foo...

As a default (1 line) option, but then 

```
<Text truncate={3}>
Foo foo foo foo
Foo foo foo foo
Foo foo foo foo
</Text>
``` 

>Foo foo foo foo
>Foo foo foo foo
>Foo foo foo foo...

To support the 3 line truncate.

It's a breaking change but something to think about for Mantine 6?  🤔


rtivital:
I'd rather have it as a separate prop `truncate` rather than merging it with the `lineClamp`

paulm17:
Have reverted the lineClamp function and created a new truncate param with a function.

rtivital:
Thanks!

PR #3178, opened 2022-12-13
Invalid initial value for "online" state in use-network hook (mantinedev#3177)

Sajarin-M:


rtivital:
This change will cause hydration mismatch if used with ssr.

Sajarin-M:
> This change will cause hydration mismatch if used with ssr.

Is there any other way to fix this

rtivital:
No

Sajarin-M:
does using `navigator.onLine` in the `useEffect` also cause a hydration mismatch ?

rtivital:
No, it does not cause it, component rerenders on client.

Sajarin-M:
then we can update it to correct state on rerender

rtivital:
Thanks!

PR #3181, opened 2022-12-13
[@mantine/spotlight] Allow overriding search input size

Strajk:
Hey, it would be awesome to have the ability to override search input size via searchInputProps on Spotlight/SpotlightProvider 🙏

image

Checked locally, also verified in Storybook, test/types passing ☑️

(Mantine is great btw!)


rtivital:
Thanks!

PR #3188, opened 2022-12-14
[@mantine/core] improve Stepper API

jvdsande:
As discussed on Discord, this PR adds some flexibility to Stepper's API:

- Adds a `allowSelectNextSteps` props on Stepper, defaults to `true` to match current behavior. When false, clicking is disabled by default on upcoming steps.
- Adds a `StepFragmentComponent` type for all props currently accepting a React.Node. This is a component that receives the current step index as props, allowing the use of reusable templates easily.

I added some demos, don't hesitate to ask for changes in them if they are not useful enough!


rtivital:
Thanks!

PR #3191, opened 2022-12-15
[core] Update yarn cache in github action

tuliren:
- Currently, the `yarn` cache in the `pull_request` always misses. The total time to install dependencies take ~2 minutes.

  Screen Shot 2022-12-15 at 01 42 44

- This PR fixes this issue by using the latest `setup-node@v3`.
- With the new cache, it takes ~30 seconds to download the cache, and ~1 minute 10 seconds to install the dependencies. So in total it saves 20 seconds. It's not a huge improvement because the cache size is very large and it takes a while to download.
- Also there are no more deprecated `set-output` / `save-state` command warnings.
- Please note that the cache will only start to work on the second run. So on this PR, the installation time will remain the same. But here is a screenshot of subsequent runs:

  Screen Shot 2022-12-15 at 01 40 46

### DIFF
--- a/.github/workflows/pull_request.yml
+++ b/.github/workflows/pull_request.yml
@@ -13,22 +13,14 @@ jobs:
   test_pull_request:
     runs-on: ubuntu-latest
     steps:
-      - name: Get Yarn cache path
-        id: yarn-cache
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

Issue #3196, opened 2022-12-16
The icon in Rich Text Editor Link cannot be changed


Avas77:
### What package has an issue

@mantine/core

### Describe the bug

We can change icons of Rich Text Editor controls by passing icon prop. This does not work for Rich Text Editor Link control and we are stuck with the default icon

### What version of @mantine/hooks page do you have in package.json?

5.9.3

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/fervent-minsky-4wv5yh?file=/src/Editor.js

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

rtivital:
Fixed in 5.9.5

PR #3205, opened 2022-12-17
[@mantine/tiptap]: Add icon prop to link control base

Avas77:
As mentioned in the issue https://github.com/mantinedev/mantine/issues/3196, I have added the icon prop to the Link Control Base and set the default icon of Link Control Base to Link Icon


rtivital:
Thanks for a PR, the issue is already fixed

PR #3232, opened 2022-12-23
[@mantine/docs] Update Container documentation to remove outdated component prop

joelzwarrington:
`padding` is not an accepted component prop for Container.


rtivital:
Thanks!

PR #3233, opened 2022-12-23
Change from "below" to "behind"

bissbr01:
It tripped me up so I thought worth fixing for others.  "below" indicates content displayed after content, while "behind" indicates content displayed under content.


Xapu1337:
both are right but it's IMO not change-worthy. in terms of zDepth it is below

rtivital:
Thanks!

PR #3237, opened 2022-12-24
[@mantine/dates] CalendarBase: implement `allowedLevels` to display only certain levels

AmruthPillai:
This PR makes changes to the CalendarBase component, and the associated story.

https://user-images.githubusercontent.com/1134738/209449008-9788f460-2c32-46ba-a598-7ccd95cd5ab8.mov

It introduces a new prop `allowedLevels` that defaults to all three levels (date, month, year). If a user passes down this array with just one or two of these options, it restricts input to just these levels.

I had a use-case for this, where I want the user to be able to select just the month and year, but not a specific date.

As a hack, I tried using a controlled component and closing the calendar popover as soon as onMonthChange was called, but this causes the date picker still to flash for a brief second.


rtivital:
These features are already implemented in 6.0 as separate components, changes to the dates package are not planned until 6.0 is released.

https://rtivital.github.io/mantine-dates-6/

AmruthPillai:
@rtivital this is actually great, I'll use this package instead. Thank you for this amazing UI library. It's really rare to find a framework that looks so much like how I would eventually design it to be, and also that provides so much customization through emotion CSS. It's really one of the best ones out there. Thank you for your time to build and support the development of Mantine!

rtivital:
Thanks for kind words!

PR #3238, opened 2022-12-25
[@mantine/core] README fix: correct readme install instructions for core

KevinVandy:


rtivital:
Updated instructions are incorrect.

PR #3293, opened 2023-01-05
[@mantine/core] Stack: Fix default justify property

stivaugoin:
There is no `top` property for `justify-content`. I search elsewhere and didn't found other occurrence.


rtivital:
Thanks!

PR #3298, opened 2023-01-06
[@mantine/hooks] use-local-storage: Fix event not called when remove …

auronsan1st:
fixes https://github.com/mantinedev/mantine/discussions/3279

it's related remove item will trigger no event on window caused localstorage just save string value. 
so we just need add custom event like setStorageValue does.

checked use-session-storage works normally.


rtivital:
Thanks!

PR #3299, opened 2023-01-06
[@mantine/dates] Fix openDropdownOnClear on DateRangePicker

maixiu:
openDropdownOnClear would still open dropdown on DateRangePicker when set to false


rtivital:
Thanks!

PR #3300, opened 2023-01-06
[@mantine/modals] Fixed closeModal async logic

### LINKED ISSUE #3289: Modals manager doesn't handle closing modals properly
### What package has an issue

@mantine/modals

### Describe the bug

When closing a modal which is not the one currently opened, causes weird behaviors.

for example when :
- open modal 1
- open modal 2
- close modal 1
Closes all the modals when it should only be removing modal 1 from the stack of the manager.

5.10.0

https://codesandbox.io/s/crazy-lake-pi78fl?file=/src/App.tsx

Yes

Yes

### Possible fix

I believe that the bit of code here assumes that we always close the last modal opened :
https://github.com/mantinedev/mantine/blob/db0551288e6fcccea5bd321c9255f61b1de61a6b/src/mantine-modals/src/reducer.ts#L35

Which I believe could be changed to something like this : 
```ts
    // ...
    case 'CLOSE': {
      const modals = state.modals.filter((m) => m.id !== action.payload)
      return {
        current: modals[modals.length - 1] || null,
        modals,
      };
    }
    // ...
```

Njiallu:
Closes #3289

Made the logic simpler and not assume that the closed modal is the last in the stack.

### DIFF
--- a/src/mantine-modals/src/ModalsProvider.tsx
+++ b/src/mantine-modals/src/ModalsProvider.tsx
@@ -122,11 +122,6 @@ export function ModalsProvider({ children, modalProps, labels, modals }: ModalsP
   };
 
   const closeModal = (id: string, canceled?: boolean) => {
-    if (state.modals.length <= 1) {
-      closeAll(canceled);
-      return;
-    }
-
     const modal = state.modals.find((item) => item.id === id);
     if (modal?.type === 'confirm' && canceled) {
       modal.props?.onCancel?.();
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

D3SOX:
When trying to close a modal like this
```ts
const submit = () => ctx.closeModal(id);
const id = ctx.openModal(...);
```
I get an `Uncaught TypeErorr: modal is undefined`

cyantree:
@D3SOX FYI, I've taken another shot at this and also investigated your reported behaviour: https://github.com/mantinedev/mantine/pull/3498#issuecomment-1417168388

PR #3303, opened 2023-01-06
[@mantine/core] InlineInput: Replace display inline-flex by flex

### LINKED ISSUE #3295: InlineInput body section is displayed as `inline-flex`
### What package has an issue

@mantine/core

### Describe the bug

Before 5.10.0, Checkbox, Switch and Radio components had a `body` div with `display: flex`. 

After https://github.com/mantinedev/mantine/commit/3f3f5d9d8c13d60d7ccb4a6f2d6eb572f623150e and the creation of `InlineInput`, the `body` div is now `display: inline-flex`.

* With `inline-flex`
CleanShot 2023-01-05 at 17 43 54@2x

* With `flex`
CleanShot 2023-01-05 at 17 45 20@2x

^5.9.5

https://codesandbox.io/s/mantine-switch-body-flex-gxvb70?file=/src/App.tsx

Yes

Yes

### Possible fix

Replace `display: inline-flex` by `display: flex` but I not sure if there are a reason to use `inline-flex`.

stivaugoin:
The problem with `inline-flex` was only visible on `Switch`. Nothing changes on `Checkbox` and `Radio`. 

Let me know if I missed something.

closes #3295

### DIFF
--- a/src/mantine-core/src/InlineInput/InlineInput.styles.ts
+++ b/src/mantine-core/src/InlineInput/InlineInput.styles.ts
@@ -17,7 +17,7 @@ export default createStyles((theme, { labelPosition, size }: InlineInputStylesPa
   root: {},
 
   body: {
-    display: 'inline-flex',
+    display: 'flex',
   },
 
   labelWrapper: {


rtivital:
Thanks!

PR #3312, opened 2023-01-07
[v6][@mantine/core] Move TagInput and PinInput to core

auronsan1st:
@rtivital 

1. moved tagInput to core. included use newest multiselect and delete stories using storiesOf and use similar like multiselect story
2. moved pinInput to core

`npm run build:all` success
`npm test` fail caused chipGroup need be solved 

regarding the UI of pinInput I think we need improvement. when input can't see the number at all.

and I have code about taginput add inputField position https://github.com/mantinedev/mantine/discussions/1056
let me know if want included on this PR too


rtivital:
Thanks!

PR #3340, opened 2023-01-11
[@mantine/core] [docs] Fix allowStepSelect regression

### LINKED ISSUE #3335: Stepper.Step "allowStepSelect" props not working
### What package has an issue

@mantine/core

### Describe the bug

allowStepSelect is not working on Mantine 5.10 (seems to be the same problem on the website doc)

5.10.0

https://mantine.dev/core/stepper/#allow-step-select

No

No

### Possible fix

Was working well on previous versions

rmzNadir:
- Fixes #3335
- Adds a few tests to prevent this from happening in the future
- Updates Stepper docs with an example of a common usecase for allowStepSelect 

This is my first contribution to the repo so feel free to roast me if something isn't up to the standards :)

### DIFF
--- a/src/mantine-core/src/Stepper/Stepper.tsx
+++ b/src/mantine-core/src/Stepper/Stepper.tsx
@@ -117,10 +117,20 @@ export const Stepper: StepperComponent = forwardRef<HTMLDivElement, StepperProps
   const items = _children.reduce<React.ReactElement[]>((acc, item, index) => {
     const state =
       active === index ? 'stepProgress' : active > index ? 'stepCompleted' : 'stepInactive';
-    const shouldAllowSelect = state === 'stepCompleted' || allowNextStepsSelect;
-    typeof item.props.allowStepSelect === 'boolean'
-      ? item.props.allowStepSelect
-      : typeof onStepClick === 'function';
+
+    const shouldAllowSelect = () => {
+      if (typeof onStepClick !== 'function') {
+        return false;
+      }
+
+      if (typeof item.props.allowStepSelect === 'boolean') {
+        return item.props.allowStepSelect;
+      }
+
+      return state === 'stepCompleted' || allowNextStepsSelect;
+    };
+
+    const isStepSelectionEnabled = shouldAllowSelect();
 
     acc.push(
       cloneElement(item, {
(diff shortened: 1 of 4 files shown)


src/mantine-core/src/Stepper/Stepper.tsx
```diff
@@ -117,10 +117,20 @@ export const Stepper: StepperComponent = forwardRef<HTMLDivElement, StepperProps
...
-    typeof item.props.allowStepSelect === 'boolean'
-      ? item.props.allowStepSelect
-      : typeof onStepClick === 'function';
+
+    const shouldAllowSelect = () => {
+      if (typeof onStepClick !== 'function') {
+        return false;
+      }
+
+      if (typeof item.props.allowStepSelect === 'boolean') {
```
rmzNadir, line 126: Giving priority to the Step's allowStepSelect prop made sense for more fine grain control over the step selection.

jvdsande: Indeed, and here was actually the typo I added in the previous code... I forgot a ` &&`  operator, with the "typeof item.props..." part staying alone on its own line for nothing.

jvdsande: In terms of readability I agree with your change; in terms of performance since this is in a loop, the added function call might not be the best idea. Then again, I don't imagine a stepper component getting tens or hundreds of steps :)

jvdsande: The following will achieve the same, without the function call:

```ts
    const defaultAllowSelect = state === 'stepCompleted' || allowNextStepsSelect;
    const shouldAllowSelect =
      defaultAllowSelect &&
      (typeof item.props.allowStepSelect === 'boolean'
        ? item.props.allowStepSelect
        : typeof onStepClick === 'function');
``` 

But it's arguably more difficult to read.

src/mantine-demos/src/demos/core/Stepper/Stepper.demo.allowStepSelect.tsx
```diff
@@ -9,29 +9,57 @@ import { Stepper, Button, Group } from '@mantine/core';
...
+    if (isOutOfBounds) {
+      return;
+    }
+
+    setActive(nextStep);
+    setHighestStepVisited((hSC) => Math.max(hSC, nextStep));
+  };
+
+  // Allow the user to freely go back and forth between visited steps.
+  const shouldAllowSelectStep = (step: number) => highestStepVisited >= step && active !== step;
```
jvdsande, line 26: I like this variant of the demo, good idea!

src/mantine-core/src/Stepper/Stepper.tsx
rmzNadir: The performance change is so negligible (even with a thousand steps) that it may as well be part of the margin of error and at that point I'd rather favor readability

jvdsande: Yeah I agree. This was definitely a nitpick.

rtivital:
Thanks!

PR #3355, opened 2023-01-13
Revaluate initialValues

jsheely:
Acts like enabled flag with `initialData` in react-query

This is useful when the `useForm` hook is used in react context in a shared parent component where initialValues are pulled from another hook. Revaluating the `initialValues` allows the form to reset gracefully to a new initial state.

This prevents race conditions triggered by attempting to use the below from outside of the form

```
useEffect(() => {
	form.setValues(myValues);
	form.resetDirty(myValues);
}, [myValues])
```

- [ ] Discuss `revaluateInitialValues` public API property with @rtivital
- [x] Storybook
- [ ] Unit test

### DIFF
--- a/src/mantine-form/src/stories/Form.initialValuesReset.story.tsx
+++ b/src/mantine-form/src/stories/Form.initialValuesReset.story.tsx
@@ -1,0 +1,53 @@
+import React, { useState } from 'react';
+import { TextInput, Code, Button, Group } from '@mantine/core';
+import { useForm } from '../use-form';
+
+export default { title: 'Form' };
+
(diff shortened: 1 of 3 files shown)


rtivital:
We had this behavior before, it was confusing and led to unexpected behaviors (mostly with data fetching).

jsheely:
@rtivital That's fair. I am doing some more investigations if there are other cleaner ways to handle this. But not being able to update the initialValues without destroying the whole component tree does always make it a little messy when triggering the reset behavior

rtivital:

PR #3356, opened 2023-01-13
Update NumberInput.demo.formatter.tsx

SrBrahma:
This PR uses a Regexp that supports more than 3 decimals digits. The current Regexp adds ',' after the '.' (invalid math!), see this example of mine:

With the current Regexp example
image

With the StackOverflow Regexp
image

Source: https://stackoverflow.com/a/2901298 (last item)

I know it's just an example and it isn't required to be followed, but took me some time to understand the issue I was having and to find this improved regexp. Seems a good change to help other users in similar situations.


rtivital:
Please fix errors

PR #3410, opened 2023-01-23
feat: allow shortcuts on contentEditable

colinricardo:
currently even if you ignoreTags to empty, the shortcut to show the spotlight will not trigger on elements which have contentEditable=true

this PR allows you to specify that behavior


rtivital:
Thanks!

PR #3431, opened 2023-01-26
[@mantine-form] zodResolver: Make schema parameter typesafe

SrBrahma:
I had a situation where a wrong & invalid data was being passed to the zodResolver as the schema argument.

This makes the zodResolver schema parameter to require the safeParse property.

I don't see a way to type test this in Mantine repo. In my codebase it didn't throw false negatives and it did complain properly.

### DIFF
--- a/src/mantine-form/src/resolvers/zod-resolver/zod-resolver.ts
+++ b/src/mantine-form/src/resolvers/zod-resolver/zod-resolver.ts
@@ -5,22 +5,24 @@ interface ZodError {
   message: string;
 }
 
-interface ZodResults {
-  success: boolean;
+interface ZodParseSuccess {
+  success: true;
+}
+
+interface ZodParseError {
+  success: false;
   error: {
     errors: ZodError[];
   };
 }
 
-interface ZodSchema {
-  safeParse(values: Record<string, any>): ZodResults;
+interface ZodSchema<T extends Record<string, any>> {
+  safeParse(values: T): ZodParseSuccess | ZodParseError;
 }
 
-export function zodResolver<T extends Record<string, any>>(schema: any) {
-  const _schema: ZodSchema = schema;
-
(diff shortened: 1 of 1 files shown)


src/mantine-form/src/resolvers/zod-resolver/zod-resolver.ts
```diff
@@ -5,30 +5,32 @@ interface ZodError {
...
+    const parsed = schema.safeParse(values);
 
     if (parsed.success) {
       return {};
     }
 
     const results = {};
 
-    parsed.error.errors.forEach((error) => {
+    (parsed as ZodParseError).error.errors.forEach((error) => {
```
SrBrahma, line 33: Required, not sure why TS don't infer this

rmzNadir: You can try narrowing the type using the `error` property, e.g. `const isError = 'error' in parsed` and then `if (!isError)` should do the trick.

src/mantine-form/src/resolvers/zod-resolver/zod-resolver.ts
```diff
@@ -5,30 +5,32 @@ interface ZodError {
...
+interface ZodParseError {
+  success: false;
   error: {
     errors: ZodError[];
   };
 }
 
-interface ZodSchema {
-  safeParse(values: Record<string, any>): ZodResults;
+interface ZodSchema<T extends Record<string, any>> {
```
rmzNadir, line 19: Just wondering, is using the exported zod types not a viable option here?

rtivital:
Thanks!

src/mantine-form/src/resolvers/zod-resolver/zod-resolver.ts
SrBrahma: No, we don't want Mantine users to have Zod as dependency/devDep just for the type.

PR #3436, opened 2023-01-27
[@mantine/core] TransferList: add `transferAllMatchingFilter` prop

justinsmid:
Adds the `transferAllMatchingFilter` prop to `TransferList` that, when true, makes it so clicking the `transfer all` button transfers only items that match the current search.

Resolves #3430

### DIFF
--- a/src/mantine-core/src/TransferList/RenderList/RenderList.tsx
+++ b/src/mantine-core/src/TransferList/RenderList/RenderList.tsx
@@ -36,6 +36,7 @@ export interface RenderListProps extends DefaultProps<RenderListStylesNames> {
   limit?: number;
   transferIcon?: React.FunctionComponent<{ reversed }>;
   transferAllIcon?: React.FunctionComponent<{ reversed }>;
+  transferAllMatchingFilter: boolean;
 }
 
 const icons = {
@@ -61,6 +62,7 @@ export function RenderList({
   listComponent,
   transferIcon: TransferIcon,
   transferAllIcon: TransferAllIcon,
+  transferAllMatchingFilter,
   searchPlaceholder,
   query,
   onSearch,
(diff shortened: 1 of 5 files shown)


src/mantine-core/src/TransferList/RenderList/RenderList.tsx
```diff
@@ -36,6 +36,7 @@ export interface RenderListProps extends DefaultProps<RenderListStylesNames> {
   limit?: number;
   transferIcon?: React.FunctionComponent<{ reversed }>;
   transferAllIcon?: React.FunctionComponent<{ reversed }>;
+  transferAllMatchingFilter: boolean;
```
justinsmid, line 39: I think `transferAllMatchingSearch` might make more sense to users, but I named it `Filter` for now since that it what the fitlering prop is called. Let me know if you have any preference.

src/mantine-core/src/TransferList/TransferList.tsx
```diff
@@ -124,8 +129,19 @@ export const TransferList = forwardRef<HTMLDivElement, TransferListProps>((props
   const handleMoveAll = (listIndex: 0 | 1) => {
     const items: TransferListData = Array(2) as any;
     const moveToIndex = listIndex === 0 ? 1 : 0;
-    items[listIndex] = [];
-    items[moveToIndex] = [...value[moveToIndex], ...value[listIndex]];
+
+    if (transferAllMatchingFilter) {
+      const query = search[listIndex];
+      const shownItems = value[listIndex].filter((item) => filter(query, item)).slice(0, limit);
```
justinsmid, line 135: Not too sure whether `.slice`-ing here makes sense. E.g: would we want this to only transfer the items that are actually **shown**, or all the items that match the filter?

rtivital:
Hi @justinsmid, is PR finished, or you want to change anything else?

justinsmid:
Hi @rtivital, the PR is finished as far as I'm concerned.

rtivital:
Okay, Thanks!

PR #3451, opened 2023-01-31
[@mantine/spotlight] Spotlight: Pass transitionDuration to GroupedTransition

DanielGibbsNZ:
This prevents `setTimeout` from being called when the Spotlight transitionDuration is 0.

This is the same fix as #1089, just for the Spotlight component.


rtivital:
Already changed in 6.0 – https://v6.mantine.dev/others/spotlight/
Spotlight now accepts all Modal component props

PR #3467, opened 2023-02-01
[@mantine/spotlight] Spotlight: Add returnFocusOnClose prop

DanielGibbsNZ:
Any objection to making the focus return behaviour of spotlight optional?


rtivital:
It is already implemented in 6.0, Spotlight component was migrated to Modal under the hood – https://v6.mantine.dev/others/spotlight/

Issue #3376, opened 2023-01-16
Switch out deprecated `@floating-ui/react-dom-interactions` for `@floating-ui/react`


mcmxcdev:
### What package has an issue

@mantine/core

### Describe the bug

`pnpm` reports the following when running install: ` WARN  deprecated @floating-ui/react-dom-interactions@0.10.3: Package renamed to @floating-ui/react`

You can also see the deprecation message on: https://www.npmjs.com/package/@floating-ui/react-dom-interactions

### What version of @mantine/hooks page do you have in package.json?

^5.10.0

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

None

rtivital:
Fixed in 5.10.3

PR #3471, opened 2023-02-02
[@mantine/core] Dependency: migrate `@floating-ui/react-dom-interactions` to `@floating-ui/react`

genffy:
issue: https://github.com/mantinedev/mantine/issues/3376

### DIFF
--- a/package.json
+++ b/package.json
@@ -134,7 +134,7 @@
     "@emotion/react": "11.9.3",
     "@emotion/server": "11.4.0",
     "@emotion/styled": "^11.10.0",
-    "@floating-ui/react-dom-interactions": "^0.10.1",
+    "@floating-ui/react": "^0.19.1",
     "@radix-ui/react-scroll-area": "1.0.2",
     "@tabler/icons": "^1.119.0",
     "@tiptap/extension-code-block-lowlight": "^2.0.0-beta.202",
--- a/src/mantine-core/package.json
+++ b/src/mantine-core/package.json
@@ -36,7 +36,7 @@
     "@mantine/styles": "5.10.2",
     "@radix-ui/react-scroll-area": "1.0.2",
     "react-textarea-autosize": "8.3.4",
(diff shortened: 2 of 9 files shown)


package.json
```diff
@@ -134,7 +134,6 @@
     "@emotion/react": "11.9.3",
     "@emotion/server": "11.4.0",
     "@emotion/styled": "^11.10.0",
-    "@floating-ui/react-dom-interactions": "^0.10.1",
```
rtivital, line 137: Dependency should not be removed from main package.json

src/mantine-demos/tsconfig.build.json
```diff
@@ -34,7 +34,8 @@
       "@mantine/nprogress": ["../mantine-nprogress/src"],
       "@mantine/labs": ["../mantine-labs/src"],
       "@mantine/tiptap": ["../mantine-tiptap/src"],
-      "@mantine/rte": ["../mantine-rte/src"]
+      "@mantine/rte": ["../mantine-rte/src"],
+      "@mantine/ds": ["../mantine-ds/src"]
```
rtivital, line 38: tsconfig files should not be changed.

genffy: when i build project at latest node LTS(v16.17.0) firstly, `tsc --project tsconfig.build.json` will throw error with `@mantine/ds`. So i think `@mantine/ds`'s path should be configured

package.json
genffy: the `@floating-ui/react` pkg just be used at `@mantine/core`. The main `package.json` should contain all workspaces dependencies?

rtivital: Yes, it should, otherwise documentation will not have access to these dependencies.

src/mantine-demos/tsconfig.build.json
rtivital: Works fine for me with current implementation, i'd rather not change that.

rtivital:
Thanks!

PR #3475, opened 2023-02-02
[@mantine/core] Skeleton: Allow overflow when not visible

### LINKED ISSUE #3442: Invisible Skeleton prevents overflowed content from showing
### What package has an issue

@mantine/core

### Describe the bug

Skeleton component has `overflow: hidden` [defined here](https://github.com/mantinedev/mantine/blob/2fe9c4ab078ec03ddcc97613a1059e7db8b4e4ed/src/mantine-core/src/Skeleton/Skeleton.styles.ts#L23), therefore it prevents shadows or other overflowed content from showing.

It should not have `overflow: visible` when `visible` prop is `false`.

^5.10.2

https://codesandbox.io/s/icy-breeze-tyophy?file=/src/App.tsx

Yes

Yes

### Possible fix

Move `overflow: 'hidden'` from `root` to `visible` class.

omerfarukdogan:
Fixes #3442

### DIFF
--- a/src/mantine-core/src/Skeleton/Skeleton.styles.ts
+++ b/src/mantine-core/src/Skeleton/Skeleton.styles.ts
@@ -20,11 +20,12 @@ export default createStyles(
       width: circle ? height : width,
       borderRadius: circle ? height : theme.fn.radius(radius),
       position: 'relative',
-      overflow: 'hidden',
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #3484, opened 2023-02-03
Fix/component default props type

congyaqwq:
First of all, I would like to apologize for submitting an invalid PR.

Last time when I submitted a PR, I made a mistake, the result of `[keyof U] extends [keyof T]` always returned `true`, so the type I wrote actually did not take effect. This time I optimized the type again, when passing in componentDefaultProps, it will deduce the result according to the parameter type without manual writing of the type of componentDefaultProps, just like this.

```typescript
const defaultProps = {
  disabled: false,
  size: 'sm' as const,
  transitionDuration: 200,
};
  const {
    className,
    disabled,
    data: _data,
    name,
    value,
    onChange,
    color,
    fullWidth,
    radius,
    size,
    transitionDuration,
    transitionTimingFunction,
    classNames,
    styles,
    defaultValue,
    orientation,
    unstyled,
    ...others
  } = useComponentDefaultProps('SegmentedControl', defaultProps, props);
```

Thus, you will get correct type.

However, it may bring some uncommon errors, for example, in the component `SegmentedControl`, because the type of the passed in `componentDefaultProps` is `Partial<Props>`, resulting in each property being processed by `useComponentDefaultProps`, causing the type of the property data to be `(string[] | SegmentedControlItem[]) & NonNullable<string[] | SegmentedControlItem[]>`, which would cause an error when calling `map` with this type. I don't know how to deal with this error, which is also why I rewrote the `SegmentedControl` component.


rtivital:
Hi @congyaqwq, thanks for a pull request! Maybe you can help me with `useInputProps` type as well? I was unable to make it strongly typed, you seem to have a decent ts experience, so perhaps you have an idea how to type it the same way as `useComponentDefaultProps`. https://github.com/mantinedev/mantine/blob/v6/src/mantine-core/src/Input/use-input-props.ts#L17

congyaqwq:
> Hi @congyaqwq, thanks for a pull request! Maybe you can help me with `useInputProps` type as well? I was unable to make it strongly typed, you seem to have a decent ts experience, so perhaps you have an idea how to type it the same way as `useComponentDefaultProps`. https://github.com/mantinedev/mantine/blob/v6/src/mantine-core/src/Input/use-input-props.ts#L17

I will try my best.

rtivital:
Thanks!

7iomka:
This PR also make strong validation for components props.

image

image

This is a good signal to fix that, and exclude props passed statically from type of mine component wrapper. Or, if I want to support this prop, I need to find some way to extend it with needed one.
Btw, thanks

PR #3486, opened 2023-02-03
fix: update useInputProps to strong type

congyaqwq:
Fix: `useInputProps`
The reason for the type error here is not clear to me, but I gave it a type through `as` and it worked.

I hope this is what you expect.


rtivital:
Thanks!

PR #3487, opened 2023-02-03
[docs] Fix an overflow issue in iOS

JayeHa:
I found that some pages have overflow issues when I access Mantine docs on my iPhone.

image
image


rtivital:
Fixed in 6.0

PR #3490, opened 2023-02-03
[@mantine/core] MultiSelect: Add missing default for dropdownPosition

jasonjmcghee:
The [documentation](https://mantine.dev/core/multi-select/#dropdown-position) claims `MultiSelect` uses `"flip"` as the default for `dropdownPosition`, but is instead `undefined`.


rtivital:
Thanks!

PR #3498, opened 2023-02-04
[@mantine/modals] Fix various issues regarding usage assumptions

cyantree:
This PR fixes some issues regarding multi modal handling.
- `closeModal()` crashed when providing an invalid id
- If only one modal was left calling `closeModal()` would close this modal regardless whether the given id is correct
- `closeModal()` assumed that the given modal is the topmost one and therefore modified the state accordingly. This might lead to a not topmost modal suddenly popping up

The fixes now introduce a more defensive approach by checking for validity and not making these assumptions.

### DIFF
--- a/src/mantine-modals/src/ModalsProvider.tsx
+++ b/src/mantine-modals/src/ModalsProvider.tsx
@@ -122,16 +122,15 @@ export function ModalsProvider({ children, modalProps, labels, modals }: ModalsP
   };
 
   const closeModal = (id: string, canceled?: boolean) => {
-    if (state.modals.length <= 1) {
-      closeAll(canceled);
+    const modal = state.modals.find((item) => item.id === id);
+    if (!modal) {
       return;
     }
 
-    const modal = state.modals.find((item) => item.id === id);
-    if (modal?.type === 'confirm' && canceled) {
-      modal.props?.onCancel?.();
+    if (modal.type === 'confirm' && canceled) {
+      modal.props.onCancel?.();
     }
-    modal?.props?.onClose?.();
+    modal.props.onClose?.();
     dispatch({ type: 'CLOSE', payload: modal.id });
   };
 
--- a/src/mantine-modals/src/ModalsProvider.tsx
+++ b/src/mantine-modals/src/ModalsProvider.tsx
@@ -1,4 +1,4 @@
-import React, { useReducer, useRef } from 'react';
+import React, { useCallback, useReducer, useRef } from 'react';
 import { Modal, getDefaultZIndex } from '@mantine/core';
 import { randomId } from '@mantine/hooks';
 import {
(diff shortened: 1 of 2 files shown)


rtivital:
We already had a similar PR that broke things #3300
I can merge it, but if any bugs are reported, then I'll roll back it.

cyantree:
Thanks for your feedback.
IMHO that PR didn't broke things because it introduced errors. The described error existed before but the changes in the PR made it to come up more often (before it would occur if you have multiple open modals and try to close an invalid id, afterwards it happened everytime you would try to close an invalid id).

My PR is roughly the same but fixes this case by checking for the modal before doing further things.
https://github.com/mantinedev/mantine/pull/3498/files#diff-1880aa9d1a0da4baa8dba2aafc0c751d4f40b60d937cfbe90e02b3fe9dd9878bR126-R128

But sure you shouldn't merge something you're not comfortable with. I'm happy to assist in getting this behaviour right. 👍 

I've also seen that the codebase currently doesn't use TS strict mode which would catch these things. Is there any specific reason to do so? If you're open to enabling it we might discuss this in another issue as it could amount in a bit of work.

rtivital:
Strict mode is not used because it does not bring any real value to a component library. It just introduces a lot of additional needless checks. It is not planned to change it. 

Regarding your PR: I'm saying that I do not have a capacity to validate these changes, as I do not use modals manager advanced features in any of my applications. If someone reports a regression, the PR will be rolled back. If not, then it will stay in the codebase.

cyantree:
Ok. Then please wait with merging until I've checked the code sample from PR #3300. I want to know whether something is wrong with it or something different is going on.
Because at first sight there `id` gets used before being assigned which might be no problem in this specific case but could be an issue in a more complex real world scenario.

cyantree:
Ok. I've checked the [example ](https://github.com/mantinedev/mantine/pull/3300#issuecomment-1385793013)out and also tested it with my PR:
With my PR the error wouldn't occur however the modal wouldn't close anymore.

Currently this works because if only one (or even none) modal is open simply `closeAll()` will be closed which bruteforcely closes everything that is. If multiple modals would be open this shortcut wouldn't apply anymore and the issue would come up again.

The reason is the following:
The sample author supposedly uses something like `const ctx = useModals()` to get hold of the `ModalsProvider`.
In the sample `const submit = () => ctx.closeModal(id);` gets passed to the children of `openModal()` and therefore never changes again after opening the modal.
However `useModals()` always returns a context object bound to the **current** state. So after running `openModal()` the state changes, the component rerenders and gets a new contex version from `useModals()` which now is aware of the new modal. But because the changed handler `submit()` won't be updated in the modal component it still uses the version where the modal hasn't been opened yet.

So when running the handler `submit()` and trying to close the dialog it uses the old context and therefore won't find the dialog. In the version of #3300 it then crashed because it expected the modal to exist. In my version it won't do anything because of the existance check.

So this leads to more questions:
Is the provided sample wrong and the behaviour to be expected? In this case I consider my PR more safe as it checks for the modal before doing anything else. However it would break setups that rely on this shortcut behaviour. So maybe it would be a good thing to postpone it to v6.

Or should the sample be a perfectly fine usecase? In this case we need to make sure the context methods always work on the most current state. This could be done by wrapping `state` with `useRef()` and therefore using the ref version of it. This additional change shouldn't break anything and therefore can be deployed in v5.

cyantree:
I've thought about this a bit more and think the example in #3300 is a fair usecase and the developer shouldn't have to pay attention to this. Therefore I implemented the suggested `useRef()` solution which makes sure all methods use the current state. I've cached the exported methods via `useCallback()` to underline this and also reduce possible rerenderings.

rtivital:
Thanks!

PR #3526, opened 2023-02-08
[@mantine/Hooks] use-focus-trap: aria-hidden Fix

### LINKED ISSUE #3515: Menu with flipping dropdown does not remove aria-hidden on close
### What package has an issue

@mantine/core

### Describe the bug

When using a Menu that is set up so that the drop down "flips", closing the menu does not remove all the `aria-hidden` attributes throughout the page, leaving the page inaccessible.

I think this has something to do with use-focus-trap, but I'm not sure why it fails when the dropdown flips.

Currently working around this by adding
```ts
onClose={() =>
  Array.from(document.querySelectorAll('body > :not(script)')).map((node) =>
    node.removeAttribute('aria-hidden')
  )}
```
to my Menu.

5.10.3

https://codesandbox.io/s/frosty-tdd-s46ff0

No

Yes

### Possible fix

_No response_

DanB-dev:
#3515

### DIFF
--- a/src/mantine-hooks/src/use-focus-trap/use-focus-trap.ts
+++ b/src/mantine-hooks/src/use-focus-trap/use-focus-trap.ts
@@ -14,6 +14,10 @@ export function useFocusTrap(active = true): (instance: HTMLElement | null) => v
       }
 
       if (node === null) {
+        if (restoreAria.current) {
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #3530, opened 2023-02-09
[@mantine/dates] Clear invalid range value on dropdown close and add labelSeprator to range value

ikapta:
1、Clear Invalid range value like  `[2023, null]`, while dropdown closed, in v5 there had this logic in DateRangePicker. (I'm not sure this is intentional by design.
image

2、Add `labelSeprator` to `DatePickerInput` \ `MonthPickerInput` \ `YearPickerInput`.
image


ikapta:
Hi, type errors not exist in this commit files
image

PR #3532, opened 2023-02-09
[@mantine/Core] Modal: fix header inappropriate margin right 

daikaiyu:
It's just a minor style problem.The original Header has ' margin-right: -9 ' and seems to be used to offset the position of the close button.This works well under the default padding. But  shrinking the padding will result in a very strange style.

Example:
1.Default padding and with ' margin-right: -9 '

image

2.padding = "sm" and with ' margin-right: -9 '
When hovering over the close button, it will be very close to the modal right.

image

3.padding = "sm" and without ' margin-right: -9 '

image

It may be a little hasty to delete it directly.It may be possible to set it dynamically based on the padding value passed in, but this requires more testing.

### DIFF
--- a/src/mantine-core/src/Modal/Modal.styles.ts
+++ b/src/mantine-core/src/Modal/Modal.styles.ts
@@ -87,7 +87,6 @@ export default createStyles(
       alignItems: 'center',
       justifyContent: 'space-between',
       marginBottom: theme.spacing.md,
-      marginRight: -9,
     },
 
     body: {


lgaspari:
Hey @daikaiyu, I've past through this a couple of weeks ago and I think the idea is to align the X with the content despite the background. Yes, it looks slightly off but kinda makes sense having both the `title` and the `x` aligned inside the same box.

In my opinion, keeping as-is looks way better than removing that negative margin. Let's see what the maintainers think about this!

PR #3535, opened 2023-02-10
[@mantine/dates] Clear invalid range value on dropdown closed and add labelSeprator to DatesProvider

ikapta:
1、Clear Invalid range value like  `[2023, null]`, while dropdown closed, in v5 there had this logic in DateRangePicker. (I'm not sure this is intentional by design.
image

2、Add `labelSeprator` to `DatePickerInput` \ `MonthPickerInput` \ `YearPickerInput`, and default value add to DatesProvider.
image

3、Fix type error `actionHighlight` missed
image


rtivital:
Thanks!

PR #3544, opened 2023-02-13
fix(core): anchor color not working

Songkeys:
The color style overwrites the `props.color`.


rtivital:
It is not recommended to use `color` prop with Text/Anchor components. Use `c` prop instead – https://codesandbox.io/s/trusting-lovelace-e5jx7c?file=/src/App.tsx

PR #3546, opened 2023-02-14
fix(core): appshell section height calculation

Songkeys:
the fallback value should have an unit (px or rem), otherwise it won't work.


rtivital:
Thanks!

PR #3550, opened 2023-02-14
[@mantine/core] Text truncateStart

nicolassanmar:
Implements `truncateStart` on `<Text />`  based on feature request: https://github.com/mantinedev/mantine/discussions/3549

### DIFF
--- a/src/mantine-core/src/Text/Text.styles.ts
+++ b/src/mantine-core/src/Text/Text.styles.ts
@@ -13,6 +13,7 @@ export interface TextStylesParams {
   size: MantineNumberSize;
   lineClamp: number;
   truncate: boolean;
+  truncateStart: boolean;
   inline: boolean;
   inherit: boolean;
   underline: boolean;
@@ -75,7 +76,16 @@ function getLineClamp(lineClamp: number): CSSObject {
   return null;
 }
 
-function getTruncate(truncate: boolean): CSSObject {
+function getTruncate(truncate: boolean, truncateStart: boolean): CSSObject {
+  if (truncateStart) {
+    return {
+      overflow: 'hidden',
+      textOverflow: 'ellipsis',
+      whiteSpace: 'nowrap',
+      direction: 'rtl',
+      textAlign: 'left',
+    };
+  }
   if (truncate) {
     return {
       overflow: 'hidden',
@@ -96,6 +106,7 @@ export default createStyles(
       size,
       lineClamp,
       truncate,
+      truncateStart,
       inline,
       inherit,
       underline,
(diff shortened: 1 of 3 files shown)


src/mantine-core/src/Text/Text.styles.ts
```diff
@@ -75,7 +76,16 @@ function getLineClamp(lineClamp: number): CSSObject {
...
 }
 
-function getTruncate(truncate: boolean): CSSObject {
+function getTruncate(truncate: boolean, truncateStart: boolean): CSSObject {
+  if (truncateStart) {
+    return {
+      overflow: 'hidden',
+      textOverflow: 'ellipsis',
+      whiteSpace: 'nowrap',
+      direction: 'rtl',
```
rtivital, line 85: This will not work for `rtl` direction. Instead of setting `direction: 'rtl'` use `direction: theme.direction === 'ltr' ? 'rtl' : 'ltr'`. In addition, please check if that works in storybook.

src/mantine-core/src/Text/Text.styles.ts
```diff
@@ -96,6 +106,7 @@ export default createStyles(
       size,
       lineClamp,
       truncate,
+      truncateStart,
```
rtivital, line 109: Instead of adding another prop, change `truncate` prop signature to `truncate?: 'end' | 'start' | true`

rtivital (changes requested):

src/mantine-core/src/Text/Text.styles.ts
nicolassanmar: @rtivital How should I make storybook use `dir: 'rtl'`?
I tried wrapping the component in `<MantineProvider theme={{dir: 'rtl'}}>` but the text is still 'ltr'.

src/mantine-core/src/Text/Text.styles.ts
nicolassanmar: That's better! Thanks for the suggestion!

src/mantine-core/src/Text/Text.styles.ts
```diff
@@ -76,17 +79,17 @@ function getLineClamp(lineClamp: number): CSSObject {
...
       textOverflow: 'ellipsis',
       whiteSpace: 'nowrap',
-      direction: 'rtl',
-      textAlign: 'left',
+      direction: theme.dir === 'ltr' ? 'rtl' : 'ltr',
+      textAlign: theme.dir === 'ltr' ? 'right' : 'left',
     };
   }
-  if (truncate) {
+  if (truncate === 'end' || truncate === true) {
```
rtivital, line 92: Can be simplified to `if (truncate)` since both `'end'` and `true` are truthy values

nicolassanmar: I can change it if you want, I dislike relying on truthy values since i think this is more clear.
If in the future we add another option for `truncate`, using truthy values could lead to problems

rtivital:
Thanks!

PR #3570, opened 2023-02-16
[@mantine/modals] Fix jumping of centered modals on close (#3409)

### LINKED ISSUE #3409: [@mantine/modals] : Centered modal position changes upon closing inside non-centered ModalsProvider
### What package has an issue

@mantine/core

### Describe the bug

When a `modal` that is set to be `centered` is closed while it is inside a non-centered `ModalsProvider`, the top position of the modal moves before it is closed. 

5.10.1

https://codesandbox.io/s/centered-modal-flicker-e45l1q?file=/src/App.tsx

No

Yes

### Possible fix

The issue may be caused by the current modal props being immediately removed upon closing the modal. This causes the default props ( which is not centered ) to take control.

cyantree:
This fixes #3409.

Originally when closing the **last** modal the action `CLOSE_ALL` got called which cleaned the `modals` state but kept the last one around in `current`. Therefore when rendering the `ModalsProvider` the last modal had been rendered but with a closed state.

In my previous PR I changed this behaviour as in my opinion when a modal has been closed it shouldn't stay in `current` - even if it's the last one. Therefore during rendering the `ModalsProvider` a closed modal with default props got rendered instead. However because closing the modal isn't immediate because of an exit transition this default modal rendering was visible.

Therefore this PR partially reverts the last one and makes sure the last modal always will be kept around (not only in `CLOSE_ALL` but also in `CLOSE`. It also documents this behaviour to make sure other devs won't trip over it again.

### DIFF
--- a/src/mantine-modals/src/reducer.ts
+++ b/src/mantine-modals/src/reducer.ts
@@ -2,6 +2,11 @@ import { ModalState } from './context';
 
 interface ModalsState {
   modals: ModalState[];
+
+  /**
+   * Modal that is currently open or was the last open one.
+   * Keeping the last one is necessary for providing a clean exit transition.
+   */
   current: ModalState | null;
 }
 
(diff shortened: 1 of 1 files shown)


cyantree:
@rtivital Do you need any more support with this?
It's come up again in #3613.

rtivital:
Thanks!

PR #3586, opened 2023-02-21
[@mantine/spotlight]: fix autoComplete showing up

Buffele:
This fixes the chrome autocomplete still showing up.


rtivital:
This does not fix the issue, chrome ignores `autocomplete="offf"`. This issue is already fixed in 6.0.

PR #3602, opened 2023-02-23
[docs] Fix import for Switch.Group example

tuckerbeauchamp:


rtivital:
Already fixed in #3603

Issue #3566, opened 2023-02-16
Mysterious `dot` prop for `Indicator`


Derrreks:
### What package has an issue

@mantine/core

### Describe the bug

In the `Indicator` component, the prop `showZero` does not correctly respond to `false` unless `dot={false}` is also provided.

In the docs ([here](https://mantine.dev/core/indicator/?t=props)), there is no description provided of what `dot` is or how to use it. However, even in the example for how to use `showZero` we can see that `dot={false}` had to be provided there.

Can we either remove this parameter if there is no other use for it (I'm happy to contribute this PR) or add a description/example for how to use it and/or why it is needed to be specified with `showZero`?

### What version of @mantine/hooks page do you have in package.json?

5.10.4

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

Remove the `dot` prop from `Indicator`.

rtivital:
I have no idea how it works, you are welcome to submit a PR with a fix to `v6` branch

cyantree:
`dot` seems to be required because if `label={null}` nothing would be displayed:
```js
dot || (label != null &&
```
https://github.com/mantinedev/mantine/blob/master/src/mantine-core/src/Indicator/Indicator.tsx#L118

Also `dot` is missing from the hook dependencies.

But maybe the whole special number handling is too much for the component and showing/hiding the dot depending on the label should be the users responsibility:
```jsx
<Indicator label={count} dot={count > 0} inline size={16}>
```
In this case `showZero` would be obsolete and the condition would get a lot simpler:
```ts
  const isShowIndicator = useMemo(
    () => !disabled && dot,
    [disabled, dot]
  );
```

This would be a breaking change.

maradney:
> the whole special number handling is too much for the component and showing/hiding the dot depending on the label should be the users responsibility

@cyantree I agree with this.
I would also change `dot` to be `showDot` or something

cyantree:
I've missed `overflowCount` which is also a number based behaviour.

So in case we would remove this special handling `overflowCount` would be removed too.

However it would be possible to reintroduce this as a utility method that will be used via object spread:
```tsx
<Indicator {...someIndicatorNumberUtility(12, {overflowCount: 100, hideZeroOrBelow: true})} />
```

> I would also change `dot` to be `showDot` or something

Good point. `showDot` would be consistent to `showZero`.
[...]

PR #3609, opened 2023-02-24
#3566 simplify indicator

cyantree:
MR for #3566. It simplifies the `Indicator` component by removing the `dot` property and number related special handling with `showZero` and `overflowCount`.


rtivital:
Thanks!

PR #3622, opened 2023-02-27
[docs] Update available color formats

Keilo75:
Noticed that I hadn't updated the documentation text when I added support for 'hexa' format in ColorPicker and ColorInput - this PR fixes that.


rtivital:
Thanks!

PR #3625, opened 2023-02-27
[@mantine/modals] Fix endless `closeAll()` loop

### LINKED ISSUE #3623: [v6] - Multiple modals briefly show empty modal when closed
### What package has an issue

@mantine/modals

### Describe the bug

When multiple modals are opened at once, and you close them using `modals.closeAll`, an empty modal is briefly shown at the top of the screen:

image

This only seems to happen when **multiple** modals are opened. It closes correctly when only one modal is opened.

Not sure if this is v6-specific or maybe even `6.0.0-alpha.5`-specific. In the codesandbox I get `RangeError: Maximum call stack size exceeded` when trying to reproduce this on **either** latest v6 (`6.0.0-alpha.7`) or latest v5 (`5.10.5`)

6.0.0-alpha.5

https://codesandbox.io/s/hungry-feather-mymohv?file=/src/App.tsx

No

None

### Possible fix

_No response_

cyantree:
Fixes #3623.

Reason was that the `onClose()` method got called before the new state has been applied therefore it resulted in an endless loop.
Swapping the method execution order unfortunately isn't enough as it's not clear when React does apply the state update.

So I moved some logic to the reducer because it will be run for each action in sequence.

In earlier versions this usecase didn't crash the whole application because the methods might not use the current modal state. Therefore an endless loop couldn't occur. However it also lead to other issues:
See this sandbox for `v5.10.0`:
https://codesandbox.io/s/modals-state-issue-3623-ppqfjr?file=/src/App.tsx

Open the console, open some modals and then close them.
Despite all modals being closed only one callback will be called.

So despite once again breaking things (sorry for that!) I'm optimistic the whole component is on a good way and looses more of its hidden issues.

I've also tested this against `v6` and can confirm that it works there too.

Also thanks for giving me the opportunity to get it right!

### DIFF
--- a/docs/src/docs/core/ColorInput.mdx
+++ b/docs/src/docs/core/ColorInput.mdx
@@ -33,8 +33,8 @@ function Demo() {
 
 ## Formats
 
-Component supports hex, rgb, rgba, hsl and hsla color formats.
-Slider to change opacity is displayed only for rgba and hsla formats:
+Component supports hex, hexa, rgb, rgba, hsl and hsla color formats.
+Slider to change opacity is displayed only for hexa, rgba and hsla formats:
 
 <Demo data={ColorInputDemos.formatsConfigurator} />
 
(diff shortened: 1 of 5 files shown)


rtivital:
Thanks!

PR #3648, opened 2023-03-02
[@mantine/dates]: CalendarHeader: Remove disabled level change button from tab order

filiptammergard:
The button for changing level is `disabled` but it's still kept in tab order when `maxLevel` should remove it. You can try it out here: https://mantine.dev/dates/date-picker/#max-level

The DatePicker at the right has `maxLevel="month"`, so the level change button is disabled, but it still has `tabIndex={0}` and is therefore kept in tab order.

image

This worked correctly in v5.x: https://v5.mantine.dev/dates/calendar/#level-change

Let me know if I should create an issue for this instead of opening a PR!

Repro here: https://codesandbox.io/s/silly-forest-fcvs01?file=/src/App.tsx


rtivital:
Thanks!

Issue #3626, opened 2023-02-27
Transition exit duration not working properly


lgaspari:
### What package has an issue

@mantine/core

### Describe the bug

Hello everybody!

I was playing around with a `Transition` in my current project and I noticed that the `exitDuration` wasn't changing as I expected when increasing or decreasing its value. At first I thought it was caused due to using a custom transition, but then I noticed it didn't work either using the built-in supported transitions (e.g. `scale-x`).

Therefore, I created two working codesandboxes to reproduce:

1. Using custom transition: https://codesandbox.io/s/magical-galileo-y4cjjo
2. Using built-in transition: https://codesandbox.io/s/sleepy-banach-cnr8ko

Note that in (2) the `EXIT_DURATION` has been set to `1 second`, but instead of the animation taking one second, the [unmount timeout](https://github.com/mantinedev/mantine/blob/master/src/mantine-core/src/Transition/use-transition.ts#L63) is taking one second but the animation lasts `5 seconds` instead (can be seen in the DOM).

I wonder if the exit duration should instead change the transition duration or if it's supposed to work this way.

Thanks in advance.

### What version of @mantine/hooks page do you have in package.json?

5.10.5

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/sleepy-banach-cnr8ko

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

I've created a fork with the fix for the issue: https://github.com/lgaspari/mantine/commit/3b0d70b8bd58041d0993f84436a8b80181c187a4. Are there any tests for this hook? I couldn't find any.

lgaspari:
Hey @rtivital, I've just noticed you removed the `review required` label. Any thoughts on this issue?

rtivital:
You are welcome to submit a PR with a fix

rtivital:
Fixed in 6.0.1

PR #3664, opened 2023-03-03
[@mantine/core] Transition: Update transition duration when unmounting

lgaspari:
Solves #3626

Stores the `transitionDuration` in `useTransition`'s state so that upon changing its value, the element's `transition-duration` style property gets updated as well.

Below you'll find a short video of the `duration` (200ms) and `exitDuration` (2000ms) after applying this fix.

https://user-images.githubusercontent.com/20128985/222743880-095fb575-18f5-4783-b639-b28fe0bf2f18.mov


rtivital:
Thanks!

Issue #3227, opened 2022-12-22
Single item in Button.Group has always sharp edges


kazlik:
### What package has an issue

@mantine/core

### Describe the bug

Single button in button group has always sharp edges. More items/single button without group are okay.

### What version of @mantine/hooks page do you have in package.json?

5.9.4

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/mantine-bad-border-radius-x0jyhn

### Do you know how to fix the issue

None

### Are you willing to participate in fixing this issue and create a pull request with the fix

No

### Possible fix

Ignore following styles if there is only 1 child element
![image](https://user-images.githubusercontent.com/4661382/209110760-10e3e0a5-2c53-446c-8e93-aec5255d07e1.png)

rtivital:
Fixed in 6.0.1

PR #3667, opened 2023-03-03
[@mantine/core] Button: fixed styles for Button.Group containing one element

dipiash:
Hi!

Fix https://github.com/mantinedev/mantine/issues/3227


dipiash:
Hi, @rtivital!

Maybe it should be included in the fix for version 5 (not only 6)?

Issue #3572, opened 2023-02-17
Floating arrow borders cause arrows to display artifacts


kfalicov:
### What package has an issue

@mantine/core

### Describe the bug

When arrows are enabled for tooltips that have a border, the arrows have artifacts extending into the bubble. This is caused by an in-line style that sets the `border-top` and `border-left` to `none`:

<img width="64" alt="image" src="https://user-images.githubusercontent.com/22381357/219714003-407d75c0-748f-42a3-988e-835a16835c3d.png">

### What version of @mantine/hooks page do you have in package.json?

5.10.3

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/gallant-saha-09ibyl?file=/src/App.tsx

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

Setting them instead to `border-top-color: transparent` and `border-left-color: transparent` improves the appearance, and allows the FloatingArrow to `inherit` the border prop from its parent:
<img width="61" alt="image" src="https://user-images.githubusercontent.com/22381357/219714338-a3dc0310-a3d3-4d74-a462-16c133c16211.png">

This is the default appearance of the arrow if the tooltip is given a border
<img width="82" alt="image" src="https://user-images.githubusercontent.com/22381357/219714654-4bf6ab36-af50-4e14-9582-3f7260b88c6a.png">

kfalicov:
https://github.com/mantinedev/mantine/blob/62926e6e71643f2d2f5cfa4e43554e7918cbf85e/src/mantine-core/src/Floating/FloatingArrow/get-arrow-position-styles.ts#L100

These are the lines that would change from `borderTop` to `borderTopColor`

rtivital:
Thanks for research! You are welcome to submit a PR with a fix.

rtivital:
Fixed in 6.0.1

PR #3693, opened 2023-03-05
[@mantine/core] Tooltip: fixed styles for Tooltip if border set for tooltip and arrow in styles

dipiash:
Fix for https://github.com/mantinedev/mantine/issues/3572


rtivital:
Thanks!

PR #3696, opened 2023-03-06
[@mantine/core] add portalProps to all components rendering a portal

jvdsande:
This PR adds spreadable `div` props to the Portal component, that get spread to the root div mounted in the portal.

It further adds a `portalProps` property to all components accepting `withinPortal`, so that spreadable props can be passed through to the rendered portal.

This allows for instance passing a custom className to the portal rendered inside a component, or add custom ID or data-attributes to the portal root for manipulation/styling hooks.

### DIFF
--- a/src/mantine-core/src/Affix/Affix.tsx
+++ b/src/mantine-core/src/Affix/Affix.tsx
@@ -2,7 +2,7 @@ import React, { forwardRef } from 'react';
 import { DefaultProps, getDefaultZIndex, useComponentDefaultProps } from '@mantine/styles';
 import { packSx } from '@mantine/utils';
 import { Box } from '../Box';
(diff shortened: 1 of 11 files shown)


rtivital:
Thanks!

jvdsande:
> I think that `portalProps` needs to be `Omit<PortalProps, "children">`; right now, using it requires something like:
> 
> ```tsx
> <Menu withinPortal portalProps={{target: "#menu-portal", children: undefined}}>
> </Menu>
> ```
> 
> because `PortalProps` requires `children`, but you don't actually want to spread that prop on the `OptionalPortal` (since explicit children are passed, instead)
> 
> Otherwise, Typescript gripes at you:
> 
> image
> 

Good catch yeah, I'll create the fix for that

PR #3710, opened 2023-03-07
[@mantine/core] DateTimePicker: Fix TimeInput not rendering sometimes

jameszhang-a:
Bug: 
DateTimePicker: time input disappears when dropdown is closed with month level
https://imgur.com/a/0UzuS0f

Fixed:
CleanShot 2023-03-07 at 15 51 46


rtivital:
Thanks!

PR #3715, opened 2023-03-08
[@mantine/core] PinInput: onComplete fires correclty - issue 3714

LeighS95:
PinInput now uses useEffect when _value changes so that onComplete if present is called on last input instead of needing an additional key stroke. Corrosponding test and storybook component have been added too.


rtivital:
Thanks!

PR #3730, opened 2023-03-09
[@mantine/core] Drawer: Fix typo from Modal to Drawer

jourmooney:
Drawer default styles props not working correctly.
I found that it was not retrieving the default styles props of Drawer, but instead retrieving the default styles of Modal.
After fixing the typo, it worked correctly.


rtivital:
Thanks!

PR #3744, opened 2023-03-10
[@mantine/core] Select: fix tab focus loss when dropdown is scrollable

### LINKED ISSUE #3734: Select: tab key causes focus loss
### What package has an issue

@mantine/core

### Describe the bug

Clicking `tab` key while focused on a searchable `Select` component with a _scrollable_ dropdown content causes complete focus loss instead of jumping to the next focusable element.

Using [native scrollbars](https://v5.mantine.dev/core/select/#native-scrollbars) doesn't help and neither does rendering [custom items](https://v5.mantine.dev/core/select/#custom-item-component).

The only solution that worked for me was to add `withinPortal` but this is not an option in my project for unrelated reasons.

The problem also goes away when searching narrows the options list and makes the container non-scrollable.

p.s. It also seems that this [large data set example](https://v5.mantine.dev/core/select/#large-data-set) uses `withinPortal` under the hood although this is not reflected in the code example.

5.10.2 (but I also tried different 5.x.x and 6.x.x versions)

https://codesandbox.io/s/elastic-monad-v5xyxz?file=/src/App.tsx

No

Yes

### Possible fix

_No response_

fbarl:
Fixes #3734.

#### Changes

- Add `tabIndex={-1}` to `SelectPopover` which fixes it for `dropdownComponent="div"`
- Add another `tabIndex: -1` to `SelectScrollArea` to address the issue for default `dropdownComponent` as well

I tested this manually by playing around with `Select.demo.large.tsx` and verified that the behaviour is as expected also without `withinPortal`.

I'm not sure if unit / component tests are needed for this change and also whether the change itself will get approved so this is all I'm submitting for now.

### DIFF
--- a/src/mantine-core/src/Select/SelectPopover/SelectPopover.tsx
+++ b/src/mantine-core/src/Select/SelectPopover/SelectPopover.tsx
@@ -45,6 +45,7 @@ function SelectPopoverDropdown({
           onMouseDown={(event) => event.preventDefault()}
           style={{ flex: 1, overflowY: component !== SelectScrollArea ? 'auto' : undefined }}
           data-combobox-popover
+          tabIndex={-1}
           ref={innerRef}
         >
           <div className={classes.itemsWrapper} style={{ flexDirection: direction }}>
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #3748, opened 2023-03-10
[@mantine/core] Anchor: prop `underline={false}` should remove the underline on hover as well

### LINKED ISSUE #3654: <Anchor /> `underline={false}` doesn't work when hovering the element
### What package has an issue

@mantine/core

### Describe the bug

Hello,

I am migrating to v6 and noticed that the underline prop in the `<Anchor />` component does not fully remove the underline

The `<a>` element gets the `text-decoration: none`, but the `:hover` pre-selector keeps the `text-decoration: underline`

v6 behavior:
https://codesandbox.io/s/mystifying-stallman-6vcuqu?file=/src/App.tsx

v5 behavior (this one is working as expected):
https://codesandbox.io/s/amazing-dream-301vme?file=/src/App.tsx

6.0.0

_No response_

No

Yes

### Possible fix

_No response_

Tronikelis:
should close #3654 by my understanding

### DIFF
--- a/src/mantine-core/src/Anchor/Anchor.story.tsx
+++ b/src/mantine-core/src/Anchor/Anchor.story.tsx
@@ -34,3 +34,15 @@ export function InheritFontSize() {
     </div>
   );
 }
+
+export function WithUnderlineProp() {
+  return (
+    <div style={{ padding: 40 }}>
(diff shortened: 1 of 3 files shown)


rtivital:
Thanks!

PR #3760, opened 2023-03-12
[@mantine/core] ScrollArea.Autosize: A missing viewportProps property

Domin-MND:
`viewportProps` was missing for ScrollArea.Autosize so I added it. Never tested it but obviously will.


Domin-MND:
Renamed the commit & tested it ^

rtivital:
Thanks!

PR #3769, opened 2023-03-13
Issue 3763 fix json input

### LINKED ISSUE #3763: JsonInput seems to be completely busted and escapes everything nonstop
Entering text into the box works fine, as you can see here:
image

Until you leave the text box. Then it gets escape characters added everywhere:
image

This happens every single time you enter the textbox and leave the textbox which makes it double the amount of escapes every time. This happens with any input including valid JSON:
image

This is happening on my local version as well as the one in the docs right now:
https://mantine.dev/core/json-input/

cyantree:
Fixes #3763 and some other issues with the component. Also adds some tests.


rtivital:
Thanks!

PR #3773, opened 2023-03-14
Fix use-hash: wrong value being set on hashchange

stijnvanderlaan:
setHashValue is being set to `hash` again instead of the `newHash` value. Sometimes it worked, but this fixes an issue where when hashchange is triggered the hash is reset to the old hash.


rtivital:
Thanks!

PR #3775, opened 2023-03-14
Improve Popover offset to include all Axis

paulm17:
This PR adds the full axis ability to popover.

See the docs here: https://floating-ui.com/docs/offset

Instead of the default Y axis, now the dev can specify both X and Y axis.  This is beneficial in case there needs to be more precise adjustments for UI.

An additional story has been added to reflect this change.

This should not be a breaking change because it retains the default number param.

If this PR is accepted, then I will also update the docs to mirror the change.


rtivital:
@paulm17 Can you also add a demo to Popover docs that shows how to use this new feature?

paulm17:
@rtivital Of course!  I'll have this done sometime today!

rtivital:
Thanks!

PR #3778, opened 2023-03-14
Update Pagination.mdx to use correct prop name.

kaeevans:
The `Pagination` component does not have a value prop as the docs show. I updated the `Pagination` component example in the docs to use the correct `page` prop.


rtivital:
Thanks!

Issue #3789, opened 2023-03-15
[use-timeout]: start and clear functions change on every render


lgaspari:
### What package has an issue

@mantine/hooks

### Describe the bug

It looks like the functions `clear` and `start` returned by the `useTimeout()` hook change on every render causing side-effects to run more than needed if set as dependency.

In the example below (see [codesanbox](https://codesandbox.io/s/hungry-herschel-sfwz4b)), we can see how the `"re-render"` log is shown whenever we change the input's value. In addition, the timeout callback gets called after 1 second delay.

```ts
import { TextInput } from "@mantine/core";
import { useInputState, useTimeout } from "@mantine/hooks";
import { useEffect } from "react";

export default function App() {
  const [value, setValue] = useInputState("");

  const { start } = useTimeout(() => console.log("callback"), 1000);

  useEffect(() => {
    console.log("re-render");
    start();
  }, [start]);

  return <TextInput onChange={setValue} value={value} />;
}
```

### What version of @mantine/hooks page do you have in package.json?

6.0.0

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/crazy-elion-i3cky9

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

A fix would be wrapping `clear` and `start` functions with `useCallback()` and store the callback within a `ref` to prevent further re-renderings (see [codesandbox](https://codesandbox.io/s/crazy-elion-i3cky9?file=/src/use-timeout-with-callback.ts))

Thanks to above fix, we would be able to add missing dependencies for auto-invoke functionality as the following:

Before:

```ts
useEffect(() => {
  if (options.autoInvoke) {
    start();
  }

  return clear;
}, [delay]);
```

After:

```ts
useEffect(() => {
  if (options.autoInvoke) {
    start();
  }

  return clear;
}, [clear, delay, start, options.autoInvoke]);
```

rtivital:
You are welcome to submit a PR with a fix

lgaspari:
This is now solved — you can check the difference between [CodeSandbox 1 (issue)](https://codesandbox.io/s/hungry-herschel-sfwz4b) and [CodeSandBox 2 (solved)](https://codesandbox.io/s/spring-thunder-gs3phf) or just change the `@mantine/hooks` version from `6.0.0` to `6.0.4` in CodeSandbox 1.

PR #3801, opened 2023-03-16
[@mantine/hooks] use-timeout: Memoize `clear` and `start` functions

lgaspari:
Solves #3789

1. Wrap both `clear()` and `start()` functions with the `useCallback()` hook so that they won't change across re-renders (unless they have to, e.g. `start()` depends on `delay`).
2. Add missing test case for `autoInvoke` effect when `delay` changes.
3. Correct `autoInvoke` effect dependencies to include `clear()` and `start()` as well as `options.autoInvoke`.

As far as I could check, the above changes shouldn't affect anyone using this hook. However, there's a separate commit (see point 3) that changes `autoInvoke` effect dependencies and I'm not 100% sure of relying on `options.autoInvoke`. People shouldn't really change that setting after the first render, so I'd think it's safe, but I'm open to suggestions.

### DIFF
--- a/src/mantine-hooks/src/use-timeout/use-timeout.test.ts
+++ b/src/mantine-hooks/src/use-timeout/use-timeout.test.ts
@@ -70,6 +70,31 @@ describe('@mantine/hooks/use-timeout', () => {
     expect(setTimeout).toHaveBeenCalled();
   });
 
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

Issue #3731, opened 2023-03-09
React warning with `theme.components.Modal.defaultProps.overlayProps`


testerez:
### What package has an issue

@mantine/core

### Describe the bug

When setting `Moal.defaultProps.overlayProps` in theme, I get this React warning:
> Warning: React does not recognize the `overlayProps` prop on a DOM element.

Here is my theme:

```ts
export const theme: MantineThemeOverride = {
  components: {
    Modal: {
      defaultProps: {
        size: 640,
        overlayProps: {
          opacity: 0.7,
          blur: 2,
        },
      },
    },
  },
}
```

### What version of @mantine/hooks page do you have in package.json?

@mantine/hooks@6.0.1

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/funny-payne-udkjxy?file=/src/App.tsx

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

cyantree:
@rtivital Is the `{...others}` spread really needed? If yes, for which kind of uses?
https://github.com/mantinedev/mantine/blob/master/src/mantine-core/src/ModalBase/ModalBase.tsx#L206
If we can't get rid of it completely is it possible to specify an allow list of supported props?
Otherwise it's quite hard to catch all invalid props and it's only a matter of time until the next issue like this comes up. Or we need to use a library like `filter-invalid-dom-props` that supposedly knows which props are valid and which not.

rtivital:
It is needed to pass additional attributes to the root element, for example data- attributes or event listeners.

rtivital:
Fixed in 6.0.3, thanks @cyantree!

Issue #3702, opened 2023-03-07
Warning: Received 'true' for a non-boolean attribute 'centered' on Modal when using defaultProps


Vahor:
### What package has an issue

@mantine/core

### Describe the bug

Since v6, Spotlight use a modal under the hood, this bug also occurs when using this package.

When I use `<Modal centered>` there are no issue. 
However, when I want to have these props shared for all modals using the MantineProvider:
```tsx
<MantineProvider
      theme={{
        components: {
          Modal: {
            defaultProps: {
              centered: true,
              withCloseButton: false
            }
          }
        }
      }}
    >
```
I get a warning saying `Warning: Received 'true' for a non-boolean attribute 'centered'.` (same issue with other boolean props)

This bug is not urgent as this is only a warning, in my case the modal is indeed centered.

### What version of @mantine/hooks page do you have in package.json?

6.0

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/snowy-fire-flgj0b?file=/src/App.tsx

### Do you know how to fix the issue

None

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

cyantree:
The issue seems to be the following:

`<ModalBase __staticSelector="Modal"` tells `ModalBase` to use the theme defaults for `Modal`.
https://github.com/mantinedev/mantine/blob/master/src/mantine-core/src/Modal/ModalRoot/ModalRoot.tsx#L55

As in many other components some props will be extracted and the rest merged in `others`.
https://github.com/mantinedev/mantine/blob/master/src/mantine-core/src/ModalBase/ModalBase.tsx#L146

Because `others` gets spread on the DOM node these unsupported props will also be passed to the DOM resulting in the described error.
https://github.com/mantinedev/mantine/blob/master/src/mantine-core/src/ModalBase/ModalBase.tsx#L206

Adding `centered` in the prop extraction won't work nicely because this component doesn't know it. So it needs to be defined in the types to get known.

This issue might apply to other props and also to other components.
I think these spreads are quite tricky but I don't have any idea for a better approach yet. Adding a filter method to only pass supported DOM props might be a way but this would also only mask the underlying issue that is that unsupported props might leak into places they shouldn't be.

But for the time being stripping the extra property in `ModalBase` might be the easiest thing to do.

strblr:
Same problem here.

rtivital:
Fixed in 6.0.3, thanks @cyantree!

PR #3802, opened 2023-03-16
[@mantine/core] Modal: Fix prop to DOM leak #3731 #3702

cyantree:
This fixes the leaks mentioned in #3702 #3731 by acknowleding that `ModalBase` can receive `Modal` props.

### DIFF
--- a/src/mantine-core/src/ModalBase/ModalBase.tsx
+++ b/src/mantine-core/src/ModalBase/ModalBase.tsx
@@ -143,7 +143,7 @@ export function ModalBase(props: ModalBaseProps) {
     styles,
     className,
     ...others
-  } = useComponentDefaultProps(props.__staticSelector, ModalBaseDefaultProps, props);
+  } = useComponentDefaultProps('ModalBase', ModalBaseDefaultProps, props);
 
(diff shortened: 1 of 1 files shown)


src/mantine-core/src/ModalBase/ModalBase.tsx
rtivital, line 1: I think these props should be filtered out in Modal.Root or Modal components. ModalBase is used as a base for both Modal and Drawer, and it should not know anything about props that are passed to these components.

rtivital: In this case the better solution will be removing `__staticSelector` from `ModalBase` `useComponentDefaultProps` and use it only for styles api. Replace it here https://github.com/mantinedev/mantine/blob/master/src/mantine-core/src/ModalBase/ModalBase.tsx#L146

to
```tsx
useComponentDefaultProps('ModalBase', ModalBaseDefaultProps, props);
```

cyantree: Great, thanks @rtivital. I reworked the PR based on your hint.

rtivital:
Thanks!

Issue #3803, opened 2023-03-16
Klona can fail to copy objects on form.setPath


benlongo:
### What package has an issue

@mantine/form

### Describe the bug

Effectively the same issue described here: https://github.com/lukeed/klona/issues/33

In my particular case, I'm using Dayjs objects in the form values and `klona` attemps to call the constructor https://github.com/iamkun/dayjs/blob/dev/src/index.js#L84 with an undefined `cfg` which fails.

![image](https://user-images.githubusercontent.com/11063448/225679920-22c9c423-6778-4de4-9192-938d03e70eea.png)

### What version of @mantine/hooks page do you have in package.json?

6.0.2

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

If I understand correctly, the purpose of klona here is to achieve copy-on-write behavior for the form state. 
Would this instead be possible to implement this with immer so that we don't have to deep copy everything? 

Switching to `klona/full` may resolve these issues at a performance and bundle size cost.

PR #3805, opened 2023-03-16
Use klona/full to clone form values.

benlongo:
potential solution for the issue described in #3803


rtivital:
Hi @benlongo, have you checked that `klona/full` resolves your issue with dayjs objects cloning? Can you include a test for it?

benlongo:
Yes I've verified it fixed my issues with it, but for some reason I wasn't getting it to reliable reproduce (was like 50% of the time). I will try to get a test up and running for it this weekend.

rtivital:
Thanks!

Issue #3758, opened 2023-03-12
v6: reversed pagination arrows in RTL


alimehasin:
### What package has an issue

@mantine/core

### Describe the bug

`Pagination` component uses wrong arrows (reversed) in RTL mode

### What version of @mantine/hooks page do you have in package.json?

v6.0.1

### Do you know how to fix the issue

None

### Are you willing to participate in fixing this issue and create a pull request with the fix

None

PR #3809, opened 2023-03-16
[@mantine/core] Pagination: Fixed reversed pagination arrows in RTL

dipiash:
Fixed https://github.com/mantinedev/mantine/issues/3758


rtivital:
Thanks!

Issue #3432, opened 2023-01-27
Not able to set default time zone for DatePicker and TimeInput


ArnaudValensi:
### What package has an issue

@mantine/dates

### Describe the bug

All the dates displayed by the DatePicker and TimeInput are in the local timezone. There is no way to display it using a custom timezone, like UTC for example.

### What version of @mantine/hooks page do you have in package.json?

5.10.2

### Some comments

Date object always store absolute dates inside, this is when we use the function toString or the functions to get any of the component that it does the shift based on the local timezone. TimeInput just uses those functions to display the dates, so it is not possible to display a date as UTC for example without creating fake dates.

Right now the only solution I have is to do this:

```js
<TimeInput
    value={subtractUtcOffset(date)}
    onChange={(value: Date) => onTimeChange(addUtcOffset(value))}
/>

function subtractUtcOffset(date: Date) {
    if (!date) {
        return date;
    }

    let result = dayjs(date);

    const utcOffsetInMinutes = dayjs().utcOffset();
    result = result.subtract(utcOffsetInMinutes, 'minutes');

    return result.toDate();
}

function addUtcOffset(date: Date) {
    if (!date) {
        return date;
    }

    let result = dayjs(date);

    const utcOffsetInMinutes = dayjs().utcOffset();
    result = result.add(utcOffsetInMinutes, 'minutes');

    return result.toDate();
}
```

Note that the dates created with the shift functions are date that are not representing the real time we manipulate. This is why having a way to select the timezone in the DatePicker and TimeInput components would be better.

Also, this code is bugged, it uses the current UTC offset, but the offset changes depending on the daylight saving in some countries. So if we are in winter and I want to select an UTC date in summer, this solution will give me an incorrect date.

benlongo:
This issue is still present with the @mantine/dates 6.0 overhaul.

If mantine is to use `Date` objects (which represent an absolute point in time), as opposed to some timezone-oblivious object, then I think it must be timezone aware to result in correct behavior when used with timezones that are different than the browsers timezone. As pointed out here, the UTC offset workaround is insufficient. 

It could be decided that dealing with timezones that are different than the browsers timezone is out of scope, but this seems like a pretty significant limitation for apps that deal with objects across timezones.

I propose adding an optional `tz` (or `timezone`) prop to anything that deals with `Date` objects and also extending `DatesProvider` to hold a default timezone.

rtivital:
[...]

PR #3823, opened 2023-03-18
Add timezone support to dates context

benlongo:
Addresses #3432 

## Current design & approach

Currently the `DatesProviderContext` has functions for getting the various settings given an optional local 'override'. Whenever a `Dayjs` object is constructed, the caller is responsible for passing these settings off to `dayjs`. It's pretty easy to forget to set the `locale`, and the addition of timezone information through the same mechanism would open the door for much less obvious bugs. 

This PR modifies `useDatesContext` so that it takes the local overrides (from props) as an argument and merges them with the values from the context. In the current code, the tasks of going from `string` to `Date` with a `format` and vice-versa were common, so I factored those tasks out into the context. Other components do varying tasks that require briefly transitioning a `Date` to a `Dayjs` object to compute something. In order to always apply the `locale` and `timezone` information, the context exposes a `dayjs: (date: Date) => Dayjs` function which always applies the `locale` and `timezone`. With this approach, components that want to use `dayjs` should always use `ctx.dayjs` instead.

There were several instances in the code where the `locale` was not applied before formatting so I believe this fixed those issues as well.

### Design choices I'm unsure of 
- naming of `ctx.dayjs`. 
  - I could see this being confusing if somebody chose to do `const { dayjs } = useDatesContext(...)`.
  - Alternative names could be like `dateToDayjs`, `asDayjs`, `toDayjs`, etc. to avoid confusion
- `dayjs.extend()` location for `utc` and `timezone` plugins
  - Is it fine to set this in the dates context file? It seems like it would be, but I'm not sure if adding extensions should be somewhere higher in scope to avoid race conditions or something.
- awkwardness in utility functions
  - several previously pure utility functions now need timezone information, but it seems suboptimal to have those not constructed with `ctx.dayjs` due to the aforementioned risks of forgetting to set locale/timezone when it's important. 

## Questions

I'm confused about this snippet of code:
```
// DateInput.tsx
const defaultDateParser = (val: string) => {
  const parsedDate = dayjs(val, valueFormat, ctx.getLocale(locale)).toDate();
  return Number.isNaN(parsedDate.getTime()) ? dateStringParser(val) : parsedDate;
};

// date-string-parser.ts
export function dateStringParser(dateString: string) {
  const date = new Date(dateString);

  if (Number.isNaN(date.getTime()) || !dateString) {
    return null;
  }

  return date;
}
```

Why is `new Date(dateString)` done if the `dayjs` parsing fails? It's unclear to me how to make this behave correctly in the presence of timezones. Can it be removed?

## Remaining work

- a few of the utility functions that live outside of components have not been made timezone aware yet
- tests
- docs

## Extensions

It would be fairly straightforward to introduce a `timezone` prop, just like the `locale` prop for local overrides of the timezone on a per-component basis. It would be really nice to have this (for me at least) since I require a lot of different timezones for date time entry which doesn't always cleanly align with a context.

### DIFF
--- a/src/mantine-dates/src/components/Calendar/Calendar.tsx
+++ b/src/mantine-dates/src/components/Calendar/Calendar.tsx
@@ -1,5 +1,4 @@
 /* eslint-disable react/no-unused-prop-types */
-import dayjs from 'dayjs';
 import React, { forwardRef } from 'react';
 import { Box, DefaultProps, Selectors, useComponentDefaultProps } from '@mantine/core';
 import { useUncontrolled } from '@mantine/hooks';
@@ -12,6 +11,7 @@ import useStyles from './Calendar.styles';
 import { MonthLevelSettings } from '../MonthLevel';
 import { YearLevelSettings } from '../YearLevel';
 import { DecadeLevelSettings } from '../DecadeLevel';
+import { useDatesContext } from '../DatesProvider';
 
 export type CalendarStylesNames =
   | Selectors<typeof useStyles>
@@ -201,6 +201,8 @@ export const Calendar = forwardRef<HTMLDivElement, CalendarProps>((props, ref) =
     ...others
   } = useComponentDefaultProps('Calendar', defaultProps, props);
 
+  const ctx = useDatesContext({ locale });
+
   const { classes, cx } = useStyles(null, {
     name: ['Calendar', __staticSelector],
     classNames,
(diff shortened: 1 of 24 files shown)


jonavila:
@benlongo Any plans to pick this up again or are you waiting for feedback before proceeding?

benlongo:
I'm still interested in getting this in, but likely won't be able to get everything done for a couple weeks on it. 

I'm also interested in feedback on the general approach. Happy to hand it off as well if you are interested in picking it up.

guptashiv123 (approved):

benlongo:
This is still on the todo list but I haven't had a chance to finish the implementation yet.

rtivital:
The feature has been implemented in 7.0.0 – https://mantine.dev/dates/dates-provider/#timezone

PR #3824, opened 2023-03-18
[@mantine/styles] Expose theme breakpoints as css variables

anri-asaturov:
Adds `--mantine-breakpoints-xs` (and the rest of the sizes) to the list of exported CSS variables.


rtivital:
Thanks!

PR #3825, opened 2023-03-19
[docs] Auto open spotlight

josiahayres:
For any page in docs site: 

- `?search` will open spotlight search.
- `?search=button` will open spotlight search, prefilled with `button`.

E.g. on local site:
image

Should work on all pages in docs site.


rtivital:
Thanks!

PR #3828, opened 2023-03-19
[@mantine/form] Update form errors on list interactions

### LINKED ISSUE #3827: [@mantine/form] Form reorderListItem doesn't update validation errors
### What package has an issue

@mantine/form

### Describe the bug

While reordering list items I would expect the errors to update accordingly. Means that the errors are also reorder. Currently, an error remains at the index where the validation failed.
In the codesandbox click in the second input to trigger the on blur validation and swap the form list items with the button afterwards

6.0.2

https://codesandbox.io/s/form-list-reorder-validation-update-yuyw1x

None

Yes

### Possible fix

Reorder the error state the same way the values change

Alf-Melmac:
Closes #3827 

This PR changes the error state in the following ways
| Function | Before | After |
|--------|--------|--------|
| reorderListItem | No changes | Changes the indices from `from` to `to` |
| removeListItem | Removed all at path | Removes the error for the given index and decrements any error with a higher index than specified to account for the removed element |
| insertListItem | No changes | Increments any error with a higher index than specified to account for the inserted element |


rtivital:
Thanks!

PR #3829, opened 2023-03-19
[@mantine/core] Modal & Drawer Header: fix content scrolling over the header 

Yandawl:
This ensures that content does not visibly scroll over the header of the Modal and Drawer components.


rtivital:
Thanks!

PR #3840, opened 2023-03-20
[docs] Update actual code to match displayed code

old-rob:
Currently the sample code says that the fourth avatar component has radius="sm", but it looks the same as the other three avatars. This should fix that


rtivital:
Thanks!

PR #3849, opened 2023-03-21
[@mantine/dates]: Consider hideOutsideDates when determining tab order

filiptammergard:
I noticed no date is added to tab order when:

- No date in month is selected
- Current date is not in month
- First non-disabled date in month is hidden because of `hideOutsideDates={true}`

Can be tested here in February 2023 for example: https://mantine.dev/dates/date-picker/#hide-outside-dates. This fixes it.

As a side-note, I was expecting be to be able to use `!getDateControlProps?.(date)?.outside` instead of `isSameMonth(date, month)`, but turns out `outside` is never set here: https://github.com/mantinedev/mantine/blob/master/src/mantine-dates/src/hooks/use-dates-state/use-dates-state.ts#L158 Maybe that could be something to look into in the future?


rtivital:
Thanks!

Issue #3877, opened 2023-03-23
MultiSelect disableSelectedItemFiltering doesn't work with searchable prop


kaureliaExlabs:
### What package has an issue

@mantine/core

### Describe the bug

Hi, today I find out that disableSelectedItemFiltering prop doesn't work with searchable prop when you are using groups. Do you have any idea how to fix that?

```tsx
  const testGroupsItems = ["2test", "lorem"];
  const testGroupsItems2 = ["test", "test3", "ipsum"];

<MultiSelect
        label={"Select"}
        placeholder={"Test select"}
        disableSelectedItemFiltering
        searchable
        data={[
          ...testGroupsItems.map((category) => {
            return {
              value: category,
              label: category,
              group: "Group 1"
            };
          }),
          ...testGroupsItems2.map((category) => {
            return {
              value: category,
              label: category,
              group: "Group 2"
            };
          })
        ]}
      />
```

### What version of @mantine/hooks page do you have in package.json?

6.0.4

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

No

rtivital:
Codesandbox with reproduction – https://codesandbox.io/s/naughty-vaughan-z5pqc9?file=/src/App.tsx

lgaspari:
Hi! I've looked into the issue on my end and it seems that when the `disableSelectedItemFiltering` feature was added, it was only implemented for the `not searchable` use case.

I think that the issue can be solved by changing how the second parameter of the `filter()` function ([filter-data:43](https://github.com/mantinedev/mantine/blob/dev/src/mantine-core/src/MultiSelect/filter-data/filter-data.ts#L43)) is calculated from:

```js
filter(
  searchValue,
  value.some((val) => val === data[i].value && !data[i].disabled),
  data[i]
)
```

to:

```js
filter(
  searchValue,
  !disableSelectedItemFiltering &&
    value.some((val) => val === data[i].value && !data[i].disabled),
  data[i]
)
```

That way, if the `disableSelectedItemFiltering` prop is enabled, the `selected` parameter of the `filter()` function will be always `false`; therefore items will never be filtered when selected. This solution should work for groups as well.

I'd create a PR with the fix if that works for you guys, let me know.

rtivital:
The fix seems fine, you are welcome to submit a PR

rtivital:
Fixed in 6.0.5

PR #3894, opened 2023-03-24
[@mantine/core] MultiSelect: Fix `disableSelectedItemFiltering` prop not working for `searchable` input

lgaspari:
Solves #3877 

It enables the `disableSelectedItemFiltering` feature for `searchable` inputs by always setting the `filter` function's `selected` parameter to `false`.

This works for the internal `defaultFilter` and should work for any custom `filter` function that relies on the `selected` parameter to remove from the list.

Lastly, I added some tests to assert:
- `disableSelectedItemFiltering` when input is `not searchable`
- `disableSelectedItemFiltering` when input is `searchable` (uses `defaultFilter`)
- `filter` function's `selected` paramater is always`false` when `disableSelectedItemFiltering` and `searchable` are enabled (generic test for any custom `filter` function)

### DIFF
--- a/docs/src/docs/core/MultiSelect.mdx
+++ b/docs/src/docs/core/MultiSelect.mdx
@@ -161,6 +161,22 @@ Apart from `itemComponent` you can customize appearance of label by providing `v
 
 <Demo data={MultiSelectDemos.maxSelectedValues} />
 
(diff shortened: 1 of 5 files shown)


rtivital:
Thanks!

Issue #3866, opened 2023-03-22
Input has no accessible name a11y error on Autocomplete


csmatt:
### What package has an issue

@mantine/core

### Describe the bug

The JAWS screen reader is not recognizing the name of the Autocomplete component. JAWS reads it as "unlabeled1". This occurs on every autocomplete example in the Mantine docs. If you do not have JAWS installed, you can see this error by downloading and installing the ARC Toolkit chrome browser extension. Go to the Autocomplete page in Mantine docs and press F12 and navigate to the ARC Toolkit tab. Click on Run Tests. The error pertaining to this issue is "Input has no accessible name".

My best guess as to what is going on is that the element with the input that is associated with the label using the for/id relationship is not the same element as the one with role="combobox" and that this is causing an issue for screen readers.

### What version of @mantine/hooks page do you have in package.json?

6.0.4

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

None

PR #3900, opened 2023-03-25
[@mantine/core] Autocomplete: Fix for input has no accessible name a11y error

csmatt:
Proposed fix for #3866

Moved aria attributes and role attribute from wrapper div to input so that JAWS and other screen readers will pick up on the Autocomplete's label correctly. I'm not at all an expert on aria stuff, so please make sure what I've done is correct.


rtivital:
Thanks!

PR #3902, opened 2023-03-25
Clarify mediaQuery usage, fix wrong cases

### LINKED ISSUE #3901: [V6] Fix mediaQueries in some cases and in docs
### What package has an issue

@mantine/core

### Describe the bug

Related old issue, that I helped with #2550
I see that some parts of the documentation need editing, and maybe also some parts of the code and the docs (like [this one](https://github.com/mantinedev/mantine/blob/fb8170d828684cf3be3e14128119e0eee18df28c/docs/src/docs/theming/functions.mdx))

Only in some places there may have wrong (old) logic (this may have been done unintentionally, but due to refactoring of a large code base)

latest

_No response_

None

None

### Possible fix

_No response_

7iomka:
Fixes #3901


rtivital:
Thanks!

Issue #3896, opened 2023-03-25
form.resetDirty is unreliable when called with partial data


btmnk:
### What package has an issue

@mantine/form

### Describe the bug

When updating the form with an async response the useForm recipe currently shows 

```tsx
  useEffect(() => {
    loadInitialValues().then((values) => {
      form.setValues(values);
      form.resetDirty(values);
    });
  }, []);
```

This works fine in the assumption that `values` contains the very same structure as the form will have after `form.setValues(values)` has updated the state. That's not always the case though.
When calling`form.setValues(values)` the form will actually just overwrite the current values with the passed object.
If `values` is just a partial of the full form values than we still keep the not updated previous values.
 
For `form.resetDirty(values)` that's not the case. Instead the form will take the partial as the new snapshot and compare it with the merged values and return true again. This kindof forces us to make absolutely sure that we always pass the full structure into `resetDirty()` and fill up any missing values that might not always be part of an API response for example.

Right now I'm solving this by using `form.resetDirty({ ...defaults, ...data })` which seems to be enough for the time being.

I think it would already help if this was explained at https://mantine.dev/form/recipes/#set-initial-values-with-async-request

### What version of @mantine/hooks page do you have in package.json?

6.0.4

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/damp-http-x9y3qg?file=/src/App.tsx

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

Maybe we could pass the data to `form.reset(data)` so it uses that instead of the initialValues to reset the form so we can get a clean form state with the response. In this case we could also just merge the initialValues with the passed data so it keeps the defaults for unset properties.

rtivital:
You are welcome to submit a pull request that will provide support for values partial in `form.resetDirty`. No need to update anything in the documentation: code itself and a test will be enough.

btmnk:
I wonder if it would make more sense to extend the setValues function...

E.g. the actual thing we want to do usually is to just update the values of the form without marking it as dirty. Using two functions after each other who both update some internal state seems a bit counter intuitive to me. Especially considering that it's not transparent what would happen if one would just call `form.resetDirty` without setValues.

Wouldn't something like
```ts
setValues = ({ values: Values, resetDirty: boolean, reset: boolean }) => ...
```
be more useful?

rtivital:
[...]

PR #3906, opened 2023-03-25
Feature/mantine form partial reset

btmnk:
fixes https://github.com/mantinedev/mantine/issues/3896

# Changes
- Renamed file resert.test.ts to reset.test.ts
- Renamed internal _dirtyValues ref to valuesSnapshot as a naming suggestion
- Changed resetDirty to spread passed values with the current values to allow passing a subset of the form value structure. (e.g. when some of the form values can be undefined)
- Added optional values parameter to `reset` to allow reliably resetting the form with new data in one go.

With these changes the resetDirty call mimics the spread behaviour of setValues and avoids unexpected dirty states when updating the form with async data that does not completely override the initialValues.

Additionally with the optional argument of reset it is possible to switch the context of the form in one reliable state update. A specific use case could be when the resource for the data is switched on the same page as the form (e.g. data for different users). Right now when using form.setValues the data would be merged and introduce unexpected form state when values from the previous data are kept (e.g. ids that might only exist for entities that have alread been persisted). 
If we just run form.reset() before form.setValues() we still have stale values from before form.reset() since React state updates work asynchronous.

### DIFF
--- a/src/mantine-form/src/tests/reset.test.ts
+++ b/src/mantine-form/src/tests/reset.test.ts
@@ -31,4 +31,18 @@ describe('@mantine/form/reset', () => {
     expect(hook.result.current.isDirty()).toBe(false);
     expect(hook.result.current.isTouched()).toBe(false);
   });
+
+  it('resets values without keeping added values', () => {
+    const hook = renderHook(() =>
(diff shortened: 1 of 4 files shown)


btmnk:
I just noticed that most of the PRs point to master. Is that the intended target branch for contributions?

rtivital:
Thanks!

PR #3921, opened 2023-03-27
[docs] Pagination: Updated pagination example

praiseisaac:
Changed `page` to `value` for pagination example to match the new `Component Props` for `6.0.4`. Ran into an error after migrating and noticed that the example was outdated.


rtivital:
Thanks!

PR #3925, opened 2023-03-28
[@mantine/dates]: Prevent page scrolling on ArrowUp/ArrowDown keyboard navigation

filiptammergard:
When switching focus between dates with ArrowUp/ArrowDown, page is scrolling. Test it here: https://mantine.dev/dates/date-picker/#usage

This prevents that page scrolling.


rtivital:
Thanks!

Issue #3929, opened 2023-03-28
Broken link


Cstrp:
Section with components and choice of locales, there is a link that leads to the description of the provider... when you click it. Came to the "secret place" now I'm here
https://mantine.dev/dates/date-provider/ - not found :3

rtivital:
On which page did you find this link?

PR #3938, opened 2023-03-28
Fix broken link to DatesProvider

noahsb:
One more broken link that fixes https://github.com/mantinedev/mantine/issues/3929

Sorry, used UI to submit PRs so I had to do separately


rtivital:
Thanks!

Issue #3861, opened 2023-03-22
'Types of DateInput is missing props from DatePicker


KilianB:
### What package has an issue

@mantine/dates

### Describe the bug

The `DateInput` typing should extend the `DatePicker` as all of the values are supposed to be supported.

e.g. at least the following currently are missing.

```typescript
    /** Max level that user can go up to (decade, year, month), defaults to decade */
    maxLevel?: CalendarLevel;
    /** Initial level displayed to the user (decade, year, month), used for uncontrolled component */
    defaultLevel?: CalendarLevel;
````

### What version of @mantine/hooks page do you have in package.json?

6.0.4

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

Fix: 

DateInput, DatePickerInput and DateTimePicker should either extend DatePickerBaseProps or be merged with it.
Same for Time Yea and Month pickers correspondingly. 

Workaround: 

1. silence ts
````typescript
<DateInput
    //@ts-expect-error incorrect typing
    defaultLevel={'decade'}
/>
````

2. declaration merging with d.ts

KilianB:
@rtivital could you please help me out here a little bit?  I would like to provide a pull request for this issue and #3862 respectively but attempting to install the dependencies on 2 different windows 64 bit machines fails while building the `lmdb-store` dependency for the doc module.

`lmdb-store` has been sunset for the new `lmdb-js` in gatsby v 5. Is the dev dependency even necessary or can I safely remove it for installation and building? 

<details>
<summary>Build error log</summary>

````
PS G:\git\adac\mantine> yarn
yarn install v1.22.19
[1/4] Resolving packages...
[2/4] Fetching packages...
[3/4] Linking dependencies...
warning "@emotion/react > @emotion/babel-plugin@11.9.2" has unmet peer dependency "@babel/core@^7.0.0".
warning "@emotion/react > @emotion/babel-plugin > @babel/plugin-syntax-jsx@7.12.13" has unmet peer dependency "@babel/core@^7.0.0-0".
warning "@emotion/styled > @emotion/babel-plugin@11.10.0" has unmet peer dependency "@babel/core@^7.0.0".
warning "@emotion/styled > @emotion/babel-plugin > @babel/plugin-syntax-jsx@7.18.6" has unmet peer dependency "@babel/core@^7.0.0-0".
[...]

PR #3951, opened 2023-03-29
[@mantine/dates] DateInput: Add missing props from DatePicker

### LINKED ISSUE #3787: DateInput is missing interface from DatePicker
### What package has an issue

@mantine/dates

### Describe the bug

In Mantine 5, you could control the default calendar level using the `initialLevel` prop. In Mantine 6, this has been replaced with the `defaultLevel` prop which is part of the [DatePicker](https://mantine.dev/dates/date-picker/?t=props) component.

The [DateInput](https://mantine.dev/dates/date-input/) component, in theory, "supports all props from DatePicker", however a few props are missing from the interface; specifically:

- `maxLevel`
- `defaultLevel`
- `level`
- `onLevelChange`

It looks like the interface for those props aren't being imported into the DateInput component. By contrast, they _are_ being properly imported into the DatePickerInput component (as [DatePickerBaseProps](https://github.com/mantinedev/mantine/blob/master/src/mantine-dates/src/components/DatePickerInput/DatePickerInput.tsx)).

6.0.2

https://codesandbox.io/s/charming-grass-t7kiqj?file=/src/App.tsx

### Possible fix

Import `DatePickerBaseProps` into `DateInput`.

jsuter:
Fixes #3787 and #3861

### DIFF
--- a/src/mantine-dates/src/components/DateInput/DateInput.tsx
+++ b/src/mantine-dates/src/components/DateInput/DateInput.tsx
@@ -19,7 +19,7 @@ import { YearLevelSettings } from '../YearLevel';
 import { MonthLevelSettings } from '../MonthLevel';
 import { HiddenDatesInput } from '../HiddenDatesInput';
 import { assignTime } from '../../utils';
-import { DateValue } from '../../types';
+import { DateValue, CalendarLevel } from '../../types';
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #3973, opened 2023-03-31
[@mantine/core] Input: Add left and right padding to all variants

dbnar2:
Add paddingLeft and paddingRight for all variants. Prevents text from overlapping when there is a rightSection.


dbnar2:
Updating with an image of the issue.
Screenshot_20230406_231832_Chrome

As you can see the text will overflow into the right section. If the text input is unstyled variant. If this is unintended then this PR will fix that

PR #3977, opened 2023-04-01
Add default values to initial state

aviemet:
It would be helpful to have an initial value set to the window dimensions rather than 0. When using this hook to make formatting decisions, the initial 0 on component load causes flashes of incorrectly formatted content until the useEffect is fired.


rtivital:
It will break ssr. window object is not defined on server.

aviemet:
Gotcha, I don't use SSR, so hadn't considered that. Would it work if we first checked for the existence of window?

rtivital:
No, it will not work. It will produce different rendering result on client and server – ssr will be broken.

Issue #3968, opened 2023-03-31
MultiSelect input and caret is shifted in Safari


Cuzart:
### What package has an issue

@mantine/core

### Describe the bug

Line height is calculated wrong so that the caret and the text input is getting shifted to the downside.

<img width="699" alt="Bildschirm­foto 2023-03-31 um 09 54 26" src="https://user-images.githubusercontent.com/60718229/229060788-de80dbcd-9b69-4d8c-a05a-9b5773c4f4b4.png">

[Also visible on the documentation](https://mantine.dev/core/multi-select/#searchable)

Safari Version 16.3

### What version of @mantine/hooks page do you have in package.json?

6.0.1

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/floral-tdd-bpkb57?file=/src/App.tsx

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

Check how line-height is being calculated and why it is wrong

line-height: calc(2.25rem - 0.125rem);

jrozbicki:
@rtivital This can be closed, got fixed within #3980 I think I just failed to tag issue properly in PR.

PR #3980, opened 2023-04-02
[@mantine/core] MultiSelect: Fix lineHeight styles affecting input on Safari

jrozbicki:
The issue seemed to be that effective lineHeight calculation resulted in `34px` for `sm` size - which was bigger than input itself (22px) which caused cursor and text to be pushed down.

Removing this style fixed the issue - as would lowering down lineHeight to 22px or below - but this seemed to not have any visual effect even when I played with lower values like 10px or 1px.

 Removal did not affect `Size XS Line Height` story and looks correct on `Multiple Rows`. Tested on both Safari and Chrome.
 
 Fixes https://github.com/mantinedev/mantine/issues/3968
 
 It's my first PR here, hopefully adhered to all standards - let me know if I could improve anything, cheers!


rtivital:
Thanks!

Issue #3982, opened 2023-04-03
NumberInput cannot obtain the latest value of the custom value in the onBlur state


xshuxin:
### What package has an issue

@mantine/core

### Describe the bug

NumberInput cannot obtain the latest value of a custom value in the onBlur state. However, ordinary input can correctly obtain the latest value

### What version of @mantine/hooks page do you have in package.json?

^6.0.5

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/kind-margulis-94682o?file=/src/App.tsx

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

  const handleChange = (event: React.ChangeEvent<HTMLInputElement>) => {
    const evt = event.nativeEvent as InputEvent;
    if (evt.isComposing) {
      return;
    }
    const val = event.target.value;
    setInputValue(val);

    onChange?.(val);
  };

 /** I found that NumberInput is packaged based on TextInput, and the onChange modification method of inputValue has been internally redefined by a layer. However, at this time, the outer layer's onChange method is not executed, but only executed during onBlur, resulting in obtaining the previous value instead of the latest value in the onBlur state*/

cyantree:
Please see the following example that uses the patched version introduced in #3984:
https://codesandbox.io/s/mutable-https-w9buhp?file=/src/App.tsx

PR #3984, opened 2023-04-03
Issue 3979 fix numberinput form validation

### LINKED ISSUE #3979: NumberInput with useForm not handling onBlur properly
### What package has an issue

@mantine/core

### Describe the bug

`<NumberInput />` not validating the first time when `validateInputOnBlur` is true. But it works the second time. Check the CSB.

6.0.5

https://codesandbox.io/s/sweet-shape-93pb1q?file=/src/App.tsx

No

No

### Possible fix

_No response_

cyantree:
Fixes #3979 and #3982.

Based on #3756 I reworked the data flow handling to be more in line with other components again.

This means `onChange()` will be triggered on input again but still outputs validated and clamped data.

I've also added a test for the expectation that `onChange()` and `onBlur()` won't get called in the same run which might lead to the `onBlur()` handler working with old data in case it's using data stored via `useState()` hook.

Maybe that's something that should be added to all input components.

I also removed the `higherPrecision` handling introduced in #3756. For controlled components it allowed display values with higher precision than allowed. Like:
```tsx
<NumberInput precision={0} value={1.23} />
```
would output `1.23`. However when changing the value manually afterwards it would fall back to the configured precision.
Because this might be confusing I dropped this behaviour again. Sorry for that.
However if you find this useful I can revert this change.

### DIFF
--- a/src/mantine-core/src/NumberInput/NumberInput.story.tsx
+++ b/src/mantine-core/src/NumberInput/NumberInput.story.tsx
@@ -40,8 +40,12 @@ export function FixedValue() {
 export function WithUseFormHook() {
   const mantineForm = useForm({
     initialValues: {
-      someNumber: 3,
+      someNumber: 0,
+    },
+    validateInputOnBlur: true,
+    validate: {
+      someNumber: (val) => (!val ? 'please enter a number greater than zero' : null),
     },
   });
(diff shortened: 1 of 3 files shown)


rtivital:
Thanks!

PR #3989, opened 2023-04-03
Issue 3919 fix datepicker type changing

### LINKED ISSUE #3919: DatePickerInput crashes on `type` prop change
### What package has an issue

@mantine/dates

### Describe the bug

The `type` prop can not be updated since it affects the value type and some other properties as well. I'm not sure this limitation is on purpose and probably can be fixed internally.

By changing  the `type` from `default` to `multiple` or `range` it throws this error:
```
Error in /turbo_modules/@mantine/dates@6.0.4/cjs/utils/get-formatted-date.js (23:17)
Cannot read properties of null (reading 'map')
```

By changing from `multiple` or `range` to `default` it won't throw error but the input entry will get change to `invalid Date`.

6.0.4

https://stackblitz.com/edit/react-ts-scbkog?file=index.tsx

Yes

None

### Possible fix

Currently I change the DatePickerInput `key` and `value` by changing `type`

cyantree:
Fixes #3919.
- For uncontrolled components it resets the internal value to `defaultValue`
- For controlled components it expects the `value` to be compatible to `type`. However it also outputs an error it that's not the case

### DIFF
--- a/src/mantine-dates/src/hooks/use-uncontrolled-dates/use-uncontrolled-dates.test.ts
+++ b/src/mantine-dates/src/hooks/use-uncontrolled-dates/use-uncontrolled-dates.test.ts
@@ -1,0 +1,145 @@
+import { renderHook } from '@testing-library/react';
+import { useUncontrolledDates } from './use-uncontrolled-dates';
+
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #3990, opened 2023-04-03
[@mantine/core] NumberInput: Add `thousandsSeparator` #3988

### LINKED ISSUE #3988: NumberInput always change the decimalSeparator to "."
### What package has an issue
@mantine/core
### Description
In the Spanish representation of numbers we separate the thousands as "." and the decimals as ",".

The problem in NumberInput is when you put decimalSeparator="," and the component at the moment of calling the parser function changes the decimalSeparator to ".".

This causes the parser function to receive values that cannot differentiate the thousands from the decimals because previously in the formatter function I changed the thousands to ".".

You could add a prop to check if you want to alter the value passed to the parser, so you can handle the value yourself.
Maybe something like this:

image

cyantree:
Fixes #3988.

### DIFF
--- a/docs/src/docs/core/NumberInput.mdx
+++ b/docs/src/docs/core/NumberInput.mdx
@@ -78,9 +78,9 @@ To use decimal steps set following props:
 
 <Demo data={NumberInputDemos.decimal} />
 
-## Decimal separator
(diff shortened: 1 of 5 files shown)


rtivital:
Thanks!

PR #3993, opened 2023-04-04
[@mantine/dates] DatePickerInput: Support withCellSpacing prop

filiptammergard:
You can see that `data-with-spacing="true"` remains when setting `withCellSpacing={false}` here: https://codesandbox.io/s/compassionate-stonebraker-zq7n52?file=/src/App.tsx

I included a test that failed before the change and passes after.


rtivital:
Thanks!

PR #4004, opened 2023-04-05
[docs] Fix missing `multiple` prop on example code at Accordion

newt239:
in the section of [Default opened items](https://mantine.dev/core/accordion/#default-opened-items), second code example was wrong. 
when set `defaultValue` as an array of strings, should be add `multiple` prop as `true`.


rtivital:
Thanks!

PR #4012, opened 2023-04-05
[@mantine/core] ColorInput: execute `onChange()` before `onBlur()` #3280

### LINKED ISSUE #3280: ColorInput is not validating
### What package has an issue

@mantine/form

### Describe the bug

I'm using the ColorInput with the zod schema validator but it doesn't validate the empty field.
Other fields like TextInput work fine.

5.10.0

https://stackblitz.com/edit/react-ts-mjdfd8?file=App.tsx

No

None

### Possible fix

_No response_

cyantree:
Fixes #3280.

`ColorInput` currently triggers an additional `onChange()` after `onBlur()`. This conflicts with form validations that have a disabled `onChange()` validation because in this case the error will be resetted immediately.

@rtivital Can you describe the reason behind this additional `onChange()` trigger and `fixOnBlur`? Maybe it can be removed completely.

### DIFF
--- a/src/mantine-core/src/ColorInput/ColorInput.test.tsx
+++ b/src/mantine-core/src/ColorInput/ColorInput.test.tsx
@@ -8,11 +8,16 @@ import {
   itSupportsProviderVariant,
   itSupportsProviderSize,
 } from '@mantine/tests';
+import { fireEvent, screen, render } from '@testing-library/react';
 import { ColorInput, ColorInputProps } from './ColorInput';
 
 const defaultProps: ColorInputProps = {
   label: 'test-label',
 };
+
+const getInput = () => screen.getByRole('textbox');
+const blurInput = () => fireEvent.blur(getInput());
+const focusInput = () => fireEvent.focus(getInput());
 
 describe('@mantine/core/ColorInput', () => {
   checkAccessibility([<ColorInput label="Color input" />, <ColorInput aria-label="Color input" />]);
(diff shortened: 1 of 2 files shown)


rtivital:
Based on the code 

```ts
const handleInputBlur = (event: React.FocusEvent<HTMLInputElement>) => {
    fixOnBlur && setValue(lastValidValue);
    onBlur?.(event);
    setDropdownOpened(false);
  };
```

`onChange` should be called once when the input is blurred. I'm not really sure what is happening there if that is not the case.

cyantree:
Thanks.
As it's also a public property removing it would be a breaking change. So maybe we'll just keep it and maybe evaluate removing it in the next major release?

rtivital:
Do you want to remove `fixOnBlur` prop? It may be useful in some cases, for example when you do not want to lose user's input even if it is invalid.

cyantree:
Ok, after reading your comment and playing around with the prop I think I know understand its naming and purpose.

For me this is about conformity across all input components. IMHO one of the key decisions is: "Should an input component output invalid data via `onChange()`".
a) `ColorInput` emits every user input regardless whether it's correct or not. To make sure a valid input remains when blurring the prop `fixOnBlur` can be used to "fix" the user input to be valid again.
b) `NumberInput` now only emits validated/sanitized user input (means interpreted and clamped one).

If haven't checked other components to see which approach is currently more used in Mantine.

Both approaches have their ups and downs:
a) Gives the user more freedom but also responsibility because the user must additionally check whether the outputted input value is valid.
b) Makes sure the outputted input value is always valid and therefore makes using the component easier.

If you want to use the best of two worlds the following interface change for `onChange()` would allow this:
`onChange({value: "#ff0000", isValid: true})`

Instead of simply outputting the validity it also could output the internal validation that failed the value:
`NumberInput.onChange({value: 123, error: 'max'})` (in case the value exceed the max boundary)
`NumberInput.onChange({value: 12, error: null})` (in case the value is valid)

For native html input elements it seems to be the following:
- You can access the raw input value via `.value`
- You can access the validity based on the configured tag via `.checkValidity()`

IMHO b) is the more natural approach as I want these components to be easy to work with and (by default) to be more high than low level.

cyantree:
> Do you want to remove `fixOnBlur` prop?

Regarding my new understanding and the currently implemented logic of the component:
No, I don't want to remove it. Previously I thought it was some kind of workaround (hence the naming "fix") but now I see it as a feature.

rtivital:
Thanks!

PR #4017, opened 2023-04-06
[@mantine/dates] DatePickerInput: fix `type` field always displaying …

welpie21:
…default in autocompletion

The `type` field in `DatePickerInput` was always showing `default` in the list.

**before change:**
image

**after change:**
image


rtivital:
Thanks!

Issue #4029, opened 2023-04-07
Using the toggle method provided internally by useDisclosure can cause Menu to rely on rendering beyond the maximum update depth


xshuxin:
### What package has an issue

@mantine/core

### Describe the bug

When Menu uses the toggle method provided internally by useClosure to control unfolding and collapsing, it can lead to dependency rendering exceeding the maximum update depth and falling into a dead loop. The following is the error message from the console: "Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either does not have a dependency array, or one of the dependency changes on every render.

### What version of @mantine/hooks page do you have in package.json?

^6.0.5

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/fast-wave-lfz8d3?file=/src/App.tsx

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

Through code retrieval, I found that Menu repeatedly called the "onClose" and "onOpen" methods internally. When "toggle" was executed, the "_opened" state became dependent, constantly triggering the closure and opening methods of Popover. "_opened" was constantly changing, leading to an infinite loop. At present, I have annotated the "onClose" and "onOpen" methods in "Popover", leaving only setOpened to control the unfolded and collapsed state. I have found that the dead loop problem has been resolved, and no new issues have been found. But I don't know if this solution will bring about other problems. Is there a better solution?

<img width="1408" alt="image" src="https://user-images.githubusercontent.com/105474651/230639719-d85a910f-645c-4aa4-b37e-4b9448bcdd66.png">

This is my PR:
https://github.com/mantinedev/mantine/pull/4030

xshuxin:
I'm very sorry, the word was written incorrectly. It should be 'useClosure' ->'useClosure'

xshuxin:
I'm very sorry, the word was written incorrectly. It should be 'useClosure' ->'use disclosure'

cyantree:
I think the issue is using `toggle()` which seems to trigger an endless loop. Try setting the state explicitly.
Here's an updated sandbox:
https://codesandbox.io/s/ecstatic-chihiro-3rvt8w

#3817 had a similar issue.

IMHO `onChange()` shouldn't be called when `opened` will be changed programatically as this violates data flow patterns involving UI interaction. It should be only called when the `Menu.target` gets clicked.
For `onOpen()` and `onClose()` that's already the case.

[...]

PR #4030, opened 2023-04-07
[@mantine/core]Menu: Removing code that may cause dead loops 

xshuxin:
**I have already raised the issue and hope it can be helpful~**
The following is a description of the problem：
When Menu uses the toggle method provided internally by useDisclosure to control unfolding and collapsing, it can lead to dependency rendering exceeding the maximum update depth and falling into a dead loop. The following is the error message from the console: "Maximum update depth exceeded. This can happen when a component calls setState inside useEffect, but useEffect either does not have a dependency array, or one of the dependency changes on every render.

### DIFF
--- a/src/mantine-core/src/Menu/Menu.story.tsx
+++ b/src/mantine-core/src/Menu/Menu.story.tsx
@@ -1,6 +1,7 @@
 import React, { useState } from 'react';
 import { IconTable, IconSearch } from '@tabler/icons-react';
 import { WithinOverlays } from '@mantine/storybook';
+import { useDisclosure } from '@mantine/hooks';
 import { Menu } from './Menu';
 import { Button } from '../Button';
 import { Tooltip } from '../Tooltip';
@@ -95,6 +96,24 @@ export function Controlled() {
       <Menu opened={opened} onChange={setOpened}>
         <Menu.Target>
           <Button>Toggle controlled menu</Button>
+        </Menu.Target>
+        <Menu.Dropdown>
+          <Menu.Item>Item 1</Menu.Item>
+          <Menu.Item>Item 2</Menu.Item>
+        </Menu.Dropdown>
+      </Menu>
+    </div>
+  );
+}
+
+export function WithUseDisclosure() {
+  const [opened, handle] = useDisclosure(false);
+
+  return (
+    <div style={{ padding: 40 }}>
+      <Menu opened={opened} onChange={() => handle.toggle()}>
+        <Menu.Target>
+          <Button>UseDisclosureToggle controlled menu</Button>
         </Menu.Target>
         <Menu.Dropdown>
           <Menu.Item>Item 1</Menu.Item>
--- a/src/mantine-core/src/Menu/Menu.tsx
+++ b/src/mantine-core/src/Menu/Menu.tsx
@@ -79,7 +79,6 @@ const defaultProps: Partial<MenuProps> = {
   openDelay: 0,
   closeDelay: 100,
 };
-
 export function Menu(props: MenuProps) {
   const {
     children,
@@ -157,7 +156,7 @@ export function Menu(props: MenuProps) {
         {...others}
         radius={radius}
         opened={_opened}
-        onChange={setOpened}
+        onChange={toggleDropdown}
         defaultOpened={defaultOpened}
         trapFocus={trigger === 'click'}
         closeOnEscape={closeOnEscape && trigger === 'click'}
--- a/src/mantine-core/src/Menu/Menu.tsx
+++ b/src/mantine-core/src/Menu/Menu.tsx
@@ -165,8 +165,6 @@ export function Menu(props: MenuProps) {
         classNames={{ ...classNames, dropdown: cx(classes.dropdown, classNames?.dropdown) }}
         styles={styles}
         unstyled={unstyled}
-        onClose={close}
-        onOpen={open}
         variant={variant}
       >
         {children}


xshuxin:
`onChange()` should not be called when the` Menu` is open, but should be called when `Menu. target`is clicked, so I attempted to fix the data flow interaction form to avoid dead loops.

#4029

cyantree:
Thanks for the PR.
However I think that's not sufficient.

IMHO there are three usecases that should be checked:
- Uncontrolled use
- Controlled use where `Menu.Target` will be used for toggling the state
- Controlled use where some other component will be used for toggling the state

I've prepared an example for these:
https://codesandbox.io/s/fast-wave-lfz8d3?file=/src/App.tsx

I think the following should happen:

## Uncontrolled
### When the menu is closed and the user clicks to open it
- the menu opens
- `onOpen()`  will be called
- `onChange(true)` will be called

### When the menu is opened and the user clicks to close it
- the menu closes
- `onClose()` will be called
- `onChange(false)` will be called

## Controlled with `Menu.Target`
### When `opened={false}` and the user clicks the trigger (or something else) to open it
- the menu **does not automatically**  open
- `onOpen()` will be called
- `onChange(true)` will be called
- `opened={true}` will be set from the outside
- the menu opens
- neither `onOpen()` nor `onChange()` will be called again

### When `opened={true}` and the user clicks the trigger (or something else) to close it
- the menu **does not automatically**  close
- `onClose()` will be called
- `onChange(false)` will be called
- the menu closes
- neither `onOpen()` nor `onChange()` will be called again

## Controlled with external component
### When `opened` gets set to `true`
- the menu opens
- neither `onOpen()` nor `onChange()` will be called

### When `opened` gets set to `false`
- the menu closes
- neither `onOpen()` nor `onChange()` will be called

---

Currently (before and after your PR) these cases do not work consistently and also `onChange()` sometimes gets triggered too often.

Also this needs to be tested for `trigger='hover'` where hovering the target opens/closes the menu.

xshuxin:
@cyantree  You're absolutely right! I have indeed overlooked many scenarios. I just updated the code and tested it again. The scenarios you mentioned are basically covered. I have relatively shallow experience and this is the first time I have mentioned `PR`. I feel that you are very experienced in this area. Thank you very much for your suggestion! If possible, do you have time to help `review` the code? Or propose a better solution?

cyantree:
Thanks @xshuxin for taking the time and initiative!

Sure, I'm happy to help reviewing your code.

However at first maybe let's check back with @rtivital whether he agrees with the proposed concept. I'm just a contributor as you and therefore not the one making the decisions.

If he agrees have a look at the unit tests in `Menu.test.tsx`. They should cover all described scenarios, succeed with your changes and (at least partially) fail with the pre-PR code.

For testing the menu tests I'd change `jest.config.js` as the following to only run the relevant test:
```
testMatch: ['**/Menu.test.tsx'],
```

I don't know if there's an easier way to achieve this?

After finalizing the tests please revert this change before committing and run the whole test suite again.

xshuxin:
@cyantree 

Thank you very much! 
Based on your proposal, I conducted a single module automation test on `Menu` locally and found that it passed successfully. At the same time, the entire automation test seems to have also passed.
I'm very sorry, I don't have a better way to conduct a single module self-test more easily.
Compared to modifying configuration files, it seems more convenient to run commands directly~

`npx jest <path-to-file> `
`npx jest --watch`

 I am excited and hope to successfully propose my first `PR`!

rtivital:
Hi @xshuxin, @cyantree proposal is correct, please let me know when your code is ready for review

xshuxin:
> 
@cyantree @rtivital 
I'm very sorry, I just saw this message.I am ready for the code to be reviewed~

cyantree:
Thanks @xshuxin , also for the Jest hint. :)
Can you extend the tests to cover all cases described in https://github.com/mantinedev/mantine/pull/4030#issuecomment-1500831605 ?
That allows avoiding regressions.
As described earlier these tests should (at least partially) fail with the old code (because it has it flaws) but succeed with the new code.

@ikapta Thanks for the insight. Yes, maybe that's also relevant here. Extending the menu tests helps getting and keeping this right.

cyantree:
Unfortunately I can't find any new tests. They would really help finding the missing pieces. Do you need help with that? It's ok if they are not perfect or fail but it would be a good next step in covering all usecases.

It's correct that the cases I described won't match the old logic. That's because I think programmatic state changes shouldn't trigger these `on*()` callbacks and the old logic does that.

cyantree:
Here are some tests for the described scenarios with `trigger="click"`.
And they fail with the old code and pass with your changes. Great!
Feel free to rework or adjust them.

So what's missing are tests for the mode `trigger="hover"`.

```ts
  it('correctly calls callbacks when opening and closing the uncontrolled menu via target click', async () => {
    const spyOnOpen = jest.fn();
    const spyOnClose = jest.fn();
    const spyOnChange = jest.fn();

    render(<TestContainer onChange={spyOnChange} onOpen={spyOnOpen} onClose={spyOnClose} />);

    expectClosed();

    await userEvent.click(getControl());

    expectOpened();
    expect(spyOnOpen).toHaveBeenCalledTimes(1);
    expect(spyOnClose).not.toHaveBeenCalled();
    expect(spyOnChange).toHaveBeenCalledTimes(1);
    expect(spyOnChange).toHaveBeenLastCalledWith(true);

    spyOnOpen.mockReset();
    spyOnClose.mockReset();
    spyOnChange.mockReset();
    await userEvent.click(getControl());

    expectClosed();
    expect(spyOnOpen).not.toHaveBeenCalled();
    expect(spyOnClose).toHaveBeenCalledTimes(1);
    expect(spyOnChange).toHaveBeenCalledTimes(1);
    expect(spyOnChange).toHaveBeenLastCalledWith(false);
  });

  it('correctly calls callbacks when opening and closing the controlled menu only via prop', async () => {
    const spyOnOpen = jest.fn();
    const spyOnClose = jest.fn();
    const spyOnChange = jest.fn();

    const ControlledMenu = (props: MenuProps) => <TestContainer onChange={spyOnChange} onOpen={spyOnOpen} onClose={spyOnClose} {...props} />;

    const rendering = render(<ControlledMenu opened={false} />);

    expectClosed();
    expect(spyOnOpen).not.toHaveBeenCalled();
    expect(spyOnClose).not.toHaveBeenCalled();
    expect(spyOnChange).not.toHaveBeenCalled();

    rendering.rerender(<ControlledMenu opened />);

    // Changing state via prop shouldn't call on* methods
    expectOpened();
    expect(spyOnOpen).not.toHaveBeenCalled();
    expect(spyOnClose).not.toHaveBeenCalled();
    expect(spyOnChange).not.toHaveBeenCalled();

    rendering.rerender(<ControlledMenu opened={false} />);

    // Changing state via prop shouldn't call on* methods
    expectClosed();
    expect(spyOnOpen).not.toHaveBeenCalled();
    expect(spyOnClose).not.toHaveBeenCalled();
    expect(spyOnChange).not.toHaveBeenCalled();
  });

  it('correctly calls callbacks when opening and closing the controlled menu via target click', async () => {
    const spyOnOpen = jest.fn();
    const spyOnClose = jest.fn();
    const spyOnChange = jest.fn();

    const ControlledMenu = (props: MenuProps) => <TestContainer onChange={spyOnChange} onOpen={spyOnOpen} onClose={spyOnClose} {...props} />;

    const rendering = render(<ControlledMenu opened={false} />);
    await userEvent.click(getControl());

    expectClosed();
    expect(spyOnOpen).toHaveBeenCalledTimes(1);
    expect(spyOnClose).not.toHaveBeenCalled();
    expect(spyOnChange).toHaveBeenCalledTimes(1);
    expect(spyOnChange).toHaveBeenLastCalledWith(true);

    spyOnOpen.mockReset();
    spyOnClose.mockReset();
    spyOnChange.mockReset();
    rendering.rerender(<ControlledMenu opened />);

    expectOpened();
    await userEvent.click(getControl());

    expectOpened();
    expect(spyOnOpen).not.toHaveBeenCalled();
    expect(spyOnClose).toHaveBeenCalledTimes(1);
    expect(spyOnChange).toHaveBeenCalledTimes(1);
    expect(spyOnChange).toHaveBeenLastCalledWith(false);

    rendering.rerender(<ControlledMenu opened={false} />);

    expectClosed();
  });
```

xshuxin:
I'm sorry, I've been too busy these past few days.

I think I misunderstood your point earlier. I only conducted testing locally and did not extend `Jest`

Thanks @cyantree . I have carefully read some of the test code you provided for the scenario described by `trigger="click"`! They're great!

But I encountered some minor issues and found that they couldn't match through the `Prettier` format. It seems that there is a problem with these few lines of code.

```
  const ControlledMenu = (props: MenuProps) => <TestContainer
      onChange={spyOnChange}
      onOpen={spyOnOpen}
      onClose={spyOnClose}
      {...props}
    />;

    const rendering = render(<ControlledMenu opened={false} />);
```


Aaccording to my humble opinion, regarding the` trigger="hover"` scenario, it seems that the original `jest` is sufficient to describe the corresponding scenario. Do you think so?

rtivital:
Thanks!

rtivital:
I'll have to revert it in 6.0.9 – https://codesandbox.io/s/nifty-curran-fy1nlu?file=/src/App.tsx
menus are opening up when clicking anywhere on the screen, instead of when the Target is clicked

PR #4031, opened 2023-04-08
[@mantine/core] ColorInput: Trigger `onChangeEnd` event with eyedropper

krzysztoff1:
This pull request adds the `onChangeEnd` event to the `ColorInput` component when the user selects a colour using the eyedropper.


rtivital:
Thanks!

Issue #751, opened 2022-01-26
Form in modal seems broken


PringlePot:
### What package has an issue

@mantine/hooks

### Describe the bug

When creating a form within a modal, the textinput seems to glitch, when using it in codesandbox it doesnt seem to open an embed at all

### In which browser did the problem occur

Chrome

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/compassionate-noether-03cd4?file=/src/App.tsx

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

No

rtivital:
- It does not work in codesandbox because you are using `useModals` hook outside the `ModalsProvider` context
- Content modals cannot be used with dynamic children – once modal is opened, snapshot is saved into internal state and cannot be updated
- There are several options to handle dynamic content with modals:
  - [Use internal component state](https://codesandbox.io/s/romantic-cray-qhi2c?file=/src/App.tsx)
  - Use Modal component instead of modals manager

If you need any clarification, feel free to reach out here or on Discord.

PR #4042, opened 2023-04-11
[docs] note re dynamic content in Modals Manager

bentron2000:
@rtivital 
Thanks for this awesome library :)

While modifying some multi step modals I was caught out by the non-obvious fact that dynamic content is not supported in content modals via the modals manager API.

see #751 and [this discussion](https://github.com/mantinedev/mantine/discussions/2527#discussioncomment-3703202)

It would be fantastic to be able to have dynamic content - but perhaps it is too difficult.
My use case would be using modals for small, multi step configuration interfaces that make sense to be in modals.

Anyway - here is a note so that others do not get caught out.

Cheers

B


rtivital:
Thanks!

PR #4054, opened 2023-04-11
[@mantine/core] Notification: Remove border

jerebtw:
Sorry, I forgot the old border because you can't really see it in dark mode


rtivital:
Thanks!

PR #4065, opened 2023-04-12
[@mantine/dropzone]: Export FileRejection type

dylnslck:
I'm currently needing to add `react-dropzone` as a dependency to my project in order to import the `FileRejection` type. I'd love to be able to import it directly from `@mantine/dropzone`, just like I'm able to import `FileWithPath`.

Cheers!


rtivital:
Thanks!

PR #4083, opened 2023-04-16
[@mantine/core] fix slider not respecting container width

jvdsande:
As discussed in #4000, this fixes the width of Slider following marks behavior update


rtivital:
Thanks!

PR #4094, opened 2023-04-18
fix(Select): add whiteSpace:pre  to SelectItem

vaynevayne:
image

When there are multiple spaces in the middle of a character, the space will be swallowed up


rtivital:
Thanks!

PR #4114, opened 2023-04-21
Fix the issue where the Menu will open when clicking anywhere on the screen other than the target

xshuxin:
When clicking anywhere on the screen that is not a target, the Menu also performs opening and closing.
​https://codesandbox.io/s/nifty-curran-fy1nlu?file=/src/App.tsx

Based on my humble opinion, it may be caused by two reasons:

image

- The `dropdown`  is only hidden and has not been completely uninstalled, resulting in its lifecycle methods and other methods still existing.

- `clickOutside` is not controlled based on the `opened` state, so both the open and closed states will trigger `onClose`.

In version 6.0.8, in order for `Menu` to be properly controlled and simultaneously call `onClose` and `onOpen`, I made `clickOutside` call the `toggleDropdown` method of `Menu`, resulting in a situation where clicking on an external button can also control the switch state.

Here are the fixes I made:
 `clickOutside` only executes the `onClose` method when `opened: true`.

```
useClickOutside(() => opened && closeOnClickOutside && popover.onClose(), clickOutsideEvents, [
    targetNode,
    dropdownNode,
  ]) 
 ```
 At the same time, I also modified the test file for `closeOnClickOutside` in `Popover`:
 ```
   it('correctly handles closeOnClickOutside={false}', async () => {
    const spy = jest.fn();
    render(<TestContainer opened closeOnClickOutside={false} onClose={spy} />);
    await userEvent.click(document.body);
    expect(spy).not.toHaveBeenCalled();
  });

  it('correctly handles closeOnClickOutside={true}', async () => {
    const spy = jest.fn();
    render(<TestContainer opened closeOnClickOutside onClose={spy} />);
    await userEvent.click(document.body);
    expect(spy).toHaveBeenCalled();
  });
  ```
Finally, I have added detailed testing files for `Menu` in both controlled and uncontrolled states. Feel free to rework or adjust them~

### DIFF
--- a/src/mantine-core/src/Menu/Menu.story.tsx
+++ b/src/mantine-core/src/Menu/Menu.story.tsx
@@ -123,3 +123,20 @@ export function WithUseDisclosure() {
     </div>
   );
 }
+
+export function WithKeepMounted() {
+  return (
+    <div style={{ padding: 40 }}>
+      <Menu id="new-notification-menu" keepMounted position="bottom-end">
+        <Menu.Target>
+          <Button>Open Menu</Button>
+        </Menu.Target>
+        <Menu.Dropdown>
+          <Menu.Item>Menu Item 1</Menu.Item>
+          <Menu.Item>Menu Item 2</Menu.Item>
+          <Menu.Item>Menu Item 3</Menu.Item>
+        </Menu.Dropdown>
+      </Menu>
+    </div>
+  );
+}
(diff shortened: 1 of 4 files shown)


rtivital:
Thanks!

rtivital:
@xshuxin This PR broke Popover component click outside logic, see #4146

Issue #4112, opened 2023-04-21
Slider does not work with `scale` and `inverted` props.


preenx:
### Links to the pages:

https://mantine.dev/core/slider/#scale
https://mantine.dev/core/slider/#inverted

### Exact quote of what is wrong

In current version (v6.0.8) of your website slider examples with `scale` and `inverted` props are don't slide by mouse dragging. However, using older version (e.g. v5) all works as fine.

All [Keyboard interactions](https://mantine.dev/core/slider/#keyboard-interactions) works as expected and slider moves.

Tested on:
Google Chrome version 112.0.5615.138
Firefox version 112.0.1

rtivital:
@jvdsande Seems like it was caused by #4000. I've tested it in `revert-8bee2e5` branch and it works correctly. Can you check this please?

jvdsande:
Hmm. Let me take a look yeah!

jvdsande:
Ok, so it's actually a bug on the demo itself and not the component, it's because the demo uses padding on the slider instead of margin, which compresses the container down to 0px height, disabling touch event.

I'll fix the track-container height to use the size instead of a 100% value, so that padding can be applied correctly. But I'm not sure what "padding" even means on such a component to be honest: the main root has a fixed height anyway, so y-padding does not work as one would expect. It did not work before either, it was messing up the DOM alignment.

jvdsande:
@rtivital if I remove the height on the main container, and keep it only on the track-container, then padding seems to work correctly, maybe that would be the best?

![image](https://user-images.githubusercontent.com/12484011/233658377-7a946f0d-245b-4b97-8f01-8717738954f8.png)

If I keep the height on the main container as before, this happens:

![image](https://user-images.githubusercontent.com/12484011/233658524-77b41196-63fd-4835-be1c-9df5f1b1f76f.png)

As you can see the slider is completely collapsed down to 0 height (but it still displays outside of its bounding box thanks to the padding)

rtivital:
Fixed in 6.0.9

PR #4116, opened 2023-04-21
[@mantine/core] fix slider styles linked to thumb size

jvdsande:
As discussed in #4112 and Discord, here are the relevant fixes to Slider:

- Remove fixed height from the `root` container to allow style props such as padding to have correct effect on the document flow
- Add fixed height to the `trackContainer` instead
- Change offsets linked to thumb size to actually use `thumbSize` instead of only `size`


rtivital:
Thanks!

PR #4118, opened 2023-04-22
[@mantine/hooks] Focus trap identity

### LINKED ISSUE #4079: `FocusTrap` doesn't remove `aria-hidden=true` attribute
### What package has an issue

@mantine/core

### Describe the bug

If a component that renders a `FocusTrap` (Modal, Drawer, etc) is opened via a click handler on a `Menu.Item` that's rendered within a portal, the `aria-hidden=true` attribute is not cleared when the Modal/Drawer is closed, and the page becomes inaccessible.

6.0.7

https://codesandbox.io/s/angry-rgb-x46ub2

None

None

### Possible fix

_No response_

Ben-Kincaid:
Fixes #4079 
---

This addresses an edge case with `useFocusTrap` that can cause the [cleanup function](https://github.com/mantinedev/mantine/blob/42df57f888be86f1fa9adb5b1ca64546c3991dfe/src/mantine-hooks/src/use-focus-trap/use-focus-trap.ts#L82-L89) returned by `createAriaHider` to run after another `useFocusTrap` instance has operated on the same set of root nodes. This is due to the `createAriaHider` function being called when the [callback ref](https://github.com/mantinedev/mantine/blob/42df57f888be86f1fa9adb5b1ca64546c3991dfe/src/mantine-hooks/src/use-focus-trap/use-focus-trap.ts#L30-L66) returned by `useFocusTrap` is invoked, which will run before the cleanup function if two elements that utilize `useFocusRoot` are toggled simultaneously. An example of this behavior can be found in the original [issue](https://github.com/mantinedev/mantine/issues/4079) with `Menu.Item`/`Modal`: https://codesandbox.io/s/angry-rgb-x46ub2 

The approach taken was to set a `data-focus-id` attribute to every root node that has it's `aria-hidden` value mutated by `createAriaHider`. In the returned restore function, we are then able to check if the nodes current `data-focus-id` attribute matches the original assigned ID. If not, we know that another `useFocusTrap` instance was rendered & we should not move forward with restoring the `aria-hidden` attributes for that given run. Additionally, a `data-hidden` attribute will also be added to these root nodes so that we are able to keep track of the elements initial `aria-hidden` state between usages of `useFocusTrap`. 

Tests for `use-focus-trap` were also added, including a test case for the above scenario.

### DIFF
--- a/src/mantine-hooks/src/use-focus-trap/create-aria-hider.ts
+++ b/src/mantine-hooks/src/use-focus-trap/create-aria-hider.ts
@@ -1,3 +1,5 @@
+import { randomId } from '../utils';
+
 type Value = {
   node: HTMLElement;
   ariaHidden: string;
@@ -7,6 +9,8 @@ export function createAriaHider(
   containerNode: HTMLElement,
   selector: string = 'body > :not(script)'
 ) {
+  const id = randomId();
+
   const rootNodes: Value[] = Array.from<HTMLElement>(document.querySelectorAll(selector)).map(
     (node) => {
       if (node?.shadowRoot?.contains(containerNode) || node.contains(containerNode)) {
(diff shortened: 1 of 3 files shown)


Ben-Kincaid:
Ugh... Even though this should be totally safe in this context, need to figure out a way to suppress these warnings in the tests. 

image

These tests must use the document body as the container element since `createAriaHider` uses `body > :not(script)` for getting the root nodes. Appending via portal also appears to throw the same warnings. 

Looks like i can reassign `global.console` for each of the tests to avoid output, but that doesn't feel safe.

rtivital:
https://github.com/mantinedev/mantine/blob/master/src/mantine-tests/src/patch-console-error.ts#L13

Ben-Kincaid:
Thanks @rtivital! Just updated with the fix

rtivital:
Thanks!

PR #4119, opened 2023-04-22
[@mantine/core] Input: Add paddingRight to unstyled variant when righ…

dbnar2:
…tSection is present

Add paddingRight to the unstyled variant only when withRightSection is true otherwise returns null styles.

Prevents text from overlapping rightSection on an unstyled variant of TextInput.

Example:
image


Ben-Kincaid:
@dbnar2 Looks like prettier check is failing on `Input.styles.ts`. Try `yarn prettier:write`.

rtivital:
Thanks!

PR #4126, opened 2023-04-23
[@mantine/dates] Fixed passing nextIcon and previousIcon props

IvanKalinin:
These properties must not be included in `...others` and should be passed to MonthLevelGroup.


rtivital:
These props should also be passed to year level group and decade level group

IvanKalinin (commented):
Added next/previous icon in other calendar levels

rtivital:
Thanks!

PR #4127, opened 2023-04-24
[@mantine/hooks] Support dynamic key for `useLocalStorage`

### LINKED ISSUE #3987: `useLocalStorage` does not update its state value on `key` change
### What package has an issue

@mantine/hooks

### Describe the bug

If a `useLocalStorage` hook's `key` value changes, the `readStorageValue` and `setStorageValue` functions are redefined, but the underlying `value` state is not reset. This causes an incorrect value to be returned from the hook if the `key` changes after initial mount.

6.0.4

https://codesandbox.io/s/prod-star-93xppn

None

Yes

### Possible fix

I was able to fix this by adding `useEffect(() => setValue(readStorageValue()), [key])` to the `useStorage` function. I'm unsure if this causes any issues with SSR as I've only tested on a client side application.

Ben-Kincaid:
Fixes #3987 
---
- Adds `key` to the dependency array for the `useEffect` responsible for setting the initial local storage value to support a dynamic `key`. Tested with both client-side and server-side(Remix) rendering. 
- Adds general tests for `useLocalStorage`, including the above use case

### DIFF
--- a/src/mantine-hooks/src/use-local-storage/create-storage.ts
+++ b/src/mantine-hooks/src/use-local-storage/create-storage.ts
@@ -113,7 +113,7 @@ export function createStorage<T>(type: StorageType, hookName: string) {
       if (getInitialValueInEffect) {
         setValue(readStorageValue());
       }
-    }, []);
+    }, [key]);
 
     return [
       value === undefined ? defaultValue : value,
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #4130, opened 2023-04-25
[docs] Clarify docs for SegmentedControl

filiptammergard:
Since you can control the orientation of `<SegmentedControl>` with the `orientation` prop, it doesn't make sense to say to describe it as horizontal in the description.

Also improve some other documentation.


rtivital:
Thanks!

Issue #4141, opened 2023-04-26
Accessibility problem in Tab component - @mantine/core


forestileao:
**Path: src/mantine-core/src/Tabs/Tab.tsx**

Inside Tab Component, exists an attribute called "aria-controls". This attribute has the id of the panel that it controls.
The problem is: When Tab is used without a <Tab.Panel/> the aria-controls still is set, but it points to nothing. So, it causes an invalid [aria-*] attribute. Is is showed in accessibility tests such as https://pagespeed.web.dev/.

A solution is to check if the panel id exists, if it's true, add "aria-controls" attribute, else, not.

PR #4142, opened 2023-04-26
[@mantine/core] Tab: set to add aria-controls with panelId conditiona…

forestileao:
…lly to the panel existance

Issue: https://github.com/mantinedev/mantine/issues/4141

Added verification inside Tab.tsx to check if the pointing Panel exists in the DOM. Else it sets to undefined.

### DIFF
--- a/src/mantine-core/src/Tabs/Tab/Tab.tsx
+++ b/src/mantine-core/src/Tabs/Tab/Tab.tsx
@@ -54,6 +54,8 @@ export const Tab = forwardRef<HTMLButtonElement, TabProps>((props, ref) => {
   );
 
   const isActive = value === ctx.value;
+  const panelId = ctx.getPanelId(value);
+  const ariaControlsValue = document.getElementById(panelId) ? panelId : undefined;
   const activateTab = (event: React.MouseEvent<HTMLButtonElement, MouseEvent>) => {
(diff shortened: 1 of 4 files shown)


src/mantine-core/src/Tabs/Tab/Tab.tsx
```diff
@@ -54,6 +54,8 @@ export const Tab = forwardRef<HTMLButtonElement, TabProps>((props, ref) => {
   );
 
   const isActive = value === ctx.value;
+  const panelId = ctx.getPanelId(value);
+  const ariaControlsValue = document.getElementById(panelId) ? panelId : undefined;
```
rtivital, line 58: This will throw an error during server side rendering.

rtivital: You can use ModalBase component as reference, it has similar logic

forestileao: I will take a look, thanks

forestileao: @rtivital I added a mounted panel list to its provider, so document.getElementById is not necessary now. Based on your recommendation of ModalBase

forestileao: Fixed eslint errors after test.

rtivital:
Thanks!

Issue #4146, opened 2023-04-26
Popover - does not dismiss with `closeOnClickOutside`


pting-me:
### What package has an issue

@mantine/core

### Describe the bug

Popover does not dismiss when clicking outside, even when `closeOnClickOutside` is set to `true`. It's already set to true by default.

This is also visible in the [documentation](https://mantine.dev/core/popover/). Issue does not exist in [v5](https://v5.mantine.dev/core/popover/).

I have also provided a `codesandbox` link that showcases the difference between `Popover`, and `useClickOutside` behavior by itself.

### What version of @mantine/hooks page do you have in package.json?

6.0.9

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/p/sandbox/mantine-react-s1xz3n

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

xshuxin:
@pting-me @rtivital   
I'm sorry, I got the wrong value, which caused this problem.
I have already raised PR to fix him. #4148

rtivital:
Fixed in 6.0.10

PR #4148, opened 2023-04-27
[@mantine/core]:Fix clickOutside trigger condition

xshuxin:
Fix `clickOutside` trigger condition.
For details：#4114 #4146


rtivital:
Thanks!

PR #4156, opened 2023-04-28
[@mantine/dates]: Support function as weekdayFormat prop

IvanKalinin:
Extended `weekdayFormat` property to support function, similar to `yeaLabelFormat`. It allows us to either render a custom node or implement a format that does not support by dayJS library, like in our case, where the labels should be just the first letter of the weekday name.


rtivital:
Thanks!

PR #4168, opened 2023-04-30
[@mantine/core] fix multi-select hovered item on last item select

jvdsande:
When `disableSelectedItemFiltering` is set to true, selecting the last item will move the focus to the second-to-last instead of keeping it on the last.

This is due to a workaround for selected-item detection needed in normal mode.

This PR changes the workaround: instead of handling it as an edge-case in `handleItemSelect`, I converted it into a rule in the render: `hovered` should never be greater than the last item's index.

This way if the item number does not change on select, the hovered value is not impacted.


rtivital:
Thanks!

PR #4180, opened 2023-05-03
[@mantine/dates] Fixed DatePicker types

IvanKalinin:
Fixed DatePicker props types by adding missing `CalendarSettings` which includes `CalendarHeaderProps`.
The component supports those props (such as `next/previousIcon`) and works correctly though. Just TS errors


rtivital:
Thanks!

PR #4183, opened 2023-05-03
[@mantine/core] ButtonGroup: Use :first-child and :last-child in css selector instead of :first-of-type and :last-of-type

### LINKED ISSUE #4182: Button.Group does not group buttons correctly when using mixed polymorphic buttons
### What package has an issue

@mantine/core

### Describe the bug

Buttons in ButtonGroup don't group together when using Button with different components, e.g. button, a, React-Router Link.

6.0.10

https://codesandbox.io/s/amazing-flower-dmm1nl?file=%2Fsrc%2FApp.tsx

Yes

Yes

### Possible fix

Use :first-child selector and :last-child selector instead of :first-of-type and :last-of-type in ButtonGroup styles.
Bad side effect would be that other non-button elements in ButtonGroup can have an effect on the button grouping.

jvllmr:
Pull request for the issue and its proposed fix in #4182

This allows buttons polymorphic buttons with different components to group. E.g. a button with normal button element can group with a button with an a element.

closes #4182

### DIFF
--- a/src/mantine-core/src/Button/ButtonGroup/ButtonGroup.styles.ts
+++ b/src/mantine-core/src/Button/ButtonGroup/ButtonGroup.styles.ts
@@ -12,7 +12,7 @@ export default createStyles(
       flexDirection: orientation === 'vertical' ? 'column' : 'row',
 
       '& [data-button]': {
-        '&:first-of-type:not(:last-of-type)': {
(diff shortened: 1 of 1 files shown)


jvllmr:
I see, so I will have to come up with another solution...

Issue #4227, opened 2023-05-11
EyeDropper button on ColorInput missing discernible text for Accessibility purposes


angelo-loria:
### What package has an issue

@mantine/core

### Describe the bug

The EyeDropper on the ColorInput component is rendered without any label, title, etc., which gets flagged as a critical issue in Deque's Axe scanner. When using a screenreader, there's no description of what the button is.

You can see the issue on the docs page for ColorInput- https://mantine.dev/core/color-input/#eye-dropper

Here is a screenshot of the error output from [Deque's Chrome extension](https://chrome.google.com/webstore/detail/axe-devtools-web-accessib/lhdoppojpmngadmnindnejefpokejbdd):
![image](https://github.com/mantinedev/mantine/assets/24941027/92053737-10e0-4dbd-8e3f-6d31a57b9775)

### What version of @mantine/hooks page do you have in package.json?

6.0.9

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

Add `aria-label` attribute to EyeDropper button (I think, I'm new to this)

rtivital:
Fixed in 6.0.11

PR #4230, opened 2023-05-11
[@mantine-core] ColorInput: Add aria-label to EyeDropper 

angelo-loria:
https://github.com/mantinedev/mantine/issues/4227


rtivital:
Thanks for a PR, it is not a correct fix, 6.0.11 will introduce `eyeDropperLabel` prop which can be used to set aria-label on the button.

Issue #4245, opened 2023-05-13
Can't add custom color in Rating Component


adarshaacharya:
### What package has an issue

@mantine/core

### Describe the bug

In rating component when I pass colors defined in docs then it works fine but if i want custom color it is fallbacking to custom color.
https://v5.mantine.dev/core/rating/#usage

Is it intended or bug?

### What version of @mantine/hooks page do you have in package.json?

^5.10.3 (latest version of mantine also have same issue)

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/new-mountain-bo97l4?file=/src/App.tsx

### Do you know how to fix the issue

None

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

rtivital:
Fixed in 6.0.11

PR #4251, opened 2023-05-15
[@mantine/core] Rating: fix add custom color

lawff:
I see the props color desc: Key of theme.colors or any CSS color value, yellow by default
so to fix to get the custom color #4245


rtivital:
Thanks!

PR #4253, opened 2023-05-15
Update MantineProvider.tsx

Inuniku:
Added an export for `MantineProviderContext`.

Sometimes accessing the MantineProviderContext is needed to make theming Mantine components work in certain reconcilers. (e.g. https://github.com/pmndrs/drei#usecontextbridge)

[react-router](https://github.com/remix-run/react-router/blob/853a9382b748466f9bf1af5c4c7e5571670f8b97/packages/react-router/index.ts#L43) marks their context exports as `UNSAFE_`, so I decided to use this pattern here. 
That said, the react-router guys make sure in their source code how unsafe and breaking using this might be.

Is it still possible to provide this change for using mantine with custom reconcilers?


Inuniku:
Sorry, have to rework files

PR #4270, opened 2023-05-18
[core] NumberInput: Adding ability to compose NumberInput.tsx rightSection with the defau…

Aram1d:
## Why ?

i would like to be able to add a custom button on the left of the default right controls of Mantine number input, this can be useful in a lot of situation:

- recompute a value from other ones in a complex form with relation between each others.
- refresh a value
- add a randomize button
- For an hour input: switch between UTC and legal time.

## How ?

Simply change the actual API so that rightSection can take both `React.ReactNode` or a function `defaultControls => React.ReactNode` allowing you to compose whatever you want, including the default controls.

## Issues :

One tricky point is the handling of the rightSectionWidth. As its computed by a function by default, composing the rightSection require the developer to keep a constant width and calculate it properly.

### DIFF
--- a/docs/src/docs/core/NumberInput.mdx
+++ b/docs/src/docs/core/NumberInput.mdx
@@ -101,6 +101,12 @@ You can use it to create custom controls:
 
 <Demo data={NumberInputDemos.handlers} />
 
+## Custom RightSection composing with default controls
+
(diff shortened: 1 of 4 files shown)


rtivital:

PR #4273, opened 2023-05-18
[@mantine/dates]: `DatePickerInput`: pass `nextIcon`, `previousIcon` props to `Calendar`

andremonteiro95:
This PR allows setting the properties `nextIcon` and `previousIcon` in `DatePickerInput`, as these are currently being ignored in the function `pickCalendarProps`.


rtivital:
Thanks!

PR #4343, opened 2023-06-01
[@mantine/modals]: hinting for mantine context modals

mpsenicka:
When using modals.openContextModal({ modal: 'something' }), something doesn't get hinted even when overrides are provided through MantineModalsOverride.

This PR fixes it, so when user defines his overrides, his own modalMap keys are being hinted, instead of a plain "string"


rtivital:

PR #4373, opened 2023-06-07
[@mantine/tiptap] LinkControl: add initialExternal to props

skyt-a:
see  [this discussion](https://github.com/mantinedev/mantine/discussions/4370)


rtivital:
Thanks!

PR #4381, opened 2023-06-09
[docs] Add pnpm and refactor out PackageInstallation

alexsurelee:
Adds `pnpm` to the tab list of package managers via the `PackagesInstallation.tsx` component.
image

Have also refactored a `PackageInstallation` component out of `PackagesInstallation`, for use across the wider MDX docs files for any given package(s).

### DIFF
--- a/docs/src/components/PackageInstallation/PackageInstallation.tsx
+++ b/docs/src/components/PackageInstallation/PackageInstallation.tsx
@@ -1,0 +1,48 @@
+import React from 'react';
+import { rem } from '@mantine/core';
+import { Prism } from '@mantine/prism';
+import { NpmIcon, PnpmIcon, YarnIcon } from '@mantine/ds';
+
+const packageManagerPrefixes = new Map([
+  ['yarn', 'yarn add'],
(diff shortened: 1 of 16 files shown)


rtivital:
https://github.com/mantinedev/mantine/pulls?q=is%3Apr+pnpm

rtivital:
You can look through other PRs to learn about the reasons pnmp is not added.

alexsurelee:
Fair enough - I do think `pnpm` is gaining popularity, particularly for monorepo tooling ([Turborepo recommends it](https://turbo.build/repo/docs/getting-started/create-new#which-package-manager-do-you-want-to-use)) due to its [non-flat `node_modules`](https://pnpm.io/motivation#creating-a-non-flat-node_modules-directory). 

This PR adds it in a way that removes the overhead of maintenance (beyond putting it into the one component), and also kind of pretties up the other docs pages by using the existing component for the `npm` and `yarn` options anyway (as in the screenshot).

I'd argue that this would just be a nice docs enhancement even without the `pnpm` option added, but to each their own.

PR #4391, opened 2023-06-11
[@mantine/core] Table: Avoid style bleed

apoco:
Ensure table stylings apply to immediate children only. This avoids things like the calendar pop-up nested inside of a `<Table/>` inheriting the styles of the parent.


apoco:
Oops, made pull request against wrong branch.

PR #4393, opened 2023-06-11
[@mantine/core] Table: Avoid style bleed

apoco:
Ensure table stylings apply to immediate children only. This avoids things like the calendar pop-up nested inside of a `<Table/>` inheriting the styles of the parent.

### DIFF
--- a/src/mantine-core/src/Table/Table.styles.ts
+++ b/src/mantine-core/src/Table/Table.styles.ts
@@ -35,14 +35,14 @@ export default createStyles(
         lineHeight: theme.lineHeight,
         border: withBorder ? border : undefined,
 
-        '& caption': {
+        '& > caption': {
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

KevinVandy:
I think this PR started causing an issue https://github.com/mantinedev/mantine/issues/4433

rtivital:
@apoco Can you fix this? Otherwise, the PR will be reverted in the next patch.

jpage-godaddy:
I found the bug. Will send a fix 😮‍💨

jpage-godaddy:
In case it'd be faster, basically this selector:

```
& > thead > th, & > tbody > td
```

...should have been:

```
& > thead > tr > th, & > tbody > tr > td
```

apoco:
Pull request to fix this in #4443; sorry about that, and thanks again for a great library.

PR #4410, opened 2023-06-15
[@mantine/core] Image: Fix image alt overflow

### LINKED ISSUE #3883: @mantine/core - Image - Long alt text appears behind placeholder
### What package has an issue

@mantine/core

### Describe the bug

(On Firefox) If an Image has alt text that is larger than the placeholder, it will "break out" from behind the placeholder image.

image

NOTE: This only seems to happen in the Firefox browser and not in the Chrome browser. Some sort of different handling with the img tag alt text?

6.0.1

https://codesandbox.io/s/kind-sanderson-jykbdz

No

Yes

### Possible fix

One solution is to use the logic that Mantine uses to determine whether to show or hide the placeholder to remove the `` tag from the DOM. But not sure if this is the desired solution since this issue only seems to affect Firefox

comphonia:
closes #3883 

image

### DIFF
--- a/src/mantine-core/src/Image/Image.test.tsx
+++ b/src/mantine-core/src/Image/Image.test.tsx
@@ -58,6 +58,25 @@ describe('@mantine/core/Image', () => {
     expect(container.querySelectorAll('.mantine-Image-placeholder')).toHaveLength(1);
   });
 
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #4423, opened 2023-06-19
Update use-window-event.ts: Fixed types

vadimkiryanov:
Changed the generic type in the function to display hints


rtivital:
I do not think that type is correct. Function must work with custom events which can be any string.

vadimkiryanov:
> I do not think that type is correct. Function must work with custom events which can be any string.

In that case, I can offer this suggestion. The developer has the ability to insert any string and still have access to the default values. What do you think about it?

`useWindowEvent<K extends string = keyof WindowEventMap>`

rtivital:
Thanks!

Issue #4432, opened 2023-06-21
PinInput delete does not shift focus


arieldotjs:
### What package has an issue

@mantine/core

### Describe the bug

When pressing delete key on PinInput, the focus isn't shifted hence requiring you to press "delete" twice in order to delete more characters in order.

https://github.com/mantinedev/mantine/assets/110123783/f5f24767-8acb-4357-aca6-83660571d006

### What version of @mantine/hooks page do you have in package.json?

6.0.14

### Do you know how to fix the issue

None

### Are you willing to participate in fixing this issue and create a pull request with the fix

None

PR #4438, opened 2023-06-21
[@mantine/core] PinInput: shift focus to prev element on backspace

mrbnclt:
This fixes https://github.com/mantinedev/mantine/issues/4432

### Scenarios:
- If the focus is on the last element and there is a value, hitting the backspace will remove the value and stay focused on that element.
- If the focus is on the last element and there is no value, hitting the backspace will shift the focus to the previous element.
- If the focus is on any input element except the last element, hitting the backspace will remove the value and shift the focus to the previous element. 

https://github.com/mantinedev/mantine/assets/45745778/eb125f31-00f0-40e8-ae09-d4861ced90d2


arieldotjs (approved):

rtivital:
Thanks!

PR #4451, opened 2023-06-22
Update use-clipboard.ts/Copying error on some devices

vadimkiryanov:
On some devices it is not possible to copy the value with navigator.clipboard.writeText(), for example on Apple devices. Sometimes it is not possible to copy a value while working in coding sandboxes.

I propose a solution that increases the probability that the value will be copied. To do this, I've added double error handling, just in case navigator.clipboard.writeText() fails, then the code will automatically run the second option to copy the value.


rtivital:

Issue #4429, opened 2023-06-20
copy button not rendering properly when using `Prism` component inside `TypographyStylesProvider`


devak997:
### What package has an issue

@mantine/core

### Describe the bug

I'm currently encountering an issue where the alignment of the copy button is incorrect when using the `Prism` component from `@mantine/prism` within the `TypographyStylesProvider`. I am rendering a markup using react-remark and have implemented a custom component that utilizes the Prism component for code rendering. However, the copy button is not properly aligned within the TypographyStylesProvider.

![image](https://github.com/mantinedev/mantine/assets/47378797/6df3cd9b-eba1-4aa2-8bba-ed106bad84ed)

### What version of @mantine/hooks page do you have in package.json?

6.0.14

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/boring-rgb-6lxz9m?file=/src/App.tsx

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

PR #4452, opened 2023-06-22
[@mantine/prism] Prism: use higher specificity for pre element margin

mrbnclt:
This fixes https://github.com/mantinedev/mantine/issues/4429

Wrapping Prism within a `TypographyStylesProvider` that has a selector with the margin values:
```
<class> pre: {
  marginTop: theme.spacing.md,
  marginBottom: theme.spacing.md
}
```
has a higher specificity than Prism's selector with the margin set to 0
```
code: {
  .......
  marginTop: 0,
  marginBottom: 0
}
```
this causes the margin styles from `TypographyStylesProvider` to be utilized by Prism's `pre` element and causes the copy button to move higher than it's normal position.


rtivital:
Thanks!

PR #4465, opened 2023-06-27
[@mantine/dates] DateTimePicker: Fix for user-provided onClick to clo…

### LINKED ISSUE #4464: DateTimePicker submitButtonProps onClick is overridden
### What package has an issue

@mantine/dates

### Describe the bug

I would like some custom behavior to be executed when the user clicks the Submit button on the `DateTimePicker` component.

It appears the that spread operation defined [here](https://github.com/mantinedev/mantine/blob/master/src/mantine-dates/src/components/DateTimePicker/DateTimePicker.tsx#L229) may override the `onClick` prop defined earlier [here](https://github.com/mantinedev/mantine/blob/master/src/mantine-dates/src/components/DateTimePicker/DateTimePicker.tsx#L220)

This causes an issue since the `onClick` function provided to the `ActionIcon` component is supposed to execute my custom `onClick` function attached  `submitButtonProps`'s and then close the associated Popover (or Modal). Instead, that spread operation takes my `onClick` function and uses that in place of the one written.

6.0.14

https://codesandbox.io/s/determined-sinoussi-9f79f5?file=/src/App.tsx

Yes

Yes

### Possible fix

Ensure that the spread operation on the submit button's props doesn't override the `onClick` handler defined earlier in the component.

wojonet:
Fixes #4464

### DIFF
--- a/src/mantine-dates/src/components/DateTimePicker/DateTimePicker.test.tsx
+++ b/src/mantine-dates/src/components/DateTimePicker/DateTimePicker.test.tsx
@@ -119,6 +119,26 @@ describe('@mantine/dates/DateTimePicker', () => {
     expectNoPopover(container);
   });
 
+  it('closes when a submit button onClick handler is provided', async () => {
+    const spy = jest.fn();
+    const { container } = render(
+      <DateTimePicker
+        {...defaultProps}
+        submitButtonProps={{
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #4469, opened 2023-06-27
[@mantine/dates] PickerInputBase: Pass input error/invalid state to Input.Placeholder

### LINKED ISSUE #4467: DatePickerInput placeholder color does not reflect error state when there is an error
### What package has an issue

@mantine/dates

### Describe the bug

The placeholder color of the DatePickerInput component does not change to red to reflect the error state when there is an error. Currently, when an error is present in the DatePickerInput, the placeholder color remains unchanged, unlike other input components which turn red.

6.0.14

https://codesandbox.io/s/mantine-dateinputpicker-placeholder-wrong-color-m56w38

Yes

Yes

### Possible fix

The error occurred due to a difference in behavior between the `DatePickerInput` component and the standard `Input` component. The `DatePickerInput` uses the `Input.Placeholder` component to implement the placeholder functionality, which requires customization to match the behavior of the `Input` component. In order to address this issue, it may be necessary to pass the error state to the `Input.Placeholder` component for consistent behavior.

alessandroaw:
This fixes #4467 by passing `inputProps.error` to `Input.Placeholder` to match the placeholder color.

### DIFF
--- a/src/mantine-dates/src/components/PickerInputBase/PickerInputBase.tsx
+++ b/src/mantine-dates/src/components/PickerInputBase/PickerInputBase.tsx
@@ -179,7 +179,12 @@ export const PickerInputBase = forwardRef<HTMLButtonElement, PickerInputBaseProp
               {...others}
             >
               {formattedValue || (
-                <Input.Placeholder className={classes.placeholder}>{placeholder}</Input.Placeholder>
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #4485, opened 2023-06-29
[@mantine/core] MultiSelect: handle portalProps

### LINKED ISSUE #4484: MutliSelect is not using portalProps prop
### What package has an issue

@mantine/core

### Describe the bug

MultiSelectProps does include a prop called _portalProps_ to pass down props to the SelectPopover.
This prop is not correctly destructured in the MultiSelect component and passed down to the input element which leads to an TS error in the console:
`React does not recognize the `portalProps` prop on a DOM element.`

Why do I need this?
It works perfectly fine in the Select component and should work the same in MultiSelect as they share the same interface and Dropdown component. My use case was to detect an outside click as described here https://mantine.dev/hooks/use-click-outside/#multiple-nodes 

Step to reproduce this with the sandbox
- Change Select to MultiSelect and see the error in the console. Also when using the useClickOutside hook the ref does not work for the dropdown

6.0.8

https://codesandbox.io/s/admiring-bose-c6wjjm?file=/src/App.tsx

Yes

Yes

### Possible fix

Destructure the _portalProps_ correctly in MultiSelect.tsx and pass it to SelectPopover like seen in the Select component.
So that is is passed down correctly to the Portal and the console disappears as it is not included in the _...rest_.
I am willing to create a PR for that.

Cuzart:
Fixed #4484 similar to Select component

### DIFF
--- a/src/mantine-core/src/MultiSelect/MultiSelect.tsx
+++ b/src/mantine-core/src/MultiSelect/MultiSelect.tsx
@@ -200,6 +200,7 @@ export const MultiSelect = forwardRef<HTMLInputElement, MultiSelectProps>((props
     onDropdownOpen,
     maxSelectedValues,
     withinPortal,
+    portalProps,
     switchDirectionOnFlip,
     zIndex,
     selectOnBlur,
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

Issue #4481, opened 2023-06-28
DateInput component with "clearable" does not clear previously selected value in UI after clicking "x" to clear until the field blurs


ipurra:
### What package has an issue

@mantine/dates

### Describe the bug

Using the `DateInput` field use the following steps to reproduce my issue:

1. Add a `DateInput` field with the `clearable` property set to `true` to your page (see sandbox for code example)
2. Navigate to the date input field in the UI
3. Click in the field, select any date, the picker closes
4. Click in the field again to open the date picker
5. When the date picker is open and there is a value in the field click the `x` button inside the input field to clear the field
6. Notice that the actual value now is set to `null` but the previously selected value is still visible in the input field until the field is blurred

### What version of @mantine/hooks page do you have in package.json?

6.0.15 (sandbox 6.0.8, same issue)

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/reverent-night-y87t8x?file=/src/App.tsx

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

None

mrbnclt:
I'd like to work on the fix for this issue

PR #4486, opened 2023-06-29
[@mantine/dates] DateInput: clear input value on clearButton click

mrbnclt:
This fixes https://github.com/mantinedev/mantine/issues/4481.

When `Calendar Dropdown` is open then the user clicks on the `clear` button, the following code won't run causing the input value to not be cleared. 

```
useDidUpdate(() => {
    value !== undefined && !dropdownOpened && setInputValue(formatValue(value));
  }, [value]);
```

So when user clicks on the `clear` button we also closes the `Calendar Dropdown` which fixes the issue. 

As other `Date` components, clicking the `clear` button closes the `Calendar Dropdown`, this should also make the behavior consistent across components.


rtivital:
Thanks!

PR #4497, opened 2023-07-03
[@mantine/core] Slider: Fix onChange prop updates being ignored

Bastian:
When updating the "onChange" property of a slider, the change was ignored and the old property was still used. This was caused by not adding the "setValue" to the dependency array.


JadeJDoucet:
This would be such a nice fix to have. Thanks @Bastian 🙌

rassie:
@rtivital would you port this to v7?

rtivital:
Yes

rassie:
Thanks, that's great!

Issue #4500, opened 2023-07-05
TypeError: Cannot assign to read only property 'area' of object '#<Object>'


vaynevayne:
### What package has an issue

@mantine/core

### Describe the bug

<img width="1112" alt="image" src="https://github.com/mantinedev/mantine/assets/124666577/0f8e7f5a-a023-461d-954d-0544952235d6">

Can you see the bug?

maybe js disables modifying const objects
```jsx
const cloned = klona(values); //  Can we solve this problem by replacing it with let?
 cloned[splittedPath[0]] = value;
```

### What version of @mantine/hooks page do you have in package.json?

latest

### Do you know how to fix the issue

None

### Are you willing to participate in fixing this issue and create a pull request with the fix

None

rtivital:
Please send a sandbox that reproduces the issue, template –https://codesandbox.io/s/mantine-template-9ze89?file=/src/App.tsx:0-6

ChiFangChen:
do you use something like redux to deal with your initial value?
maybe you can try to destructure the object by this:
```
    const form = useForm({
        initialValues: { ...valueFromReduxOrImmutableValue },
    });
```

vaynevayne:
no, i don't use any lib, just do this
```jsx
 const form = useForm({
        initialValues: { ...props.form },
    });
```

rtivital:
This issue is closed for one of these reasons:

- It was fixed in 7.0
- It is no longer reproducible in 7.0
- It is no longer applicable in 7.0
- It does not have a reproduction

If you think that this issue was closed by mistake, and it is still an issue in version 7.0, please reopen it.

PR #4501, opened 2023-07-05
fix(Form): fixed  https://github.com/mantinedev/mantine/issues/4500

vaynevayne:


vaynevayne:
@rtivital  hello, What's up?

rtivital:
You've submited a PR with incorrect code.

vaynevayne:
Have you looked at this question carefully?

I made a mistake with const, but ts won't pass with let

You want to deal with it? Do you have any good ideas? @rtivital

PR #4512, opened 2023-07-06
[@mantine/core] Spoiler: Fix control button flickering on rerender

KurtGokhan:
When the Spoiler component is initially rendered, it doesn't know its content's size, so it doesn't show the "Show more" button. This was causing flickering and layout shifts.

https://github.com/mantinedev/mantine/assets/6034931/f253d153-4559-4a28-b7dc-d4334f9d5621

It is possible to know the initial size by implementing `useResizeObserver` hook slightly different. That is what I did.

Some remarks:

- Removed `useEffect` from the Spoiler component which was used for a computed property
- Having `ref.current` in the dependency array of `useEffect` is quite unreliable.
- `useResizeObserver` now returns a callback ref instead of mutable ref. Normally, this is an implementation detail and should not count as a breaking change.
- I had to modify some code in the `SegmentedControl` because it was expecting `useResizeObserver` to return a mutable ref.

Tested with the following code. If you click the rerender button quickly, you will notice the flickering in the old implementation.

```tsx
export function RerenderTest() {
  const [count, setCount] = useState(1);
  const incrementCount = () => setCount((x) => x + 1);

  const content = Array(5)
    .fill(0)
    .map((_, index) => (
      <p key={index}>
        Lorem ipsum dolor sit amet consectetur adipisicing elit. Autem officiis, incidunt libero,
        itaque, quaerat labore quis odio culpa tempore quisquam porro unde omnis tempora nostrum
        nihil eligendi distinctio. Animi, consectetur! des
      </p>
    ));

  return (
    <div style={{ padding: 40 }}>
      <button type="button" onClick={incrementCount}>
        Rerender
      </button>

      <Spoiler maxHeight={120} showLabel="Show more" hideLabel="Hide" key={count}>
        {content}
      </Spoiler>
    </div>
  );
}
```

### DIFF
--- a/src/mantine-core/src/SegmentedControl/SegmentedControl.tsx
+++ b/src/mantine-core/src/SegmentedControl/SegmentedControl.tsx
@@ -148,6 +148,7 @@ export const SegmentedControl = forwardRef<HTMLDivElement, SegmentedControlProps
   const uuid = useId(name);
   const refs = useRef<Record<string, HTMLLabelElement>>({});
   const [observerRef, containerRect] = useResizeObserver();
+  const rootRef = useRef<HTMLDivElement>(null);
 
   useIsomorphicEffect(() => {
     if (!mounted.current) {
@@ -159,7 +160,7 @@ export const SegmentedControl = forwardRef<HTMLDivElement, SegmentedControlProps
   });
 
   useEffect(() => {
-    if (_value in refs.current && observerRef.current) {
+    if (_value in refs.current && rootRef.current) {
       const element = refs.current[_value];
       const elementRect = element.getBoundingClientRect();
       const scaledValue = element.offsetWidth / elementRect.width;
@@ -211,7 +212,7 @@ export const SegmentedControl = forwardRef<HTMLDivElement, SegmentedControlProps
     </div>
   ));
 
-  const mergedRef = useMergedRef(observerRef, ref);
+  const mergedRef = useMergedRef(observerRef, ref, rootRef);
 
   if (data.length === 0) {
     return null;
--- a/src/mantine-core/src/Spoiler/Spoiler.tsx
+++ b/src/mantine-core/src/Spoiler/Spoiler.tsx
@@ -1,4 +1,4 @@
-import React, { useState, useEffect, forwardRef } from 'react';
+import React, { useState, forwardRef } from 'react';
 import { DefaultProps, Selectors, useComponentDefaultProps, rem } from '@mantine/styles';
 import { useElementSize } from '@mantine/hooks';
 import { Anchor } from '../Anchor';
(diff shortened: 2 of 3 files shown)


rtivital:
This will make both component and use-resize-observer incompatible with ssr – there will be hydration mismatch

KurtGokhan:
> This will make both component and use-resize-observer incompatible with ssr – there will be hydration mismatch

I think there won't be a hydration mismatch, as the initial size for both client and server will still be 0. The real size is assigned in ref callback (essentially at the same phase as a `useLayoutEffect`). 

But I have to try this. I couldn't set up an actually working Mantine dev workspace yet. Will tell you when I test it.

KurtGokhan:
Tried this in Next.js (v13 but pages dir). It doesn't cause hydration errors. As I said, the initial size is `0`, so it doesn't cause hydration errors.

The "Show more" button doesn't exist in the HTML sent by the server.

image

rtivital:
Thanks!

mehimanshupatil:
@rtivital 
This change should have been released as major instead of patch. because previously the ref was accessible from outside the hook now its not.

```js
 const { ref, height } = useElementSize();

  useEffect(() => {
  // version < 6.0.18 logs element. version 6.0.18 logs undefined
    console.log("element",  ref.current);
  }, [height]);
```

KurtGokhan:
@mehimanshupatil  sorry for that. You can use `useMergedRef` as a workaround for now. The type seems correct to me though, so Typescript should warn you about this.

@rtivital  I could introduce a hook which handles both ref callbacks and mutable refs in the same variable. What do you think?

mehimanshupatil:
@KurtGokhan 
Yes, we can have a workaround, but the current changes are breaking the existing functionality. According to semantic versioning, breaking changes should be released as a major version. 
Regarding the types change, I mean the documentation should be updated, and typeScript will warn about the property 'current' not existing in the ref variable. 
In my case, I had gitignored the package-lock.json file, and after deploying via CI/CD, the website was breaking. It took me around 2+ hours to pinpoint the problem.

KurtGokhan:
@mehimanshupatil yes, I wouldn't want to break semver.

@rtivital  the bug is in `useResizeObserver` though. `Spoiler` uses that but I also use that hook in other places. Ideally, the hook should be fixed. `Spoiler` can't be fixed without that.

I am going to revert the code of `useResizeObserver`, but use `useIsomorphicEffect` instead of `useEffect`. That should resolve the issue for now. I hope that is ok.

icflorescu:
This is a breaking change that should have been released as a major.
The signature of `useElementSize` changed in an unpredictable manner and because of that, `v6.0.18` introduces bugs like this one: https://github.com/icflorescu/mantine-datatable/issues/398

rtivital:
This will be fixed in next patch

icflorescu:
@rtivital Thanks! People were getting anxious about this...

icflorescu:
Thanks a lot!

Issue #4516, opened 2023-07-07
Mobile web modal height display bug


connershoop:
PR: #4517 

### What package has an issue

@mantine/core

### Describe the bug

As seen in the attached screenshot taken from the mantine website, on mobile web browsers the mantine modal does not take into account the height lost to the mobile web header, which causes the modal to display off the bottom of the screen.

![Screenshot_20230707-090549](https://github.com/mantinedev/mantine/assets/31523466/c571d021-27ac-41ed-bf85-545ff03dd57b)

### What version of @mantine/hooks page do you have in package.json?

6.0.16

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

`vh` -> `dvh` when `dvh` is adequately supported

https://caniuse.com/viewport-unit-variants

PR: #4517

PR #4517, opened 2023-07-07
[@mantine/core] Modal: Fix mobile web modal height display bug. vh ->…

connershoop:
#4516 

Fix for mobile web modal display bug by replacing `vh` unit with newer `dvh` unit.  

(Can redo with branch off of dev if needed)

### DIFF
--- a/src/mantine-core/src/Modal/ModalContent/ModalContent.tsx
+++ b/src/mantine-core/src/Modal/ModalContent/ModalContent.tsx
@@ -25,7 +25,7 @@ export const ModalContent = forwardRef<HTMLElement, ModalContentProps>((props, r
 
   return (
     <ModalBase.Content ref={ref} radius={ctx.radius} {...others}>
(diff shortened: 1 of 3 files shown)


rtivital:
Have you checked on a mobile device that the fix works?

connershoop:
> Have you checked on a mobile device that the fix works?

I have only manually overwritten the modal styles in my own project.  I will try and test manually in this repo when I get a chance.
It is important to note that this `dvh` unit is not as universally supported as `vh`, which could cause problems.
https://caniuse.com/viewport-unit-variants

hulloitskai:
Our team ran into this too, and I can confirm that the `dvh` units fix it on iOS mobile safari.

Our current solution:

```ts
export const APP_THEME: MantineThemeOverride = {
  # ...snip...
  components: {
    Modal: {
      content: {
        maxHeight: `calc(100dvh - (5dvh * 2))`,
      },
    }
  }
}
```

rtivital:
Thanks!

PR #4518, opened 2023-07-08
[@mantine/core] NumberInput: Fix precision parser behavior when the value is nullable

mlshvdv:


rtivital:
NumberInput value cannot be null, the library does not support it. It is explicitly defined in component props type – value can either be an empty string or number.

mlshvdv:
@rtivital Yeah, but sometimes it's difficult to control nullable values which passed to the mantine/form's initialValues. So, I think it will be better to have a fix for the precision parser, to avoid the error of the toFixed() method. 

```js
if (val === '' || value === null) return '';

let result = val.toFixed(precision);
```
instead of:

https://github.com/mantinedev/mantine/blob/d03cb692eedd7e525275bdedcc9b077b141f0b78/src/mantine-core/src/NumberInput/NumberInput.tsx#L175-L177

mlshvdv:
@rtivital  This fix helps to avoid this bug when using mantine forms:

Screenshot 2023-07-08 at 13 22 48

rtivital:

PR #4527, opened 2023-07-11
[@mantine/dates] DateInput: Separate clearable and allowDeselect logic

### LINKED ISSUE #4524: allowDeselect is ignored when clearable is enabled
### What package has an issue

@mantine/dates

### Describe the bug

When using a `DateInput` with the `clearable` prop set, the value of the `allowDeselect` prop is ignored.

While I can see why this may be intended behaviour, it would be nicer for the user to be able to control deselection using the `allowDeselect` prop even when `clearable` is set.

My use case is this:

* When typing a date in the date field, some users click the date in the date picker to confirm the date however this then deselects the date, which isn't a very nice UX
* The `clearable` prop already enables the ability to clear the date field by clicking the clear button or by deleting all the text from the text field, so being able to deselect by clicking a date isn't really necessary

My proposed solution doesn't change the default behaviour when `allowDeselect` is not explicitly given, but it does allow `allowDeselect` to have an effect when `clearable` is set. Is this acceptable?

6.0.16

_No response_

Yes

Yes

### Possible fix

The simplest fix would be to change:
```
const _allowDeselect = clearable || allowDeselect;
```
to something like
```
const _allowDeselect = allowDeselect !== undefined ? allowDeselect : clearable;
```

DanielGibbsNZ:
Fixes #4524.

The change is a bit larger than I intended but I believe it makes the `clearable` and `allowDeselect` props more understandable:

* `clearable` controls whether the DateInput can be cleared (via the text field or the date picker)
* `allowDeselect` controls whether clicking on the existing date in the date picker clears the field (only applicable if clearable is set), defaults to true if `clearable` is set

If this is too much of a change, left me know and I'll revert it back to my original idea of just allowing `allowDeselect` to be manually set. The downside with this approach is that it can result in some weirdly inconsistent behaviour where the value can be cleared by clicking the date when `allowDeselect` is set and `clearable` isn't set.

### DIFF
--- a/docs/src/docs/dates/DateInput.mdx
+++ b/docs/src/docs/dates/DateInput.mdx
@@ -42,12 +42,12 @@ and must return `Date` object:
 
 <Demo data={DateInputDemos.parser} />
 
-## Allow deselect
+## Allow clear
 
-Set `allowDeselect` prop to allow removing value from the input. Input will be cleared if
+Set `clearable` prop to allow removing value from the input. Input will be cleared if
 user selects the same date in dropdown or clears input value:
 
-<Demo data={DateInputDemos.deselect} />
+<Demo data={DateInputDemos.clearable} />
 
 ## Min and max date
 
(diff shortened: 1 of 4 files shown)


rtivital:
Thanks!

PR #4532, opened 2023-07-14
fix: date picker input default date

congyaqwq:
fix: props `defaultDate` in `DatePickerInput` not work.


rtivital:
Thanks!

ikapta:
Hi, @congyaqwq does you check when value no empty, the initial opened value should be selected date other than defaultDate. The defaultDate maybe  a low priority.

congyaqwq:
fixed with https://github.com/mantinedev/mantine/pull/4624.

> Hi, @congyaqwq does you check when value no empty, the initial opened value should be selected date other than defaultDate. The defaultDate maybe a low priority.

PR #4556, opened 2023-07-21
[@mantine/form] Add setInitialValues function, allow modify initialValues after useForm was called

Ryan2128:
## Description

When use a form to process two operations, such as create and update, initialValues can only be set once, after `useForm` has been called, there is no way to modify initialValues.

The reason why need to modify initialValues is when use a form to process two operations, such as create and update, initialValues can only be set once, after `useForm` has been called, there is no way to modify initialValues.

This leads to switching to the update operation after the form is initialized, and the initialValues of the update operation form cannot be set.

## `use-form`
- Add setInitialValues function, allow modify initialValues after useForm was called

## Work

- [x] Code Changed
- [x] Lint
- [x] Test
- [x] Docs

### DIFF
--- a/docs/src/docs/form/use-form.mdx
+++ b/docs/src/docs/form/use-form.mdx
@@ -77,6 +77,9 @@ form.setFieldValue('path', value);
 // Set value of nested field
 form.setFieldValue('user.firstName', 'Jane');
 
+// set initial values
+form.setInitialValues(values);
+
 // Resets form.values to initialValues,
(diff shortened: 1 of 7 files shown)


Ryan2128:
@rtivital Please have a look.😋

Ryan2128:
@rtivital Hello, I would like to ask is there anything to discuss about this PR? Or is there anything else to do about this PR?

rtivital:

Issue #4574, opened 2023-07-26
Pagination: clicking Link buttons still change the URL when they are disabled


HoHieuLuc:
### What package has an issue

@mantine/core

### Describe the bug

When using Pagination with links, clicking the previous, first, next and last button still change the URL when they are disabled because disabled attribute doesn't work on anchor element.

### What version of @mantine/hooks page do you have in package.json?

6.0.17

### If possible, please include a link to a codesandbox with the reproduced problem

https://mantine.dev/core/pagination/#controls-as-links

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

No

saadsiddiqui07:
Could you please provide a code example of your issue?

HoHieuLuc:
> Could you please provide a code example of your issue?

Go to https://mantine.dev/core/pagination/#controls-as-links and try clicking on previous or first button

ItaiAxelrad:
The Pagination element is missing `pointerEvents: 'none'` in its disabled style. I'd be more than happy to submit a PR

verheyenkoen:
@ItaiAxelrad Any specific reason you chose to add this fix to the `[data-disabled]` rule and not `:disabled` pseudo-class or `[disabled]` rule? Anyhow I don't see this working in Mantine 7 (with Next.js Link as the component and the `disabled` prop set: `<Button as={Link} href="url" disabled={true}>Link text</Button>`). Or is that another issue and is this one pagination specific?

ItaiAxelrad:
@verheyenkoen I was following the package convention [https://mantine.dev/styles/data-attributes/](https://mantine.dev/styles/data-attributes/). This was a pagination-specific issue but I'm happy to check it out if you open a new issue for it

PR #4578, opened 2023-07-26
[@mantine/core] Pagination: Add disabled styles

ItaiAxelrad:
Fixes issue #4574. As a side note, `cursor: 'not-allowed'` becomes obsolete with the presence of `pointerEvents: 'none'` but I left it in because that seems to be the convention for the package.


rtivital:
Thanks!

Issue #4580, opened 2023-07-27
Dropzone.Fullscreen stays open after first drop


esheliko:
### What package has an issue

@mantine/dropzone

### Describe the bug

This can be reproduced in the mantive.dev website! [here](https://mantine.dev/others/dropzone/#dropzonefullscreen-component)
(I'm using Chrome 115)

Steps to reproduce:
- Activate fullscreen dropzone
- Drag&Drop any file
- (if the file was an image, you need to re-activate the dropzone)
- Drag an image over the dropzone but cancel (i.e. press ESC or drag back outside of window)

You will see that the dropzone stays open and in fact renders the user stuck unless they drop a file.

### What version of @mantine/hooks page do you have in package.json?

6.0.17

### If possible, please include a link to a codesandbox with the reproduced problem

https://mantine.dev/others/dropzone/#dropzonefullscreen-component

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

The problem seems to be with the inner counter of drag events inside the Dropzone.Fullscreen component.
When you drop a file you don't get a 'dragleave' event so the counter stays at 1.
I think resetting the counter to 0 when the user drops a file could solve the problem.

rtivital:
The issue is resolved in [7.0.1](https://github.com/mantinedev/mantine/releases/tag/7.0.1)

PR #4581, opened 2023-07-27
[@mantine/dropzone] Fix Dropzone.Fullscreen not disappearing correctly after first file drop (#4580)

esheliko:
PR for issue #4580


rtivital:
Thanks for submitting a PR, the issue was resolved another way

Issue #4576, opened 2023-07-26
Cannot resize Avatar placeholderIcon due to hardcoded size


mattaningram:
### What package has an issue

@mantine/core

### Describe the bug

The `placeholderIcon` in the `Avatar` component has a hardcoded `rem(15)` size on the SVG even though Mantine itself tries to set a `70%` size on it that gets overridden by this hardcoded size.

Hardcoded size is here: https://github.com/mantinedev/mantine/blob/master/src/mantine-core/src/Avatar/AvatarPlaceholderIcon.tsx#L11

Failing override is here: https://github.com/mantinedev/mantine/blob/master/src/mantine-core/src/Avatar/Avatar.styles.ts#L125

<img width="494" alt="image" src="https://github.com/mantinedev/mantine/assets/955092/62cf1c18-2047-4b97-b7c0-f2c97d3a0769">

This makes it impossible to customize the `placeholderIcon` size. 

### What version of @mantine/hooks page do you have in package.json?

6.0.17

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

Remove the `rem(15)` hardcoded style on the `AvatarPlaceholderIcon` svg. It's unclear why it is there in the first place.

rtivital:
You are welcome to submit a PR with a fix

mattaningram:
Happy to do so, but was curious to get context on if there is a reason for the `rem(15)` that I might not be aware of (the PR with a fix would basically be just removing that style).

rtivital:
The purpose of the PR is to check whether `rem(15)` is required, if it is not breaking anything, then it can be removed

atumas-bananamilk:
After updating from `6.0.13` to `6.0.17` started getting this error with no other changes in code:

![Screenshot 2023-07-29 at 07 07 17](https://github.com/mantinedev/mantine/assets/22228431/81707feb-9335-4945-9021-4afc74a46112)

PR #4600, opened 2023-08-01
[@mantine/core] AvatarPlaceholderIcon: Remove width and height

mattaningram:
Removes hardcoded `rem(15)` on width and height of `AvatarPlaceholderIcon` to allow it to scale properly with `Avatar` size and override with custom styles.

Addresses https://github.com/mantinedev/mantine/issues/4576


anuragdaksh7 (approved):

rtivital:
Thanks!

PR #4624, opened 2023-08-07
fix: date input `defaultDate` displace `value`

congyaqwq:
fix https://github.com/mantinedev/mantine/pull/4532
Sorry for submitting a pr with an error. 
props `defaultDate` should have a lower priority than `value`.


rtivital:
Thanks!

PR #4632, opened 2023-08-08
[@mantine/hooks] useResizeObserver: convert to mutable ref and fix useEffect

KurtGokhan:
Reverts most of the changes in #4512 because it introduced a breaking change to the useResizeObserver hook. Instead of using callback ref, the hook will keep using mutable ref. But I replaced `useEffect` with `useIsomorphicEffect` so that the size is calculated as soon as element is mounted.


rtivital:
Thanks!

Issue #4616, opened 2023-08-05
DatePickerInput, DateTimePicker label click does not focus input


Plesin:
### What package has an issue

@mantine/dates

### Describe the bug

Clicking on the input label does not focus the input.

### What version of @mantine/hooks page do you have in package.json?

6.0.16

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/silly-elgamal-6hlm59?file=/src/App.tsx

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

I have not looked at the code, but should be an easy fix. The `<label for=""...` does not match the `<button id...`
If I have time in the following days, I'd like to create a PR.

rtivital:
You are welcome to investigate the issue and send a PR with a fix

Plesin:
Had a look and it seems the correct id is overwritten by accesibleProps.id from the [PopoverTarget](https://github.com/mantinedev/mantine/blob/cf0f85faec56615ea5fbd7813e83bac60dbaefb7/src/mantine-core/src/Popover/PopoverTarget/PopoverTarget.tsx#L45)

So in this case there are three elements involved which needs to be in sync, label, input, dropdown as opposed to a regular input, where there is only label and input.

I'm not sure how to fix it right now. If I have more time, I'll investigate more but the free time available is very limited on my side.
Btw, thanks for such great component library with accesibility, localisation support and more!

rtivital:
Fixed in 6.0.20

Plesin:
Awesome @rommelmamedov Thanks a lot for fixing this! 🚀

PR #4634, opened 2023-08-09
[@mantine/core] & [@mantine/dates] Fix: DatePickerInput, DateTimePicker label click does not focus input #4616

rommelmamedov:
This PR introduces a fix for this [open issue](https://github.com/mantinedev/mantine/issues/4616) https://github.com/mantinedev/mantine/issues/4616.

### DIFF
--- a/src/mantine-dates/src/components/PickerInputBase/PickerInputBase.tsx
+++ b/src/mantine-dates/src/components/PickerInputBase/PickerInputBase.tsx
@@ -133,6 +133,8 @@ export const PickerInputBase = forwardRef<HTMLButtonElement, PickerInputBaseProp
     dropdownHandlers.close();
   };
 
+  const hasLabel = wrapperProps.label;
+
   return (
     <>
       {dropdownType === 'modal' && !readOnly && (
@@ -155,12 +157,12 @@ export const PickerInputBase = forwardRef<HTMLButtonElement, PickerInputBaseProp
           opened={dropdownOpened}
           onClose={handleClose}
           disabled={dropdownType === 'modal' || readOnly}
-          trapFocus
+          trapFocus={!hasLabel}
           returnFocus
           unstyled={unstyled}
           {...popoverProps}
         >
-          <Popover.Target>
+          <Popover.Target shouldOverrideDefaultTargetId={!hasLabel}>
             <Input
               data-dates-input
               data-read-only={readOnly || undefined}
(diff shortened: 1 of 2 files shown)


rommelmamedov:
If this fix looks good I will add the same logic for the `DateTimePicker` component as well.

src/mantine-dates/src/components/PickerInputBase/PickerInputBase.tsx
```diff
@@ -155,12 +157,12 @@ export const PickerInputBase = forwardRef<HTMLButtonElement, PickerInputBaseProp
           opened={dropdownOpened}
           onClose={handleClose}
           disabled={dropdownType === 'modal' || readOnly}
-          trapFocus
+          trapFocus={!hasLabel}
```
rtivital, line 160: Why focus is trapped only when the input has label?

rommelmamedov: Actually, it's the other way around. We don't want to focus to be trapped when we have a label.

So if we will have `trapFocus: true` by default focus will switch instantly from the input field and `Popover.Target` will try to trap focus on the calendar element. (You can see a small glitch of blue outline around input for a moment). 
 
By adding `trapFocus={!hasLabel}` we are trying to disable that behaviour and keep the focus on the input field. If there is no label provided to the input field then it's okay to trap the focus on the calendar.
 
Here is the recording:
https://github.com/mantinedev/mantine/assets/38986496/c55c9705-9e84-4a75-9128-a58e2d41123e

rtivital: Focus trap is the correct behavior in both cases – if it is not trapped, then keyboard controls do not work correctly and thus UX suffers.

rommelmamedov: What do you mean by keyboard controls do not work correctly? If we don't have a focus trap we will be focused on the input field and if the user will press `Tab` to navigate to the next interactive element it will be the direction buttons of `Calendar`.

```jsx
trapFocus={!hasLabel}
// trapFocus
```

image

```jsx
// trapFocus={!hasLabel}
trapFocus
```
image

rtivital: - you can still escape the dropdown when pressing tab key
- when the input is clicked focus is not shifted to the dropdown – users with screen readers will be confused, they've performed an action and there was no announced result

Disabling focus trap is accessibility violation

rommelmamedov: Sorry very busy couple of weeks, couldn't get back to you. Yeah, I guess you are right, I pushed a change that reverts `trapFocus`.

rtivital:
Thanks!

PR #4641, opened 2023-08-10
Fix Gboard paste action when rises from the clipboard section of keypad

saeidalidadi:
I tried to find a native event for the paste action when rises from Gboard's clipboard, but there was not any event related to that action. currently, the `onChange` event is the right place to handle this action.


src/mantine-core/src/PinInput/PinInput.tsx
```diff
@@ -190,16 +190,20 @@ export const PinInput = forwardRef<HTMLDivElement, PinInputProps>((props, ref) =
 
   const handleChange = (event: React.ChangeEvent<HTMLInputElement>, index: number) => {
     const inputValue = event.target.value;
-    const nextChar =
-      inputValue.length > 1 ? inputValue.split('')[inputValue.length - 1] : inputValue;
+    const nextCharOrValue =
+      inputValue.length === 2 ? inputValue.split('')[inputValue.length - 1] : inputValue;
```
rtivital, line 194: Would not it break the case when we paste values from regular keyboard, for example when `1234` is pasted?

saeidalidadi: As the regular keyboard is fully handled by the `onPaste` event handler, logically it should not have any breaks. 
I also tried it out using `Ctrl + v` and `RightClick  + paste` that was ok too.

rtivital:
Thanks!

PR #4666, opened 2023-08-17
[@mantine/core] Grid: Pass `id` prop to dom element

lscheibel:
This PR will allow users to pass the `id` prop to the `Grid` and `Grid.Col` components, as mantine is already doing for other components.

I couldn't find a good reason why the `id` prop was removed from the props that are passed to the DOM element. If there is, feel free to disregard this PR, although in that case the TS types should probably reflect that.


rtivital:
Thanks!

PR #4667, opened 2023-08-17
[@mantine/tiptap] ColorPickerControl: fix incorrect color displayed

HoHieuLuc:
I added parentheses to fix the incorrect expression for `currentColor`. In the previous code, the expression will always returns `theme.colors.dark[1]` in dark mode and `theme.black` in light mode.


rtivital:
Thanks!

Issue #4678, opened 2023-08-21
use-focus-trap doesn't handle Radio Groups well


Devon-Dickson:
### What package has an issue

@mantine/hooks

### Describe the bug

When you shift + tab backwards into a Radio Group, [the last Radio button in the Group is selected](https://www.w3.org/wiki/RadioButton).
But [useFocusTrap() always calculates finalTabbable as the first Radio button in the Group](https://github.com/mantinedev/mantine/blob/33f51e1a7e755ef189d3e87ce17bd9d78a8a0742/src/mantine-hooks/src/use-focus-trap/scope-tab.ts#L9C9-L9C22).
This means it fails to `preventDefault()` or manually set the focus to the last element in `tabbable`, and the user is able to escape the focus trap.

### What version of @mantine/hooks page do you have in package.json?

6.0.19

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/nervous-surf-8xdhy7?file=/src/App.tsx

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

In `scope-tab.ts` we could check if `root.activeElement` is in a RadioGroup, and if so, check if the `finalTabbable` is in the same RadioGroup. If so, set `leavingFinalTabbable = true` as if the `activeElement` was the `finalTabbable`.

This approach is specific to Radio buttons, and I appear to be using properties not available on the `HtmlElement` interface, so it is just serving as a proof of concept. I'm open to feedback.

```typescript
export function scopeTab(node: HTMLElement, event: KeyboardEvent) {
    const tabbable = findTabbableDescendants(node)
    if (!tabbable.length) {
        event.preventDefault()
        return
    }
    const finalTabbable = tabbable[event.shiftKey ? 0 : tabbable.length - 1]
    const root = node.getRootNode() as unknown as DocumentOrShadowRoot
    let leavingFinalTabbable = finalTabbable === root.activeElement || node === root.activeElement

    // Handle the case of the active element being in a RadioGroup with the finalTabbable element
    if (root.activeElement.type === "radio") {
        const activeRadioGroup = tabbable.filter(
            (element) => element.type === "radio" && element.name === root.activeElement.name
        )
        leavingFinalTabbable = activeRadioGroup.includes(finalTabbable)
    }

    if (!leavingFinalTabbable) {
        return
    }

    event.preventDefault()

[...]

PR #4679, opened 2023-08-21
[@mantine/hooks] use-focus-trap: shift + tab bug

Devon-Dickson:
When using shift + tab to go back into a Radio Group, the last option will be selected, which doesn't line up with `finalTabbable`, and breaks out of the focus trap.

To fix, we can check to see if the active element is a radio button, find its radio group, and then check to see if the `finalTabbable` is in the same group. If so, set `leavingFinalTabbable` to `true`.

Addresses #4678

### DIFF
--- a/src/mantine-hooks/src/use-focus-trap/scope-tab.ts
+++ b/src/mantine-hooks/src/use-focus-trap/scope-tab.ts
@@ -8,7 +8,19 @@ export function scopeTab(node: HTMLElement, event: KeyboardEvent) {
   }
   const finalTabbable = tabbable[event.shiftKey ? 0 : tabbable.length - 1];
   const root = node.getRootNode() as unknown as DocumentOrShadowRoot;
-  const leavingFinalTabbable = finalTabbable === root.activeElement || node === root.activeElement;
+  let leavingFinalTabbable = finalTabbable === root.activeElement || node === root.activeElement;
+
+  // Handle the case of the active element being in a RadioGroup with the finalTabbable element
+  if (root.activeElement instanceof HTMLInputElement && root.activeElement.type === 'radio') {
+    const activeRadioGroup = tabbable.filter(
+      (element) =>
+        element instanceof HTMLInputElement &&
+        root.activeElement instanceof HTMLInputElement &&
+        element.type === 'radio' &&
+        element.name === root.activeElement.name
+    );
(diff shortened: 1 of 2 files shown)


src/mantine-hooks/src/use-focus-trap/scope-tab.ts
```diff
@@ -8,7 +8,19 @@ export function scopeTab(node: HTMLElement, event: KeyboardEvent) {
...
+  if (root.activeElement instanceof HTMLInputElement && root.activeElement.type === 'radio') {
+    const activeRadioGroup = tabbable.filter(
+      (element) =>
+        element instanceof HTMLInputElement &&
+        root.activeElement instanceof HTMLInputElement &&
+        element.type === 'radio' &&
+        element.name === root.activeElement.name
+    );
+    leavingFinalTabbable = activeRadioGroup.includes(finalTabbable);
+  }
```
Devon-Dickson, line 23: This, obviously, is only a solution for Radio Groups, but I suspect there's similar behavior for other elements.

Another idea I had was to get the next target, and check if it's in `tabbable` at all, but I wasn't sure if there was an API exposed to do that easily, or if we'd need to do a lot of work to calculate it on our own.

rtivital:
@Devon-Dickson are you planning to finish working on this PR?

Devon-Dickson:
> @Devon-Dickson are you planning to finish working on this PR?

I believe this is an accessibility bug, and I'd like to see the fix merged. However I'm a bit busy at the moment and don't have the bandwidth to debug the RTL issue (`userEvent.tab({ shift: true })` doesn't seem to work).

I may be able to revisit in a few weeks, unless you're familiar with the issue and can recommend an immediate solution?

Devon-Dickson:
> I believe it should be resolved by https://github.com/mantinedev/mantine/pull/4856

Looks like another route to the same solution 👍.

No tests yet though, which is the blocker on this PR.

Devon-Dickson:
I see #4856 has been merged, but is still missing tests.

If I get the time in the next couple of weeks, I can look at this again and get some tests added and passing. In the meantime, we can close this PR.

rtivital:
Thanks!

Devon-Dickson:
Issue with the tests ended up being a small thing, but comes with a performance cost
https://github.com/mantinedev/mantine/pull/4969

PR #4703, opened 2023-08-29
[docs] Add prepend option to Emotion Cache example

marwinburesch:
When implementing Emotion Cache in our project, the `prepend` option caused some confusion and some missed visual bugs. The docs don't mention that you need to set `prepend: true`, if you need the custom cache to behave like the cache automatically generated by Mantine. This behavior only made sense when reading between the lines of the docs, this change aims to make it a little more obvious by adjusting the example code provided.


rtivital:
Thanks!

Issue #4695, opened 2023-08-25
Anchor's span prop does not work


arel:
### What package has an issue

@mantine/core

### Describe the bug

Using the `span` prop on an `<Anchor />` has no effect. The resulting component is still an `<a/>`.

Contrast this to setting the component explicitly, `<Anchor component="span">`, which does behave correctly.

Based on the documentation, I would expect that setting `span` or `component="span"` would behave the same.

### What version of @mantine/hooks page do you have in package.json?

6.0.19

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/naughty-star-m98lv5?file=/src/App.tsx

### Do you know how to fix the issue

None

### Are you willing to participate in fixing this issue and create a pull request with the fix

None

jishat:
But I see that you need to declare "component" as a key. Please share the link.

arel:
@jishat, I'm not sure I understood. When you set the `component` prop of an `Anchor`, it behaves correctly, but the issue happens when you set the `span` prop. It does not work as advertised.

```
        <Anchor span>This should be a span element but is not</Anchor>
```


The link to the documentation is: https://mantine.dev/core/anchor/?t=props

> `span`, boolean, shorthand for `component="span"`

The link to the codesandbox is: https://codesandbox.io/s/naughty-star-m98lv5?file=/src/App.tsx

jishat:
Yes, you are right. It is a bug.

jishat:
I will can fix it

arel:
That works, @rtivital, but doesn't that break the semantic versioning?

PR #4704, opened 2023-08-29
[@mantine/core] Anchor: fix span prop #4695

jishat:
Using the span prop on an Anchor as in <Anchor span> has no effect. Now its work


rtivital:
Thanks for sending a PR, this is not a correct solution, instead span prop should be removed from the Anchor component.

Issue #4713, opened 2023-08-30
BackgroundImage doesn't worth with URLs containing parenthesis


stavfx:
### What package has an issue

@mantine/core

### Describe the bug

Passing a url with `(` or `)` in the path causes BackgroundImage to not load.
The same url in the `Image` component works fine.

A workaround is to wrap the url in quotes:
```
<BackgroundImage src={`"${urlWithParens}"`} />
```

### What version of @mantine/hooks page do you have in package.json?

6.0.19

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/mantine-bgimage-parens-79skx2

### Do you know how to fix the issue

None

### Are you willing to participate in fixing this issue and create a pull request with the fix

None

PR #4715, opened 2023-08-31
[@mantine/core]: Add double quotes to wrap src for backgroundImage property

blueagler:
Fix Issue: https://github.com/mantinedev/mantine/issues/4713


rtivital:
Thanks!

Issue #4746, opened 2023-09-11
The <Radio/> component does not receive padding when a value of 0 is passed into the label prop.


qaynam:
### What package has an issue

@mantine/core

### Describe the bug

When the number 0 is passed to the label props of the Radio component, it is treated as false and the padding does not work.

```tsx
import { Box, Flex, Group, MantineProvider, Radio } from "@mantine/core";

export default function App() {
  return (
    <MantineProvider withGlobalStyles withNormalizeCSS>
      <Box sx={{
        padding: 10
      }}>
      <Group>{Array.from({length: 10}).map((_, i) => (
        <Radio key={i} label={i} value={i}/>
      ))}
      </Group>
      </Box>
    </MantineProvider>
  );
}
```

![image](https://github.com/mantinedev/mantine/assets/33139713/5e371071-db3f-4fdc-8d77-f04f892b9376)

### What version of @mantine/hooks page do you have in package.json?

6.0.20

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/blissful-jerry-l29nwm

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

change this line 

https://github.com/mantinedev/mantine/blob/0455a0f3c47938c0dcae69e5093ce6c28a6bc73b/src/mantine-core/src/InlineInput/InlineInput.tsx#L55

to this

```tsx
{label != null && (
  <label className={classes.label} data-disabled={disabled || undefined} htmlFor={id}>
    {label}
  </label>
 )}
```

PR #4755, opened 2023-09-13
[@mantine/core] InlineInput: Fix InlineInput component does not receive padding when a value of 0 is passed into the label prop. 

qaynam:
#4746


rtivital:
Thanks!

PR #4760, opened 2023-09-13
[@mantine/core] Remove incompatible use of replaceAll

### LINKED ISSUE #4759: NumberInput broken in Edge 84 - Uncaught TypeError: t.replaceAll is not a function
### What package has an issue

@mantine/core

### Describe the bug

Supported versions of Chromium browsers (e.g., Edge 84) break when using the `NumberInput` component because `replaceAll()` is not supported and results in the error:

```
Uncaught TypeError: t.replaceAll is not a function
```

Yet, version 84 [should be supported](https://mantine.dev/pages/about):

> All Mantine components and hooks were tested to work in the following browsers:
> ...
> Chromium browsers 84+ – Chrome, Edge, Chrome for Android, etc. (released July 2020)

6.0.19

_No response_

Yes

Yes

### Possible fix

Use `replace` with a global regex instead of `replaceAll`.

See, for instance: https://www.designcise.com/web/tutorial/how-to-fix-replaceall-is-not-a-function-javascript-error

arel:
Fixes #4759.

This removes `replaceAll()` that is not supported by all of the browsers that Mantine intends to support.

Instead, `replace` is called with a global regex.

Note that it's necessary to escape the potential regex string (e.g., in the case that `thousandsSeparator` is a special character like `.`). Fortunately, this escape function was already written.

### DIFF
--- a/src/mantine-core/src/Highlight/highlighter/highlighter.ts
+++ b/src/mantine-core/src/Highlight/highlighter/highlighter.ts
@@ -1,6 +1,4 @@
-function escapeRegex(value: string) {
-  return value.replace(/[-[\]{}()*+?.,\\^$|#]/g, '\\$&');
-}
+import { escapeRegex } from '@mantine/utils';
 
 export function highlighter(value: string, _highlight: string | string[]) {
   if (_highlight == null) {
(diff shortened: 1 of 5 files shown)


rtivital:

PR #4778, opened 2023-09-18
Fix links in README and Footer in docs

jason-lee88:
It appears the `href` for many pages on the Mantine docs have changed with Mantine v7. Ex:

- `/pages/contributing` --> `/contribute`
- `/pages/about` --> `/about`
- ...

These haven't been updated in the README or footer in the Mantine documentation, so users clicking on these links would navigate to the 404 page. These are now updated in this PR.

It appears the root `/changelog` page does not exist anymore with Mantine v7, so the Previous Versions page `/changelog/previous-versions` is now used in place of that.

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,9 +11,9 @@
 ## Links
 
 - [Documentation](https://mantine.dev/)
-- [Contributing](https://mantine.dev/pages/contributing/)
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #4779, opened 2023-09-18
Update changelog link

chrisbendel:
I don't know if there is an internal link to @latest anywhere, perhaps could use that instead of `previous-versions`. Feels wrong to hardcode `/changelog/7-0-0` somewhere

Changelog link in footer currently `404`s
Screenshot 2023-09-18 at 11 57 27 AM


rtivital:
Thanks for a PR, the issue was resolved by the other contributor already

PR #4801, opened 2023-09-20
[@mantine/core] Update dots loader component

jonavila:
Updated Loader component 'dots' variant to match animation and style as close as possible to v6

https://github.com/mantinedev/mantine/assets/5589281/fe0f4bc7-7d13-40f3-9c98-7ab620d5338b


rtivital:
Thanks!

PR #4817, opened 2023-09-20
[@mantine/core] Fix table highlight doesn't work on striped dark rows

mateuscqueiros:
Problem was on specificity, darker rows have styles that override the hover background change. Adding `!important` rule solved the problem by prioritizing hover.


rtivital:
Thanks!

PR #4822, opened 2023-09-21
[@mantine/core] Select: Fix search text not updating

DoubleJ-G:
The select input isn't watching the `data` prop change after a value is already set, this will cause the underlying input to remain unchanged/blank. A simple case could be seen with something like this:

```ts
  const value = "Foo";
  const [options, setOptions] = useState([]);

  useEffect(() => {
    const optionsSetter = setTimeout(() => {
      setOptions(["Foo", "Bar"]);
    }, 1500);
    return () => clearTimeout(optionsSetter);
  });

  return (
      <Select data={options} value={value} />
  );
  ```
  
By watching the optionsLockup in the useEffect dependencies it should catch updates to that data.


rtivital:
Thanks!

Issue #4827, opened 2023-09-21
useLocalStorage type undefined when defaultValue present


DoubleJ-G:
### What package has an issue

@mantine/hooks

### Mantine v7

The value side of `useLocalStorage` seems to be returning a type of T | undefined even when defaultValue is present. I have created two code sandboxes with the different versions of mantine, you can over [value] to see the different types. 

v6 Sandbox: https://codesandbox.io/s/laughing-tharp-nykr28
v7 Sandbox: https://codesandbox.io/s/kind-rhodes-k3yrsd

This is probably a side effect of v7 bumping the typescript version, the `use-local-storage.d.ts` file output is different

v6
```ts
import { IStorageProperties } from './create-storage';
export declare function useLocalStorage<T = string>(props: IStorageProperties<T>): readonly [T, (val: T | ((prevState: T) => T)) => void, () => void];
//# sourceMappingURL=use-local-storage.d.ts.map
```

v7
```ts
import { IStorageProperties } from './create-storage';
export declare function useLocalStorage<T = string>(props: IStorageProperties<T>): readonly [T | undefined, (val: T | ((prevState: T) => T)) => void, () => void];
```
 

### What version of @mantine/* page do you have in package.json? (Note that all @mantine/* packages must have the same version in order to work correctly)

7.0.0

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/kind-rhodes-k3yrsd

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

I believe the fix to get it behaving the same as before is to also cast the return type in `createStorage.ts`

```ts
return [
   value === undefined ? (defaultValue as T) : value,
   setStorageValue,
   removeStorageValue,
] as const;
 ```
 
There are already similar type casts in the inner `useStorage` function. But this still hides the possibility the data is undefined if the localstorage is missing the key and a defaultValue isn't provided.
 
 It should be possible to have some conditional return type based on T being undefined in the arguments, but I am not enough of a typescript wizard to get that working locally

mateuscqueiros:
It looks like the `T | undefined` is coming directly from the type `StorageProperties<T>` since it's an optional property, meaning it could be undefined if not passed as a prop. But what would happen if no defaultValue was passed to the hook in your solution? Would the cast transform `undefined` into `T`?

DoubleJ-G:
> Would the cast transform undefined into T?

[...]

PR #4828, opened 2023-09-21
[@mantine/hooks] Restore return type of useLocalStorage

DoubleJ-G:
Related: #4827


rtivital:
The type is incorrect, it will be false in the following case:

```tsx
const [colorScheme] = useLocalStorage<'light' | 'dark'>({ key: '1' });
```

Here the `colorScheme` type is `'light' | 'dark'` although its value is `undefined`.

PR #4816, opened 2023-09-20
[@mantine/core] make MantineThemeColorsOverride an interface instead of type

matheusvellone:
This is needed to allow `MantineThemeColorsOverride` to be overwritten

PR #4831, opened 2023-09-21
[@mantine/core] Fix broken other theme override type

Ryiski:
Similar to the `MantineThemeColorsOverride` issue  #4816, changed `MantineThemeOther` to interface to be overwritten


rtivital:
Thanks!

PR #4842, opened 2023-09-22
[@mantine/core] Fix AppShell.Main not taking up full height

### LINKED ISSUE #4830: AppShell.Main not taking up full height in v7
### What package has an issue

@mantine/core

### Describe the bug

Currently migrating an app to v7 and I noticed that the `AppShell.Main` component doesn't take up the full height of the viewport. That means if you want to change the `background-color` it will not have the desired effect. Instead, it will look something like this:

image

The component is not getting sized properly:

image

### What version of @mantine/* page do you have in package.json? (Note that all @mantine/* packages must have the same version in order to work correctly)

7.0.0

_No response_

Yes

Yes

### Possible fix

I'm not deep enough into the library code but to me it seems this could simply be fixed by changing the `min-height` of the element to `100dvh` instead of subtracting header/footer height which causes the problem. Relevant code line:

https://github.com/mantinedev/mantine/blob/5449e029ad0f47bca24be0d490c10854f314ec38/src/mantine-core/src/components/AppShell/AppShell.module.css#L110C2-L112C5

Since the footer/header/aside position is fixed and are on a higher layer this doesn't affect anything. Screenshot of element with `min-height: 100dvh`.

image

The element takes on full height/width of viewport now:

image

RobinTTY:
Fixes #4830

### DIFF
--- a/src/mantine-core/src/components/AppShell/AppShell.module.css
+++ b/src/mantine-core/src/components/AppShell/AppShell.module.css
@@ -107,9 +107,7 @@
   );
   padding-top: calc(var(--app-shell-header-offset, 0px) + var(--app-shell-padding));
   padding-bottom: calc(var(--app-shell-footer-offset, 0px) + var(--app-shell-padding));
-  min-height: calc(
-    100dvh - var(--app-shell-header-offset, 0px) - var(--app-shell-footer-offset, 0px)
-  );
+  min-height: 100dvh;
   transition-property: padding;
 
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

ldalves:
It seems this could create another issue if a user tries to use `AppShell.Footer `?

rtivital:
If you have found an issue, please report it in GitHub issues with a reproduction

PR #4892, opened 2023-09-26
[@mantine/core] TagsInput: Call onOptionSubmit on enter key in input

### LINKED ISSUE #4891: TagsInput onOptionSubmit is not called on submitting with Enter key
### What package has an issue

@mantine/core

### Describe the bug

TagsInput `onOptionSubmit` is not called on submitting with Enter key with focus on the input

What calls `onOptionSubmit`
- clicking on menu item
- pressing enter while the item is highlighted in the dropdown

What doesn't work
- Pressing enter to submit a new tag not in the list
- Pressing enter to submit an already created tag (with focus still on the input)

tested with
`<TagsInput data={["test"]} onOptionSubmit={(val) => console.log(val)} />`

\-

On a related note, is it possible to bring back the `onCreate` prop to handle the specific case? It was very helpful along with `getCreateLabel` also, but it seems they are not there anymore.

^7.0.1

_No response_

Yes

Yes

### Possible fix

_No response_

Yhprum:
fixes #4891

### DIFF
--- a/src/mantine-core/src/components/TagsInput/TagsInput.tsx
+++ b/src/mantine-core/src/components/TagsInput/TagsInput.tsx
@@ -234,6 +234,7 @@ export const TagsInput = factory<TagsInputFactory>((_props, ref) => {
       }
 
       if ((!isDuplicate || (isDuplicate && allowDuplicates)) && _value.length < maxTags!) {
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

Issue #4596, opened 2023-08-01
DateTimePicker not respecting minDate for the TimeInput


manuel-rw:
### What package has an issue

@mantine/dates

### Describe the bug

DateTimePicker does not respect ``minDate`` and ``maxDate`` for the TimeInput component.
You can select a time outside the allowed range.

### What version of @mantine/hooks page do you have in package.json?

6.0.8

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/twilight-darkness-pkdmwq?file=/src/App.tsx

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

No

### Possible fix

- Add a new property for minTime and maxTime on the TimeInput
- Pass the minDate and maxDate from the DateTimePicker to the TimeInput

rajeevdodda:
@rtivital  I would like to work on this issue.

rtivital:
@rajeevdodda You are welcome to submit a PR with a fix

chewhx:
What I understand about the problem from my own experience trying out the component.

Example:
```typescript
 const minDateTime = new Date('2024-01-01T01:00:00.000Z');  
 const maxDateTime = new Date('2024-01-02T01:00:00.000Z'); 

    <DateTimePicker
        placeholder="Date time picker"
        clearable
        minDate={minDateTime} // This is 01 Jan 2024, 9:00 AM on my local time (GMT+8)
        maxDate={maxDateTime} // This is 02 Jan 2024, 9:00 AM on my local time (GMT+8)
      />

```

Expectations: 
- The user able to select 01 Jan 2024 and not be able to input any time before 9:00 AM
- The user able to select 02 Jan 2024 and not be able to input any time after 9:00 AM

Actual behaviour:
- The user able to select 01 Jan 2024 ✅  and input time before 9:00AM ❌ 
- The user able to select 02 Jan 2024 ✅ and input time after 9:00AM ❌ 
- The user able to input any time range 

---

**The `<TimeInput />` component uses the native html input tag with `type="time"`.**

As I understand from [MDN docs](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time#setting_maximum_and_minimum_times):
- We are able to set `min` and `max` values for time inputs but it only triggers time validation when the user tries to submit the form
- It does not prevent the user from inputting time values outside of the min and max range
- We can only do so programmatically

---

[...]

PR #4900, opened 2023-09-27
[@mantine/dates]: DateTimePicker minTime & maxTime fix

ShaifArfan:
Possible fix for #4596

### DIFF
--- a/src/mantine-dates/src/components/TimeInput/TimeInput.tsx
+++ b/src/mantine-dates/src/components/TimeInput/TimeInput.tsx
@@ -15,6 +15,10 @@ import classes from './TimeInput.module.css';
 export interface TimeInputProps extends InputBaseProps, ElementProps<'input', 'size'> {
   /** Determines whether seconds input should be rendered */
   withSeconds?: boolean;
+  /** Minimum possible time */
+  minTime?: string;
+  /** Maximum possible time */
+  maxTime?: string;
 }
 
 export type TimeInputFactory = Factory<{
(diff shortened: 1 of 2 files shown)


src/mantine-dates/src/components/TimeInput/TimeInput.tsx
```diff
@@ -45,6 +49,8 @@ export const TimeInput = factory<TimeInputFactory>((_props, ref) => {
       step={withSeconds ? 1 : 60}
       type="time"
       __staticSelector="TimeInput"
+      min={props.minTime}
```
rtivital, line 52: There is no reason to define custom props for this feature as you just pass it to the input wihout any modifications. Please remove `minTime` and `maxTime` props. In DatePickerInput component use `min` and `max` props on the TimeInput instead

ShaifArfan: got it

rtivital:
It does not work. When min date is set, it is impossible to enter any time if selected date is min or max date. If date is between min and max dates it is impossible to enter minutes. 

https://github.com/mantinedev/mantine/assets/10353856/170de793-29c6-4499-b9ee-071e9dada817

Story: 

```tsx
export function WithTimeMinMaxDate() {
  return (
    <div style={{ padding: 40, maxWidth: 400 }}>
      <DateTimePicker
        placeholder="Date time picker"
        defaultValue={new Date(2022, 3, 11)}
        clearable
        minDate={new Date(2022, 3, 10, 10, 50)}
        maxDate={new Date(2022, 3, 12, 10, 50)}
      />
    </div>
  );
}
```

PR #4914, opened 2023-09-28
[@mantine/hooks] use-timeout: Memoize returned callbacks

Raicuparta:
Currently, `start` and `clear` have unstable references, so they change every render. This makes them hard to use in effects, and cause unnecessary rerenders. This PR fixes that with `useCallback`.

### DIFF
--- a/src/mantine-hooks/src/use-timeout/use-timeout.ts
+++ b/src/mantine-hooks/src/use-timeout/use-timeout.ts
@@ -32,7 +32,7 @@ export function useTimeout(
     }
 
     return clear;
-  }, [clear, delay, options.autoInvoke, start]);
+  }, [clear, options.autoInvoke, start]);
 
   return { start, clear };
 }
--- a/src/mantine-hooks/src/use-timeout/use-timeout.ts
+++ b/src/mantine-hooks/src/use-timeout/use-timeout.ts
@@ -16,7 +16,7 @@ export function useTimeout(
         }, delay);
       }
     },
-    [callback, delay],
+    [callback, delay]
   );
 
   const clear = useCallback(() => {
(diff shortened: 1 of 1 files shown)


src/mantine-hooks/src/use-timeout/use-timeout.ts
```diff
@@ -7,29 +7,32 @@ export function useTimeout(
...
+  }, []);
 
   useEffect(() => {
     if (options.autoInvoke) {
       start();
     }
 
     return clear;
-  }, [delay]);
+  }, [clear, delay, options.autoInvoke, start]);
```
rtivital, line 35: There is no reason to add options.autoInvoke to the dependencies array. `start` should be called exactly once when hook is initialized.

rtivital: @Raicuparta you did not resolve my comment.

Raicuparta: Do we really want an effect here then?

The way it was before, changing the delay would also autoInvoke again, but changing the callback wouldn't. Was that intended? Because it seems very counterintuitive to me.

In this PR, changing the callback, the delay, or the autoInvoke option would trigger the autoInvoke again.

If the intention is to really only make it run exactly once, we should just have no dependencies at all, but then maybe it shouldn't be in an effect at all.

I have removed autoInvoke from the dependencies anyway to avoid making changes outside the scope of the PR, but this is pretty clearly against the intended useEffect usage as per the docs: https://react.dev/learn/removing-effect-dependencies.

(I also removed delay from the dependencies since it is now redundant)

Raicuparta: Oh and yeah I didn't mean to make that first push that only removed the delay, sorry about the notification spam

rtivital:
Thanks!

PR #4915, opened 2023-09-28
[@mantine/core] Fix searchable select with controlled value

### LINKED ISSUE #4909: Controlled Select not working with searchable
### What package has an issue

@mantine/core

### Describe the bug

I want to migrate this code from Mantine 6 to Mantine 7:

```
const [value, setValue] = useState("React");
return (
    <Select
      value={value}
      onChange={(value) => setValue(value || "")}
      data={["React", "Angular", "Svelte", "Vue"]}
      searchable
    />
);
```

In Mantine 6 this worked as expected, but with Mantine 7 (latest release) it is impossible to enter a search value because the component is overwriting the search with `value` on every render.

Now I am not sure if this is a bug or if this is on purpose and I have do something else.

Thanks for any help!

7.1.0

_No response_

No

Yes

### Possible fix

_No response_

wuifdesign:
fixes #4909

`optionsLockup` needs to be memorized as they are used in a effect further down. otherwise the effect will be triggered each render - preventing search changes to take effect.

https://github.com/mantinedev/mantine/blob/7dc4a72ece31650761c8aae96756c10e628dc9c7/src/mantine-core/src/components/Select/Select.tsx#L179-L187

### DIFF
--- a/src/mantine-core/src/components/Select/Select.story.tsx
+++ b/src/mantine-core/src/components/Select/Select.story.tsx
@@ -86,6 +86,19 @@ export function ControlledSearch() {
   );
 }
 
+export function SearchControlledValue() {
+  return (
+    <div style={{ padding: 40 }}>
+      <Select
+        value="React"
+        data={['React', 'Angular', 'Svelte']}
+        placeholder="Select something"
+        searchable
+      />
+    </div>
+  );
+}
+
 export function AllowDeselectFalse() {
   return (
     <div style={{ padding: 40 }}>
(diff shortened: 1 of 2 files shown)


wuifdesign:
Other ideas to fix this, i can do so if wanted:
- rework `optionsLockup` to `useRef` so it will not trigger the useEffect
- rework the `useEffect` -> this may have the effect that external value changes are not displayed correctly

@rtivital

rtivital:
I think current solution is perfectly fine, thanks!

rtivital:
Thanks!

Issue #4912, opened 2023-09-28
NumberInput bug on Smaprtphone


Kenzo-Wada:
### What package has an issue

@mantine/core

### Describe the bug

- When entering a number in NumberInput, the input value becomes incorrect if you input more than two digits on a smartphone using japanese keyboard. Specifically, it occurs in the following way:

10 -> 11010
23 -> 22323

- Olso, is entering number larger than 1111111111111111, the rest of the number will become 0. (ex. 1111111111111111200000000000). when this occurs, increment or decrement will now work at all. this occurs not only japanese keyboard.

### What version of @mantine/* packages do you have in package.json? (Note that all @mantine/* packages must have the same version in order to work correctly)

7.0.0

### Do you know how to fix the issue

None

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

### Possible fix

There is a possibility that there may be something wrong with the event handler or state management, so those should be corrected. Additionally, it might be worthwhile to consider whether the NumberInput should really have type="number" as the default.

Kenzo-Wada:
this only happens on japanese keybord...

saeidalidadi:
This issue also exists on all devices. numbers with a length of 16 digits and more have this issue.
I can work on this if.

rtivital:
You are welcome to submit a PR with a fix if #4916 does not resolve the issue

saeidalidadi:
#4916 fixes this one too.

rtivital:
Great, thanks!

Issue #4901, opened 2023-09-27
NumberInput sets value to 0 when first character (i.e. "1"), is deleted (in "1000" value)


ivanjuric:
### What package has an issue

@mantine/core

### Describe the bug

When editing existing value, i.e. 1000, if we delete only first number, in this example 1, when trying to change it to 2000, whole value is set to 0, and caret stays at 0 index. Then we enter desired value (2) and final value in input is 20 instead of leaving 000 in first step, resulting with 2000 as expected final value. 

Example of actual behavior:
![numberinput_actual](https://github.com/mantinedev/mantine/assets/7100678/3dea9223-5cf2-4e30-8508-2b7891341ff9)

Example of expected behavior from previous versions (this is Mantine v6, before react-number-format):
![numberinput_expected](https://github.com/mantinedev/mantine/assets/7100678/41bcfdf7-29ff-4600-b814-0564348b833b)

### What version of @mantine/* packages do you have in package.json? (Note that all @mantine/* packages must have the same version in order to work correctly)

7.0.2

### Do you know how to fix the issue

None

### Are you willing to participate in fixing this issue and create a pull request with the fix

None

ShaifArfan:
I guess this is expected behavior if your initial value is a `number`. You can set initial value as `string` to achieve this behavior.
But right now, even if you set the initial value as `string`, the returned value from the component will be a `number`.

A possible fix could be, if the type of the `value` is `number` then use `payload.floatValue` otherwise we can use `payload.value` which will be a `string`.
```tsx
  const handleValueChange: OnValueChange = (payload, event) => {
    setValue(
      typeof _value === 'number' && isValidNumber(payload.floatValue)
        ? payload.floatValue
        : payload.value
    );
    onValueChange?.(payload, event);
  };
```

@rtivital what you think?

rtivital:
Sounds right, you are welcome to submit a PR with a fix (if it is needed)

ivanjuric:
Hi, thank you for the fix!

However, there are still left some cases that are not covered by this fix.

For example, if you look at this case in v7: [https://mantine.dev/core/number-input/#thousand-separator](https://mantine.dev/core/number-input/#thousand-separator), you'll see it's setting the value again to 0.

But (almost the) same example from v6 ([https://mantine.dev/core/number-input/#thousand-separator](https://mantine.dev/core/number-input/#thousand-separator)) works fine and leaves zeroes until focus is moved from input.

PR #4916, opened 2023-09-28
[@mantine/core] NumberInput value type fix

ShaifArfan:
`NumberInput` value type will be based on initial value type. 

Fixes: #4912
Fixes: #4901


rtivital:
Thanks!

PR #4939, opened 2023-10-01
[@mantine/core] Button: fix loading overlay default border radius

Raicuparta:
Fixes a problem you can see easily in the Mantine docs in dark mode, where the border radius of the overlay doesn't match the button's: https://mantine.dev/core/button/#loading-state

image


rtivital:
Thanks!

PR #4953, opened 2023-10-02
[docs] Progress: Fix animated prop in Storybook demo

spicybackend:
Update the Progress component's Storybook to use the `animated` prop instead of `animate` from v6 which is [currently broken in the docs](https://mantine.dev/core/progress/).

progress-animated


rtivital:
Thanks!

PR #4956, opened 2023-10-03
add other size to type Contianer Component css varibales

mehdiraized:
we need change other size in createTheme > components > Container
after change size i see we have problem in typescript

### DIFF
--- a/src/mantine-core/src/components/Container/Container.tsx
+++ b/src/mantine-core/src/components/Container/Container.tsx
@@ -16,7 +16,7 @@ import classes from './Container.module.css';
 
 export type ContainerStylesNames = 'root';
 export type ContainerCssVariables = {
-  root: '--container-size';
+  root: '--container-size' | '--container-size-xs' | '--container-size-sm'| '--container-size-md' | '--container-size-lg' | '--container-size-xl';
 };
 
 export interface ContainerProps
@@ -42,6 +42,11 @@ const defaultProps: Partial<ContainerProps> = {};
 const varsResolver = createVarsResolver<ContainerFactory>((_, { size, fluid }) => ({
   root: {
     '--container-size': fluid ? undefined : getSize(size, 'container-size'),
+    '--container-size-xs': getSize('xs', 'container-size'),
+    '--container-size-sm': getSize('sm', 'container-size'),
+    '--container-size-md': getSize('md', 'container-size'),
+    '--container-size-lg': getSize('lg', 'container-size'),
+    '--container-size-xl': getSize('xl', 'container-size'),
   },
 }));
 


rtivital:
You can find documentation on how to change size here – https://mantine.dev/styles/variants-sizes/#sizes-with-components-css-variables

mehdiraized:
i have issue
```
Type '() => { root: { '--container-size-xs': string; '--container-size-sm': string; '--container-size-md': string; '--container-size-lg': string; '--container-size-xl': string; }; }' is not assignable to type 'PartialVarsResolver<{ props: ContainerProps; ref: HTMLDivElement; stylesNames: "root"; vars: ContainerCssVariables; }>'.
  Call signature return types '{ root: { '--container-size-xs': string; '--container-size-sm': string; '--container-size-md': string; '--container-size-lg': string; '--container-size-xl': string; }; }' and 'PartialTransformVars<ContainerCssVariables>' are incompatible.
    The types of 'root' are incompatible between these types.
      Type '{ '--container-size-xs': string; '--container-size-sm': string; '--container-size-md': string; '--container-size-lg': string; '--container-size-xl': string; }' has no properties in common with type 'Partial<Record<"--container-size", string | undefined>>'.
 ```
 
Code:
```
'use client';

import { Container, createTheme } from '@mantine/core';

export const theme = createTheme({
	fontFamily: 'Optimistic Display',
	breakpoints: {
		xs: '30em',
		sm: '60em',
		md: '74em',
		lg: '82em',
		xl: '90em',
	},
	components: {
		Container: Container.extend({
			vars: () => ({
				root: {
					'--container-size-xs': 'calc(30em*var(--mantine-scale))',
					'--container-size-sm': 'calc(60em*var(--mantine-scale))',
					'--container-size-md': 'calc(74em*var(--mantine-scale))',
					'--container-size-lg': 'calc(82em*var(--mantine-scale))',
					'--container-size-xl': 'calc(90em*var(--mantine-scale))',
				},
			}),
		}),
	},
});
```

**Comment by rtivital** [comment]
The issue reported by TypeScript is correct, with CSS variables resolver you are supposed to resolve these values with given component props as shown in example. If you want to add static styles, you should use `classNames` prop instead.

**Comment by mehdiraized** [comment]
> The issue reported by TypeScript is correct, with CSS variables resolver you are supposed to resolve these values with given component props as shown in example. If you want to add static styles, you should use `classNames` prop instead.

Can you write an example for this piece of code?

**Comment by rtivital** [comment]
https://codesandbox.io/s/cool-darkness-yctpdx?file=/src/App.tsx

**Comment by mehdiraized** [comment]
> https://codesandbox.io/s/cool-darkness-yctpdx?file=/src/App.tsx

i have problem without size props container component
i need change default sizes container

```
      <Container bg="gray">
        without size props container
      </Container>
```

rtivital:
I've updated the sandbox to include container without size.

mehdiraized:
> I've updated the sandbox to include container without size.

Without the size of the page, the container does not change the size of the various displays

rtivital:
https://mantine.dev/core/container/#responsive-max-width

PR #4968, opened 2023-10-03
[@mantine/hooks] useTimeout callback not running after state change

### LINKED ISSUE #4965: [@mantine/hooks] useTimeout callback not running after state change
### What package has an issue

@mantine/hooks

### Describe the bug

Setting a state before the useTimeout callback has been called causes the useTimeout to "reset". The callback is not being called.

Very similar to #3789

When comparing with V6 implementation of the useTimeout hook, it looks like we use the callback as a dependency of the internal `useCallback`. [V6](https://github.com/mantinedev/mantine/blob/cdf1179358c6d4591f5de76a2a41935ff4c91ec6/src/mantine-hooks/src/use-timeout/use-timeout.ts#L20) vs [V7](https://github.com/mantinedev/mantine/blob/34e52affc566161481d21e58d9926a9aeb52ec40/src/mantine-hooks/src/use-timeout/use-timeout.ts#L19C21-L19C21)

7.1.2

https://codesandbox.io/s/funny-snow-nzjv8p

Yes

Yes

### Possible fix

Simply remove the callback from the dependency of the `useCallback`, just like the V6 implementation.

For the moment, I have to wrap all of my timeout callbacks inside a useCallback and everything works like before. I don't know if this was an intended change in V7

Jacouille:
fixes #4965

### DIFF
--- a/docs/yarn.lock
+++ b/docs/yarn.lock
@@ -133,10 +133,10 @@
   version "0.0.0"
   uid ""
 
-"@mantine/store@7.1.1":
-  version "7.1.1"
-  resolved "https://registry.yarnpkg.com/@mantine/store/-/store-7.1.1.tgz#859071535c97fc54862ffe946f58d95cb0455777"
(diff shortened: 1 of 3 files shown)


rtivital:
Thanks!

PR #4969, opened 2023-10-04
[@mantine/core] Add tests for <FocusTrap /> using shift + tab

Devon-Dickson:
Fixes the tests that blocked #4679, and #4856 was merged without. Ended up being a silly thing that blocked my PR for weeks 🙄 .

Unfortunately, `useFocusTrap` has a `setTimeout` before actually focusing on an element, so we need to add a `wait()` before each of these assertions. These two test cases will add about 60ms to the total run time of the test suite, so I'll leave it up to @rtivital to decide on their worth.


rtivital:
Thanks!

PR #4976, opened 2023-10-05
[@mantine/core] Rating: Prevent emulated mouse events in touch devices

### LINKED ISSUE #4399: Rating component does not take fractional values in mobile
### What package has an issue

@mantine/core

### Describe the bug

https://github.com/mantinedev/mantine/assets/60938164/087203ba-07b8-4137-a5fc-706bbbae8b62

In the video you can see that when I put a fractional rating, it works fine. But as soon as I click on a different place, the fractional part becomes whole (3.5 stars -> 4 stars).

This happens only in mobile.

6.0.8

Here is the code: https://github.com/IgnisDa/ryot/blob/fad4950777eefac2094a9c44a238deecf272f6a6/apps/frontend/src/pages/media/post-review.tsx#L143

No

Yes

### Possible fix

_No response_

thealmarques:
Fixes #4399 

When using a touch device browsers will dispatch emulated mouse events which, in this case, will cause the `handleMouseLeave` callback to be called as soon as we click elsewhere. One of the ways to fix this is to prevent emulated events by preventing this behaviour in `onTouchStart` and `onTouchEnd`.
Read more about emulated events [here](https://developer.mozilla.org/en-US/docs/Web/API/Touch_events/Using_Touch_Events).

Issue fixed:

https://github.com/mantinedev/mantine/assets/7091198/0c2b8566-b162-4d60-a09b-c1e291dd7966

I also noticed that when using the keyboard we are changing the rating when using arrow navigation and not waiting for the spacebar which IMO is not right, because we might just be passing by 🤷 ? I added this fix here but I can remove it if it doesn't fit.

Issue fixed:

https://github.com/mantinedev/mantine/assets/7091198/cdf8cf43-fafd-47a7-8679-cfe628cd7a5b

### DIFF
--- a/src/mantine-core/src/components/Rating/Rating.tsx
+++ b/src/mantine-core/src/components/Rating/Rating.tsx
@@ -128,6 +128,8 @@ export const Rating = factory<RatingFactory>((_props, ref) => {
     onMouseMove,
     onHover,
     onMouseLeave,
+    onTouchStart,
+    onTouchEnd,
     size,
     variant,
     getSymbolLabel,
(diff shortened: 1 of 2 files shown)


src/mantine-core/src/components/Rating/Rating.tsx
```diff
@@ -214,8 +216,38 @@ export const Rating = factory<RatingFactory>((_props, ref) => {
     hovered !== -1 && onHover?.(-1);
   };
 
+  const handleTouchStart = (event: React.TouchEvent<HTMLDivElement>) => {
+    event.preventDefault();
+
+    const { touches } = event;
+    if (touches.length !== 1) {
```
thealmarques, line 223: For multitouch devices it's a bit farfetched so we might be safe here? 🤔

rtivital:
Thanks!

PR #4979, opened 2023-10-05
[core] Update storybook to 7.4.6

wuifdesign:
Since you're still using Storybook v6, I've upgraded it to v7 for you.

As part of the upgrade, I've enhanced the linting for stories by adding the `eslint-plugin-storybook`. Additionally, I've replaced the deprecated `@storybook/addon-postcss` with the latest `@storybook/addon-styling-webpack`.

Furthermore, I've modernized the demos by transitioning from the deprecated `storiesOf` format to the new Component Story Format (CSF). To facilitate this, I've introduced a new command, `npm run generate-demo-stories`. This command scans the `mantine-demos/src/**/*.demos.story.tsx` files and generates stories based on the `index.ts` file within the same folder.


rtivital:
Thanks for the PR! I've been trying it myself but did not have much luck.

It seems like Jest tests are no longer running.

wuifdesign:
yeah, i noticed it, they are running locally, but not if i do a clean reinstall with yarn. i do some testing and try to provide a fix. maybe i have to upgrade jest too.

wuifdesign:
@rtivital tests are running now

wuifdesign:
i hope switching from storiesOf to the CSF is ok. running storiesOf with storybook 7 legacy mode just lead to problems for me and i think currently there is no way to dynamically load stories. this i the reason why i went with the npm command to generate them.

rtivital:
Yeah, it's for the better as `storiesOf` is deprecated anyway

rtivital:
I've tested how it works, and it is awful (not related to PR). The performance is at least 3x time worse than with storybook v6 – both start time and reload when code changes (takes ~5-7 seconds to apply changes compared to ~1-2 seconds with v6). Currently, will stay on v6 and probably wait for perf improvements, please the PR opened for reference.

wuifdesign:
I'll try to take a look at performance point, but didn't notice that much of a difference. Maybe i can tweak some things.

Would you prefere to switch to vite instead of webpack for storybook, as this may increase performance. I didn't do the switch as vite is currently not used in the rest of the repo.

rtivital:
If it will provide better performance, then Vite is better

wuifdesign:
@rtivital i found a workaround do improve performance. The startup time is now about half of the startup time of storybook v6. (for me about 13-16sec until i see the components in the browser)

seams like the builded files influence the build time for storybook v7, i made a more specific list of imports which improve performance by a big amount.

I had do to a bit of workaround to create the list of imports as just using a `glob` is also really slow currently (https://github.com/storybookjs/storybook/issues/19812)

rtivital:
Thanks! Works great now

PR #4981, opened 2023-10-06
[@mantine/core] Add createable back into to MultiSelect and Select components

drumnation:
I updated @mantine/core from 6 to 7 and noticed `createable` didn't exist anymore and I was using it in my project. I studied the old version and tried to add it back to function as close to the old version as I could. I split off a chunk of my code into a 	`createable.ts` file to keep it separate. No ts errors.

I'm not sure why it was removed from 7, maybe there is a reason other than the feature just didn't make it in? If it just didn't make it in, here it is.  Otherwise I'm curious to know why it was removed. I'd rather not use a fork, so hoping this is useful. 

Happy to make any requested modifications. Love this library.

image

image

UPDATED: MultiSelect Component 
`createable` prop to determine if new values can be created by the user
- Added `onCreate` callback when a new value is created
- Added `handleCreate` function to handle the creation of new values
- Modified `handleOptionsSubmit` function to handle the submission of options, including the handling of createable options
- Added `dataWithPotentialNewItem` to include the potential new createable item
- Modified `noOptionsFound` to consider whether picked options should be hidden or not
- Added `displayCreateOption` to check if createable and search value are present
- Modified `optionsData` to include the createable item if present
- Added `handleCreate` function to handle the creation of new options
- Added MultiSelect storybook story to include `Createable` component

Did the same for Select.  Added a storybook story, no TS errors. Tried to follow a similar format.

image

image

UPDATED: Select component now supports creating new options
- Added a new prop `createable` to the `Select` component which determines if new values can be created by the user. It is set to `false` by default.
- Added a new prop `onCreate` to the `Select` component which is a callback function that is called when a new value is created by the user.
- Added a new function `handleCreate` which adds a new option to the list of options and sets it as the selected value.
- Added a new function `addNewOptionToData` which appends the new option to the list of data if `createable` is `true`.
- Added a new function `getOptionExists` which checks if the new option already exists in the list of data.
- Updated the `getParsedComboboxData` function to return the parsed data along with the new option if it exists.
- Updated the `handleOptionSubmit` function to handle creating a new option if `createable` is `true`.
- Updated the `OptionsDropdown` component to use the new parsed data.
- Added a new file `createable.ts` which contains the helper functions used for creating new options.

### DIFF
--- a/src/mantine-core/src/components/MultiSelect/MultiSelect.tsx
+++ b/src/mantine-core/src/components/MultiSelect/MultiSelect.tsx
@@ -1,30 +1,33 @@
-import React from 'react';
 import { useId, useUncontrolled } from '@mantine/hooks';
+import React from 'react';
 import {
   BoxProps,
-  StylesApiProps,
-  factory,
   ElementProps,
-  useProps,
   Factory,
+  StylesApiProps,
   extractStyleProps,
-  useStyles,
+  factory,
+  useProps,
   useResolvedStylesApi,
+  useStyles,
 } from '../../core';
+import { __CloseButtonProps } from '../CloseButton';
 import {
   Combobox,
-  OptionsDropdown,
-  useCombobox,
-  getParsedComboboxData,
-  getOptionsLockup,
+  ComboboxItem,
   ComboboxLikeProps,
   ComboboxLikeStylesNames,
+  ComboboxParsedItem,
+  OptionsDropdown,
+  getOptionsLockup,
+  getParsedComboboxData,
+  useCombobox,
 } from '../Combobox';
 import { __BaseInputProps, __InputStylesNames } from '../Input';
-import { PillsInput } from '../PillsInput';
-import { Pill } from '../Pill';
 import { InputBase } from '../InputBase';
-import { __CloseButtonProps } from '../CloseButton';
+import { Pill } from '../Pill';
+import { PillsInput } from '../PillsInput';
+import { addNewOptionToData, createLabel } from './createable';
 import { filterPickedValues } from './filter-picked-values';
 
 export type MultiSelectStylesNames =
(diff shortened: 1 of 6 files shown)


src/mantine-core/src/components/Select/Select.tsx
```diff
@@ -146,6 +157,8 @@ export const Select = factory<SelectFactory>((_props, ref) => {
     onChange,
   });
 
+  const optionsLockup = useMemo(() => getOptionsLockup(parsedData), [parsedData, _value]);
```
drumnation, line 160: Is calling this "Lockup" a mistake? Seems like it should be named "Lookup", I was going to change it, but thought I'd ask first.

src/mantine-core/src/components/MultiSelect/MultiSelect.tsx
```diff
@@ -101,70 +110,72 @@ const defaultProps: Partial<MultiSelectProps> = {
...
-    name,
-    form,
-    id,
-    clearable,
-    clearButtonProps,
-    hiddenInputProps,
+    withScrollArea,
+    wrapperProps,
     ...others
   } = props;
```
drumnation, line 180: alphabetized props

src/mantine-core/src/components/Select/Select.tsx
```diff
@@ -90,53 +99,55 @@ const defaultProps: Partial<SelectProps> = {
...
+    selectFirstOptionOnChange,
+    size,
+    styles,
+    unstyled,
+    value,
+    vars,
+    withCheckIcon,
+    withScrollArea,
     ...others
   } = props;
```
drumnation, line 147: alphabetized props

rtivital:
Hi @drumnation, thanks for showing interest in contributing to Mantine! `creatable` prop was removed from components on purpose:
- It is much easier to maintain components without extra features
- Its API was confusing and there were a lot of questions about how it works from the community (mostly about async operations)
- It was hard to modify any logic because of the countless features combinations that components provided (it almost impossible to write good unit tests for these components)

In v7 this feature and other features were not added because of these reasons and this functionality was moved to Combobox component.

drumnation:
Do you have an example anywhere of how I would add createable to Select and MultiSelect using Combobox? I just dove through the docs and I didn't see anything.

rtivital:
Combobox provides 50+ examples that include features that were previously baked in other components – https://mantine.dev/combobox/?e=MultiSelectCreatable

From v7, if you need these features, you should build a custom component with Combobox

volkandkaya:
I just ran into this issue along with grouping.

Might be worth highlighting on changelog.

volkandkaya:
@drumnation how far did you get?

Just spent a while trying to extend <Select> and went terrible due to rerendering.

Issue #4985, opened 2023-10-06
Carousel previousControlProps not worked


kolobok001:
### What package has an issue

@mantine/carousel

### Describe the bug

Good afternoon. Carousel previousControlProps not worked. Moreover, for the right control everything works as expected. 
I'll do a PR a little later

### What version of @mantine/* packages do you have in package.json? (Note that all @mantine/* packages must have the same version in order to work correctly)

7.1.2

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/funny-poitras-j4tx6z?file=/src/App.tsx

### Do you know how to fix the issue

Yes

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

rtivital:
The issue has been resolved in [7.1.3](https://github.com/mantinedev/mantine/releases/tag/7.1.3)

PR #4986, opened 2023-10-06
Carousel previousControlProps not woked

kolobok001:
Fix issue https://github.com/mantinedev/mantine/issues/4985


rtivital:
Thanks for showing interest in contributing to Mantine! Unfortunately, it is not a valid fix. The correct fix was implemented in https://github.com/mantinedev/mantine/commit/ce23377ae7ce03b736258d52ee6bc478fc2a0350

Issue #4995, opened 2023-10-07
Select v7 search not working if you make searchValue and select's value controlled


VugarAhmadov:
### What package has an issue

@mantine/core

### Describe the bug

In mantine Select component, when you add searchable prop and make the searchValue controlled via React state it is not working. 

Use case:
what I want to do is based on the searchValue I am filling the options. this is my whole code

```
"use client";

import { useState } from "react";
import { Select } from "@mantine/core";
import { useDebouncedValue } from "@mantine/hooks";
import { useQuery } from "@tanstack/react-query";

import { api } from "@/api/client";

export default function ChooseCity() {
  const [citySearchValue, setCitySearchValue] = useState("");
  const [debouncedCitySearchValue] = useDebouncedValue(citySearchValue, 1000);

  const [cityValue, setCityValue] = useState<string | null>(null);

  const { data: cities } = useQuery({
    queryKey: ["cities", debouncedCitySearchValue],
    queryFn: () => api.getCities("az", debouncedCitySearchValue),
  });

  return (
    <Select
      data={cities?.map((c) => ({ label: c.name, value: c.id }))}
      clearable
      label="City"
      filter={(f) => f.options}
      onChange={(value) => setCityValue(value)}
      value={cityValue}
      searchable
      searchValue={citySearchValue}
      onSearchChange={(value) => setCitySearchValue(value)}
    />
  );
}
```

### What version of @mantine/* packages do you have in package.json? (Note that all @mantine/* packages must have the same version in order to work correctly)

7.1.2

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/mantine-v7-select-forked-fwklwp?file=/src/App.jsx

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

No

DoubleJ-G:
This is an issue with some of the Select code having a useEffect on state derived from the `data` (which in this case and many others is not stable) and unnecessarily running too many times. I have improved the useEffect dependency that was causing the issue in #4998.

DoubleJ-G:
In the meantime if you wrap the data in a `useMemo` it should work as expected

```js
const data = useMemo(() => cities?.map((c) => ({ label: c.name, value: c.id })), [cities])


return (
    <Select
      data={data}
      clearable
      label="City"
      filter={(f) => f.options}
      onChange={(value) => setCityValue(value)}
      value={cityValue}
      searchable
      searchValue={citySearchValue}
[...]

PR #4998, opened 2023-10-09
[@mantine/core] Select: More stable useEffect dependency for search

DoubleJ-G:
I was tracking down the issue mentioned in #4995 and discovered it was due to multiple renders caused by the useEffect hook.

Often times the `data` prop isn't a stable reference so using the `optionsLockup` which is derived from it in a useEffect was causing unnecessary renders. 

Instead, I have used the `selectedOption` in the dependency array which seems to be more stable even when `data` changes.


rtivital:
Thanks!

VugarAhmadov:
Thanks! works like a charm.

PR #5019, opened 2023-10-11
[Templates]: Added T3 Stack + Mantine Template

WillKirkmanM:
# Mantine + T3 Stack Template
> https://github.com/WillKirkmanM/mantine-t3-template

Added T3 Icon, changed respective types, and updated templates section.

*WillKirkmanM*


rtivital:
Thanks for showing interest in contributing to Mantine! Currently, we do not list community maintained templates. Later, an additional page will be added where such templates can be shared.

PR #5032, opened 2023-10-12
[@mantine/core] TagsInput: Fix required prop behavior

### LINKED ISSUE #5016: TagsInput: required prop doesn't work as intended.
### What package has an issue

@mantine/core

### Describe the bug

When `required` prop is given to the `TagsInput` component, after entering some tags and then submitting the form, the "Please fill out this field" form validation is displayed and the form cannot be submitted.

I think this is because `_searchValue` is empty immediately after pressing `EnterKey` or `splitChars` key.

In addition, the asterisk is not displayed when I give a `required` prop

7.1.2

https://codesandbox.io/s/headless-smoke-dp3qg4?file=/src/App.tsx

Yes

Yes

### Possible fix

- Passing `required` to `PillInput` component to display asterisks when `required` prop is given.
- Set `required` prop in `PillsInput.Field` to `required && values.length === 0`.

The following video shows the behavior of the above fix in local.
It appears to be working as intended.

https://github.com/mantinedev/mantine/assets/57832553/da982daf-d7d6-4022-90c6-1ee0f73c5310

sunadoi:
Fixes #5016 

## 1. display asterisk when `required` prop is passed
Pass `required` prop to `PillInput` component so that an asterisk is displayed when `required` prop is passed to `TagsInput`.

## 2. fix form validation behavior when `required` prop is passed
The `value` prop of `PillsInput.Field` is `_searchValue`, which is empty immediately after splitting.
Therefore, setting `required` will enable form validation even if tag (`_value`) is present.

To fix this, the condition of `required` in `PillsInput.Field` is set to `required && _value.length === 0`.

### DIFF
--- a/src/mantine-core/src/components/TagsInput/TagsInput.tsx
+++ b/src/mantine-core/src/components/TagsInput/TagsInput.tsx
@@ -144,6 +144,7 @@ export const TagsInput = factory<TagsInputFactory>((_props, ref) => {
     inputContainer,
     inputWrapperOrder,
     withAsterisk,
+    required,
     labelProps,
     descriptionProps,
     errorProps,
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #5045, opened 2023-10-15
[@mantine/core] Highlight: sort highlight options by longest first

### LINKED ISSUE #5044: Incorrect highlighting when one word is a substring of another
### What package has an issue

@mantine/core

### Describe the bug

In the Highlight component, if you want to highlight a string where one word you want to highlight is a substring of another word you want to highlight, only the shorter part of the word gets highlighted.

Please see the code sandbox to see a minimal repro - the words Mr and Mrs should be highlighted, but only the Mr part of both words gets highlighted. 

latest

https://codesandbox.io/s/gifted-wiles-zn4l4g?file=/src/App.tsx

Yes

Yes

### Possible fix

The file highlight.js creates a regex by `|` concatenating all words you are wanting to be highlighted (in `_highlight`), and then splits the input on this regex. This list of split chunks is then tested one by one on the regex to see if it should be highlighted or not. 

If you are trying to highlight the string: `"Welcome Mr and Mrs Smith"`, with the words: `["Mr", "Mrs"]`

Then the string you are trying to highlight gets split first on `Mr`, resulting in this: `['Welcome ', 'Mr', ' and ', 'Mr', 's ', 'Smith', '']`

Then when this is looped over to see if it should be highlighted, `Mr` gets highlighted twice, and the `s` of `Mrs` does not get highligheted.

A quick fix would be first to order `_highlight` by longest word first, so that matches to the longer options get matched first before the shorter ones, so your string split results in this:
`['Welcome ', 'Mr', ' and ', 'Mrs', ' ', 'Smith', '']`

Which gets highlighted correctly.

wgpsutherland:
Fixes #5044

### DIFF
--- a/src/mantine-core/src/components/Highlight/highlighter/highlighter.ts
+++ b/src/mantine-core/src/components/Highlight/highlighter/highlighter.ts
@@ -25,6 +25,7 @@ export function highlighter(value: string, _highlight: string | string[]) {
       : highlight
           .filter((part) => part.trim().length !== 0)
           .map((part) => part.trim())
+          .sort((a, b) => b.length - a.length)  // sorts by longest string first
           .join('|');
 
   const re = new RegExp(`(${matcher})`, 'gi');


rtivital:
Please fix errors

wgpsutherland:
I will fix - thank you

rtivital:
Thanks!

PR #5051, opened 2023-10-16
[docs] Sort order of other packages entries alphabetically

santorial:


rtivital:
Why do these packages have to be sorted alphabetically?

santorial:
To be consistent with the other packages. The components in core seems to be ordered asc.
In my opinion, you can find the components more quickly.

rtivital:
I think that it's better to keep it as is

PR #5052, opened 2023-10-16
[docs] Sync contribution guide in docs with CONTRIBUTING.md

santorial:


rtivital:
Thanks for pointing the issue out. The fix is the opposite – to remove the commit convention section from the contributing.md.

PR #5054, opened 2023-10-16
[core] Add children to CardSectionProps

phungleson:
From the document, it looks like that `CardSectionProps` should have children.

https://mantine.dev/core/card/


phungleson:
I see there are some components missing children as well, should I update them as well?

rtivital:
CardSection as well as other polymorphic components have dynamic types that are calculated based on `component` prop. `children` prop comes from the element or component props, it should not be defined explicitly.

PR #5067, opened 2023-10-18
Fix callback types

nmay231:
As discussed in Discord, this changes all callback types from methods to function attributes (`{ callback(a: A): T }` => `{ callback: (a: A) => T }`). This prevents type variance bugs, such as the following where the real type of `onChange` is `(value: string | null) => void`.

```typescript
  function expectsAString(value: string) {
    // null error is only caught at runtime, currently
    console.log(value.length); // TypeError: Cannot read properties of null (reading 'length')
  }

  return (
      <Select
        onChange={expectsAString}
        data={["React", "Vue", "Angular"]}
        clearable
      />
  );
}
```

<hr />

Also synchronized `.nvmrc` and `.tool-versions`.


rtivital:
Thanks!

nmay231:
And thank you!

Issue #4971, opened 2023-10-05
Focus is not set inside DateTimePicker upon open in Modal with minDate applied


anthony-dandrea:
### What package has an issue

@mantine/core and/or @mantine/dates

### Describe the bug

This may occur with other types of fields and attribute combinations. But I was able to isolate it to having a `minDate` set in the `DateTimePicker` inside of an open `Modal`. I am unable to use the keyboard to navigate and go through the date popover or modal.

See isolated [example here](https://codesandbox.io/s/datepicker-in-modal-mantine-jwnnx2?file=/src/App.js). 

![modal](https://github.com/mantinedev/mantine/assets/5413624/8c52ec25-4f79-41d6-8a19-67539821a69e)

### What version of @mantine/* packages do you have in package.json? (Note that all @mantine/* packages must have the same version in order to work correctly)

7.1.2

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/datepicker-in-modal-mantine-jwnnx2?file=/src/App.js

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

None

PR #5078, opened 2023-10-19
[@mantine/dates] Calendar: Fix focusing on disabled element

teh23:
fix: https://github.com/mantinedev/mantine/issues/4971
first element for focus (.focus() method in use-focus-trap) was disabled causing this bug


rtivital:
Thanks!

PR #5091, opened 2023-10-21
[@mantine/hooks]: Fix storage access error in create-storage

### LINKED ISSUE #5089: useSessionStorage will throw if sessionStorage is blocked by window
### What package has an issue

@mantine/hooks

### Describe the bug

If `useSessionStorage` is used on a blocked context (e.g. iframed 3rd party site embedded on another website in incognito mode), `window['sessionStorage']` will throw when executed

https://github.com/mantinedev/mantine/blob/05e26d79c050f29838985ff6875739b60ddd6a63/src/mantine-hooks/src/use-local-storage/create-storage.ts#L87

image

I imagine it'd be the same for localStorage as well

7.1.3

_No response_

Yes

Yes

### Possible fix

Wrap the checks in try-catch block

yongwee:
Fixes #5089

### DIFF
--- a/docs/yarn.lock
+++ b/docs/yarn.lock
@@ -133,10 +133,10 @@
   version "0.0.0"
   uid ""
 
-"@mantine/store@7.1.3":
-  version "7.1.3"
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #5123, opened 2023-10-25
[@mantine/form] Reset should set values to current initialValues

### LINKED ISSUE #5122: useForm: reset() does not respect updated initialValues
### What package has an issue

@mantine/form

### Describe the bug

Calling `form.reset()` sets the values to the `initialValues` that were defined when initializing `useForm()`.
However, if I update initial values with `form.setInitialValues()` and call `form.reset()`, the form is still reset to the original initialValues, not the updated ones.

7.1.5

_No response_

Yes

Yes

### Possible fix

[Here](https://github.com/mantinedev/mantine/blob/e006b48cc990a96bdee0d174e9170db5d502cd4e/src/mantine-form/src/use-form.ts#L74), the `initialValues` needs to be changed to `valuesSnapshot.current`.

dodas:
Fixes #5122

### DIFF
--- a/src/mantine-form/src/tests/reset.test.ts
+++ b/src/mantine-form/src/tests/reset.test.ts
@@ -45,4 +45,15 @@ describe('@mantine/form/reset', () => {
     expect(hook.result.current.isDirty()).toBe(false);
     expect(hook.result.current.values).toStrictEqual({ a: 1, b: 2 });
   });
+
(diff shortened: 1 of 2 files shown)


rtivital:
Can you please add a test for this?

dodas:
@rtivital sure, test added.
lmk if it needs any tweaking.

rtivital:
Thanks!

PR #5126, opened 2023-10-26
[@mantine/core] Spoiler: Avoid leaving a gap when label is null

sartak:
I'd like to use Spoiler _just_ to reveal, not to hide again. However, passing `hideLabel={null}` leaves an unsightly gap below the Spoiler.

https://codesandbox.io/s/epic-waterfall-xkkzmv?file=/src/App.tsx

Spoiler leaves a gap when label is null

This PR fixes this issue by checking whether `spoilerMoreContent` is `null`.


rtivital:
Thanks!

PR #5131, opened 2023-10-26
DOC: slightly improve language style in styles-api.mdx

yogabonito:
The documentation is already really good. I just tried to further improve it a little bit by, for example, adding articles here and there.


rtivital:
Thanks!

PR #5133, opened 2023-10-26
DOC: slightly improve language style in data-attributes.mdx

yogabonito:
As mentioned in #5131, the docs are really good and probably only require a few more articles in some places.


rtivital:
Thanks! If you have time, your contributions to update other documentation files are very welcome!

yogabonito:
Thank you @rtivital!
Unfortunately, I won't be working with react for the next few months. But if I am going to use it again, I will definitely use it with mantine and take another look at the docs :slightly_smiling_face:
Thank you for this awesome library! :raised_hands:

Issue #5144, opened 2023-10-28
Embla carousel version is missing in Docs


kavinda1995:
### Link to the page where something nasty is located

https://mantine.dev/others/carousel/

### Exact quote of what is wrong

In the docs (I've referenced docs for [v5](https://v5.mantine.dev/others/carousel/), but should be relevant for other versions as well) - it haven't mentioned the exact embla version needed to make the carousel works. When I just copied and pasted the code given in the docs to install embla - it gave few issues including https://github.com/mantinedev/mantine/issues/5039.

Hence, it's better to define the exact version we need to embla in the docs as well. Will create a PR for this

### Are you willing to create pull request with the fix?

Yes

kavinda1995:
Fixed by https://github.com/mantinedev/mantine/pull/5145

PR #5145, opened 2023-10-28
Denote needed embla version for carousel

kavinda1995:
Fixing the issue https://github.com/mantinedev/mantine/issues/5144


rtivital:
Thanks!

PR #5194, opened 2023-11-05
[@mantine/core]: Text: Fix inline prop in css (mantinedev#5190)

gorkunov:
Fixes mantinedev#5190.

I compared with other components and it seems var `--_text-line-height` is missed in core text css configuration.


rtivital:
Thanks!

PR #5213, opened 2023-11-07
Add "size" popover middleware to fix overflow issues in Popover.Dropdown

ranquild:
Currently, if the `Popover.Dropdown` content is tall, there is no way to add scrollbars (overflow) properly as the dropdown overflows the page. Reproduction https://codesandbox.io/s/amazing-nobel-s6zqy8

This issue is supposed to be fixed by [size](https://floating-ui.com/docs/size) middleware in `floating-ui`. This PR fixes the issue by adding `size` option in `PopoverMiddlewares` which enables the `size` middleware, which in turn uses `availableWidth` and `availableHeight` per `floating-ui` docs.

I decided not to set `overflow: auto` automatically in `Popover.Dropdown` because in some places there might be a requirement to shrink the content and not to add scrollbars.

How to verify:
- `yarn storybook`
- Open http://localhost:6006/?path=/story/popover--size
- Click on the button and verify that it's possible to scroll the dropdown content properly

Issue (codesandbox):
Screenshot 2023-11-07 at 12 43 33

Fix (storybook):
Screenshot 2023-11-07 at 12 32 46

### DIFF
--- a/src/mantine-core/src/components/Popover/use-popover.ts
+++ b/src/mantine-core/src/components/Popover/use-popover.ts
@@ -9,6 +9,7 @@ import {
   Middleware,
   inline,
   limitShift,
+  UseFloatingReturn,
 } from '@floating-ui/react';
 import { FloatingAxesOffsets, FloatingPosition, useFloatingAutoUpdate } from '../Floating';
 import { PopoverWidth, PopoverMiddlewares } from './Popover.types';
@@ -29,7 +30,7 @@ interface UsePopoverOptions {
   arrowOffset: number;
 }
 
-function getPopoverMiddlewares(options: UsePopoverOptions) {
+function getPopoverMiddlewares(options: UsePopoverOptions, getFloating: () => UseFloatingReturn) {
   const middlewares: Middleware[] = [offset(options.offset)];
 
   if (options.middlewares?.shift) {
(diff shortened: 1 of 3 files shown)


src/mantine-core/src/components/Popover/use-popover.ts
```diff
@@ -74,22 +102,9 @@ export function usePopover(options: UsePopoverOptions) {
     }
   };
 
-  const floating = useFloating({
+  const floating: UseFloatingReturn<Element> = useFloating({
     placement: options.position,
-    middleware: [
-      ...getPopoverMiddlewares(options),
-      ...(options.width === 'target'
```
ranquild, line 81: Moved to `getPopoverMiddlewares`

src/mantine-core/src/components/Popover/use-popover.ts
```diff
@@ -74,22 +102,9 @@ export function usePopover(options: UsePopoverOptions) {
     }
   };
 
-  const floating = useFloating({
+  const floating: UseFloatingReturn<Element> = useFloating({
```
ranquild, line 105: Without an explicit type TS throws an error

rtivital:
Thanks!

Issue #5021, opened 2023-10-11
Mantine v7: highlight.js is huge


sparanoid:
### What package has an issue

@mantine/code-highlight

### Describe the bug

v6:

<img width="473" alt="image" src="https://github.com/mantinedev/mantine/assets/96356/3aa4ee35-6bfd-4603-bfd5-6ddb5d078857">

v7:

<img width="473" alt="image" src="https://github.com/mantinedev/mantine/assets/96356/bc1f6ecf-7ca8-4a50-9a87-dc550920cffe">

After some investigations with next-bundle-analyzer:

<img width="705" alt="image" src="https://github.com/mantinedev/mantine/assets/96356/7ffad55c-ecd2-44b8-9892-eb917c30ce2e">

### What version of @mantine/* packages do you have in package.json? (Note that all @mantine/* packages must have the same version in order to work correctly)

7.1.2

### Do you know how to fix the issue

None

### Are you willing to participate in fixing this issue and create a pull request with the fix

None

### Possible fix

According to [highlight.js's documentation](https://highlightjs.org/#usage):

```js
// Using ES6 import syntax
import hljs from 'highlight.js/lib/core';
import javascript from 'highlight.js/lib/languages/javascript';

// Then register the languages you need
hljs.registerLanguage('javascript', javascript);
```

I think we could only import the syntax language I need to reduce the bundle size.

rtivital:
Duplicate of #4879

Issue #4879, opened 2023-09-26
@mantine/code-highlight huge bundle size


wuifdesign:
### What package has an issue

@mantine/code-highlight

### Describe the bug

As `@mantine/code-highlight` includes all files of higlight.js it increases the bundle size by about 300kb. 

### What version of @mantine/* packages do you have in package.json? (Note that all @mantine/* packages must have the same version in order to work correctly)

7.0.1

### Do you know how to fix the issue

None

### Are you willing to participate in fixing this issue and create a pull request with the fix

None

### Possible fix

May introduce some kind of lazy-loading or config which languages should be included.

ibedwi:
Hi! I proposed a package that might address this issue. Ended up with smaller bundle size. In the given example, I use `typescript` and `javascript`. The gzipped size for `highlight.js` in the bundle is around `12.23 KB`.

Could you please give feedback? I would love to hear your suggestions! Thanks! @rtivital @wuifdesign 😁

wuifdesign:
it think this will work, but would be a breaking change if added to the main package. 2 same packages is somewhat strange.

i also tried to implement some lazy loading, but haven't found a way that works nice with all bundlers.

jvllmr:
I think the reason are in the following lines:
https://github.com/mantinedev/mantine/blob/00a94ceefc22c80ad08af73acb943f8f0c01d662/packages/%40mantine/code-highlight/src/use-highlight.ts#L2

https://github.com/mantinedev/mantine/blob/00a94ceefc22c80ad08af73acb943f8f0c01d662/packages/%40mantine/code-highlight/src/InlineCodeHighlight.tsx#L2


In this case hljs imports a lot of language presets and registers them.

The original management object originates from `highlight.js/lib/core`.

To solve this issue one must import and register the plugins by themselves and mantine works with a highlight.js provided by the user instead of the default one. This would also allow users to inject their own highlight.js extensions. Or mantine still manages the highlight.js instance and a user provides the extensions. Anyway, it would be breaking change needed.

jvllmr:
Is there a specific reason why this is not planned? Not even in 8.0 when looking at the closed PR? I really only want to display JSON in my app and don't need highlight.js to load all extensions for it to do that.

JustinGrote:
[...]

PR #5236, opened 2023-11-12
[@mantine/code-highlight] Attempt on reducing bundle size

ibedwi:
# Description

This is an attempt to reduce the size of `code-highlight`.

Probably address https://github.com/mantinedev/mantine/issues/4879, https://github.com/mantinedev/mantine/issues/5021

## Approach

Instead of install `highlight.js` as the library's dependencies, it's installed as `peerDependencies`. The project that use code highlight library must install `highlight.js` as its `dependencies` and pass it to this library via `HiglightProvider`

```tsx
// /the-project/page.tsx
import { CodeHighlight, HighlightProvider } from "@mantine/code-highlight-lite";

// Only import the core module
import hljs from "highlight.js/lib/core";

// Only registere the used languages
hljs.registerLanguage(
  "plaintext",
  require("highlight.js/lib/languages/plaintext")
);
hljs.registerLanguage(
  "javascript",
  require("highlight.js/lib/languages/javascript")
);
hljs.registerLanguage(
  "typescript",
  require("highlight.js/lib/languages/typescript")
);

// Create example typescript code
const exampleCode = "...";

export default function Home() {
  return (
    // pass hljs as props to the provider
    <HighlightProvider highlightInstance={hljs}>
        <CodeHighlight code={exampleCode} language="typescript" />
    </HighlightProvider>
  );
}
```

## Result

This approach tested using Next.js `14.0.2` (app router) and Webpack Bundle Analyzer. The `highlight.js`'s chunk is very small and only include the registered language:

```
Stat size: 120.91 KB
Parsed size: 38.16 KB
Gzipped size: 12.23 KB
```

CleanShot 2023-11-12 at 13 53 48@2x

## Demo Project

See the [demo project here](https://github.com/ibedwi/mantine-code-highlight-lite-test)


rtivital:
Hi @ibedwi, thanks for the PR! I would rather not maintain two separate versions of the same package. It also seems that it is not possible to solve this issue without breaking changes. I'll keep this PR opened until 8.0 is out for reference. Now the PR cannot be merged.

ibedwi:
Hi @rtivital ! I agree to not maintain two separate versions of the same package, since it serves the same purpose.

>  It also seems that it is not possible to solve this issue without breaking changes.
Yeap! Since this approach exclude `highlight.js` from the package, it definitely introduce breaking changes.

>  I'll keep this PR opened until 8.0 is out for reference. Now the PR cannot be merged.
Thanks for considering this PR! I would love to contribute! Feel free to ping me 😁

Issue #5211, opened 2023-11-07
Select Component did not reset to default value when using form.reset()


tranhoanglinh-dev:
### What package has an issue?

@mantine/core

### Describe the bug

I'm using Mantine form , or event React hook form. I think it worked fine with v6. But I got the error when upgrading to v7
Description: 

- Init Form : value & label of Select is undefined/ blank
- I started change the value of Select 
- I use form.reset() to clear all values & expect the form return to Init state (all fields is blank)
- but Select component did not reset. It keep selected value

Note that this issue only happen if it is the form to create new object (which initialValues is "" or undefined)
It did happen in Edit form (having initialValues)

Thanks

### What version of @mantine/* packages do you have in package.json? (Note that all @mantine/* packages must have the same version in order to work correctly)

@mantine/core 7.2.0

(https://codesandbox.io/s/determined-ben-jzz84k)

_No response_

### Do you know how to fix the issue

No

### Do you want to send a pull request with a fix?

None

rtivital:
Please provide a sandbox with a minimal reproduction

dodas:
@tranhoanglinh-dev does this happen in v7.2.1 as well?

tranhoanglinh-dev:
@rtivital @dodas Please see code sandbox here https://codesandbox.io/s/determined-ben-jzz84k
Case 1: 
1/ First , you set **default value of Select is ""** in the code -> Submit to see form values
2/ Change the Select to Car. Then I reset => form UI should be back to initial state with no value selected & form values should be reset to default
Case 2
1/ First , you set **default value of Select is "car"** in the code -> Submit to see form values
2/ Change the Select to House. Then I reset => Select changes to Car as expected on UI but values of form did not change
![image](https://github.com/mantinedev/mantine/assets/138766999/110e009c-ea44-4ddc-a62a-fc0b66280d49)

Hope it help.

tranhoanglinh-dev:
> Please provide a sandbox with a minimal reproduction

https://codesandbox.io/s/determined-ben-jzz84k @rtivital

tranhoanglinh-dev:
> @tranhoanglinh-dev does this happen in v7.2.1 as well?

it happened on latest version @dodas

nadbad:
@rtivital The issue has been reproduced successfully and I have started investigating the issue

gursahibsahni:
If anyone is still looking for a fix for this - you need to set the initial value of the form select input to null. The reset doesn't work without a default null value.

PR #5248, opened 2023-11-14
[@mantine/core] Select: fix reset issue when

nadbad:
Fixes https://github.com/mantinedev/mantine/issues/5211


nadbad:
fixed issue (#5211)

I also do apologise for the mess in the prevousily close pr, i had some issues with it 😅

PR #5283, opened 2023-11-18
[@mantine/core] Tabs: Add `keepMounted` prop to TabsPanel

aadityabhusal:
When `keepMounted` prop is passed to Tabs component, it is applied to all the panels. This change allows to set `keepMounted` property to specific panels.

**A scenario I faced:**
I have 3 panels, a normal, a lazy loaded, and one with iframe. I want to keep the normal and iframe panels mounted but want the lazy loaded to only mount when opened. If `keepMounted` is set to false, the iframe keeps refreshing and if it is set to true, the component is not lazy loaded.


rtivital:
Thanks!

PR #5284, opened 2023-11-19
[docs: Group component] Mention default gap and vertical centering to show it's a higher level abstraction and more opinonated than `Flex`

slk333:
I'm not sure if talking about the default values in the `Usage` section is something desirable or not.

Having a slight glance of the gap and vertical centering is useful here, but also increases verbosity.

I considered also talking about the wrap behaviour that is activated by default, as opposed to a `Flex` where it is does not wrap by default.


rtivital:
Thanks for showing interest in contributing to Mantine! The documentation already has a comparison table and demos for Group and Flex, this change is not necessary.

PR #5287, opened 2023-11-20
[@mantine/spotlight] Fix input bug with Japanese IME

RE-yura:
related: https://discord.com/channels/854810300876062770/887408088800448553/1176006456332529714

---

Spotlight is not submitted when the enter key is pressed during IME input.
Tested on macOS 14.0.


rtivital:
Thanks!

PR #5326, opened 2023-11-25
[docs] fix style api overview example

elpddev:
The leading text says:

```mdx
Example of [TextInput](/core/text-input) component selectors:
```

But the example shows button style prop api

```mdx
<SelectorsTable
  data={TextInputStylesApi}
  component="Button"
  withTableBorder={false}
  fixedLayout={false}
/>
```

This change the example to `TextInput`.

Result:

image


rtivital:
Thanks!

PR #5299, opened 2023-11-22
Add Menu trigger="both" option

mg-aceik:
https://github.com/orgs/mantinedev/discussions/2389#discussioncomment-7626296

PR #5335, opened 2023-11-27
Implement options to make the Menu compatible with WAI-ARIA recommendations for navigation

mg-aceik:
Following on from https://github.com/mantinedev/mantine/pull/5299

I've added some further configuration to allow the Menu to be compatible with WAI-ARIA recommendations for navigation, in the case that it is being used as a navigation instead of a menu.

I've also updated the doc site but note that the behaviour on the doc site seems different than on storybook, so not sure what's going on there, perhaps some kind of focus trap in place on the doc site?


rtivital:
Thanks!

Issue #5250, opened 2023-11-14
RTL issues


CripyIce:
### What package has an issue?

@mantine/core

### Describe the bug

- Icons and text have no space between them - https://mantine.dev/core/segmented-control/#react-node-as-label
![image](https://github.com/mantinedev/mantine/assets/26387823/6cbb7285-530f-4251-8d18-64e6820c2873)

- No space between text and divider - https://mantine.dev/core/autocomplete/#group-options
![image](https://github.com/mantinedev/mantine/assets/26387823/d1168847-6a9f-4879-ac0a-aa3d8eecc6b1)

- Text is too close to the edge - https://mantine.dev/core/pill/
![image](https://github.com/mantinedev/mantine/assets/26387823/d7ef0fcf-699e-49bd-aa02-92bbf0126f88)

- Text stays on the left (also in nested) - https://mantine.dev/core/nav-link/
- No padding for the content - https://mantine.dev/core/tabs/#vertical-tabs-placement
![image](https://github.com/mantinedev/mantine/assets/26387823/9f10de32-df06-4f93-b601-0454856695d9)

- Extra padding on the right - https://mantine.dev/core/drawer/
![image](https://github.com/mantinedev/mantine/assets/26387823/beae6c4a-3fd2-4e69-bee9-70979c3dc25f)

- Extra padding on the right and too close to the edge on the left - https://mantine.dev/core/modal/#fullscreen
![image](https://github.com/mantinedev/mantine/assets/26387823/321db790-c44c-4748-88b2-795eb0ebb6d8)
![image](https://github.com/mantinedev/mantine/assets/26387823/c3393597-4ea7-45d1-8206-30c3f52613f4)

- Positions start/end are not truly working - https://mantine.dev/core/indicator/
![image](https://github.com/mantinedev/mantine/assets/26387823/890b2e1d-940b-435e-87b5-f149ac8cfee0)

- Order is incorrect (should be command->shift->m rather then command->m->shift) - https://mantine.dev/core/kbd/
- “Show more” stays in the left - https://mantine.dev/core/spoiler/
![image](https://github.com/mantinedev/mantine/assets/26387823/c3b00cf4-c8bb-44fc-990c-6c00de577980)

- Icons and text are too close - https://mantine.dev/others/spotlight/
![image](https://github.com/mantinedev/mantine/assets/26387823/72819406-1593-4242-ae92-f4e9a06a99cc)

[...]

PR #5347, opened 2023-11-29
[@mantine/core] [@mantine/spotlight] [@mantine/demos] Fixes for RTL

mmounirf:
This PR includes fixes for most of the issues listed here https://github.com/mantinedev/mantine/issues/5250 except for Indicator component which the scope of the fix would require changes in how the position property apply correct styles.


rtivital:
Thanks!

PR #5351, opened 2023-11-30
Remove obsolete prop from example in Pagination.demo.icons.tsx

julamb:
The docs still contain a reference to the `position` prop on the `Pagination` component (in the [`change-icons`](https://mantine.dev/core/pagination/#change-icons) example), but apparently it's been removed. I updated the text to match the actual code.


rtivital:
Thanks!

PR #5375, opened 2023-12-06
[@mantine/core] Fix line-height resolver on style props

riettsruff:
I think previously there was a typo referring to `theme.fontSizes`


rtivital:
Thanks!

PR #5377, opened 2023-12-06
[@mantine/core] Select component breaks if user translates page using google translate

### LINKED ISSUE #5260: Error when using <Select /> and Google Chrome with translate

@mantine/core

### Describe the bug

When using Google Chrome and using the "translate this page" functionality, and the user selects an option on the Select or MultiSelect component (perhaps other components as well?), then the error from the screenshot will occur: 

Tested on the newest version of Chrome (I think), version 119 and Mantine 7.2.2 (but I think this might have been an issue for a longer period of time).

Both happens on my website and on the Mantine docs.

error

7.2.2

### If possible, please include a link to a codesandbox with a minimal reproduction

_No response_

None

### Do you want to send a pull request with a fix?

None

### Possible fix

Seems like you can disable Google translate option entirely, but of course nice to allow on your website: https://stackoverflow.com/a/76004257/8963287

Also, the problem might be that the Select (and similar components) uses React Fragment instead of a div or similar. This answer says switching away from Fragments solved a similar issue: https://stackoverflow.com/a/74627339/8963287

manuelbosi:
fixes #5260

### DIFF
--- a/packages/@mantine/core/src/components/Combobox/OptionsDropdown/OptionsDropdown.tsx
+++ b/packages/@mantine/core/src/components/Combobox/OptionsDropdown/OptionsDropdown.tsx
@@ -47,7 +47,7 @@ function Option({ data, withCheckIcon, value, checkIconPosition, unstyled }: Opt
         aria-selected={isValueChecked(value, data.value)}
       >
         {checkIconPosition === 'left' && check}
-        <label>{data.label}</label>
+        <span>{data.label}</span>
         {checkIconPosition === 'right' && check}
       </Combobox.Option>
     );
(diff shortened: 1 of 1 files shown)


packages/@mantine/core/src/components/Combobox/OptionsDropdown/OptionsDropdown.tsx
```diff
@@ -47,7 +47,7 @@ function Option({ data, withCheckIcon, value, checkIconPosition, unstyled }: Opt
         aria-selected={isValueChecked(value, data.value)}
       >
         {checkIconPosition === 'left' && check}
-        {data.label}
+        <label>{data.label}</label>
```
rtivital, line 50: Should be span element, not label

PR #5400, opened 2023-12-09
[@mantine/core] add non-empty check for portal className

jvdsande:
As discussed on Discord, filter out the tokens passed to `classList.add` in modal to avoid the following DOM exception:

```
Uncaught DOMException: Failed to execute 'add' on 'DOMTokenList': The token provided must not be empty.
```

I added two non-regression tests regarding this.


rtivital:
Thanks!

PR #5403, opened 2023-12-11
Menu trigger=click-hover updates for UX and accessibility

mg-aceik:
This is a further change following on from https://github.com/mantinedev/mantine/pull/5335

A UX review of click-hover showed that if a menu has been clicked then this should override the hover until an item is clicked or a click-outside has happened.

I've also added unit tests to cover this functionality.

### DIFF
--- a/packages/@docs/demos/src/demos/core/Menu/Menu.demo.navigation.tsx
+++ b/packages/@docs/demos/src/demos/core/Menu/Menu.demo.navigation.tsx
@@ -9,18 +9,20 @@ import { Group, Menu } from '@mantine/core';
 function Demo() {
   return (
     <Group>
-      {Array(4).fill(0).map((e, i) => (
-        <Menu
-          key={i}
-          trigger="click-hover"
-          loop={false}
-          withinPortal={false}
-          trapFocus={false}
-          menuItemTabIndex={0}
-        >
-          <DemoMenuItems />
-        </Menu>
-      ))}
+      {Array(4)
+        .fill(0)
+        .map((e, i) => (
+          <Menu
+            key={i}
+            trigger="click-hover"
+            loop={false}
+            withinPortal={false}
+            trapFocus={false}
+            menuItemTabIndex={0}
+          >
+            <DemoMenuItems />
+          </Menu>
+        ))}
     </Group>
   );
 }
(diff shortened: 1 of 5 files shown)


packages/@docs/demos/src/demos/core/Menu/Menu.demo.navigation.tsx
```diff
@@ -4,43 +4,45 @@ import { MantineDemo } from '@mantinex/demo';
...
-      <Menu
-        trigger="click-hover"
-        loop={false}
-        withinPortal={false}
-        trapFocus={false}
-        menuItemTabIndex={0}
-      >
-        <DemoMenuItems />
-      </Menu>
+      {[...Array(navMenus)].map((e, i) => (
```
rtivital, line 34: - Do not use map inside jsx, create a separate variable.
- There is no reason to store `navMenus` in a variable, it is used only once
- Instead of `[...Array(navMenus)]` syntax, use `Array(4).fill(0)` to be consistent with other demos.

packages/@docs/demos/src/demos/core/Menu/Menu.demo.navigation.tsx
```diff
@@ -4,17 +4,25 @@ import { MantineDemo } from '@mantinex/demo';
...
+        .map((e, i) => (
+          <Menu
+            key={i}
+            trigger="click-hover"
+            loop={false}
+            withinPortal={false}
+            trapFocus={false}
+            menuItemTabIndex={0}
+          >
+            <DemoMenuItems />
```
rtivital, line 23: This will throw an error if copy pasted into a project, replace `<DemoMenuItems />` with `{/* ... menu items */}`

packages/@docs/demos/src/demos/core/Menu/Menu.demo.navigation.tsx
```diff
@@ -4,17 +4,25 @@ import { MantineDemo } from '@mantinex/demo';
...
 function Demo() {
   return (
     <Group>
-      <Menu trigger="click-hover" loop={false} withinPortal={false} trapFocus={false} menuItemTabIndex={0}>
-        {/* ... menu items */}
-      </Menu>
-      <Menu trigger="click-hover" loop={false} withinPortal={false} trapFocus={false} menuItemTabIndex={0}>
-        {/* ... menu items */}
-      </Menu>
+      {Array(4)
```
rtivital, line 12: Do not use map inside jsx, create a new variable with the results of map function

rtivital (changes requested):

rtivital:
Thanks!

Issue #5411, opened 2023-12-12
From: zodResolver provide last error message instead of first one


GuiLec:
### Versions

```
    "@mantine/core": "^7.3.1",
    "@mantine/form": "^7.2.2",
    "@mantine/hooks": "7.3.0",
``` 

### Context

When using a zodResolver on a form, if a field has 2 checks failing, the result of the zodResolver for the field, should be the first error.

```typescript
const check = z.object({
  field: z
    .string()
    .refine((value) => value.length > 0, {
      message: 'message 1',
    })
    .refine((value) => value.length > 0, {
      message: 'message 2',
    }),
});

zodResolver(check)

``` 

**Expected behaviour**: if field is "", we expect an error message 'message 1' for 'field'
**Actual behaviour**: if field is "", the actual error message is 'message 2' for 'field'

### Sandbox

Find [here](https://stackblitz.com/edit/react-kef6dq?file=src%2FApp.js) a minimal sandbox with the example herebefore

### Suggested fix

Please look my pull request: https://github.com/mantinedev/mantine/pull/5412

PR #5412, opened 2023-12-12
fix(form/zod-resolver): fix error message priority

GuiLec:
This PR is to solve the following issue: https://github.com/mantinedev/mantine/issues/5411

### DIFF
--- a/packages/@mantine/form/src/resolvers/zod-resolver/zod-resolver.ts
+++ b/packages/@mantine/form/src/resolvers/zod-resolver/zod-resolver.ts
@@ -30,7 +30,7 @@ export function zodResolver<T extends Record<string, any>>(schema: ZodSchema<T>)
 
     const results: Record<string, any> = {};
 
-    (parsed as ZodParseError).error.errors.forEach((error) => {
+    (parsed as ZodParseError).error.errors.toReversed().forEach((error) => {
       results[error.path.join('.')] = error.message;
     });
 


packages/@mantine/form/src/resolvers/zod-resolver/zod-resolver.ts
```diff
@@ -30,7 +30,7 @@ export function zodResolver<T extends Record<string, any>>(schema: ZodSchema<T>)
 
     const results: Record<string, any> = {};
 
-    (parsed as ZodParseError).error.errors.forEach((error) => {
+    (parsed as ZodParseError).error.errors.toReversed().forEach((error) => {
       results[error.path.join('.')] = error.message;
```
GuiLec, line 34: If a field has multiple errors, the first ones are overwritten by the last one. Reversing errors order makes sure that the first error message is returned in the result

rtivital:
`zodResolver` is `@mantine/form` package was deprecated in the previous minor version, if you need to fix this issue, send a PR to https://github.com/mantinedev/mantine-form-zod-resolver

rtivital:
I also think that it is opinionated and should be configurable with resolver option.

GuiLec:
> `zodResolver` is `@mantine/form` package was deprecated in the previous minor version, if you need to fix this issue, send a PR to https://github.com/mantinedev/mantine-form-zod-resolver

@rtivital done here: https://github.com/mantinedev/mantine-form-zod-resolver/pull/1

PR #5452, opened 2023-12-20
[@mantine/vanilla-extract] Make themeToVars build time compatible

### LINKED ISSUE #5451: themeToVars isn't build time compatible
### Dependencies check up

7.3.2

@mantine/vanilla-extract

Next.js

### Describe the bug

When referencing `vars` resulting from `themeToVars` in a Next.js server component I get the following error:
```
Error: _mantine_vanilla_extract__WEBPACK_IMPORTED_MODULE_2__.themeToVars is not a function
```

The problem stems from the `theme-to-vars` module (and some of the imported `@mantine/core` modules) being marked as `'use_client'`. The modules in question don't need the `'use_client'` directive.

As far as I understand this means `themeToVars` (and the resulting `vars` object) cannot be referenced in build time code, which defeats the purpose of using vanilla-extract.

_No response_

### Possible fix

Add the following files to [scripts/build/rollup/rollup-exclude-use-client.ts](https://github.com/mantinedev/mantine/blob/master/scripts/build/rollup/rollup-exclude-use-client.ts):
- `theme-to-vars`
- `core/MantineProvider/merge-mantine-theme/merge-mantine-theme`
- `core/utils/deep-merge/deep-merge`

### Self-service

godiagonal:
Closes #5451.

### DIFF
--- a/scripts/build/rollup/rollup-exclude-use-client.ts
+++ b/scripts/build/rollup/rollup-exclude-use-client.ts
@@ -1,6 +1,7 @@
 // files names that should not have use client directive at the top of the output file
 export const ROLLUP_EXCLUDE_USE_CLIENT = [
   'index',
+  'core/utils/deep-merge/deep-merge',
   'core/utils/units-converters/rem',
   'core/utils/units-converters/px',
   'core/factory/create-polymorphic-component',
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

Issue #5427, opened 2023-12-15
Combobox onClose does not work


crabvk:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.3.2

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Firefox

### Describe the bug

I took the first code example from the [Combobox examples page](https://mantine.dev/combobox/?e=BasicSelect) called "Basic select". Added `onClose={() => alert('onClose')}` to Combobox component. The event does not trigger when I close dropdown. `onOpen` works just fine though.

### If possible, include a link to a codesandbox with a minimal reproduction

https://codesandbox.io/p/sandbox/infallible-leakey-5w99kk

### Self-service

- [ ] I would be willing to implement a fix for this issue

kartikm02:
I would be willing to fix this issue

ID-JA:
@rtivital  
I've reviewed the problem, and it appears that the onClose prop isn't getting triggered, even though we've passed it to the Combobox component.

I've made a small adjustment to the `Combobox.tsx` file. With these changes, only one of the close handlers will be triggered (`onClose` or `onDropdownClose`).

`Combobox.tsx`
```diff
 <Popover
        opened={store.dropdownOpened}
        {...others}
-       onClose={store.closeDropdown}
+      onClose={onClose || store.closeDropdown}
        withRoles={false}
        unstyled={unstyled}
      >
        {children}
</Popover>

```

ID-JA:
@rtivital 
I've looked at the issue, and it seems like the `onClose` prop isn't being triggered even if we passed it as a prop for the `Combobox` component.

I made a minor modification to the `Combobox.tsx` file. 
```diff 
 <Popover
        opened={store.dropdownOpened}
        {...others}
-        onClose={store.closeDropdown}
+       onClose={onClose || store.closeDropdown}
        withRoles={false}
        unstyled={unstyled}
      >
        {children}
</Popover>
```
I've check the `Select` component behaviour (it closed on clicking outside).

rtivital:
You are welcome to submit a PR with a fix

PR #5473, opened 2023-12-24
Fix: Combobox onClose does not work #5427

kartikm02:
[https://github.com/mantinedev/mantine/issues/5427](url)

onClose function was getting called directly without going through the onChange events. So, onClose was removed directly and now the dropdown is closing through onChange. Because of which events are passing and alerts are working fine,


rtivital:
This PR introduces incorrect behavior – Select dropdown is not closed on click outside

PR #5481, opened 2023-12-26
[@mantine/core] ScrollAreaAutosize: fix overflow

ayovev:
While using the `ScrollAreaAutosize` component I noticed that when used in conjunction with the `Table` component the data would overflow out of the viewport, particularly on smaller screen widths and mobile devices (screenshot 1 and 2).

**Screenshot 1 (Issue)**
Screenshot 2023-12-26 at 14 08 01

**Screenshot 2 (Issue)**
Screenshot 2023-12-26 at 14 08 13

I was able to fix this by adding `overflow: auto` to the `style` prop of the `ScrollAreaAutosize` component which renders it in the outermost parent `div` of the `ScrollArea` component (screenshot 3 and 4). I don't think this is an issue specific to my project so I'm making this update in hopes that it helps the library as a whole.

**Screenshot 3 (Fix)**
Screenshot 2023-12-26 at 14 12 11

**Screenshot 4 (Fix)**
Screenshot 2023-12-26 at 14 16 27

### DIFF
--- a/packages/@mantine/core/src/components/ScrollArea/ScrollArea.tsx
+++ b/packages/@mantine/core/src/components/ScrollArea/ScrollArea.tsx
@@ -206,7 +206,7 @@ export const ScrollAreaAutosize = factory<ScrollAreaFactory>((props, ref) => {
   } = useProps('ScrollAreaAutosize', defaultProps, props);
 
   return (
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #5509, opened 2024-01-03
[@mantine/core] Combobox: Fix #5427

### LINKED ISSUE #5427: Combobox onClose does not work

ID-JA:
Fixes #5427

### DIFF
--- a/packages/@mantine/core/src/components/Combobox/Combobox.tsx
+++ b/packages/@mantine/core/src/components/Combobox/Combobox.tsx
@@ -123,6 +123,7 @@ export function Combobox(_props: ComboboxProps) {
     store: controlledStore,
     vars,
     onOptionSubmit,
+    onClose,
     size,
     dropdownPadding,
     resetSelectionOnOptionHover,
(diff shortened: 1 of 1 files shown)


packages/@mantine/core/src/components/Combobox/Combobox.tsx
```diff
@@ -159,7 +160,7 @@ export function Combobox(_props: ComboboxProps) {
       <Popover
         opened={store.dropdownOpened}
         {...others}
-        onClose={store.closeDropdown}
+        onClose={onClose || store.closeDropdown}
```
rtivital, line 163: I suppose that it would be better to call both functions:

`() => {onClose?.(); store.closeDropdown()}`

ID-JA: I agree it's a good idea to call both functions. I've updated it to trigger both close handlers

rtivital:
Thanks!

Issue #5540, opened 2024-01-09
Anchor in RightSection of NumberINput not Clickable When Input is Disabled


elibroftw:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.4.1

### What package has an issue?

@mantine/core

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

None

### Describe the bug

When I have a disabled input with an Anchor in the right section, the cursor remains as non-clickable when the input is disabled. This used to work in v6 of mantine.

### If possible, include a link to a codesandbox with a minimal reproduction

[...]

PR #5544, opened 2024-01-11
Fix: RGBA calculations && issue 5540

kjk7034:
## Fix 1 : add "/" to RGBA calculations

When calculating RGBA, errors occur in the following cases:

Success: rgb(255, 0, 255, 1)
Failure: rgb(255, 0, 255 / 1)

Upon executing 'rgb(255, 0, 255/ 1),' the expected result is {r: 255, g: 0, b: 255, a: 1}, but the actual result is {r: 255, g: 0, b: 2551, a: 1}.

commit :  https://github.com/mantinedev/mantine/pull/5544/commits/3caefc4768fb62924ac90d078333a9a7365502b1

## Fix 2 : error in the expression caused by missing parentheses

Resolved the issue: https://github.com/mantinedev/mantine/issues/5540

commit : https://github.com/mantinedev/mantine/pull/5544/commits/b941c67c3c5f1cde17dd917a280841112d13e90c


soham2k06 (approved):

rtivital:
Thanks!

PR #5587, opened 2024-01-17
use `leftSection` in NavLink.demo.active.tsx

craigmichaelmartin:
With mantine version 7, [components that previously had an `icon` prop now use `leftSection` instead of `icon`](https://mantine.dev/changelog/7-0-0/#left-and-right-section).

This PR updates a demo example that still uses `icon`.

Thanks for all your work on Mantine!


rtivital:
Thanks!

PR #5588, opened 2024-01-17
Allow useComputedColorScheme defaultValue to be undefined

c0nd3v:
If we have getInitialValueInEffect, we might as well be able to set defaultValue to undefined. Preferably I'd like to completely omit the parameter...


rtivital:
Thanks!

PR #5597, opened 2024-01-18
[docs] Fix version select to appear in front of nav menu

anthony-healthstreet:
Code version select to go over navbar


rtivital:
Thanks for showing interest in contributing to Mantine, the issue has been fixed in another PR

PR #5605, opened 2024-01-21
[Docs] Grammar fix in the bug report issue template

sxflynn:
Fixed a minor grammar issue with the bug report issue template. I added a period instead of the comma to separate the two independent clauses.


rtivital:
Thanks!

PR #5608, opened 2024-01-22
feat(CodeHighlightTabs): add withCopyButton prop for consistency with CodeHighlight

sxflynn:
The `CodeHighlight` has a prop called `withCopyButton` which allows you to hide the copy button if set to `false`. This PR adds the `withCopyButton` prop to `CodeHighlightTabs` so it can also have that capability.


rtivital:
Thanks!

Issue #5545, opened 2024-01-11
Can't change the name of series in Charts


hitfm00:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.4.1

### What package has an issue?

@mantine/core

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

For example:
**Data**
`[{date: "00:00", income: 2}, {date: "01:00", income: 3}, {date: "02:00", income: 95},…]`

**Component:**
```
<LineChart
    h={300}
    data={data}
    dataKey="date"
    series={[{ name: 'income', color: 'white' }]}
/>
```
And this name in the `series` prop must necessarily match the name in the `data`, I would like to suggest that you can add your own name, and not just by the key in the object

**Example**:
`series={[{ name: 'income', color: 'white', label: 'All money' }]}
`
Because for now I can't translate this name with my `i18next` library 

### Possible fix

Add new prop to series object
```
export interface ChartSeries {
    name: string;
    color: MantineColor;
    label?: string // Will show this if exists in tooltip
}
```

### Self-service

- [ ] I would be willing to implement a fix for this issue

soham2k06:
Hello,

Is your problem about not being able to see the name you have added or you are not able to see the whole chart.

Also, please show your whole code because I tried code and data that you shared with us.

My code:

import { LineChart } from "@mantine/charts";

export const data = [
  { date: "00:00", income: 2 },
  { date: "01:00", income: 3 },
  { date: "02:00", income: 95 },
];

export default function Demo() {
  return (
    <LineChart
      h={300}
      data={data}
      dataKey="date"
      series={[{ name: "income", color: "yellow" }]}
    />
  );
}

Thank you.

hitfm00:
@soham2k06  Hey, everything works, I just suggest that we can change the name in the tooltip, at the moment we can only show what comes from the request, and this is `income` (as shown in the screenshot), but instead of income I want to see a different name, which is what I I'll write `t('translatedText')` from `i18next`, for example

<img width="309" alt="Снимок экрана 2024-01-13 в 17 31 24" src="https://github.com/mantinedev/mantine/assets/21322208/cdca3eca-2aed-4fd1-a8da-31f7a6e112b6">

At the moment, in the name of tooltip we only see the key in the object, and not what I need.
The keys may not have the name needed to be displayed to the user

[...]

PR #5617, opened 2024-01-22
Add data formatter for charts

OlfillasOdikno:
This PR is implementing formatters for charts, as in [recharts](https://github.com/recharts/recharts/blob/a8a26db1df5353ea66d94e3509894d8165781428/src/component/DefaultLegendContent.tsx#L49).

This should also address [#5545](https://github.com/mantinedev/mantine/issues/5545).


rtivital:
`valueFormatter` feature has already been added in 7.5.0 release – https://mantine.dev/charts/area-chart/#value-formatter

OlfillasOdikno:
`valueFormatter` only formats the values, but not the labels.

rtivital:
Series labels can be changed in the `series` array – https://mantine.dev/charts/area-chart/#series-labels

Issue #5621, opened 2024-01-23
Pills inside MultiSelect component are removable even if the entry itself is disabled


lukasbash:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.4.2

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

All

### Describe the bug

Data entries of the MultiSelect component have an optional `disabled` flag. Setting this flag correctly reflects inside the actual Popover but not on the Pill-Input if the entry has been selected already.
I would expect disabled items to be immutable in both parts of the component.

### Possible fix

Either pass the disabled property to the pill and remove/disable the close button, or provide a way to pass down pill and/or pill-input props.

### Self-service

- [ ] I would be willing to implement a fix for this issue

Phirb:
```diff
diff --git a/packages/@mantine/core/src/components/MultiSelect/MultiSelect.tsx b/packages/@mantine/core/src/components/MultiSelect/MultiSelect.tsx
index 2c0e0ffd1..ecbb556bc 100644
--- a/packages/@mantine/core/src/components/MultiSelect/MultiSelect.tsx
+++ b/packages/@mantine/core/src/components/MultiSelect/MultiSelect.tsx
@@ -239,7 +239,7 @@ export const MultiSelect = factory<MultiSelectFactory>((_props, ref) => {
   const values = _value.map((item, index) => (
     <Pill
       key={`${item}-${index}`}
-      withRemoveButton={!readOnly}
+      withRemoveButton={!readOnly && !(optionsLockup[item]?.disabled)}
       onRemove={() => setValue(_value.filter((i) => item !== i))}
       unstyled={unstyled}
       {...getStyles('pill')}
```
I believe this could be fixed with this change. May I make a PR for this issue?

PR #5630, opened 2024-01-24
[@mantine/core] fix: multiselect disabled functionality for pills

Phirb:
Description:
Makes it so that pills in the input field are not removable if its respective option is disabled.

Issues addressed:
#5621 

Video of implementation:

https://github.com/mantinedev/mantine/assets/39493240/295c6c73-e772-44cd-8df5-c1f88f9a4fee


rtivital:
Thanks!

PR #5641, opened 2024-01-26
[@mantine/core] Removed flex property for Image css

msv96:
### Description:
Removed flex property as it is enabled when the parent uses `display: flex;`.

### Issue:
Screenshot 2024-01-26 131952


rtivital:
Why it should be removed?

msv96:
Because the parent element has `display: flex;`, the image is not showing since `flex-basis: 0%;`.

msv96:
Or we can make `flex: 1`

rtivital:
It works as intended inside a flexbox container. I do not see any issue with styles on codesandbox – https://codesandbox.io/p/sandbox/cranky-golick-p7t4td?file=%2Fsrc%2FApp.tsx%3A6%2C22

msv96:
Hi @rtivital, found it. When we have `* { min-width: 0px; }`, it behaves like I said earlier. min-width removes the default values for an element and applies min-width - 0px.

https://codesandbox.io/p/sandbox/headless-cookies-6py3mw?file=%2Fsrc%2FApp.tsx%3A16%2C18

Source - https://www.youtube.com/watch?v=cH8VbLM1958

rtivital:
Mantine does not include `* { min-width: 0px; }`, thus this case is not handled by the library.

yhranik-vp:
Hello @rtivital , I'm experiencing the same issue, but only on Firefox.

You can reproduce with the sanbox link you shared.

Chrome
Screenshot 2024-06-06 at 13 35 31

Firefox
Screenshot 2024-06-06 at 13 35 22

PR #5649, opened 2024-01-26
[@mantine/form] call onValuesChange with previous form values

mariansimecek:
Hi,
based on #5644, I tried to implement that. 
I think it is convenient to have the previous values in that callback function. You can then check which values changed and how.


rtivital:
Thanks!

PR #5655, opened 2024-01-27
[@mantine/core] Fixed Pagination autoContrast given by MantineProvider

dswbx:
The `Pagination` component's `autoContrast` worked when property is added to the component. But if the global `autoContrast` from theme object was given, it was ignored.

Worked:
```jsx
<Pagination total={45} autoContrast />
```

Didn't work:
```jsx
<MantineProvider theme={{ autoContrast: true }}>
  <Pagination total={45} />
</MantineProvider>
```

Also added a pagination story to show the case along with a button (which can be removed of course).


rtivital:
Thanks for reporting the issue! The fix that you've provided is not valid, I've implemented it differently. The fix will be available with the next patch.

PR #5665, opened 2024-01-28
[@mantine/hooks] change useDebouncedState setter to accept set function

cristianghita24:
doesn't change current behavior - it enables the option to set the state via a SetStateAction that gives you access to its previous state value


rtivital:
Thanks!

PR #5668, opened 2024-01-28
fix: Make explicit imports of `index.js` in main library files.

### LINKED ISSUE #4991: Imports fail with TS "moduleResolution": "nodenext"
### What package has an issue

@mantine/core

### Describe the bug

When I set "moduleResolution": "nodenext" in tsconfig, TS throws an error for all Mantine imports. I am using the Epic Remix Stack from Kent C. Dodds, and if I go back to "moduleResolution": "node", Mantine imports work fine, but all kinds of other errors pop up all over the project. Now I'm just adding a @ts-ignore for each Mantine import, but that makes auto-import impossible, which is not great.

Is there a better workaround? What would it take to fix this?

7.1.2

_No response_

No

Yes

### Possible fix

_No response_

waweber:
Fixes #4991 by making imports of subpackages reference the `index.js` file explicitly.

The newer `"nodenext"` resolution rules require relative imports to include the file extension, and implicit index imports are not supported. Under these rules, exports like `export * from './core';` fail. This change works around these rules by changing these imports to `export * from './core/index.js';` which resolves correctly under both `"nodenext"` and the default `"node"` resolution rules. It seems sufficient to only update the imports in the main `index.ts` files, TypeScript appears to use the default rules with deeper imports.

This is likely the less hacky solution, but this might break people's bundlers if they don't handle imports that include extensions.


rtivital:
Thanks!

rtivital:
Unfortunately, it breaks Storybook, I had to revert this change

PR #5673, opened 2024-01-29
[@mantine/core] Expose useTabsContext to enable reusability

vizath:
Not sure if this lives in the spirit of the library, but it would allow more customization of the Tabs component, allowing users to have full control over the component being used, allowing for better reusability.

eg

```
import { useTabsContext, SegmentedControl, SegmentedControlItem } from "@mantine/core";

type Props = {
  items: SegmentedControlItem[];
};

const TabSegmentedControl = ({ items }: Props) => {
  const ctx = useTabsContext();

  return (
    <SegmentedControl
      value={ctx.value}
      onChange={(value) => ctx.onChange(value)}
      data={items}
    />
  );
};

export default TabSegmentedControl;
```


rtivital:
Thanks for showing interest in contributing to the library! This code is not a part of the public API and was never intended to be. It includes implementation that can have breaking changes in any minor/patch releases. It is not planned to make it public.

PR #5677, opened 2024-01-30
[@mantine/core] OptionsDropdown: Remove redundant styles

### LINKED ISSUE #5676: Redundant styles for `ScrollArea` in `Autocomplete` and `Select` dropdown lists
### Dependencies check up

7.5.0

@mantine/core

All

### Describe the bug

`Autocomplete` and `Select` components use `ScrollArea` component for dropdown lists by default, but `padding` is displayed incorrectly for dropdown lists. This is most noticeable in [this example](https://mantine.dev/core/autocomplete/#dropdown-padding).

It looks like [these styles](https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/components/Combobox/Combobox.module.css#L192) are redundant for [this component](https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/components/Combobox/OptionsDropdown/OptionsDropdown.tsx#L137).

_No response_

### Possible fix

_No response_

### Self-service

yverby:
Fixes #5676

#### Before:
dropdown-before

#### After:
dropdown-after

### DIFF
--- a/packages/@mantine/core/src/components/Combobox/Combobox.module.css
+++ b/packages/@mantine/core/src/components/Combobox/Combobox.module.css
@@ -189,15 +189,6 @@
 }
 
 /* ------- OptionsDropdown ------- */
-.optionsDropdownScrollArea {
-  margin-right: calc(var(--_combobox-padding) * -1);
-
-  @mixin rtl {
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

Issue #5680, opened 2024-01-30
Indicator with "processing" flag looks very jittery on a lower-resolution (96 dpi) monitor


matthiasfeist:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.5.0

### What package has an issue?

@mantine/core

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

When enabling the `processing` property for the indicator, the animation looks fine on my HD mac book display but once I move it to an older monitor with a lower resolution, the animation looks very "jittery".
I reproduced the issue on the docs page for the indicator: https://mantine.dev/core/indicator/

I think the problem is that the animation is relatively subtle and therefore doesn't move many pixels on a regular 96 dpi monitor, resulting in a jittery "jumping" effect.

I tried to do a screen recoding of the effect and it's a bit less pronounced in the recording than on screen, but it's still visible.
https://github.com/mantinedev/mantine/assets/1589740/d34e2b0c-267b-4320-9b5e-f406fef7392f

### Possible fix

I can think of some fixes:
1. make the animation a bit larger in general
2. make it depending on the dpi of the screen (not sure if that's even possible)
3. introduce a prop that enables the developer to control the size (effectively making it the developers problem 😆)

### Self-service

- [x] I would be willing to implement a fix for this issue

rtivital:
I do not have a low resolution monitor to validate this. You are welcome to send a PR with non-breaking changes.

matthiasfeist:
cool, I'll try to see if I can fix this. 
I have no experience with the codebase so any pointers that you can give me are appreciated :)

rtivital:
https://mantine.dev/contribute/#get-started-with-mantine-locally

PR #5682, opened 2024-01-30
fixed jittery animation on lower-res monitors

matthiasfeist:
As reported in isssue #5680, this PR changed the Indicator component's "processing" animation to be smoother on monitors with a 96dpi resolution.
This is done by removing the blurring of the shadow and expanding the radius of the shadow slightly. It looks a little different (larger) than before on high-res monitors (Macbooks etc) but now also works on older screens.


rtivital:
Thanks!

matthiasfeist:
my pleasure and thanks for such an amazing component lib!

PR #5684, opened 2024-01-30
Fix slow rerendering of large tables

Smert:
Hi! I have lags with rerendering of large tables with memoized rows:

```typescript
const TestRow = memo(function Row() {
  return (
    <Table.Tr>
      {new Array(10).fill(0).map((_, index) => (
        <Table.Td key={index}>Cell</Table.Td>
      ))}
    </Table.Tr>
  );
});

function TestTable() {
  const [counter, setCounter] = useState(0);
  return (
    <>
      <p>Counter: {counter}</p>
      <button onClick={() => setCounter(counter + 1)}>Increment</button>
      <Table>
        <Table.Tbody>
          {new Array(1000).fill(0).map((_, index) => (
            <TestRow key={index}/>
          ))}
        </Table.Tbody>
      </Table>
    </>
  );
}
```

Eventloop lags for 200ms when I click on the button in this simple example. But all rows are memoized.
I noticed then `TableContext` have a new value every render. It leads to rerender of `Table.Tr` and `Table.Td`:

Screenshot 2024-01-30 at 19 52 06

Memoizing of the context and `useStyle` can improve performance:
Screenshot 2024-01-30 at 19 48 48

I can suggest my decision of this problem in the pull request. Maybe you have another ideas.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mantine-a91763c0e2",
-  "version": "7.10.1",
+  "version": "7.10.2",
   "description": "Mantine Components Monorepo",
   "packageManager": "yarn@4.2.2",
   "license": "MIT",
--- a/packages/@mantine/carousel/package.json
+++ b/packages/@mantine/carousel/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mantine/carousel",
-  "version": "7.10.1",
+  "version": "7.10.2",
   "description": "Embla based carousel",
   "homepage": "https://mantine.dev/x/carousel/",
   "license": "MIT",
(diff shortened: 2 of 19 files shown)


packages/@mantine/core/src/core/styles-api/use-styles/use-styles.ts
```diff
@@ -46,36 +59,58 @@ export function useStyles<Payload extends FactoryPayload>({
 }: UseStylesInput<Payload>): GetStylesApi<Payload> {
   const theme = useMantineTheme();
   const classNamesPrefix = useMantineClassNamesPrefix();
-  const themeName = (Array.isArray(name) ? name : [name]).filter((n) => n) as string[];
+  const themeName = useMemo(
+    () => (Array.isArray(name) ? name : [name]).filter((n) => n) as string[],
+    [String(name)]
```
2jordan3, line 64: Putting **String(name)** in the dependency array may not work as intended.
How about putting the value of **name** directly in the dependency array?

Smert: Sorry, reopened in #6422

PR #5689, opened 2024-01-31
[@mantine/core] Allow value reset in SegmentedControl

gl3nn:
Currently it's not possible to deselect/reset the value of the `SegmentedControl`. You can start out with an empty value, if you take for instance the empty string as value and make sure that in the `data` property it doesn't map to anything. If however you then select a value, and then change the value state back to the empty string, the `SegmentedControl` will still visually show your last selected value to be selected (however `data-selected=true` will no longer be present).

My PR here fixes this behavior and allows you, with a controlled value, to make the component unselected / without anything selected.


rtivital:
Thanks!

PR #5694, opened 2024-02-01
[@mantine/core] Import useIsomorhicEffect instead of creating new one

qweered:


qweered:
Shoupd we replace most useEffects in codebase with useIsomorphicEffects? I can do that

rtivital:
useIsomorphicEffects replaces useLayoutEffect, not useEffect.

qweered:
Also there is creation of useIsomorphicEffect in mantine/form, but I believe it should bot be dependent in mantine/hooks

rtivital:
Thanks!

PR #5696, opened 2024-02-01
`form.setFieldValue` based on previous value

twiddler:
Hey everyone! :wave: 

I would like to set the value of a field `foo` based on the value it currently has. This is currently possible by using 

```ts
form.setValues(prev => ({ foo: prev.foo + 1 }))
```

but (to the best of my knowledge) not by something like

```ts
form.setFieldValue("foo", prev => prev + 1)
```

I thought I'd give it a shot to make this a proper pull request, but I am facing type issues in the test I added:

> Parameter 'prev' implicitly has an 'any' type. ts(7006)

I would really appreciate if someone could take it from here. Thanks! :pray:


rtivital:
Please fix test errors

twiddler:
I ran `npm run prettier:write` to fix the prettier errors. The type error that now shows is the one that I do not know how to solve. I'd appreciate any help. :pray:

twiddler:
Thanks @Pr0gramWizard! :pray:

rtivital:
Thanks!

PR #5697, opened 2024-02-01
fix: Make explicit imports of `index.js` in main library files.

### LINKED ISSUE #4991: Imports fail with TS "moduleResolution": "nodenext"

waweber:
Fixes #4991

Updated https://github.com/mantinedev/mantine/pull/5668 to make Storybook aware that .js files come from the corresponding .ts files.

This change shouldn't affect users of the library since they are already only consuming the transpiled files.


rtivital:
Thanks!

PR #5698, opened 2024-02-01
[@mantine/core]: Init Composite component

ktunador:
A new component to create `Composite` components with focus and selection management.

The initial documentation can be found under `Mantine Core > Misc > Composite`

Demo: https://dev2069.d2npuq2dwaqoyp.amplifyapp.com/


rtivital:
Thanks for showing interest in contributing to Mantine! Currently, I do not have the capacity to maintain this component in the library. You are welcome to publish a separate package that implements this component as a separate extension. In future releases, there will be a separate section on mantine.dev website for community extensions. If you want to get started with a template, you can use mantine-extension-template – https://github.com/mantinedev/extension-template

PR #5704, opened 2024-02-02
[@mantine/core]: PinInput - some improvements

kkaplita:
- typing and pasting code problems
- now you can enter a value not only from the first input (previously, entering a value in, for example, the third input would change the value in the first input, but the focus would shift to the next one, i.e., the fourth input)
- when deleting a character in inputs, the values of subsequent inputs remain unchanged (previously, deleting a character in the middle of a PIN, for instance, would shift values from right to left in the inputs)
- paste a value from the clipboard that has any whitespace or line breaks - fixed
- added a few more tests


mellrebook:
I've had some issues with this component before, I was considering ditching it and writing my own from scratch because of it.
I've checked out this PR and resolved all that weirdness I had a problem with.

Thanks!

rtivital:
Thanks!

PR #5714, opened 2024-02-04
Transition: Fix skew-up transition is not working

### LINKED ISSUE #5614: Skew-up animation in <Modal> doesn't work
### Dependencies check up

7.4.2

@mantine/core

Next.js

Chrome

### Describe the bug

Just no effect when I set `skew-up` in **Modal** component, only see `fade` animation when set `skew-up`

`<Modal transitionProps={{ transition: 'skew-up' }}/> `

When I set `skew-down` it works and I see some animation effect for skew

_No response_

### Possible fix

_No response_

### Self-service

Maetes:
Fix #5614


rtivital:
Thanks!

PR #5724, opened 2024-02-06
[@mantine/nprogress]: added nprogress export

smyja:
nprogress was not exported


rtivital:
NavigationProgress file does not have `nprogress` export. It is already exported correctly from `nprogress.store.ts`.

PR #5730, opened 2024-02-06
Fix for #5722 by exporting PolymorphicComponentProps, PolymorphicRef

### LINKED ISSUE #5722: PolymorphicComponentProps/PolymorphicRef types are not exported, causing errors when using "NodeNext" resolution
### Dependencies check up

7.5.1

@mantine/core

### Describe the bug

When creating a library with TS settings `"declaration": true, "moduleResolution": "NodeNext"`, exporting a component created with `createPolymorphicComponent` throws an error:

> The inferred type of 'MyComponent' cannot be named without a reference to './node_modules/@mantine/core/lib/core/factory/create-polymorphic-component.js'. This is likely not portable. A type annotation is necessary.

From what I gather, this is because TypeScript decides it is necessary to annotate the exported type in the declarations file using `PolymorphicComponentProps`, and this type is not reachable via the package exports.

_No response_

### Possible fix

Just adding

`export type { PolymorphicComponentProps, PolymorphicRef } from './create-polymorphic-component'`

to `@mantine/core/src/core/factory/index.ts` should fix the issue.

Example project:
[example.zip](https://github.com/mantinedev/mantine/files/14173145/example.zip)

### Self-service

waweber:
Fixes #5722 by exporting these types so they can be referenced in emitted .d.ts files

### DIFF
--- a/packages/@mantine/core/src/core/factory/index.ts
+++ b/packages/@mantine/core/src/core/factory/index.ts
@@ -7,4 +7,5 @@ export type {
   MantineComponent,
   MantineComponentStaticProperties,
 } from './factory';
+export type { PolymorphicComponentProps, PolymorphicRef } from './create-polymorphic-component';
 export type { Factory, PolymorphicFactory } from './create-factory';


rtivital:
Thanks!

PR #5731, opened 2024-02-06
fix view source code 404 in useElementSize documentation page

jgravois:
hey there! 👋 

https://mantine.dev/hooks/use-element-size/ has a broken github link right now. here's one option (of many) to fix the issue.


rtivital:
Thanks for reporting the issue! It has been fixed in other commit.

PR #5746, opened 2024-02-09
Fix switch accessibility

### LINKED ISSUE #5699: Mantine Switch should follow accessibility best practices
Current Switch component fails accessibility checks for using an empty label to land user clicks on a checkbox. If user also uses the label prop a second label is created with the same `for` target which is also not allowed. 

Consider reworking the component to behave more like the [W3C APG example ](https://www.w3.org/WAI/ARIA/apg/patterns/switch/examples/switch/)

rilrom:
Fixes #5699.

Things to note:

- Due to the implementation of `role="switch"`, some tests needed to be adjusted.
- Screen readers no longer read out the `onLabel` and `offLabel` values, however I consider this to be appropriate as the `label` or `aria-label` should be used to provide enough context that these values are not needed.
- Screen reader testing was conducted across macOS VoiceOver and Windows NVDA.
- Open to suggestions on types for `bodyElement` and `labelElement`.


packages/@mantine/core/src/components/Switch/Switch.tsx
```diff
@@ -208,12 +210,12 @@ export const Switch = factory<SwitchFactory>((_props, ref) => {
...
         ref={ref}
         type="checkbox"
+        role="switch"
         {...getStyles('input')}
       />
 
       <Box
-        component="label"
-        htmlFor={uuid}
+        aria-hidden="true"
```
ocshawn, line 218: i would argue that the aria-hidden is not needed, as this will not be read if empty, but this will prevent any text from be read if onLabel  or offLabel are used

ocshawn (changes requested):
please explain reasoning for aria- hidden, otherwise everything looks great

packages/@mantine/core/src/components/InlineInput/InlineInput.tsx
```diff
@@ -36,6 +36,8 @@ export interface InlineInputProps
   error: React.ReactNode;
   size: MantineSize | (string & {}) | undefined;
   labelPosition?: 'left' | 'right';
+  bodyElement?: any;
+  labelElement?: any;
```
ocshawn, line 40:
```suggestion
  bodyElement?: string;
  labelElement?: string;
```
should also be changed from any

ocshawn (changes requested):

packages/@mantine/core/src/components/Switch/Switch.tsx
ocshawn: after re reading the accessibility documentation found at https://www.w3.org/WAI/ARIA/apg/patterns/switch/examples/switch-checkbox/ i retract my question

ocshawn (approved):

rtivital:
Thanks!

PR #5748, opened 2024-02-11
[@mantine/core] PinInput: Fix changing of "length" prop not rerendering Component

Maetes:
Should only rarely affect production deployments, but at least it affects the docs. PinInput is not rerendered when you change the length prop after the initial rendering.

@rtivital tell me if you also want a test, then I'll create one. Didn't cause I think it's very rare that the length change dynamically in prod. Did this fix mainly for the docs (and because I dont find peace when I know it should update but it doesn't :D).


rtivital:
Thanks for submitting the PR! I've already resolved the issue in another commit.

PR #5753, opened 2024-02-12
[@mantine/core] NumberInput: Fix double border between controls

justinsmid:
**Before:**

image

**After:**

image


rtivital:
Thanks!

PR #5755, opened 2024-02-12
[@mantine/core] Switch: fix accessibility V6

ocshawn:
Fixes https://github.com/mantinedev/mantine/issues/5699.

Backport of https://github.com/mantinedev/mantine/pull/5746

Things to note:

- Due to the implementation of role="switch", some tests needed to be adjusted.
- Screen readers no longer read out the onLabel and offLabel values, however I consider this to be appropriate as the label or aria-label should be used to provide enough context that these values are not needed.
- Screen reader testing was conducted across ChromeVox and Ubuntu Orca.
- Open to suggestions on types for bodyElement and labelElement.


rtivital:
Thanks!

PR #5757, opened 2024-02-13
fix: modal props update

abdulbasithqb:
Description:

docs: modal props update

Issue Details:
This model props causes readability problem, so now default is removed.

Changes:

it could be true, false or undefined
Checklist:

 Code follows the project's coding guidelines.


rtivital:
It is not correct. `keepMounted` must be false by default in ModalBase component.

PR #5762, opened 2024-02-13
[@mantine/core] ScrollArea: fix horizontal offsetScrollbars

### LINKED ISSUE #5758: ScrollArea's offsetScrollbars does not work if it is horizontal
### Dependencies check up

7.5.2

@mantine/core

Vite

Chrome

### Describe the bug

ScrollArea's offsetScrollbars does not work when the scrollbar is horizontal.

It works if you specify `className="pb-3"`

In 7.2.2, it worked without specifying `className="pb-3"`.

Below is an example code to reproduce.
```
      <ScrollArea w={100} offsetScrollbars>
        aaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaaa
      </ScrollArea>
```

_No response_

### Possible fix

_No response_

### Self-service

rilrom:
Fixes #5758.

### DIFF
--- a/packages/@mantine/core/src/components/ScrollArea/ScrollArea.module.css
+++ b/packages/@mantine/core/src/components/ScrollArea/ScrollArea.module.css
@@ -29,7 +29,7 @@
   }
 
   &:where([data-scrollbars='xy'], [data-scrollbars='x']) {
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #5766, opened 2024-02-14
[@mantine/hooks] Fixed initial value of the `online` attribute returned by `useNetwork()`

### LINKED ISSUE #5765: Initial value of `online` returned by `useNetwork()` is always true (even when offline)
### Dependencies check up

7.5.2

@mantine/hooks

Firefox

### Describe the bug

When using the online attribute, its initial value is true, even while being offline.
The actual code is only updating the value using window events, but the initial value is a constant `true`.

_No response_

### Possible fix

Replace `true` by `navigator?.onLine || true`

### Self-service

jalik:
Fix #5765

### DIFF
--- a/packages/@mantine/hooks/src/use-network/use-network.ts
+++ b/packages/@mantine/hooks/src/use-network/use-network.ts
@@ -35,7 +35,7 @@ function getConnection(): NetworkStatus {
 
 export function useNetwork() {
   const [status, setStatus] = useState<{ online: boolean } & NetworkStatus>({
-    online: true,
+    online: typeof navigator?.onLine === 'boolean' ? navigator.onLine : true,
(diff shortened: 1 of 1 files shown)


rtivital:
It is not a valid solution. It will produce a hydration mismatch if the hook value is used to render content in frameworks with ssr.

jalik:
@rtivital you are right, I changed the code so the initial value is set in the `useEffect()` callback, thus only executed on the client.
The problem with the current code is that the `online` value relies on the presence of `navigator.connection`, but on Firefox and many other browsers this is not available (https://caniuse.com/netinfo), resulting in not setting the online value, even when available.

rtivital:
Thanks!

PR #5786, opened 2024-02-18
[@mantine/dates] Fix: minimum-date is not in current month and value is not provided

### LINKED ISSUE #5695: The minDate in Dates should be the default date, when date is not given
### Dependencies check up

7.5.1

@mantine/dates

Next.js

Chrome

### Describe the bug

When the DateInput property (minDate) is set to, for example, 2024-09-01, and the DateInput is controlled through the form but the date is undefined, I expect the calendar to automatically navigate to the correct month of the minDate. It can be confusing for the user when the min date is in the future (and the month is greyed out). Additionally, the defaultDate does not work properly when using the mantine/forms plugin.

_No response_

### Possible fix

Start the default month (date) with the minDate given.

### Self-service

msv96:
Fixes #5695.

This PR solves:
* When the `value` is null and `minDate` is in past or future, then this fix will take `minDate` as default and show the calendar.

### DIFF
--- a/packages/@mantine/dates/src/components/DateInput/DateInput.tsx
+++ b/packages/@mantine/dates/src/components/DateInput/DateInput.tsx
@@ -167,7 +167,7 @@ export const DateInput = factory<DateInputFactory>((_props, ref) => {
 
   useEffect(() => {
     if (controlled) {
-      setDate(value!);
+      setDate(value ? value! : minDate!);
     }
   }, [controlled, value]);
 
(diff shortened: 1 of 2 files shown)


msv96:
Hi @rtivital, If this approach is good, then I will be checking for other date inputs and update those as well. Else, whatever the aproach is suggested, I will follow the same for every date inputs.

packages/@mantine/dates/src/components/DateInput/DateInput.tsx
```diff
@@ -167,7 +167,7 @@ export const DateInput = factory<DateInputFactory>((_props, ref) => {
 
   useEffect(() => {
     if (controlled) {
-      setDate(value!);
+      setDate(value ? value! : minDate!);
```
rilrom, line 170: When `clearable: true` and you remove a selected value, visually it's removed from the UI however `_date` still gets set to the `minDate` instead of `null`.

rtivital:

PR #5789, opened 2024-02-19
docs: fix Grid overflow default value in 7.3.0 changelog

### LINKED ISSUE #5784: Grid overflow "visible" by default
### Dependencies check up

7.5.2

@mantine/core

Next.js

All

### Describe the bug

According to v7.3.0 changelog -
[Grid](https://mantine.dev/changelog/7-3-0/#:~:text=Grid%20component%20now%20supports%20overflow%20prop%2C%20which%20allows%20controlling%20overflow%20CSS%20property%20on%20the%20root%20element.%20It%20is%20hidden%20by%20default.) component now supports overflow prop, which allows controlling overflow CSS property on the root element. It is hidden by default.

But on the latest Grid docs -
[Grid docs](https://mantine.dev/core/grid/?t=props) Sets overflow CSS property on the root element, 'visible' by default

Which one is correct? The latest behavior is that overflow is set to visible by default, but I would have expected the changelog to match the current docs.

_No response_

### Possible fix

_No response_

### Self-service

rilrom:
Fixes #5784.


rtivital:
Thanks!

PR #5806, opened 2024-02-22
[@mantine/core] Checkbox: Fix `indeterminate` styles when `disabled` for `outlined` variant

### LINKED ISSUE #5805: Checkbox's `outline` variant doesn't look `disabled` when `indeterminate`
### Dependencies check up

7.5.3

@mantine/core

Vite

Chrome

### Describe the bug

It looks like when `Checkbox` is `indeterminate` as well as `disabled`, you cannot interact with the checkbox but the styles show otherwise.

Reproduction:

- Go to https://mantine.dev/core/checkbox/#usage
- Change variant from `filled` to `outline`
- Turn on `indeterminate`
- Turn on `disabled`
- Note that checkbox style remains blue-ish like enabled instead of disabled

https://codesandbox.io/p/sandbox/mantine-react-template-forked-hlhvfc?file=%2Fsrc%2FApp.tsx%3A10%2C26

### Possible fix

IMO it would be as simple as override `indeterminate` styles for `outline` variant ONLY when **not disabled**.

### Self-service

lgaspari:
Fixes #5805

Reproduction:

- Go to http://localhost:7545/core/checkbox/#usage
- Change variant from `filled` to `outline`
- Turn on `indeterminate`
- Turn on `disabled`
- **Note that checkbox style is now gray-ish**

Screenshots _(taken from localhost)_:

image

### DIFF
--- a/packages/@mantine/core/src/components/Checkbox/Checkbox.module.css
+++ b/packages/@mantine/core/src/components/Checkbox/Checkbox.module.css
@@ -94,7 +94,7 @@
     --_checkbox-icon-color: var(--checkbox-color);
   }
 
-  &[data-indeterminate],
+  &[data-indeterminate]:not(:disabled),
   &:checked:not(:disabled) {
     [data-mantine-color-scheme] & {
       --_checkbox-bg: transparent;


rtivital:
Thanks!

PR #5850, opened 2024-03-01
Allow using colorScheme in MantineProvider's theme

Fumaz:
This PR allows you to pass a function to MantineProvider's theme, which takes as an argument the current `colorScheme`. This allows for easier customization of components based on the color scheme.


rtivital:
Thanks for showing interest in contributing to Mantine! This feature is not planned to be added as it is not compatible with SSR.

PR #5860, opened 2024-03-03
[docs] Add CssVariableColorSwatch column to CSS variables list

timkrins:
I was going through the documentation to find some CSS variables to use for some default theme colors, and noticed that it was a bit difficult to figure out.

This PR adds a `ColorSwatch` column to the docs with the corresponding color if the CSS variable value matches a regex.

### Screenshots
| Before | After |
| -------|------|
|color-variables-1-before|color-variables-1-after
|color-variables-2-before|color-variables-2-after


rtivital:
Thanks!

Issue #5866, opened 2024-03-04
IMAGE_MIME_TYPE does not include image/heic and image/heif


redachl:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

v7.6.0

### What package has an issue?

@mantine/dropzone

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Not applicable – issue is not related to the browser

### Describe the bug

# Bug description
If I use `IMAGE_MIME_TYPE` in the `accept` prop of a Dropzone, `.heic` images are rejected.
This is because `IMAGE_MIME_TYPE` does not include `image/heic`.

# How to reproduce
Can be reproduced in your documentation: https://mantine.dev/x/dropzone/

### Possible fix

Add `image/heic` to `IMAGE_MIME_TYPE`.

### Self-service

- [X] I would be willing to implement a fix for this issue

redachl:
I took the freedom to push a pull request because it's a very small change.

FYI, HEIC is used by all iOS devices after iOS 11, so this could be useful in many cases.

PR #5867, opened 2024-03-04
[@mantine/dropzone] Add image HEIC to IMAGE_MIME_TYPE in @mantine/dropzone

redachl:
Fixes https://github.com/mantinedev/mantine/issues/5866.

HEIC is used by Apple devices after iOS 11. It can be very useful.


rtivital:
Thanks!

Issue #5849, opened 2024-03-01
Transitions still randomly animate instantly


macjuul:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.6.1

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

Possibly related to #5193 and #3126 however I am still experiencing transition issues on the latest version of Mantine, in my case when using Modals, Drawers, etc.

I can reproduce this issue on all chromium browsers including on Android, however Safari appears to be unaffected.

I'm experiencing this issue in both my main application as well as fresh reproduction project, however the frequency of it happening seems to vary wildly, sometimes occurring 50% of the time while other times it's significantly less frequent.

Here is a recording showing the issue in action

https://github.com/mantinedev/mantine/assets/7606171/96d8e440-bf8b-4a26-97c2-1b31f52af3bb

### Self-service

- [ ] I would be willing to implement a fix for this issue

rtivital:
Fixed in [7.6.2](https://github.com/mantinedev/mantine/releases/tag/7.6.2)

Issue #5193, opened 2023-11-05
Sometimes `Transition` renders child instanly


dokugo:
### What package has an issue

@mantine/core

### Describe the bug

Sometimes `Transition` renders child instanly without actually transitioning, this happens when `Transition` status is currently `exited` and then its `mounted` prop value changes from `false` to `true`.

Here is the code to reproduce the problem:
https://codesandbox.io/p/devbox/mantine-transition-test-next-4dfvdc

Here is the screen recording. I toggle `isVisible` boolean value by clicking the first button. `isVisible` is used as the value of `mounted` `Transition` prop. After a few toggles the hidden button appears instantly.

https://github.com/mantinedev/mantine/assets/7773972/6d6ecade-7c68-4bc9-af80-5dee2ef9f059

Simulation of fast toggling:

https://github.com/mantinedev/mantine/assets/7773972/f02d2584-e769-486c-b322-dc8c1db66700

I'm not sure why exactly this happens. I tried experimenting with `flushSync` inside the `handleStateChange` function here, but without success.
https://github.com/mantinedev/mantine/blob/master/src/mantine-core/src/components/Transition/use-transition.ts#L41

### What version of @mantine/* packages do you have in package.json? (Note that all @mantine/* packages must have the same version in order to work correctly)

7.1.7

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/p/devbox/mantine-transition-test-next-4dfvdc

### Do you know how to fix the issue

No

### Are you willing to participate in fixing this issue and create a pull request with the fix

None

ID-JA:
Hi @rtivital, I've tackled the issue and implemented a solution. Would it be possible for me to create a PR for your review?

ghana7989:
Sandbox link is not working

dokugo:
@ghana7989 thanks, I fixed the link.

cyantree:
I think that's the same issue as #3126.

cyantree:
Please have a look at [this comment](https://github.com/mantinedev/mantine/issues/3126#issuecomment-1962409265) where I've provided an example using a patched version of the library.

Could you please check whether this resolves the issue for you?

rtivital:
Fixed in [7.6.1](https://github.com/mantinedev/mantine/releases/tag/7.6.1)

fauxparse:
> Fixed in [7.6.1](https://github.com/mantinedev/mantine/releases/tag/7.6.1)

[...]

PR #5873, opened 2024-03-05
[@mantine/core] next try fixing transition issues

### LINKED ISSUE #3126: Transition resolves instantly sometimes
### What package has an issue

@mantine/core

### Describe the bug

`Transition` sometimes shows its content instantly. I encountered this error in my app and was able to reproduce it on [Mantine's documentation page on Transitions](https://mantine.dev/core/transition/):

[Transition bug on try 4 and 7](https://user-images.githubusercontent.com/13364827/205909417-4475ed94-2915-46ac-bb9f-2055417ce157.webm)

In the video, the dropdown is shown instantly when opening it the 4th and the 7th time.

I cannot reproduce this bug every time. It seems to happen randomly.

I am running Fedora 37. I encounter this bug in Mozilla Firefox 107.0, Chromium 105.0.5195.125 and Chrome 110.0.5449.0. I did not try it in other browsers or versions.

5.9.0

https://mantine.dev/core/transition/

No.

I don't believe I can, so no.

cyantree:
(Hopefully) fixes #3126 #5193 #5849.

As I found out [earlier](https://github.com/mantinedev/mantine/issues/3126#issuecomment-1426761731) it's probably related to the automatic batching in React. Therefore this fix uses `flushSync()` to enforce the state update at the right moment.

I hope that this will fix the issue.

### DIFF
--- a/packages/@mantine/core/src/components/Transition/use-transition.ts
+++ b/packages/@mantine/core/src/components/Transition/use-transition.ts
@@ -1,4 +1,5 @@
 import { useEffect, useRef, useState } from 'react';
+import ReactDOM from 'react-dom';
 import { useDidUpdate, useReducedMotion } from '@mantine/hooks';
 import { useMantineTheme } from '../../core';
 
@@ -43,7 +44,6 @@ export function useTransition({
     const preHandler = shouldMount ? onEnter : onExit;
     const handler = shouldMount ? onEntered : onExited;
 
-    setStatus(shouldMount ? 'pre-entering' : 'pre-exiting');
     window.clearTimeout(timeoutRef.current);
 
     const newTransitionDuration = reduceMotion ? 0 : shouldMount ? duration : exitDuration;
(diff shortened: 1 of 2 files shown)


rilrom (commented):
Nice work @cyantree! I can no longer reproduce the issue.

I'm curious why the changes to modal tests were necessary?

cyantree:
I think that's because at first the component is unmounted and therefore the state is `exited`. This leads to the children not being rendered at all. This in effects leads to the test failing.
https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/components/Transition/Transition.tsx#L71

This is because with my changes the state transition from `exited` to `pre-entering` doesn't happen in the same frame anymore (this is done because `ReactDOM.flushSync()` can't be used within a running render pass).

So setting `transitionDuration: 0` disables the transition and renders the children directly. IMHO that's an acceptable behaviour.

However I'm happy for any suggestions as this (at least for me) is a finicky bug.

macjuul:
I can also confirm it solves the problem on my end, thanks so much!

rtivital:
Thanks everyone for validation! @cyantree, thanks for the fix!

robmonie:
This change was made quite some time ago but hopefully someone can help with an issue i'm facing whicn is preventing me from upgrading past 7.6.1.

I've found that since this change was introduced, any code that incorporates `useTransition` such as modals and popovers fail to render the content of the Modal or popover in tests when using testing library and vitest / jsdom. It may be the same with  jest but i'm not sure.

I've been able to isolate it to the moving of `setStatus(shouldMount ? "pre-entering" : "pre-exiting");` into `ReactDOM.flushSync`.  When I revert this, everything works again.

I'm currently looking into whether this is something I can work around within vitest / react testing library etc but I suspect i'm not the only person to face this and maybe someone else has figured it out?

PR #5880, opened 2024-03-06
[@mantine/dropzone] Fix: added inputProps for dropzone internal input component

AdarshJais:
This is in refernece to this [dicussion](https://github.com/orgs/mantinedev/discussions/5874).

tested the same by passing the inputProps via the storybook.


rtivital:
Thanks!

PR #5881, opened 2024-03-07
[@mantine/core]: update global.css

mhmdjaw:
Modify universal box sizing to use inheritance which is considered a better practice.

reasons why you would need the inheritance version:
[https://css-tricks.com/box-sizing/](https://css-tricks.com/box-sizing/)
[https://css-tricks.com/inheriting-box-sizing-probably-slightly-better-best-practice/](https://css-tricks.com/inheriting-box-sizing-probably-slightly-better-best-practice/)


rtivital:
It is not a valid approach. Mantine can be used in shadow DOM where html selector is not available.

PR #5890, opened 2024-03-08
[@mantine/core] Timeline: Fix `autoContrast` being passed to the dom node as attribute (#5846)

### LINKED ISSUE #5846: Timeline `autoContrast` prop not recognised by React
### Dependencies check up

7.6.1

@mantine/core

Next.js

Chrome

### Describe the bug

It appears that the issue that was affecting the Indicator component also affects the Timeline component. Refer to the related issue here #5508.

https://codesandbox.io/p/sandbox/mantine-react-template-forked-lrjk9c?file=%2Fsrc%2FApp.tsx

### Possible fix

The same change applied in this commit: https://github.com/mantinedev/mantine/commit/b71a976dec1896867c41674f4ad19fd98ad1109d could be applied to the Timeline component here: https://github.com/mantinedev/mantine/blob/7b1457f9eb084f0fb663c25793d08877cc957db0/packages/%40mantine/core/src/components/Timeline/Timeline.tsx#L108 

### Self-service

Zachary-Kaye:
Fixes #5846

### DIFF
--- a/packages/@mantine/core/src/components/Timeline/Timeline.tsx
+++ b/packages/@mantine/core/src/components/Timeline/Timeline.tsx
@@ -106,6 +106,7 @@ export const Timeline = factory<TimelineFactory>((_props, ref) => {
     lineWidth,
     reverseActive,
     mod,
+    autoContrast,
     ...others
   } = props;
 


rtivital:
Thanks!

PR #5922, opened 2024-03-14
[@mantine/form] Add resetField function

moimael:
fix https://github.com/orgs/mantinedev/discussions/5023

It'd be super useful to be able to reset only one field, especially since we don't have any access to the `initialValues` snapshot of mantine form.

Currently we need to wrap a Mantine form with a custom context that keeps track of the `initialValues` to be able to reset the value for a specific field. 

Introducing a `resetField` function makes this a lot easier.

There is one test that fails, the reason is that I'm not sure how to handle dynamically added fields. Should it reset to an empty value, or remove the field entirely? I want some input on that.

Small related comment regarding the tests, we shouldn't need to wrap all the `renderHook` actions in an act. It's already done at the testing library level.

cc @dedalusium


rtivital:

PR #5935, opened 2024-03-16
[@mantine/core] Fix conbobox keyDown handling during IME input conversion

### LINKED ISSUE #5934: Combobox Submits Focused Option on Enter Key with IME Conversion
### Dependencies check up

7.6.2

@mantine/core

Next.js

All

### Describe the bug

Related: #1565 

When using a Combobox and confirming a conversion with the Enter key while using an IME, the focused option is being submitted. I am a Japanese user, but I believe this issue also affects users who type in Chinese.

This becomes a particular problem with examples like ["Select first option on type."](https://mantine.dev/combobox/?e=AutocompleteSelectFirstOption) Here is a reference to the issue demonstrated: 

https://github.com/mantinedev/mantine/assets/37701077/bd00ee52-1f15-4971-b42b-53a9c223743e

In the video, I am using an IME called ATOK, but the issue occurs with the standard Apple IME as well.

I think the ideal behavior would be as demonstrated in this video.

https://github.com/mantinedev/mantine/assets/37701077/35a1fc96-545a-421c-a0c6-d2bfbcc70416

_No response_

### Possible fix

This can be resolved by checking the `isComposing` property of the KeyboardEvent.
https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/isComposing

If there are no objections, I will proceed to create a pull request.
https://github.com/elecdeer/mantine/commit/b7a12045ec920602e88ddddd77d09faef5e3cf9e

### Self-service

elecdeer:
Fixes #5934.

Fixes an issue in the Combobox where confirming an IME conversion incorrectly submits the focused option.

I confirmed to work as expected in Safari.

### DIFF
--- a/packages/@mantine/core/src/components/Combobox/use-combobox-target-props/use-combobox-target-props.ts
+++ b/packages/@mantine/core/src/components/Combobox/use-combobox-target-props/use-combobox-target-props.ts
@@ -29,6 +29,9 @@ export function useComboboxTargetProps({
     }
 
     if (withKeyboardNavigation) {
+      // Ignore during composition in IME
+      if (event.nativeEvent.isComposing) return;
+
       if (event.nativeEvent.code === 'ArrowDown') {
         event.preventDefault();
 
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

elecdeer:
Thank you for the review and merge!

PR #5939, opened 2024-03-18
UPDATE: Import `useForm` from `@mantine/form` instead of `core``

brkcrdk:
The old doc was importing `useForm` from `@mantine/core` instead of `@mantine/form`


rtivital:
Thanks for reporting the issue, it was fixed in another commit.

PR #5953, opened 2024-03-19
[docs] Normalize the Contribute page

wasamistake:
Hi :wave:

Besides changing some wording, this commit adds an instruction to set up Yarn, the package manager used in the project, as well as changes all references from `npm run` to `yarn`.


rtivital:
Thanks for showing interest in contributing to Mantine! I would prefer to keep the page as is.

Issue #5959, opened 2024-03-20
Disabled TagsInput Pills can be removed


snturk:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.6.1

### What package has an issue?

@mantine/core

### What framework do you use?

Other, I will specify in the bug description

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

Disabled TagsInput pills can be removed by just clicking them on the side where the "cross" button appears when not disabled.

Example video on Mantine website:

https://github.com/mantinedev/mantine/assets/41567673/dce35323-c912-4a12-b139-42468dbf3c44

### Self-service

- [ ] I would be willing to implement a fix for this issue

rtivital:
Fixed in [7.1.1](https://github.com/mantinedev/mantine/releases/tag/7.7.1)

PR #5962, opened 2024-03-20
[@mantine/core] TagsInput: Fix disable RemoveButton when Input is `disabled`

Kenzo-Wada:
Fixes: https://github.com/mantinedev/mantine/issues/5959


rtivital:
It is not a suitable solution, the issue will persist when TagsInput is used within a disabled fieldset without `disabled` prop.

Kenzo-Wada:
I see. Thank you for the review!
Ill keep working on issue by different solution.

PR #5963, opened 2024-03-20
[@mantine/core] PinInput: Fix paste on Mac, Tab behavior

joshua-webdev:
This PR fixes two issues with the PinInput component I noticed.
1. Pasting on MacOS was not working, as the default paste shortcut is `Command + V` instead of `Ctrl + V`.
2. Default browser tab navigation was prevented, which caused problems when a PinInput component appeared in the middle of a form. Now, if you press `tab` while in the last input segment, focus will move to the next focusable element. If you press `shift + tab` on the first input segment, focus will move to the previous focusable element.


rtivital:
Thanks!

PR #44, opened 2021-05-07
[Docs] Sort the props in alphabetical order

khushal87:
Fixes #42

PR #5964, opened 2024-03-20
[@mantine/core] Avatar: Set theme.primaryColor as default color

cristiancundari:
it was gray before


rtivital:
It is intended to be gray

cristiancundari:
Oh, ok then. I was thinking it was intended to be primary color by looking at [row #44](https://github.com/mantinedev/mantine/blob/1fe944fadcbf4dcb5ef994c22e1656815e4368a4/packages/%40mantine/core/src/components/Avatar/Avatar.tsx#L44C2-L45C24) of `Avatar.tsx`
```javascript
/** Key of `theme.colors` or any valid CSS color, default value is `theme.primaryColor`  */
color?: MantineColor;
  ```

PR #5971, opened 2024-03-22
[core] Fix typo on the esm folder exclude for the `@docs` package

wasamistake:
I was looking around and spotted what appears to be a typo on an exclude for the "esm" folder (it doesn't seem to produce any errors, though).


rtivital:
Thanks!

PR #5975, opened 2024-03-23
DataTable concept

esavitskiy:
I would like to propose the concept of a DataTable component via TanStack Table. I implemented some basic features like grouping, sorting, infinity scroll, etc.
Is it worth continuing or is the community making such components on its own and a general implementation is not needed?

#1057

### DIFF
--- a/package.json
+++ b/package.json
@@ -63,6 +63,7 @@
   "dependencies": {
     "@floating-ui/react": "^0.26.9",
     "@tabler/icons-react": "^2.40.0",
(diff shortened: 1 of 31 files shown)


rtivital:
Thanks for showing interest in contributing to Mantine! DataTable component is not planned to be added as a part of the library any time soon. If you need it in your application, use one of the available third-party libraries – https://help.mantine.dev/q/data-grid-i-need

PR #5976, opened 2024-03-24
[@mantine/core] Fix `Pill` to respect disabled state (#5959)

snturk:
Fixes: #5959


packages/@mantine/core/src/components/Pill/Pill.tsx
```diff
@@ -128,6 +128,7 @@ export const Pill = factory<PillFactory>((_props, ref) => {
           tabIndex={-1}
           aria-hidden
           unstyled={unstyled}
+          disabled={disabled}
```
snturk, line 131: Not mandatory for this issue, but it could be needed in the future.

rtivital:
As noted in https://github.com/mantinedev/mantine/pull/5962, it does not fix the issue entirely. The issue will persist when TagsInput is used in a disabled fieldset. I'll fix the issue.

PR #5988, opened 2024-03-26
[docs] Update CssVariableColorSwatch table regex to match `rgba` variables

timkrins:
I noticed the table did not show a color swatch for the `rgba` values.
This PR adjusts the regex so the `rgba` variables are matched.


rtivital:
Thanks!

PR #5989, opened 2024-03-27
Ensure fallback image has ref

s-cork:
The ref is not set on the Image when there is an error or no src

this PR fixes that issue


rtivital:
Thanks!

PR #5993, opened 2024-03-27
[mantine/tiptap] set align-items to center for toolbar

xiaohanyu:
In mantine v6, by default Toolbar will render into a `<Group>` component and the `align-items` value is `center` by default. V7 breaks this behavior.


rtivital:
Thanks!

PR #5081, opened 2023-10-19
[@mantine/dates] Add range type to DateTimePicker

wuifdesign:
Added `type="range"` to `DateTimePicker` as it is currently not implemented, but my be a nice feature in my opinion.

PR #5999, opened 2024-03-29
Export hooks from `@mantine/dates` and parse `allowSameDay` to `useDatesState` & `DatePicker`

wuifdesign:
As it seams like https://github.com/mantinedev/mantine/pull/5081 will not be merged in near future i prepared a new PR to allow someone to implement their own DateTimePickerInput with this functionality.

Currently the hooks needed to implement it are not exported by the `@mantine/dates` packages, so i added it to the exported functions.

Also i added `allowSameDay` setting to `useDatesState` & `DatePicker` to allow configuration of this value in the custom component.


wuifdesign:
ahh. you are totally right! totally missed this 😅

i will update the PR to only include the hooks export. or do you think there will be a problem exporting the hooks?

rtivital:
Hooks is internal api, there are not supposed to be exported from the package. API can change at any time without any notice.

wuifdesign:
so what would be the preferred way to implement a custom date input without the `useUncontrolledDates` hook? should the hook and dependencies be copied over to the custom repo and updated each time mantine updates them? or is there som other way i didn't think of?

`useUncontrolled` from the cores repo is exported too.

i currently just want to build a datetimepicker with a cutom time input and range support, but copying the hooks feels somewhat wrong. 

i'm open for other ideas :)

rtivital:
Copy useUncontrolledDates in your codebase

Issue #6003, opened 2024-03-30
`NumberInput`: position of cursor moves weird


Kenzo-Wada:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.7.1

### What package has an issue?

@mantine/core

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

All

### Describe the bug

on `NumberInput`, the cursor moves weird way when decrementing Nagative numbers
[capture - screencast](https://github.com/mantinedev/mantine/assets/79452224/756948ed-3789-4a99-bf69-7e53da1795f7)

### If possible, include a link to a codesandbox with a minimal reproduction

https://mantine.dev/core/number-input/#usage

### Possible fix

I'll submit PR :)

### Self-service

- [X] I would be willing to implement a fix for this issue

PR #6004, opened 2024-03-30
[@mantine/core] fix: NumberInput cursor position

Kenzo-Wada:
Fixes: https://github.com/mantinedev/mantine/issues/6003

after
[Screencast](https://github.com/mantinedev/mantine/assets/79452224/202647b8-e51e-428a-a995-0c75f0b994f7)

### DIFF
--- a/packages/@mantine/core/src/components/NumberInput/NumberInput.tsx
+++ b/packages/@mantine/core/src/components/NumberInput/NumberInput.tsx
@@ -1,4 +1,4 @@
-import React, { useEffect, useRef } from 'react';
+import React, { useRef } from 'react';
 import cx from 'clsx';
 import { NumberFormatValues, NumericFormat, OnValueChange } from 'react-number-format';
 import { assignRef, clamp, useMergedRef, useUncontrolled } from '@mantine/hooks';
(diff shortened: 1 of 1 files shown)


packages/@mantine/core/src/components/NumberInput/NumberInput.tsx
```diff
@@ -358,6 +358,13 @@ export const NumberInput = factory<NumberInputFactory>((_props, ref) => {
     stepCountRef.current = 0;
   };
 
+  useEffect(() => {
+    if (inputRef.current) {
+      const { length } = inputRef.current.value;
+      inputRef.current.setSelectionRange(length, length);
```
rtivital, line 364: What if user wants to edit value in the middle, for example `251{cursor}213`, when anything is entered the selection range is moved to the end – this is unexpected behavior.

Kenzo-Wada: I see. I've fixed this issue at [f49af28](https://github.com/mantinedev/mantine/pull/6004/commits/f49af28d6573c801a05be93e7b5dd138c8acd560). by this change, the position of cusor only changes when decrement/increment. I'll be happy if you check this change is Ok :)

rtivital:
Thanks!

Issue #6010, opened 2024-03-31
TagsInput IME issue for `Backspace` key


xiaohanyu:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.7.1

### What package has an issue?

@mantine/core

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

None

### Describe the bug

When you use IME to input CJK words and types a `Backspace` key to delete, not only the CJK words got deleted by one char, but the whole, previous tag got deleted.

Demo:

https://github.com/mantinedev/mantine/assets/1164623/eabb2329-fec0-4b85-ae18-62fb7e0a9b47

### Possible fix

https://github.com/mantinedev/mantine/pull/6011

### Self-service

- [X] I would be willing to implement a fix for this issue

PR #4947, opened 2023-10-02
[@mantine/core] TagsInput: fix IME #4946

a-ryang:
- fixes #4946
- It looks like the same problem can be solved with the following PR.
https://github.com/mantinedev/mantine/pull/1570
### before

https://github.com/mantinedev/mantine/assets/105474635/d0a1566c-a629-4009-a836-4387d29c0f45

### after

https://github.com/mantinedev/mantine/assets/105474635/54c44288-9ca5-4ead-a48f-c4dd0eafc988

PR #6011, opened 2024-03-31
[@mantine/core] TagsInput: Fix incorrect IME keyboard input handling for `Backspace` (#6010)

xiaohanyu:
When people use IME to type in CJK words, a `Backspace` key would delete both a char in the CJK words and the previous tag before this fix.

Ref PR: https://github.com/mantinedev/mantine/pull/4947


xiaohanyu:
Before fix demo check #6010 .

After fix demo:


https://github.com/mantinedev/mantine/assets/1164623/6d2b5567-f71c-4d32-91ea-6d3baaeb4c09

rtivital:
Thanks!

PR #6014, opened 2024-04-01
Update use-isomorphic-effect.ts comment

abeisleem:
comment is message is other way around


rtivital:
The change is wrong, the original message is correct.

abeisleem:
really? https://nextjs.org/docs/messages/react-client-hook-in-server-component

Issue #6017, opened 2024-04-02
useLocalStorage causes infinite re-renders


NimmLor:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.7.1

### What package has an issue?

@mantine/hooks

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

v7.7.1 break my useLocalStorage states. The following PR updates the memo dependencies and causes infinite re-renders.

https://github.com/mantinedev/mantine/pull/5910/files

I suspect that the useEffect on [Line 145](https://github.com/israelins85/mantine/blob/148705be0ffbfed2a7c2174b649c0c808f80a86c/packages/%40mantine/hooks/src/use-local-storage/create-storage.ts#L145) causes this.

![image](https://github.com/mantinedev/mantine/assets/32486857/c601a759-ee98-4f46-b211-fb6a2c74f0be)

### Possible fix

Revert memo dependencies back to how they were in v7.7.0

### Self-service

- [ ] I would be willing to implement a fix for this issue

wuifdesign:
i was just typing the same issue when this popped up. if you use simple values like string this doesn't happen, but when you use an object and the deserialize is triggered it runs into an infinite loop,

MathiasGr:
👍 We're seeing the same here

wuifdesign:
i prepared a fix here: https://github.com/mantinedev/mantine/pull/6019

LuckyLukas98:
> ### Dependencies check up
> * [x]  I have verified that I use latest version of all @mantine/* packages
> 
> ### What version of @mantine/* packages do you have in package.json?
> 7.1.1
> 
> ### What package has an issue?
> @mantine/hooks
> 
> ### What framework do you use?
> Vite
> 
> ### In which browsers you can reproduce the issue?
> Chrome
> 
> ### Describe the bug
> v7.1.1 break my useLocalStorage states. The following PR updates the memo dependencies and causes infinite re-renders.
> 
> https://github.com/mantinedev/mantine/pull/5910/files
> 
> I suspect that the useEffect on [Line 145](https://github.com/israelins85/mantine/blob/148705be0ffbfed2a7c2174b649c0c808f80a86c/packages/%40mantine/hooks/src/use-local-storage/create-storage.ts#L145) causes this.
> 
[...]

PR #6019, opened 2024-04-02
[@mantine/hooks] fix infinite re-render of `useLocalStorage` hook

wuifdesign:
as the `serialize` prop was not memorized the `setStorageValue` function was created on each re-render - resulting in an infinite re-render loop.

After this fix it is still needed to memorize the `serialize` and `deserialize` if overwriting them in your own code.

fixes https://github.com/mantinedev/mantine/issues/6017

### DIFF
--- a/packages/@mantine/hooks/src/use-local-storage/create-storage.ts
+++ b/packages/@mantine/hooks/src/use-local-storage/create-storage.ts
@@ -77,7 +77,7 @@ export function createStorage<T>(type: StorageType, hookName: string) {
     defaultValue,
     getInitialValueInEffect = true,
     deserialize = deserializeJSON,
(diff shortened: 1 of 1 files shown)


israelins85:
@rtivital this commit fixes the real problem, please don't redo my commit, change the key is another real bug, the hook have to reload the value when key is changed.

kenzaflow:
Please, merge this

noxan:
@rtivital could we please merge this fix? the latest published version `7.7.1` is literally broken and mantine is a such a great library :)

rtivital:
It is not a valid solution – it breaks rules of hooks. If `serialize` option changes from undefined to other function, it will cause an error of the amount of hook rendered by the parent component. The issue is resolved by rolling back previous PR.

PR #6032, opened 2024-04-03
Pin Input: add rootRef prop

s-cork:
addresses #6031

not currently possible to access the ref of the pin input container


rtivital:
Thanks!

PR #6039, opened 2024-04-04
Fix: PinInput length property set to null causes infinite render cycle

s-cork:
length probably won't be undefined because it has a default value of 4
but if it is `null` then the PinInput finds itself in an infinite render loop
(I had the misfortune of doing this in a demo 🙃)

See the second use effect in the PR diff
if length is null we call setValues with an empty array
which changes the `_value` which causes the effect to re run


rtivital:
`null` is not a supported value for the length property. Properly configured TypeScript will catch this error, there is no need to support this use case in the library.

PR #6040, opened 2024-04-04
[@mantine/charts] fix overflow of chart

Kenzo-Wada:
Related discussion: https://discord.com/channels/854810300876062770/1225096473130041374/1225096473130041374

after:
Screenshot from 2024-04-04 22-58-21


rtivital:
It is not a valid fix, this way styles are added to all svg elements, even those that are not related to `@mantine/charts` package. I've fixed the issue in another commit.

Kenzo-Wada:
@rtivital thank you for your fix!

PR #6055, opened 2024-04-09
[@mantine/core] ScrollArea: Fix viewport `onScroll` prop not working

zo-ly:
ScrollArea `viewportProps.onScroll` prop not working


rtivital:
Thanks!

PR #6062, opened 2024-04-11
Update Table/index.ts to export components

davidpett:
The TableScrollContainer props and the TableDataRenderer were not being exported from the Table package.

### DIFF
--- a/packages/@mantine/core/src/components/Table/index.ts
+++ b/packages/@mantine/core/src/components/Table/index.ts
@@ -32,4 +32,3 @@ export type {
   TableScrollContainerCssVariables,
   TableScrollContainerFactory,
 } from './TableScrollContainer';
-export type { TableDataRendererProps } from './TableDataRenderer';
--- a/packages/@mantine/core/src/components/Table/index.ts
+++ b/packages/@mantine/core/src/components/Table/index.ts
@@ -28,7 +28,7 @@ export type {
 } from './Table.components';
 export type { 
   TableScrollContainerProps,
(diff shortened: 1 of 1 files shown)


packages/@mantine/core/src/components/Table/index.ts
```diff
@@ -26,3 +26,10 @@ export type {
   TableTfootProps,
   TableTheadProps,
 } from './Table.components';
+export type { 
+  TableScrollContainerProps,
+  TableScrollContainerStyleNames,
+  TableScrollContainerCssVariables,
+  TableScrollContainerFactory,
+} from './TableScrollContainer';
+export type { TableDataRendererProps } from './TableDataRenderer';
```
rtivital, line 35: This should not be exported from the package. TableDataRenderer is an internal component.

packages/@mantine/core/src/components/Table/index.ts
```diff
@@ -9,6 +9,7 @@ export {
   TableThead,
 } from './Table.components';
 export { TableScrollContainer } from './TableScrollContainer';
+export { TableDataRenderer } from './TableDataRenderer';
```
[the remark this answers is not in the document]
davidpett: Updated to remove that, but kept the type export. Does that work for you?

packages/@mantine/core/src/components/Table/index.ts
davidpett: Removed

rtivital:

PR #6064, opened 2024-04-11
[bugfix]: reorder onChange and onChangeEnd events for Slider arrow keys

### LINKED ISSUE #6063: Slider onChangeEnd event fires before onChange when using arrow keys
### Dependencies check up

7.7.2

@mantine/core

Vite

All

### Describe the bug

When using a `Slider` component with `onChange` and `onChangeEnd` events, using arrow keys to change the value will cause the `onChangeEnd` event to fire before `onChange`

_No response_

### Possible fix

Reorder the events at https://github.com/mantinedev/mantine/blob/4c9a875a823daf8132c034e62f247a6960c84b93/packages/%40mantine/core/src/components/Slider/Slider/Slider.tsx#L239-L240

### Self-service

kgarner7:
Resolves #6063 by changing the function call order, and adds a test to verify the timing.

It's unclear to me why `onChangeEnd` is called twice (this behavior was there previously), but importantly the second `onChangeEnd` happens *after* `onChange`.

### DIFF
--- a/packages/@mantine/core/src/components/Slider/Slider/Slider.test.tsx
+++ b/packages/@mantine/core/src/components/Slider/Slider/Slider.test.tsx
@@ -129,4 +129,27 @@ describe('@mantine/core/Slider', () => {
     const { container } = render(<Slider defaultValue={120} max={100} />);
     expectInputValue('100', container);
   });
+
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #6072, opened 2024-04-12
[@mantine/core] fix: NumberInput cursor no jump to end when pressing backspace on pos 0

### LINKED ISSUE #6056: NumberInput: Cursor jumps to the end when pressing backspace with cursor on the very left
### Dependencies check up

7.7.1

@mantine/core

Vite

### Describe the bug

When moving the cursor to the start on NumberInput, then pressing backspace, the cursor moves to the end of the input.

Tested with Chrome and Firefox.

https://codesandbox.io/p/sandbox/mantine-react-template-forked-c4yvtd?file=%2Fsrc%2FApp.tsx%3A5%2C66

### Possible fix

_No response_

### Self-service

jvllmr:
fixes #6056 

See this more or less as a draft.
The solution is a bit ... and I don't really know why the accessibility tests don't accept type="number" :(
Maybe it would be a better issue to solve for react or the browsers.

### DIFF
--- a/packages/@mantine/core/src/components/NumberInput/NumberInput.tsx
+++ b/packages/@mantine/core/src/components/NumberInput/NumberInput.tsx
@@ -190,6 +190,7 @@ export const NumberInput = factory<NumberInputFactory>((_props, ref) => {
     allowDecimal,
     decimalScale,
     onKeyDown,
+    onKeyDownCapture,
     handlersRef,
     startValue,
     disabled,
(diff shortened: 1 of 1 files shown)


jvllmr:
I found a better solution! 🥳 

Note that it wouldn't work with the onKeyDown event handler because it  takes place after the cursor pos changed.

rtivital:
Thanks!

PR #6078, opened 2024-04-13
Spacing: fix resolver for empty strings, and strings with spaces

### LINKED ISSUE #6077: expected output of an empty string in spacing properties
### Dependencies check up

7.7.2

@mantine/core

All

### Describe the bug

I don't think it's intended behvaior
but an empty string for something like `mt` in the style props leads to the output `marginTop: calc(0px * (...))`
I would expect an empty string to be treated like undefined

(Similar to how it is in react, and css generally)

(set mt="" on any component and inspect the element)

### Possible fix

in `createConverter`
 this line is true for empty strings
```javascript
      // Number("") -> 0
      if (!Number.isNaN(Number(replaced))) {
```

so we should explicitly check for empty string

There's also a corollary of this, that if you have extra spaces in your prop, then you get weird things in the output e.g. `mt=" 4px"`
`calc(0rem * var(--mantine-scale)) calc(0.25rem * var(--mantine-scale))`

The str.split.join function in the same place could be a bit cleverer
(getting blocks of non-space characters, rather than splitting on spaces)

### Self-service

s-cork:
closes #6077 

(test that failed in the first commit was always broken)


rtivital:
Thanks!

PR #6134, opened 2024-04-26
[@mantine/core] feat: adding some default variants to notification

sindrefl:
Hello,

First time contributor, but long time user of the mantine library.

I wanted to get some familiarity with the mantine codebase, and add something of value at the same time, so I took a look at adding some of the standard variants in the color-variants-resolver to the Notification component.

I looked at Alert for inspiration, and have tried not to introduce any breaking changes. Updated storybook and documentation highlighting the three variants:

Outline - same as Notification base component
Filled - proposal for filled version with autocontrast support
Light - proposal for light version similar to Alert

Let me know what you think and if any changes are required

### DIFF
--- a/apps/mantine.dev/src/pages/core/notification.mdx
+++ b/apps/mantine.dev/src/pages/core/notification.mdx
@@ -11,6 +11,10 @@ Build your own or use [@mantine/notifications](/x/notifications/) package.
 
 <Demo data={NotificationDemos.configurator} />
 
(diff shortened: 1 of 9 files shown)


rtivital:
Thanks for showing interest in contributing to Mantine! Variants are not planned to be introduced to the Notification component. If you need to change component styles, you can do that with Styles API.

PR #6144, opened 2024-04-29
Storybook v8 upgrade

aliozinan:
- storybook upgraded to `v8.0.9`, along with relevant addons & plugins
- `storybook-dark-mode` upgraded to v4.0.1
- `storybook` config updated
- a few TS errors that caused the `build all` command fail were resolved

### DIFF
--- a/packages/@mantine/core/src/components/Tooltip/use-tooltip.ts
+++ b/packages/@mantine/core/src/components/Tooltip/use-tooltip.ts
@@ -5,7 +5,6 @@ import {
   inline,
   offset,
   shift,
-  // useDelayGroup,
   useDelayGroupContext,
   useDismiss,
   useFloating,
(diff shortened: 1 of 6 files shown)


packages/@mantine/core/src/components/Tooltip/use-tooltip.ts
```diff
@@ -90,7 +89,6 @@ export function useTooltip(settings: UseTooltip) {
     useRole(context, { role: 'tooltip' }),
     // cannot be used with controlled tooltip, page jumps
     useDismiss(context, { enabled: typeof settings.opened === 'undefined' }),
-    useDelayGroup(context, { id: uid }),
```
rtivital, line 93: Removing this will break functionality of Tooltip. It should not be removed.

aliozinan: It has no effect. group delay is already defined here: 

`const { delay: groupDelay, currentId, setCurrentId } = useDelayGroupContext();`

and applied to the tooltip (/ group) here : 

`delay: withinGroup ? groupDelay : { open: settings.openDelay, close: settings.closeDelay },`

I tested this change on Storybook, works without any issues. 

The reason I removed this line is the TS error I had shared in the screenshot in our conversation on the support channel.

rtivital: I cannot merge PR with this change, please remove it and keep PR specific to storybook

Issue #6162, opened 2024-05-02
[@mantine/core] Number Input With allowNegative Attribute Allows Negative Form Values


AustinWildgrube:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.9.0

### What package has an issue?

@mantine/core

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

All

### Describe the bug

The `NumberInput` component incorrectly allows negative numbers despite setting the allowNegative attribute. When clicking the down arrow, it correctly reaches 0, but subsequent clicks result in the UI incrementing while the form value, managed by `useForm`, continues to decrement. I have tested this with both controlled and uncontrolled forms. I am not sure when this issue first started occurring, but I noticed it after upgrading to the latest version of both the core package and the forms package.

https://github.com/mantinedev/mantine/assets/40448411/87c92b73-5426-43d8-b9be-233e827773c0

### If possible, include a link to a codesandbox with a minimal reproduction

https://stackblitz.com/edit/stackblitz-starters-a4f4lm?file=app%2Fpage.tsx

### Self-service

- [X] I would be willing to implement a fix for this issue

PR #6170, opened 2024-05-03
[@mantine/core] NumberInput: Fix ignored allowNegative prop (#6162)

AustinWildgrube:
After looking into #6162, I found that commit 081a47a removed the `allowNegative` prop. This PR brings back support for it and fixes the bug where the `NumberInput` could go negative.


rtivital:
Thanks!

PR #6180, opened 2024-05-04
[@mantine/core] NavLink: Use active styles when aria-current is present

corydeppen:
Based on the discussion in #5788.


rtivital:
Thanks!

PR #6188, opened 2024-05-06
Fix inline code highlight examples

OliverWales:
Just fixing something I noticed when viewing the docs on inline code highlighting.

Before:

Screenshot 2024-05-06 at 08 26 49

Screenshot 2024-05-06 at 08 23 09

After:

Screenshot 2024-05-06 at 08 26 39

Screenshot 2024-05-06 at 08 22 18


rtivital:
Thanks!

PR #6206, opened 2024-05-09
[@mantine/hooks] Added useDebouncedLocalStorageState

EdwardEB:
Local Storage state that is also debounced. 
From use-debounced-state and use-local-storage-state.


warkanum:
This will be a nice addition, very useful hook since writing to local storage can be slow and excessive.

rtivital:

EdwardEB:
Will follow this route in future. Thanks

Issue #6215, opened 2024-05-10
Hook use-fetch throwing error to component crashes whole application


lucastzmov:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.9.8

### What package has an issue?

@mantine/hooks

### What framework do you use?

create-react-app (CRA)

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

In the use-fetch, when returning an error, the actual flow catches the error but then throw it again, it is difficult to catch it again at component level making the whole application crash.

With the actual use-fetch hook is not possible to code component as follows:

```javascript
if (loading) {
  return <Text>Loading...</Text>;
}

if (error) {
  return <Text>Oops...</Text>;
}

return (
  <>
    <div>{JSON.stringify(data)}</div>
  </>
);
```

### If possible, include a link to a codesandbox with a minimal reproduction

https://codesandbox.io/p/devbox/react-vite-ts-mantine-hook-use-fetch-6fpzrs

### Possible fix

I think it must returns the catched error to the consumer, and not throwing it again. The way it is now, when having an error in the promise, it is not returning inside the error's field but in the data one.

- The data's field must be only when has success;
- The error one must be the Error object returned from the promise;

Actual code handling error:

```javascript
.catch((err) => {
  setLoading(false);

  if (err.name !== 'AbortError') {
    setError(err);
  }

  throw err;
});
```

Suggestion:

```javascript
.catch((err) => {
  setLoading(false);

  if (err.name !== 'AbortError') {
    setError(err);
  }

  return err;
});
```

### Self-service

- [X] I would be willing to implement a fix for this issue

axelkovacki:
I got the same problem

rtivital:
Fixed in [7.10.1](https://github.com/mantinedev/mantine/releases/tag/7.10.1)

PR #6217, opened 2024-05-10
feat: use fetch not depends on fetch function and better error handling

ElTupac:
Based on the [issue #6215](https://github.com/mantinedev/mantine/issues/6215), that use-fetch hook throws error and crash whole application. Proposal is a better error handling model, and also to not depend the flow in the fetch function, leaving developer to use whatever promise handler he wants. Example of use:

- Component level:
```javascript
const { data, error, loading, fetch, abort } = useFetch(getInstallmentsList);

  useEffect(() => {
    fetch();
  }, []);
```

Service level:
```javascript
export const getInstallmentsList = async () => {
  const { data } = await axios.get<PaymentListEntity[]>(
    "https://xchjpacsyl.execute-api.us-east-1.amazonaws.com/prod/bill_order/v1/installments"
  );

  return data;
};
```

If the service function throws an error, it will be catched in the use-fetch hook and returned inside the `error` field. The function passed to the hook can receive the abort controller so you continue having the control to abort the request whenever needed.

### DIFF
--- a/packages/@mantine/hooks/src/use-fetch/use-fetch.ts
+++ b/packages/@mantine/hooks/src/use-fetch/use-fetch.ts
@@ -1,63 +1,39 @@
-import { useCallback, useEffect, useRef, useState } from 'react';
+import { useCallback, useMemo, useRef, useState } from 'react';
 
-export interface UseFetchOptions extends RequestInit {
-  autoInvoke?: boolean;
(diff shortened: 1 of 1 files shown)


rtivital:
It is a breaking change. The fix should not be implemented this way.

PR #6231, opened 2024-05-13
[@mantine/charts] BarChart: added waterfall type

Maetes:
This is related to #6224 

**Enhancement to ChartTooltip.tsx: Support for number[] in valueFormatter**

This pull request introduces support for an array of numbers (number[]) as input to the valueFormatter property in ChartTooltip.tsx. Previously, valueFormatter was strictly typed as (value: number) => string. However, the Recharts documentation indicates that both bar and area charts may utilize an array of numbers as input (see: [AreaChart API](https://recharts.org/en-US/api/AreaChart#data) and [BarChart API](https://recharts.org/en-US/api/BarChart#data)).

Modifying the valueFormatter type globally would impact other charts adversely. Therefore, I have refrained from altering the type but have extended the functionality in ChartTooltip.tsx to accommodate a tuple of two numbers, which is essential for the waterfall chart implementation. This update ensures compatibility without affecting existing chart functionalities.

**Enhancement to `ChartLegend.tsx`**

Accepts now a showColor prop to turn on and of the ColorSwatch (true) by default

**Review questions**
Please check that the special keys `color` and `standalone` are semantically okay for you to use as a data input.

`color`: beside of the series adds the functionality to change the color of a single Bar.
`standalone`: Only used for waterfall if you want to render a Bar which does not depend on the previous one.

### DIFF
--- a/apps/mantine.dev/src/pages/charts/bar-chart.mdx
+++ b/apps/mantine.dev/src/pages/charts/bar-chart.mdx
@@ -28,6 +28,16 @@ contribution of each series in terms of percentages.
 
 <Demo data={BarChartDemos.percent} />
 
+## Waterfall bar chart
+
+Set `type="waterfall"` to render a waterfall bar chart. This chart type illustrates how an
+initial value is influenced by subsequent positive or negative values,
+with each bar starting where the previous one ended.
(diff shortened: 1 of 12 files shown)


rtivital:
Thanks!

PR #6235, opened 2024-05-14
Update docs with description on how to use with PostCSS plugins as an…

brochington:
… Array

This PR add a small update to the docs to help describe how to update a PostCSS config file if the plugins are declared as an Array instead of an Object.

This is not a difficult change to understand, but it was a bit of a foot-gun in a current project that does use an Array for plugins.


rtivital:
Thanks for showing interest in contributing to Mantine! PostCSS configuration can be provided in very different formats. Mantine documentation provides only format that is compatible with all frameworks/bundlers listed in the documentation. Other formats are not planned to be added.

Issue #6220, opened 2024-05-10
use-throttled-value does not always issue an update with the last value


dfaust:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.9.0

### What package has an issue?

@mantine/hooks

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

All

### Describe the bug

Currently `useThrottledValue` (and the other throttled hooks) don't emit an update for the last value change if that change occurred while the timer was active. `lodash` calls this the trailing edge [1].
Ideally the hooks would support arguments to configure this behavior. But in absence of such arguments, I think emitting on the trailing edge is the more intuitive behavior.

I changed the implementation of `useThrottledValue` to fit my need.
Let me know if you are interested in a PR. You may also just take my code if you want to.

[1] https://lodash.com/docs/4.17.15#throttle

### Possible fix

```ts
import { useCallback, useEffect, useRef, useState } from 'react';

export function useThrottledValue<T>(value: T, wait: number) {
  const [throttledValue, setThrottledValue] = useState(value);
  const valueRef = useRef(value);
  const changedValueRef = useRef(value);
  const active = useRef(true);
  const waitRef = useRef(wait);
  const timeoutRef = useRef<number>(-1);

  const updateThrottledValue = useCallback((value: T) => {
    setThrottledValue(value);
    valueRef.current = value;
    changedValueRef.current = value;
    active.current = false;
  }, []);

  const timerCallback = useCallback(() => {
    if (changedValueRef.current !== valueRef.current) {
      updateThrottledValue(changedValueRef.current);

      window.clearTimeout(timeoutRef.current);
      timeoutRef.current = window.setTimeout(timerCallback, waitRef.current);
    } else {
      active.current = true;
    }
  }, [updateThrottledValue]);

  useEffect(() => {
    if (valueRef.current !== value) {
      if (active.current) {
        updateThrottledValue(value);

        window.clearTimeout(timeoutRef.current);
        timeoutRef.current = window.setTimeout(timerCallback, waitRef.current);
      } else {
        changedValueRef.current = value;
      }
    }
  }, [timerCallback, updateThrottledValue, value]);

  useEffect(() => () => window.clearTimeout(timeoutRef.current), []);

  useEffect(() => {
    waitRef.current = wait;
  }, [wait]);

  return throttledValue;
}
```

### Self-service

- [X] I would be willing to implement a fix for this issue

[...]

PR #6257, opened 2024-05-19
[@mantine/hooks] use-throttled-*: Emit updates on trailing edges

dfaust:
Make sure the throttled value/state/callback always has the last value it was called with after the timeout has expired.

- I refactored the hooks to all depend on a single hook (`useThrottledCallbackWithClearTimeout`) that contains the throttling logic. This hook is not exported.

- Mantine doesn't use the `react-hooks/exhaustive-deps` lint. So there is no warning for the missing `wait` in the old use-throttled-value implementation. This is why I introduced the `waitRef`. I later saw that use-throttled-state and use-throttled-callback did include the `wait` dependency. So maybe it was just a mistake. But anyway I kept the `waitRef` for now.

- I also adjusted the tests to reflect the new behavior and unified the "should clear timeout on unmount" tests.

Fixes: #6220


rtivital:
Thanks!

PR #6265, opened 2024-05-21
Fix multiple typos

a-kon:
Hey, thanks for the project. I've found a few typos while working with the library, so I think it is time to open a PR to fix them.


rtivital:
Thanks!

PR #6278, opened 2024-05-23
fix: avoid re throwing use-fetch error

ElTupac:
Based on the https://github.com/mantinedev/mantine/issues/6215, that use-fetch hook throws error and crash whole application. Proposal is just not re-throwing again the error to component level, since in React catching an error of a hook is not that easy as doing a try-catch.


rtivital:
Thanks!

PR #6279, opened 2024-05-23
feat: use-promise hook

ElTupac:
Proposal is to not depend the flow in the fetch function as occurs in the use-fetch hook, leaving developer to use whatever promise handler he wants. Example of use:

- Component level:
```javascript
const { data, error, loading, fetch, abort } = usePromise(getInstallmentsList);

useEffect(() => {
    fetch();
}, []);
```

Service level:
```javascript
const getInstallmentsList: (abortController: AbortController) => Promise<InstallmentListEntity[]> = async () => {
    await new Promise((resolve) => {
      setTimeout(resolve, 300);
    });
    return payments;
};
```

The function passed to the hook can receive the abort controller so you continue having the control to abort the request whenever needed.

### DIFF
--- a/packages/@mantine/hooks/src/use-promise/use-promise.ts
+++ b/packages/@mantine/hooks/src/use-promise/use-promise.ts
@@ -1,0 +1,39 @@
+import { useCallback, useMemo, useRef, useState } from 'react';
+
+export const usePromise = <T>(fn: (abortController: AbortController) => Promise<T>) => {
+  const [data, setData] = useState<T | null>(null);
+  const [loading, setLoading] = useState<boolean>(false);
+  const [error, setError] = useState<Error | null>(null);
+
+  const controller = useRef<AbortController | null>(null);
+
+  const fetch = useCallback(async () => {
+    if (!fn) throw new Error('Missing function parameter');
+
+    if (controller.current) controller.current.abort();
+
+    setError(null);
+    setLoading(true);
+
+    const newController = new AbortController();
+    controller.current = newController;
+    fn(newController)
+      .then(setData)
+      .catch((err) => {
+        setError(err);
+      })
+      .finally(() => {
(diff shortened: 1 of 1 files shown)


jvllmr:
I think this doesn't really fit into the scope of mantine.
If you need this functionality you can use tanstack/react-query, which covers this exact use-case.

Quick start:
https://tanstack.com/query/latest/docs/framework/react/quick-start

Cancellation:
https://tanstack.com/query/latest/docs/framework/react/guides/query-cancellation#query-cancellation

ElTupac:
> I think this doesn't really fit into the scope of mantine. If you need this functionality you can use tanstack/react-query, which covers this exact use-case.
> 
> Quick start: https://tanstack.com/query/latest/docs/framework/react/quick-start
> 
> Cancellation: https://tanstack.com/query/latest/docs/framework/react/guides/query-cancellation#query-cancellation

If it's not the scope of Mantine it's ok, just decline the PR. But I am proposing this as a better interface as the actual use-fetch hook, that is not really useful since it relies on fetch. Indeed Mantine is a UI library, and the hooks are mainly focused on UI development. I also think the correct state management of the Promises is part of that development, and since the interface given by the use-fetch hook is something really great(error, data, abort, loading), I reuse that in something more "useful".
And for little projects, the need to add more layers of complexity as react-query or similar approachs, seems like a terrible overkill. More if the projects are being mantained for lots of people that barely understands javascript at all.
And since this is not the first time someone says "if you need this functionality", I am opening a PR just because I want some specific code for a case of my own. I repeat again, Mantine already has the use-fetch hook, that is too simple and has no way of going out of it, and this is a proposal to have something with the same interface provided by it, but with a lot more of possibilities.

rtivital:
Thanks for submitting the PR, I'd rather not add such hook to the library

PR #6282, opened 2024-05-24
[@mantine/core] Fix: not applied bd style prop

ryuhangyeong:
## Description

After installing `7.10.0`, it does not work properly when using `bd` Style props.

## Example

```tsx
export function Usage() {
  return (
    <div style={{ padding: 40 }}>
      <Text bd={`1px solid ${'#333'}`}>Just text</Text>
    </div>
  );
}
```

## Before

image

## After

image


crosskayla:
Thank you 💯 Updated for this feature and would love to see it working

rtivital:
Thanks!

ryuhangyeong:
@rtivital 

According to my use case, I think I also need a style prop like below. 🤔

- `bdt` (border-top)
- `bdr` (border-right)
- `bdl` (border-left)
- `bdb` (border-bottom)

May I suggest an MR that adds that attribute?

ryuhangyeong:
Thank you for answer

PR #6307, opened 2024-05-29
[@mantine/hooks] use-eye-dropper.ts Fix: #6306

### LINKED ISSUE #6306: EyeDropper doesn't work in Opera browser
### Dependencies check up

7.10.0

@mantine/hooks

Next.js

### Describe the bug

Found out that EyeDropper hook does not work correctly in Opera browser. Despite the fact that there is EyeDropper key in window object and under the hood Opera uses Chromium. There will be no errors in the console, because there is a key in window and there are methods. 

Opera-05-29-2024-15-36-04

Opera-05-29-2024-15-46-18

_No response_

### Possible fix

As a solution I suggest not to show this dropper in Opera browser by adding a check:

```
function isOpera() {
    return navigator.userAgent.includes('OPR');
}
```

### Self-service

viktorkasap:
Added additional check for Opera browser, if it is Opera browser then do not show EyeDropper.

Fixes #6306

image

### DIFF
--- a/packages/@mantine/hooks/src/use-eye-dropper/use-eye-dropper.ts
+++ b/packages/@mantine/hooks/src/use-eye-dropper/use-eye-dropper.ts
@@ -10,7 +10,7 @@ export interface EyeDropperOpenReturnType {
 }
 
 function isOpera() {
-    return navigator.userAgent.includes('OPR');
+  return navigator.userAgent.includes('OPR');
 }
 
 export function useEyeDropper() {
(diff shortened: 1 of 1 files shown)


rtivital:
Please fix errors

viktorkasap:
> Please fix errors

Thank you for noticing it.
Fixed 👍

viktorkasap:
> Please fix errors

Hi Vitaly.
Does the error still exist?

rtivital:
Thanks!

PR #6319, opened 2024-06-01
[typo] Update copy-button.mdx

slk333:
children -> child


rtivital:
In this context, `children` is referencing React children prop.

PR #6325, opened 2024-06-02
FIX - Small glitch when using a confirm modal without children #6324

### LINKED ISSUE #6324: Small glitch when using a confirm modal without children
### Dependencies check up

7.10.1

@mantine/modals

Next.js

Chrome

### Describe the bug

Small visual glitch on the button focus when the confirm modal is not using children:

image

Sample code:

```
  const openResetConfirmModal = () => modals.openConfirmModal({
    centered: true,
    labels: {
      cancel: "Cancel",
      confirm: "Reset",
    },
    onConfirm: () => reset(),
    title: "Reset the form?",
    withCloseButton: false,
  });
```

_No response_

### Possible fix

_No response_

### Self-service

Prasiddha22:
Fixes #6324

Implement a condition within the **Group** inside **ConfirmModal.tsx** that dynamically sets the marginTop style property to 'md' if the component contains children, otherwise set it to '0'.

Before: 
image

After
image

### DIFF
--- a/packages/@mantine/modals/src/ConfirmModal.tsx
+++ b/packages/@mantine/modals/src/ConfirmModal.tsx
@@ -46,7 +46,7 @@ export function ConfirmModal({
     <>
       {children && <Box mb="md">{children}</Box>}
 
-      <Group justify="flex-end" {...groupProps}>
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #6326, opened 2024-06-03
[@mantine/core] Fix: correct a small typo referencing colorsTuple in border-resolver

djdduty:
This is not a reported issue, I just found what I believe to be a small typo when reading through the code base.

It seems this may have been an auto-complete mistake that also automatically imported the colorsTuple function from MantineProvider. Calling `length` on a function returns the number of arguments in the function definition, in this case colorsTuple takes a parameter, so its length is `> 0`.

It just so happens there's no bug or failing test case, as colorResolver will ultimately return an empty string for an empty string input, so this typo doesn't appear to really matter.


rtivital:
Thanks!

PR #6328, opened 2024-06-03
[@mantine/charts] fix: PieChart inner label not using valueFormatter 

Jones-Griffin:
Fixes PieChart using the valueFormatter for the tooltip but not for the label


rtivital:
Thanks!

Issue #6175, opened 2024-05-03
Modal ScrollArea hides scrollbar


nreisingercres:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.9.0

### What package has an issue?

@mantine/core

### What framework do you use?

create-react-app (CRA)

### In which browsers you can reproduce the issue?

All

### Describe the bug

ScrollArea's scroll bar hides behind header of Modal.
I also see this issue in the docs itself.
![image](https://github.com/mantinedev/mantine/assets/141150843/7e8570a9-06c3-41c3-a236-b13ea7c3715c)

### If possible, include a link to a codesandbox with a minimal reproduction

https://mantine.dev/core/modal/#usage-with-scrollarea

### Self-service

- [ ] I would be willing to implement a fix for this issue

rtivital:
Fixed in [7.11.2](https://github.com/mantinedev/mantine/releases/tag/7.11.2)

PR #6353, opened 2024-06-07
Increase scrollbar zIndex to overlap sticky elements

stefanzmf:
This fixes [6175](https://github.com/mantinedev/mantine/issues/6175).

Increase scrollbar zIndex to 1001 to overlap sticky elements


rtivital:
It should be a part of the Modal/Darwer component, not part of ScrollArea.

stefanzmf:
> It should be a part of the Modal/Darwer component, not part of ScrollArea.

Uh, now I see it. This means we have to decouple the `Scroll` logic from `ModalContent` and move it to `ModalBody` in order to have only that area scrollable, without the title part. It is a straighforward job, but I wonder if you think it can bring some unexpected behaviour in certain cases. 

Thank you

PR #6357, opened 2024-06-08
Update NavLink.demo.autoContrast.tsx

sarthak09walia:
changed color to c prop, to change color


rtivital:
The change is not correct.

PR #6370, opened 2024-06-11
[@mantine/core] add `withScrollArea` to `ModalContext` #6360

GZTimeWalker:
TL;DR:

Allows users to bypass the `Scroll` component to provide more flexibility for handling complex layouts.

Related feature request: https://github.com/orgs/mantinedev/discussions/6360

### DIFF
--- a/packages/@mantine/core/src/components/Modal/Modal.context.ts
+++ b/packages/@mantine/core/src/components/Modal/Modal.context.ts
@@ -6,6 +6,7 @@ export type ScrollAreaComponent = React.FC<any>;
 interface ModalContext {
   fullScreen: boolean | undefined;
   yOffset: string | number | undefined;
+  withScrollArea: boolean | undefined;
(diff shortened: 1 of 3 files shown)


rtivital:
I do not see a reason to add this prop. If you do not want styles from the library, you can replace scroll component with `scrollAreaComponent={({ children }) => <>{children}</>}`

GZTimeWalker:
> I do not see a reason to add this prop. If you do not want styles from the library, you can replace scroll component with `scrollAreaComponent={({ children }) => <>{children}</>}` 

Oh, sorry, I didn't realize that.

GZTimeWalker:
> If you do not want styles from the library, you can replace scroll component with `scrollAreaComponent={({ children }) => <>{children}</>}`

**Note:** Avoid using anonymous functions as components. This practice can trigger unnecessary re-renders for the modal, causing input fields to lose focus, which disrupts user interaction.

**Solution:** Use a static function or `useCallback` to store the component.

Issue #6103, opened 2024-04-18
Double-tap on select/multi-select inputs on mobile triggers unwanted tooltip


AndrejNemec:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.8.0

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

All

### Describe the bug

I am encountering a bug on mobile devices related to select/multi-select inputs. When I double-tap a select input that does not allow text entry, a mobile tooltip is triggered, offering autofill/paste options. This behavior is quite annoying for the users.

It seems that the issue is caused by an internal `input` element within the selector (`mantine-Select-input`), which is set to readonly. I have attached a video to demonstrate the issue.

https://github.com/mantinedev/mantine/assets/32060067/30a30e50-e7ec-46e8-a8b1-d79b1e8c354c

### Possible fix

Replace the input element with another element.

### Self-service

- [ ] I would be willing to implement a fix for this issue

Prasiddha22:
@rtivital I had made some changes and made a PR that resolved this issue. You can see that in the PR description that I had created. But that implementation failed some test cases and I closed the PR for now. Can you please check it once at your convenience? If that felt the right direction, we can further continue working on that.

minosss:
Em... I have tried many methods. Clicking on a **focused** input field that is read-only on a mobile device still brings up autofill.
#7012 also has the same problem.

AndrejNemec:
@tejas-1206 This issue occurs on all browsers on iOS devices. It is likely because all browsers on iOS use Safari’s WebKit engine under the hood, but I am not entirely certain about this claim. Importantly, the issue does not happen with other frameworks implementing custom selects (e.g., Radix Select); it is specific to Mantine.

tejas-1206:
@AndrejNemec That's interesting. I just made an observation as I re-installed firefox on my personal mobile and checked this. Quite weirdly, I can see a static text cursor right next to the selected value in the Select component on the doc. It's like firefox is treating the Select component as a Text Input box (and hence places a text cursor, but no keyboard pops up) instead of treating it as a read-only Input component.

You also pointed out that Radix Select does not have this issue on iOS devices, irrespective of the browser used. So, it can also be an issue from Mantine as well.

[...]

PR #6386, opened 2024-06-13
FIX - ISSUE #6103 - Double-tap on select/multi-select inputs on mobile triggers unwanted tooltip

Prasiddha22:
Double-tap on select/multi-select inputs on mobile triggers unwanted tooltip #6103

https://github.com/mantinedev/mantine/assets/54320725/9c945070-307d-43e2-b5b4-a822b96ba97f


Prasiddha22:
@rtivital I had made some changes and made a PR that resolved this issue. You can see that in the PR that I had created. But that implementation failed some test cases and I closed the PR for now. Can you please check it once at your convenience? If that felt the right direction, we can further continue working on that.

PR #6388, opened 2024-06-13
[@mantine/hooks] use-local-storage Update document to specify the getInitialValueInEffect default value

toanxyz:
The code is specifying this property as optional.
https://github.com/mantinedev/mantine/blob/29f24e31d9c33316395b317cd152648d55c65ef9/packages/%40mantine/hooks/src/use-local-storage/create-storage.ts#L15C3-L15C37

The document didn't specify it as optional:
https://github.com/mantinedev/mantine/blob/master/apps/mantine.dev/src/pages/hooks/use-local-storage.mdx#definition

The code also assign default value as "true" when not specified.
https://github.com/mantinedev/mantine/blob/29f24e31d9c33316395b317cd152648d55c65ef9/packages/%40mantine/hooks/src/use-local-storage/create-storage.ts#L78

It created a confusing when I was working with this hook at the 1st time and caused unexpected behavior.

### DIFF
--- a/apps/mantine.dev/src/pages/hooks/use-local-storage.mdx
+++ b/apps/mantine.dev/src/pages/hooks/use-local-storage.mdx
@@ -141,7 +141,7 @@ interface UseLocalStorage<T> {
   /** Default value that will be set if value is not found in local storage */
   defaultValue?: T;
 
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #6404, opened 2024-06-18
[@mantine/core] conver-css-variables for host selector

yyytae0:
[:host() mdn](https://developer.mozilla.org/en-US/docs/Web/CSS/:host_function)
css variables doesn't work in shadow dom even though set `cssVariablesSelector` to `:host` on `MantineProvider`
cause `:host` has another convention unlike `:root`

add ternary operator when `cssVariablesSelector` is `:host`


rtivital:
Thanks!

PR #6422, opened 2024-06-20
Fix slow rerendering of large tables (reopened)

Smert:
Hi! I have lags with rerendering of large tables with memoized rows:

```typescript
const TestRow = memo(function Row() {
  return (
    <Table.Tr>
      {new Array(10).fill(0).map((_, index) => (
        <Table.Td key={index}>Cell</Table.Td>
      ))}
    </Table.Tr>
  );
});

function TestTable() {
  const [counter, setCounter] = useState(0);
  return (
    <>
      <p>Counter: {counter}</p>
      <button onClick={() => setCounter(counter + 1)}>Increment</button>
      <Table>
        <Table.Tbody>
          {new Array(1000).fill(0).map((_, index) => (
            <TestRow key={index}/>
          ))}
        </Table.Tbody>
      </Table>
    </>
  );
}
```

Eventloop lags for 200ms when I click on the button in this simple example. But all rows are memoized.
I noticed then `TableContext` have a new value every render. It leads to rerender of `Table.Tr` and `Table.Td`:

Screenshot 2024-01-30 at 19 52 06

Memoizing of the context and `useStyle` can improve performance:
Screenshot 2024-01-30 at 19 48 48

I can suggest my decision of this problem in the pull request. Maybe you have another ideas.


Smert:
@2jordan3 sorry, I closed #5684 by accident.

> Putting String(name) in the dependency array may not work as intended.
How about putting the value of name directly in the dependency array?

I think value of name couldn't be putting directly in the dependency array, because this array can have different size in renders.

How about  use `JSON.stringify(name)`?

packages/@mantine/core/src/core/styles-api/use-styles/use-styles.ts
```diff
@@ -36,6 +37,18 @@ export type GetStylesApi<Payload extends FactoryPayload> = (
...
+function omitObjectProps(props: Record<string, any>): Record<string, any> {
+  const result: Record<string, any> = {};
+  for (const [key, value] of Object.entries(props)) {
+    if (!value || (typeof value !== 'object' && typeof value !== 'function')) {
+      result[key] = props[key];
+    }
+  }
+
+  return result;
+}
```
2jordan3, line 50: How about using Record<string, unknown>? It helps to improve type safety compared to any.

Also, is there a reason you used props[key]? Would it be problematic to just use value?

2jordan3:
> @2jordan3 sorry, I closed #5684 by accident.
> 
> > Putting String(name) in the dependency array may not work as intended.
> > How about putting the value of name directly in the dependency array?
> 
> I think value of name couldn't be putting directly in the dependency array, because this array can have different size in renders.
> 
> How about use `JSON.stringify(name)`?

I see, that method seems better!

Smert:
Fixed

jvllmr:
Wow, that is actually a significant performance improvement. Maybe this can get merged someday? Would help a table heavy app I have a lot.

WojakGra:
Is there anything preventing this from being added to the codebase?

rtivital:
Thanks for the PR and for the detailed investigation – the issue was real! Memoized table rows were rerendered every time `Table` rerendered because `getStyles` function and Table context value were recreated on every render.

The issue has been fixed differently in 3db26123ab: instead of changing `useStyles` and `createSafeContext` behavior for all components, `useStyles` now supports an opt-in `stable` option that keeps `getStyles` reference until its output changes, and `Table` memoizes its context value. With 1000 memoized rows × 10 cells, rerendering the parent component went from ~100ms (all rows rerendered) to ~0.5ms (no rows rerendered).

The approach in this PR was not merged because it removes object and function props from props passed to theme and component `classNames`/`styles`/`vars` functions, can return stale styles when only an object prop changes, and changes context behavior of all compound components. The fix will be available in the next patch release.

PR #6424, opened 2024-06-20
Add setValues  to dependency list in use-form-values.ts (fixes #6392)

### LINKED ISSUE #6392: onValuesChange not used in dependency list of setFieldValue (use-form-values.ts)
### Dependencies check up

7.10.2

@mantine/form

create-react-app (CRA)

All

### Describe the bug

If you change passed onValuesChange, some events will be fired with old handler

example
    const form = useForm({
        mode: 'uncontrolled',
        initialValues: selectedRecord as Cred,
        validate: {
            name: (value) => {
                return true
            }
        },
        // new handler every rerender
        onValuesChange: (values) => {
            // change some scoped value
            setValue({
              ...oldScopedValue,
              ...values
            }),
        },
    });

Let's see use-form-values.ts
https://github.com/mantinedev/mantine/blob/29f24e31d9c33316395b317cd152648d55c65ef9/packages/%40mantine/form/src/hooks/use-form-values/use-form-values.ts#L56

const setValues = useCallback(...,[onValuesChange])
 
setValues has onValuesChange in dependency list

const setFieldValue = useCallback(..., [])
const initialize = useCallback(..., [])
const resetValues = useCallback(..., [])

setFieldValue, initialize and resetValues has empty dependency list and will use ONLY first calculated setValues

_No response_

### Possible fix

_No response_

### Self-service

rguryanov:
Fixes https://github.com/mantinedev/mantine/issues/6392

### DIFF
--- a/packages/@mantine/form/src/hooks/use-form-values/use-form-values.ts
+++ b/packages/@mantine/form/src/hooks/use-form-values/use-form-values.ts
@@ -91,7 +91,7 @@ export function useFormValues<Values extends Record<PropertyKey, any>>({
           subscriber!({ path: payload.path, updatedValues, previousValues })
         );
     }
-  }, []);
+  }, [setValues]);
 
   const setValuesSnapshot = useCallback((payload: Values) => {
     valuesSnapshot.current = payload;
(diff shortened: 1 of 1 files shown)


rtivital:
Please fix errors in tests

rtivital:
I've implemented the fix on my side, thanks for the PR

PR #6432, opened 2024-06-21
[REF] more easy work with category icon

shurshilov:
It seems to me that transferring components at once instead of (text with mapping) looks easier


rtivital:
Thanks for submitting the PR! I'd rather keep the current logic as it keeps pages data serializable to JSON.

PR #6442, opened 2024-06-24
[@mantine/hooks]: use-os tests

marcobiedermann:
Add `useOs` unit tests

Fixtures taken from [MDN documentation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent)


rtivital:
Thanks!

Issue #6438, opened 2024-06-23
Modal overrides type augmentation not working anymore with TypeScript 5.5


lukasbash:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.10.2

### What package has an issue?

@mantine/modals

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Not applicable – issue is not related to the browser

### Describe the bug

Apparently the TypeScript team fixed a bug with regard to module augmentation of type aliases. Within the @mantine/modals package, the override is exported via type declaration. In order to make augmentation possible, it must be exported as an interface.

Reference issue: https://github.com/microsoft/TypeScript/issues/58961

### Self-service

- [ ] I would be willing to implement a fix for this issue

stevezhu:
For anyone running into this before the fix is merged, you can temporarily run `pnpm patch` or use https://www.npmjs.com/package/patch-package if your package manager doesn't support patching.

Steps for `pnpm patch`:
1. Run `pnpm patch @mantine/modals`
2. Modify the definition for `MantineModalsOverride`
```diff
-export type MantineModalsOverride = {};
+export interface MantineModalsOverride {}
```
3. Follow the instructions in terminal from pnpm to complete the patch

ihmpavel:
Same issue with [Custom color types](https://mantine.dev/theming/colors/#add-custom-colors-types)

florisdg:
@stevezhu Created a PR with your exact fix: https://github.com/mantinedev/mantine/pull/6443

ihmpavel:
Created PR for v6 https://github.com/mantinedev/mantine/pull/6445

PR #6443, opened 2024-06-24
Fix Modal overrides type augmentation not working anymore with TypeScript 5.5 #6438 

florisdg:
First open source PR I ever created 🙌 , hope this goes well

This PR should fix a new issue caused by TypeScript 5.5 https://github.com/mantinedev/mantine/issues/6438

PR #6445, opened 2024-06-24
fix: Typescript 5.5 compatibility for Mantine v6

ihmpavel:
TypeScript type augmentation does work with TypeScript 5.5

Fixes https://github.com/mantinedev/mantine/issues/6438 for Mantine v6

Same as https://github.com/mantinedev/mantine/pull/6443 for Mantine v7


rtivital:
Thanks!

PR #6449, opened 2024-06-25
Changes leftSection / rightSection to leftIcon / rightIcon

rafaelcacilhas:
leftSection and rightSection do not exist anymore. The correct props are leftIcon and rightIcon


rtivital:
The documentation is correct. In your application you have outdated Mantine packages.

PR #6466, opened 2024-06-27
Added defaultOpened prop on tooltip

jpranays:
A defaultOpened (false by default ) prop for tooltip component, which will show tooltip by default.
Added story and test for it.


jpranays:
Fixed prettier warnings

rtivital:
Thanks!

PR #6477, opened 2024-06-30
[@mantine/core] Select & MultiSelect: Show not found message even when not searchable

AustinWildgrube:
This PR ensures the `notFoundMessage` shows up even when no data is available. It can be confusing for users to click on the select and see nothing happen. The search functionality for both select and multi-select remains unchanged. The docs have also been updated to reflect this change.

References: #6474


colecoleman:
Ooh, love this change

rtivital:
Thanks!

PR #6490, opened 2024-07-05
[@mantine/emotion] Added back classes names to createStyles

Njiallu:
Added back classes names from `createStyles` to work like Mantine 6.


rtivital:
Thanks!

Issue #6497, opened 2024-07-07
floatingStrategy="fixed" has no effect in Tooltip


laugri:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.11.1

### What package has an issue?

@mantine/core

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

This is the same as the issue for Popover which was fixed last week: https://github.com/mantinedev/mantine/issues/6419

`floatingStrategy="fixed"` has no effect of Tooltips. It's still `absolute`. I expected it to `fixed`.

How to reproduce:

Use a tooltip anywhere with `floatingStrategy="fixed"` and `withinPortal={false}` and inspect the tooltip.

```tsx
import "@mantine/core/styles.css";
import { MantineProvider, Button, Tooltip } from "@mantine/core";

export default function App() {
  return (
    <MantineProvider>
      <>
        App
        <Tooltip
          withinPortal={false}
          floatingStrategy="fixed"
          label="Hello"
          opened={true}
        >
          <Button>Hello</Button>
        </Tooltip>
      </>
    </MantineProvider>
  );
}
```

### Self-service

- [ ] I would be willing to implement a fix for this issue

Issue #6419, opened 2024-06-20
floatingStrategy="fixed" has no effect in Popover


laugri:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.10.2

### What package has an issue?

@mantine/core

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

Chrome

### Issue

`floatingStrategy="fixed"` has no effect of Popover dropdown element. It's still `absolute`. I expected it to `fixed`. From what I can tell, nothing changes with the option.

How to reproduce:

Use a popover anywhere with `floatingStrategy="fixed"` and `withinPortal={false}` and inspect the dropdown.

```tsx
import "@mantine/core/styles.css";
import {
  MantineProvider,
  Button,
  Popover,
  PopoverDropdown,
  PopoverTarget,
} from "@mantine/core";

export default function App() {
  return (
    <MantineProvider>
      <>
        App
        <Popover
          width={400}
          position="top"
          shadow="md"
          withinPortal={false}
          floatingStrategy="fixed"
          radius="8"
        >
          <PopoverTarget>
            <Button>Hello</Button>
          </PopoverTarget>
          <PopoverDropdown style={{ pointerEvents: "none" }}>
            Content
          </PopoverDropdown>
        </Popover>
      </>
    </MantineProvider>
  );
}
```

![image](https://github.com/mantinedev/mantine/assets/7806174/5ebd6e7b-ef45-4c77-8c70-5430ec6bac8f)

### Context

Because I am developing a chrome extension content script UI, my app is in a shadow root and I can't use Portals. For this reason, I am using Popovers with `withinPortal={false}`. I have a clipping problem (as shown in the screenshot below), so I tried `floatingStrategy="fixed"` as [recommended by floating-ui ](https://floating-ui.com/docs/misc#clipping)for clipping issues.
![image](https://github.com/mantinedev/mantine/assets/7806174/a6908397-dc08-49b1-8112-883b40130b68)

### If possible, include a link to a codesandbox with a minimal reproduction

[...]

PR #6502, opened 2024-07-09
[@mantine/core] Tooltip floatingStrategy=fixed updates position to fixed

lwkchan:
Fixes issue raised here: https://github.com/mantinedev/mantine/issues/6497 

I followed what was done in https://github.com/mantinedev/mantine/issues/6419, which fixes the same issue for the Popover component.


rtivital:
Thanks!

PR #6505, opened 2024-07-10
[@mantine/tiptap]: Add a text field that allows user to supply displa…

sindrefl:
…y text for inserting a link. If user has selected a text, use this text as a default


rtivital:
Thanks for showing interest in contributing to Mantine, this change is not planned to be implemented – I consider the current implementation to be good enough. If you are interested in using this feature on your side, you can build a custom tiptap extension for this.

TannerJuby1:
@rtivital In the current implementation, when the cursor is at the end of the text but "inside" a link, there is no clear, intuitive way to "exit" out of the link to continue typing non-link text. Is there a configuration to change this behavior?

Issue #6508, opened 2024-07-10
useOs() is incorrect for iPad Safari


justinkao:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.11.1

### What package has an issue?

@mantine/hooks

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Safari

### Describe the bug

`useOs()` is incorrect for iPad Safari -- it reports as `macos`. Easy to reproduce by opening https://mantine.dev/hooks/use-os/ on an iPad and looking at the first example.

The problem is that `useOs()` only looks at the user agent, where iPad Safari has reported itself as `Macintosh` for some time now. (E.g. see https://stackoverflow.com/questions/56934826/distinguish-between-ipad-and-mac-on-ipad-with-ipados)

### Self-service

- [ ] I would be willing to implement a fix for this issue

justinkao:
@rtivital @Kenzo-Wada I don't think the linked PR actually fixes anything, as the user agent on iPad Safari doesn't contain the string `iPad` at all. For example on my iPad, the user agent is:
```
Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/17.5 Safari/605.1.15
```

Kenzo-Wada:
Oh... So sorry for this...
Ill keep on moving to fix this bug

PR #6511, opened 2024-07-11
fix(@mantine/hooks): useOS returns macos when ipad

Kenzo-Wada:
fixes: https://github.com/mantinedev/mantine/issues/6508
 

reason of bug  this is that ipad also contains "Machintosh".
So that I've changed to check for IOS first!


rtivital:
Thanks!

PR #6517, opened 2024-07-12
MANTINE-6336: Fix DateInput to setInputValue to _value which has time zone applied

### LINKED ISSUE #6336: DateInput flickers the wrong date when used with timezones.
### Dependencies check up

7.10.1

@mantine/dates

Next.js

Chrome

### Describe the bug

When using `DateInput` as a controlled input and using a timezone, the displayed date will flicker to your localized time before switching to the timezone provided.  The time value is correct and does not change, merely the displayed date.

I was able to reproduce this in a sandbox and in my local mantine storybook.

https://codesandbox.io/p/sandbox/dateinput-timezone-issue-ywwqtq?file=%2Fsrc%2FApp.tsx

### Possible fix

Not 100% sure.  One way would be to set the `input` to use the provided timezone, or prevent it from changing the provided value.

### Self-service

Sergio16T:
Fixes #6336 

Reviewed code and discovered that the useDidUpdate hook was not setting the state value with timezone when applicable. 

Before: It was running and setting the inputValue to the initial onChange Value without applying the updates from the useUncontrolledDates Hook. 

After: _value is the source of truth to display in input and will display with updates applied by useUncontrolledDates Hook.

### DIFF
--- a/packages/@mantine/dates/src/components/DateInput/DateInput.tsx
+++ b/packages/@mantine/dates/src/components/DateInput/DateInput.tsx
@@ -245,8 +245,8 @@ export const DateInput = factory<DateInputFactory>((_props, ref) => {
     ) : null);
 
   useDidUpdate(() => {
-    value !== undefined && !dropdownOpened && setInputValue(formatValue(value!));
-  }, [value]);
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #6535, opened 2024-07-16
[@mantine/hook]: fix use-os returns `macos` when ipad

Kenzo-Wada:
fixes: https://github.com/mantinedev/mantine/issues/6508

added `ontouchend` to determine ipad air and pro.


rtivital:
Thanks!

PR #5886, opened 2024-03-07
[@mantine/charts] Fix: `Chartlegend` and `ChartTooltip` not displaying values when they are nested

ID-JA:
This PR fixes  #5885

PR #6536, opened 2024-07-16
[@mantine/charts]: Fix LineChart Legend and Tooltip to support nested names

### LINKED ISSUE #6529: Graphs with nested values ignore "label" property
### Dependencies check up

7.11.2

@mantine/charts

Vite

All

### Describe the bug

Graphs that are using nested values now ignore the "label" property for tooltips and legends and instead use the last part of the "name" property.
Image from 7.11.1:
image

Image from 7.11.2:
image

The included code sandbox shows the issue and can be fixed by reverting the version of all mantine packages to 7.11.1.

I believe this is related to #5885 and its associated PR #5886.

https://codesandbox.io/p/sandbox/mantine-react-template-forked-cxqmf8

### Possible fix

The following files had new functions added as part of the mentioned PR:
packages/@mantine/charts/src/ChartLegend/ChartLegend.tsx  updateChartTooltipPayload
packages/@mantine/charts/src/ChartTooltip/ChartTooltip.tsx  updateChartLegendPayload

These override the name with the nested property name, I assume these should instead be using label property with the fallback being the nested property name.

### Self-service

Sergio16T:
Fixes #6529 
Relates to #5886 

Update tooltip payload by unpacking item's not related to the nested name. Then unpack nested details into same object. 

Updates getSeriesLabels to account for nested name. 

Adds Unit Tests for displaying legend and tooltip label and data.

### DIFF
--- a/packages/@mantine/charts/src/ChartTooltip/ChartTooltip.tsx
+++ b/packages/@mantine/charts/src/ChartTooltip/ChartTooltip.tsx
@@ -1,4 +1,3 @@
-import _ from 'lodash';
 import {
   Box,
   BoxProps,
@@ -21,7 +20,7 @@ function updateChartTooltipPayload(payload: Record<string, any>[]): Record<strin
     const matchFound = item.name.search(/\./);
     if (matchFound >= 0) {
       const newDataKey = item.name.substring(0, matchFound);
-      const nestedPayload = _.clone(item.payload[newDataKey]);
+      const nestedPayload = { ...item.payload[newDataKey] };
       const shallowPayload = Object.entries(item.payload).reduce((acc, current) => {
         const [k, v] = current;
         return k === newDataKey ? acc : { ...acc, [k]: v };
--- a/packages/@mantine/charts/src/ChartTooltip/ChartTooltip.tsx
+++ b/packages/@mantine/charts/src/ChartTooltip/ChartTooltip.tsx
@@ -1,3 +1,4 @@
+import _ from 'lodash';
 import {
   Box,
   BoxProps,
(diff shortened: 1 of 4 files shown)


packages/@mantine/charts/src/ChartTooltip/ChartTooltip.tsx
```diff
@@ -1,3 +1,4 @@
+import _ from 'lodash';
```
rtivital, line 1: `@mantine/charts` package does not depend on `lodash`, this dependency cannot be used.

packages/@mantine/charts/src/LineChart/LineChart.test.tsx
```diff
@@ -11,6 +24,82 @@ const defaultProps: LineChartProps = {
   style: { width: 200, height: 200 },
 };
 
+const nestedData = [
+  {
+    date: 'Mar 22',
```
rtivital, line 29: Too much data, it is not needed for tests, reduce it to 2 items per data array.

packages/@mantine/charts/src/LineChart/LineChart.test.tsx
```diff
@@ -1,6 +1,19 @@
-import { patchConsoleWarn, tests } from '@mantine-tests/core';
+import React from 'react';
+import { patchConsoleWarn, render, tests } from '@mantine-tests/core';
 import { LineChart, LineChartProps, LineChartStylesNames } from './LineChart';
 
+jest.mock('recharts', () => {
```
rtivital, line 5: Why is this required? Other tests work correctly without it.

rtivital (changes requested):

packages/@mantine/charts/src/LineChart/LineChart.test.tsx
Sergio16T: Hi @rtivital! Thank you for the PR review. 

This mock enables LineChart to be rendered in the DOM. Without this mock we are not able to do assertions that elements are in the DOM and rendered correctly. Prior to this the unit tests did not assert that any elements were rendering correctly for LineChart. 

Before tests.itSupportsSystemProps were asserting that system properties were supported and this particular test did not rely on LineChart mounting to be able to test elements rendering correctly.

rtivital:
Thanks!

PR #6539, opened 2024-07-17
fix: source link of Box was 404

hajimism:
old: https://github.com/mantinedev/mantine/blob/master/packages/@mantine/core/src/components/Box/Box.tsx

new: https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/core/Box/Box.tsx


rtivital:
Thanks!

PR #6541, opened 2024-07-17
[@mantine/core]: add ChromeOS to useOS

Kenzo-Wada:
# What
add chromeos to useOS

# Why
I realized that useOS returns `undetermined` on ChromeOS.


rtivital:
It is a breaking change, cannot be merged until 8.0 release. Changes of OS type may introduce unexpected errors, for example, when it is used as a key of a record: `type OSLabels = Record<OS, ReactNode>`.

Kenzo-Wada:
@rtivital 

I see, should I close this PR for now? or is it ok to keep here opened until next major release?

rtivital:
Let's keep it open

rtivital:
Thanks!

PR #6569, opened 2024-07-22
Allow overwrite of default onKeyDown behaviour of the TagsInput component

### LINKED ISSUE #6568: The default onKeyDown behaviour of the TagsInput component cannot be overwritten
It would be nice if the default onKeyDown behaviour of the TagsInput component could be overwritten in certain cases, for example to implement a custom Backspace behaviour for the deletion of Tags.

This would just need one additional line after callign the onKeyDown handler from the props, to cancel the default behaviour if the custom handler cancels the event propagation (https://github.com/mantinedev/mantine/pull/6569):

    if (event.isPropagationStopped()) return;

inx-fldu:
Fixes #6568
Allow disabling of default key down behaviour by stopping the event propagation.


inx-fldu:
--> https://github.com/mantinedev/mantine/issues/6568

rtivital:
Thanks!

PR #6583, opened 2024-07-24
fix: Mantine 6.x MantineThemeOther support for module augmentation for TypeScript 5.5

jacobtoye:
Mantine 6.0.22 introduced some fixes to support module augmentation. This is documented on the [Mantine.dev](https://v6.mantine.dev/theming/theme-object/#other) website as a way to support typing the `MantineThemeOther` object. This PR specifically address `MantineThemeOther`, which was missed.

The issue was a result of Typescript fixing a bug, see https://github.com/microsoft/TypeScript/issues/58961

See https://github.com/mantinedev/mantine/pull/6445 for previous fixes.


rtivital:
Thanks!

jacobtoye:
@rtivital do you think we could get a new release for this please?

PR #6586, opened 2024-07-25
Fix checked chip padding at XL size

david-szabo97:
Fixes the incorrect padding for checked Chip at XL size.

image

Formula: 32-(18+(22/1.5))/2


rtivital:
Thanks!

PR #6598, opened 2024-07-28
[@mantine/core]: Allow passing custom size apply function for popover middleware

Jones-Griffin:
Currently the Popover's middleware accepts an `apply` function under the size prop, however it will never be run and a default apply function is run instead.


rtivital:
Thanks!

PR #6607, opened 2024-07-31
[@mantine/core]: Remove invalid and redundant parentheses in Slider

bastiankistner:
This PR removes unnecessary parentheses in the `Slider.module.css` that can lead to an error when imported.


rtivital:
Thanks!

PR #6616, opened 2024-08-01
[@mantine/charts] Add valueFormatter support on outside labels

### LINKED ISSUE #6614: Mantine Charts PieChart ValueFormatter not working with outside label position
### Dependencies check up

7.11.2

@mantine/charts

Vite

All

### Describe the bug

When displaying a pie chart and using a value formatter, the label is not formatted when the `labelPosition="outside"`

Ex
``` js
  <PieChart
    data={chartData}
    labelsPosition="inside"
    labelsType="value"
    mx="auto"
    valueFormatter={numberFormatter}
    withLabels
    withTooltip
  />
```
image

making no other change than setting `labelPosition:"outside"` the formatter is no longer used on the labels:

image

Note that changing the labelPosition prop does not effect the valueFormatter being used on the tooltip

_No response_

### Possible fix

_No response_

### Self-service

stefanzmf:
Fixes #6614

### DIFF
--- a/packages/@mantine/charts/src/PieChart/PieChart.story.tsx
+++ b/packages/@mantine/charts/src/PieChart/PieChart.story.tsx
@@ -21,24 +21,32 @@ export function Usage() {
 
 export function WithLabels() {
   return (
-    <div style={{ padding: 40 }}>
-      <PieChart data={data} withLabels size={200} />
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #6617, opened 2024-08-01
Show warnings on unload when dirty

jessedijkstra:
Checks if the form is dirty when the user unloads the window and shows a warning.

I've build this several times when using form to ensure the user doesn't loose changes when the form is dirty. 

This dirty check is reset after using setDirty or resetting the form.

Steps to reproduce:
- Start storybook
- Go to /?path=/story/form--demo-show-warning-on-unload-when-dirty
- Edit the field
- Close the window
- Alert is shown for unsaved changes


rtivital:
Thanks for showing interest in contributing to Mantine! Unfortunately, this feature is not planned to be added to the library and the PR cannot be merged.

PR #6622, opened 2024-08-02
Fix SegmentedControl FloatingIndicator Bug

### LINKED ISSUE #6527: SegmentedControl displays incorrect value after opening or collapsing AppShell Navbar
### Dependencies check up

7.11.2

@mantine/core

Next.js

All

### Describe the bug

When a non-default option is selected in `SegmentedControl`, the incorrect value is **displayed** in SegmentedControl when the collapsed state of `AppShell.Navbar` changes. The actual state does not change, just the rendered value.

Steps to reproduce
1. See the attached CodeSandbox
2. Select 'Option 2' in the `SegmentedControl`
3. Open the `AppShell.Navbar` using the burger icon
4. Observe that the incorrect value is displayed

https://codesandbox.io/p/sandbox/mantine-react-template-forked-lg4dq7
### Possible fix

_No response_

### Self-service

michaelperros:
Fixes #6527 

The problem was that the transitionend event listener was added to the entire document, so it was picking up transitions from any element. 

When the user clicked the burger, it opened the Navbar which changed the offsetLeft of the FloatingIndicator parent. This triggered the useEffect and called isParent with the transition event from the Navbar burger. This made isParent return true and then call updatePositionWithoutAnimation() which caused the bug.

By attaching the event listener directly to the parent element, the handler only responds to transitions within the parent and its children, resulting in correct behavior.

Please let me know if this good. Happy to revise based on your feedback.

Thanks,
Michael

### DIFF
--- a/packages/@mantine/core/src/components/FloatingIndicator/use-floating-indicator.ts
+++ b/packages/@mantine/core/src/components/FloatingIndicator/use-floating-indicator.ts
@@ -108,9 +108,9 @@ export function useFloatingIndicator({
         }
       };
 
-      document.addEventListener('transitionend', handleTransitionEnd);
+      parent.addEventListener('transitionend', handleTransitionEnd);
       return () => {
-        document.removeEventListener('transitionend', handleTransitionEnd);
+        parent.removeEventListener('transitionend', handleTransitionEnd);
       };
     }
 


rtivital:
Thanks!

PR #6625, opened 2024-08-02
[@mantine/core] SegmentedControl: Add missing disabled data-attribute

risen228:
Hello. I want to customize the `root` styles of segmented control in `disabled` state, but it's not possible without `data-disabled` attribute on `root`, because the only thing you can use in CSS in that case is `:disabled` pseudo-class on nested inputs, but there are no parent selectors implemented in CSS right now, so basically it's just not possible to change `root` styles without JS when it's child inputs are disabled. But JS logic doesn't allow to customize this behavior once and for all (I mean in theme), and I have to create my own wrapper-component.

I suggest adding this attribute to make customization easier


rtivital:
Thanks!

PR #6629, opened 2024-08-03
FIX: badge dot width fix without breaking rounded badge styles

ddoemonn:
I encountered an issue with text overflow in badges inside tables at my company. The dot variant was particularly problematic, with the dot's width shrinking when the overflow was too extensive. I have fixed this issue and will post before and after images to illustrate the changes.

Note: The new styles added to the badge root caused a problem with rounded badges. I also addressed this issue and ensured all tests passed. I checked all badge formats, and everything is now working correctly.

### Before:
WhatsApp Image 2024-08-03 at 17 56 10
WhatsApp Image 2024-08-03 at 17 56 31

### After:
WhatsApp Image 2024-08-03 at 17 56 45

#### Also you can see rounded badges are working correctly:
WhatsApp Image 2024-08-03 at 17 59 39(1)

### DIFF
--- a/packages/@mantine/core/src/components/Badge/Badge.module.css
+++ b/packages/@mantine/core/src/components/Badge/Badge.module.css
@@ -33,7 +33,8 @@
   line-height: var(--badge-lh);
   text-decoration: none;
   padding: 0 var(--badge-padding-x);
-  display: inline-flex;
+  display: grid;
+  grid-template-columns: auto 1fr auto;
   align-items: center;
   justify-content: center;
   width: fit-content;
@@ -54,6 +55,7 @@
 
   &:where([data-circle]) {
     padding-inline: 2px;
+    display: flex;
     width: var(--badge-height);
   }
 }


rtivital:
The screenshot of rounded badges that you've sent has issues with centering. Same for the regular badge – the label is not centered vertically.

ddoemonn:
Yes, you're right. The badges do have centering issues. I'd like to fix these issues in this pull request. Can I proceed with that? @rtivital

rtivital:
I'll run your code later and see if additional changes are required, might be the issue only with your screenshots.

ddoemonn:
Actually, I think the problem was with my browser (I was using Firefox). I compared mantine.dev and my localhost, and I can confirm the results are the same. 
WhatsApp Image 2024-08-03 at 21 16 46
WhatsApp Image 2024-08-03 at 21 17 13

rtivital:
I've checked, it works fine, thanks!

PR #6632, opened 2024-08-03
[@mantine/spotlight] FIX #6620: spotlight limit prop

### LINKED ISSUE #6620: Spotlight limit prop behaves strangely with grouped actions
### Dependencies check up

7.11.2

@mantine/spotlight

Next.js

All

### Describe the bug

Limit prop does not correctly limit actions when the actions are grouped using `SpotlightActionGroupData[]`.

See codesandbox. Change the limit prop and observe that an incorrect number of actions is displayed.

https://codesandbox.io/p/sandbox/mantine-react-template-forked-66k5jn

### Possible fix

Fix in `limit-actions.ts` (tested working as intended via `pnpm patch` in my project)

```ts
import { isActionsGroup } from './is-actions-group';
import type { SpotlightActionData, SpotlightActions } from './Spotlight';

export function limitActions(actions: SpotlightActions[], limit: number) {
  if (!Array.isArray(actions)) {
    return [];
  }

  let count = 0;
  return actions.reduce<SpotlightActions[]>((acc, item) => {
    if (count >= limit) {
      return acc;
    }

    if (isActionsGroup(item)) {
      const groupActions = limitActions(item.actions, limit - count);
      acc.push({
        group: item.group,
        actions: groupActions as SpotlightActionData[],
      });
      count += groupActions.length;
    } else {
      acc.push(item);
      count++;
    }

    return acc;
  }, []);
}
```

### Self-service

yorkeJohn:
Fixes #6620

### DIFF
--- a/packages/@mantine/spotlight/src/limit-actions.ts
+++ b/packages/@mantine/spotlight/src/limit-actions.ts
@@ -2,30 +2,28 @@ import { isActionsGroup } from './is-actions-group';
 import type { SpotlightActionData, SpotlightActions } from './Spotlight';
 
 export function limitActions(actions: SpotlightActions[], limit: number) {
-  const result: SpotlightActions[] = [];
-
   if (!Array.isArray(actions)) {
     return [];
   }
 
-  for (let i = 0; i < actions.length; i += 1) {
-    const item = actions[i];
-
-    if (result.length >= limit) {
-      return result;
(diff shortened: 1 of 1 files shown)


yorkeJohn:
Fixed eslint error.

rtivital:
Thanks!

PR #6639, opened 2024-08-06
[@mantine/form] Call `form.watch` subscribers on `form.initialize`

### LINKED ISSUE #6638: form.watch does not trigger when form.initialize is called
### Dependencies check up

7.12.0

@mantine/form

Vite

All

### Describe the bug

When using and uncontrolled form and form.watch will not trigger when form.initialize is called. In this case, the value will correctly update but the change will not be transmitted to the form.watch subscribers.

The expected behavior is the subscribers being called every time the value changes, including by the form initialization using form.initialize.

In the provided codesandbox it is shown an example of a staled value (empty) being rendered because of this problem.

https://codesandbox.io/p/sandbox/elegant-saha-h5cs6f

### Possible fix

_No response_

### Self-service

raulfpl:
Fixes #6638

This PR adds tests covering the problem of `form.watch` subscribers not being called when the value changes because of `form.initialize()` as described in #6638 and fixes it.

The way it fixes it is by notifying the subscribers the same way `setValues` does it.

Let me know if this is not the right way to fix this.

### DIFF
--- a/packages/@mantine/form/src/use-form.ts
+++ b/packages/@mantine/form/src/use-form.ts
@@ -61,23 +61,33 @@ export function useForm<
     mode === 'uncontrolled' && setFormKey((key) => key + 1);
   }, []);
 
-  const initialize: Initialize<Values> = useCallback((values) => {
-    const previousValues = $values.refValues.current;
-    $values.initialize(values, () => mode === 'uncontrolled' && setFormKey((key) => key + 1));
-    clearInputErrorOnChange && $errors.clearErrors();
-    mode === 'uncontrolled' && setFormKey((key) => key + 1);
+  const handleValuesChanges = useCallback(
+    (previousValues: Values) => {
+      clearInputErrorOnChange && $errors.clearErrors();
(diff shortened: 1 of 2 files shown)


packages/@mantine/form/src/use-form.ts
```diff
@@ -62,8 +62,22 @@ export function useForm<
   }, []);
 
   const initialize: Initialize<Values> = useCallback((values) => {
+    const previousValues = $values.refValues.current;
     $values.initialize(values, () => mode === 'uncontrolled' && setFormKey((key) => key + 1));
-  }, []);
+    clearInputErrorOnChange && $errors.clearErrors();
+    mode === 'uncontrolled' && setFormKey((key) => key + 1);
+
+    Object.keys($watch.subscribers.current).forEach((path) => {
```
rtivital, line 70: This logic must not be duplicated, it is already present in setValues handler.

rtivital (changes requested):

rtivital:
Thanks!

PR #6646, opened 2024-08-08
Fix: DateInput does not default to minDate month

### LINKED ISSUE #5695: The minDate in Dates should be the default date, when date is not given

michaelperros:
Fixes #5695 

If you need any adjustments, please let me know. Happy to revise based on your feedback.

Thanks

### DIFF
--- a/packages/@mantine/dates/src/components/Calendar/Calendar.tsx
+++ b/packages/@mantine/dates/src/components/Calendar/Calendar.tsx
@@ -261,8 +261,10 @@ export const Calendar = factory<CalendarFactory>((_props, ref) => {
   const ctx = useDatesContext();
 
   const _columnsToScroll = columnsToScroll || numberOfColumns || 1;
-  const currentDate =
-    _date || shiftTimezone('add', new Date(minDate || Date.now()), ctx.getTimezone());
+
+  const now = new Date();
+  const fallbackDate = minDate && minDate > now ? minDate : now;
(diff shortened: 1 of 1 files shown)


michaelperros:
Sorry this has failed checks. I will revise and update

michaelperros:
Checks passed! Let me know if you need anything else here.

Additionally, I wanted to note that I found a separate bug in both the master and my pr. The bug is that when minDate is set, the selectable days in the DateInput Calendar start one day before the minDate. The expected behavior is that the selectable days would start on the minDate. Since it's happening in the master and is not directly related to my pr or this issue, I did not fix it here. I will create a new issue shortly to address it.

rtivital:
I'd prefer this logic to be a little different to avoid accidental breaking changes. It should check whether the minDate is after the current date and only if it is, use it as the default date.

michaelperros:
Good call. Code has been update. Let me know if any further tweaks are needed. Thanks

rtivital:
Thanks!

PR #6665, opened 2024-08-14
fix: let discordLink prop be optional

Ives7:


rtivital:
Thanks for showing interest in contributing to Mantine, this change is not planned to be implemented as it makes code less type safe.

Ives7:
> Thanks for showing interest in contributing to Mantine, this change is not planned to be implemented as it makes code less type safe.
Thanks for your reply first.
But this behavior weird,when i want to disabled discordLink , i need to pass a false to withDiscord .
And then, this componet ask me i pass a discordLink .

image

Ives7:
image
@rtivital It's pass now.

PR #6680, opened 2024-08-16
fix: badge content is not centered with custom width

Meierschlumpf:
Hey this morning I came across the following badges in our application:

image

As you can see they are no longer centered. I took a look in the history and found the pull request #6629 from last week. The changes I made allow the content to still be centered, but also allow the dot to be it's full size when having it small:

image
image
image
image
image
image
image


rtivital:
Thanks for submitting a PR, display was changed to grid, it is not planned to change it to inline-flex back. The issue will be resolved differently.

PR #6683, opened 2024-08-16
[@mantine/hooks] use-idle: Start timer immediately instead of waiting for an event (#6682)

### LINKED ISSUE #6682: useIdle doesn't start tracking until after the first event
### Dependencies check up

7.12.1

@mantine/hooks

Vite

All

### Describe the bug

The `useIdle` hook does not start its timer until after the first event occurs. For example, with the code below, if there are no click or keypress events after the page loads, `idle` will _never_ be true.

```ts
const idle = useIdle(5000, {
    initialState: false,
    events: ['click', 'keypress'],
});
```

I found one report of this in Discord, but it was posted in November 2023 and has no replies to the initial posting.

https://codesandbox.io/p/sandbox/useidle-mantine-bug-tx5znz?file=%2Fsrc%2FApp.tsx

### Possible fix

Start the timer immediately in the `useIdle` hook instead of waiting for an event first.

### Self-service

chimericdream:
This came up when I was implementing automatic logout in our application. If a given event _never_ fires, the current implementation of `useIdle` will never return `true`.

This PR starts the timer immediately after setting up the event listeners.

Fixes #6682

### DIFF
--- a/packages/@mantine/hooks/src/use-idle/use-idle.test.tsx
+++ b/packages/@mantine/hooks/src/use-idle/use-idle.test.tsx
@@ -11,6 +11,20 @@ describe('@mantine/hooks/use-idle', () => {
     expect(hook.result.current).toBe(true);
   });
 
+  it('Starts the timer immediately instead of waiting for the first event to happen', () => {
+    const spy = jest.spyOn(window, 'setTimeout');
+    expect(spy).not.toHaveBeenCalled();
+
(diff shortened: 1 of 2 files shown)


rtivital:
Please fix errors in tests

chimericdream:
Fixed.

rtivital:
Thanks!

PR #6697, opened 2024-08-19
[@mantine/core] Tree: Add optional initialCheckedState property

fredgig:
Motivation: This PR aims to add an optional `initialCheckedState` prop to `useTree,` akin to `initialExpandedState` and `initialSelectedState`, which addresses #6641.

- Adds `initialCheckedState` property to `useTree` hook;
- Adds default property to `checkedState` hook;
- Renames internal `getInitialState` function to `getInitialExpandedState`;
- Adds internal `getInitialCheckedState` function: This takes the current tree checkedState and ensures only leaf nodes are present, which allows users to input parent nodes in the `initialCheckedState` property as it outputs their leaf nodes;
- Adds `setCheckedState` instruction and dependency to `initialize` callback to update `checkedState` with behaviour indicated in previous point.

### DIFF
--- a/packages/@mantine/core/src/components/Tree/use-tree.ts
+++ b/packages/@mantine/core/src/components/Tree/use-tree.ts
@@ -27,7 +27,9 @@ function getInitialExpandedState(
   return acc;
 }
 
-function getInitialCheckedState(initialState: string[], data: TreeNodeData[], acc: string[] = []) {
+function getInitialCheckedState(initialState: string[], data: TreeNodeData[]) {
+  const acc: string[] = [];
+
   initialState.forEach((node) => acc.push(...getChildrenNodesValues(node, data)));
 
   return acc;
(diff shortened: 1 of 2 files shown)


packages/@mantine/core/src/components/Tree/use-tree.ts
```diff
@@ -20,20 +20,29 @@ function getInitialState(
...
     if (Array.isArray(node.children)) {
-      getInitialState(initialState, node.children, value, acc);
+      getInitialExpandedState(initialState, node.children, value, acc);
     }
   });
 
   return acc;
 }
 
+function getInitialCheckedState(initialState: string[], data: TreeNodeData[], acc: string[] = []) {
```
rtivital, line 30: If acc is not part of recursive function, it should be declared in the function body.

rtivital:
Thanks!

PR #6701, opened 2024-08-20
[@mantine/dropzone] Added 7z & rar mime types

Streusel:
Added 7z & rar mime types for simpler usage in Dropzone and more aesthetically pleasing usage.

Note: rar can supposedly also be represented as [application/vnd.rar](https://developer.mozilla.org/en-US/docs/Web/HTTP/Basics_of_HTTP/MIME_types/Common_types), [application/x-rar-compressed](https://en.wikipedia.org/wiki/RAR_(file_format)) and [application/octet-stream](https://simple.wikipedia.org/wiki/RAR_(file_format))
*Source: Mozilla & Wikipedia


rtivital:
Please fix errors

PR #6747, opened 2024-08-31
fix: Unknown event handler property `onTopReached`. It will be ignored.

zrll12:
`onBottomReached` has been fixed in 7.12.2, but `onTopReached` has not been fixed until now.


rtivital:
Thanks!

PR #6750, opened 2024-09-01
Widen `description` type to include JSX nodes

chopfitzroy:
Is there any reason this can't be a JSX component?

Have tested locally and it appears to render fine but there may be some more wide reaching concerns.


rtivital:
Action description is used for search, search function cannot process all react nodes, only strings.

chopfitzroy:
Makes total sense, will close out.

PR #6754, opened 2024-09-02
Allow SegmentedControlProps to accept custom string literals as gener…

emipmttt:
Allow SegmentedControlProps to accept custom string literals as generic type for data, value, defaultValue and onChange prop.

- Updated `SegmentedControlProps` to support a customizable `onChange` handler that can accept specific string literal types.
- Introduced a generic type `T extends string = string` to constrain the value prop while allowing users to define their own set of valid string options.
- The change maintains backward compatibility by defaulting to the general `string` type if no specific literals are provided.


rtivital:
This will not work because SegmentedControl is not a generic component and there is no way to supply generic arguments to it.

emipmttt:
So even when the user is able to set what the options for the component, we are not able to type those by defualt? (we still can check types later from the usage of the component and not internally.)

rtivital:
No, types cannot be inferred this way.

PR #6780, opened 2024-09-05
Jsdoc annotations

blakenetz:
This is a POC for a quick way to add TS annotations to a component's functional declaration. This is intended to increase DX by providing a quick way to link back to the Mantine's docs. If you would like to move forward with this, there's a couple enhancements I would like to make:

* Make it applicable to other `@mantine` packages
* Add a check to make sure file doesn't already contain annotations
* Iterate through components instead of specifying as an argv

I pulled the `description` from the Frontmatter data. I'm not sure how valuable this is to a developer, so we can scratch that. 

The approach might be a little excessive because I'm reading the content of each file as utf-8 and then splitting and joining by linebreak. I figured since A. this is internal and B. it's a POC this small performance flaw wouldn't be too impactful. 

Here's the command I used to run the script:

```
npx tsx scripts/utils/generate-annotation.ts [COMPONENT NAME]
```


blakenetz:
Hi @rtivital, what do you think?

rtivital:
Thanks for submitting the PR, I've decided not to have such feature as it will increase maintanance efforts

PR #6782, opened 2024-09-05
Update use-hover.ts to react to node changes

max-degterev:
Currently if DOM node isn't immediately available or changes the hook doesn't react to that change. This fixes the issue.


rtivital:
Please fix errors

max-degterev:
Copied my local version without checking your linter, my bad. Updated

rtivital:
Thanks!

Issue #6764, opened 2024-09-03
Type checking on `ParseThemeColorOptions["color"]`


dylangrandmont:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.12.2

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

None

### Describe the bug

Currently `ParseThemeColorOptions["color"]` is typed as `unknown`. Calling the following code will not cause a typescript compile error
```ts
parseThemeColor({ color: undefined, theme });
```

However, it will result in the following runtime error

```
[@mantine/core] Failed to parse color. Expected color to be a string, instead got undefined
```

### If possible, include a link to a codesandbox with a minimal reproduction

https://codesandbox.io/p/sandbox/7xtwqt?file=%2Fsrc%2FApp.tsx%3A5%2C3-5%2C18

### Possible fix

Should `ParseThemeColorOptions` be updated to

```ts
interface ParseThemeColorOptions {
    color: string;
    theme: MantineTheme;
    colorScheme?: MantineColorScheme;
```

?

### Self-service

- [X] I would be willing to implement a fix for this issue

rtivital:
It is not planned to be changed, the type is designed to be flexible to simplify and reduce types of other components.

PR #6795, opened 2024-09-09
[@mantine/core] fix parse color option props

Kenzo-Wada:
fixes: https://github.com/mantinedev/mantine/issues/6764

Sorry if you had any reasons to make `color` unknown.


Kenzo-Wada:
ahh... this was actually hard to fix...

PR #6807, opened 2024-09-11
[mantine.dev] fix: dependency of `@mantine/tiptap` was wrong

Kenzo-Wada:
SEE: https://mantine.dev/getting-started/#get-started-without-framework

## describe BUG
when you check `@mantine/tiptap` , tabler-icons will be imported instead of `@tiptap/pm`


rtivital:
Thanks!

PR #6809, opened 2024-09-11
[mantine.dev] fix: remove autofocus from searchinput at Prop tab

Kenzo-Wada:
# Not Good UX

On SmartPhones, when you arrive Prop tab, keyboard shows and I thought it was not good at UX.

https://github.com/user-attachments/assets/acc1fc42-5bb7-4dcb-bd70-1eb843056c84


rtivital:
I'd rather keep it as is. 92% of traffic on the docs website is coming from desktops, and autofocus is the better option on desktop.

Kenzo-Wada:
no problems!

PR #6820, opened 2024-09-14
 [@mantine/chart] fix: typing on chart props

Jones-Griffin:
Make `[Xchart]Props` take a partial like line and bar chart do


rtivital:
Thanks!

PR #6826, opened 2024-09-16
[@mantine/chart] fix: exception when chart tooltip contains period

### LINKED ISSUE #6825: Exception when PieChart or DonutChart tooltip label contains a `.`
### Dependencies check up

7.12.2

@mantine/charts

Next.js

All

### Describe the bug

- When PieChart or DonutChart data has a `name` that contains a `.` it will throw the following exception:
- `TypeError: Cannot read properties of undefined` from `updateChartTooltipPayload` in `ChartTooltip`

repro:
```typescript
import { PieChart } from '@mantine/charts';

function Demo() {
  return (
    <PieChart
      data={[
        { name: 'USA.com', value: 400, color: 'blue' },
        { name: 'Other', value: 200, color: 'gray.6' },
      ]}
      withTooltip
      w={200}
      h={200}
    />
  );
}
```

I also tried using `label` with a `.` and a sanitized `name` on the data but the `ChartTooltip` doesn't use the `getSeriesLabels` util since there's no "series" passed in.

https://codesandbox.io/p/sandbox/busy-wind-6p7vh4

### Possible fix

_No response_

### Self-service

Jones-Griffin:
fixes #6825

### DIFF
--- a/packages/@mantine/charts/src/ChartTooltip/ChartTooltip.tsx
+++ b/packages/@mantine/charts/src/ChartTooltip/ChartTooltip.tsx
@@ -17,7 +17,7 @@ import classes from './ChartTooltip.module.css';
 
 function updateChartTooltipPayload(payload: Record<string, any>[]): Record<string, any>[] {
   return payload.map((item) => {
-    if(!item.payload || item.payload[item.name]) {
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

Issue #6827, opened 2024-09-16
Size prop is sometimes ignored in Title component


kimfucious:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.12.12

### What package has an issue?

@mantine/core

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

All

### Describe the bug

When using the Title component the [size prop](https://mantine.dev/core/title/#size) does not seem to be respected.

I'm specifically trying to use an H2 Header with a smaller font-size.

```ts
<Title order={2} size="h5">H2 with h5 font-size</Title>
```

I've listed the below as a demonstration, knowing that some of these would never be used.

I could very well be doing something wrong, or misunderstanding what the size prop does, but this seems pretty straight forward.

```ts
<Title order={1} size="h5">H1 with h5 font-size</Title>
<Title order={1}>H1 without size prop</Title>
<Title order={2} size="h5">H2 with h5 font-size</Title>
<Title order={2}>H2 without size prop</Title>
<Title order={3} size="h5">H3 with h5 font-size</Title>
<Title order={3}>H3 without size prop</Title>
<Title order={4} size="h5">H4 with h5 font-size</Title>
<Title order={4}>H4 without size prop</Title>
<Title order={5} size="h5">H5 with h5 font-size</Title>
<Title order={5}>H5 without size prop</Title>
<Title order={6}>H6 with h5 font-size</Title>
<br/>
<Title order={2} size="h1">H2 with h1 font-size</Title>
<Title order={2} size="h2">H2 with h2 font-size</Title>
<Title order={2} size="h3">H2 with h3 font-size</Title>
<Title order={2} size="h4">H2 with h4 font-size</Title>
<Title order={2} size="h5">H2 with h5 font-size</Title>
<Title order={2} size="h6">H2 with h6 font-size</Title>
<br/>
<Title order={2} size="xs">H2 with xs font-size</Title>
<Title order={2} size="sm">H2 with sm font-size</Title>
<Title order={2} size="md">H2 with md font-size</Title>
<Title order={2} size="lg">H2 with lg font-size</Title>
<Title order={2} size="xl">H2 with xl font-size</Title>
```

![image](https://github.com/user-attachments/assets/f5bce54e-9bd7-46a8-b5ac-de7f1f24d22c)

### Self-service

- [ ] I would be willing to implement a fix for this issue

[...]

PR #6833, opened 2024-09-17
[@mantine/core] chore: add sizes to `Title`

Kenzo-Wada:
Fixes https://github.com/mantinedev/mantine/issues/6827

# What
when MantineSize is used for `size` Props, --tile-fz will be like `xl` which is invalid. this PR will fix it.


rtivital:
Thanks!

PR #6842, opened 2024-09-17
[@mantine/spotlight] createSpotlight : Reset the selected index when using the toggle function

maclementED:
## Bug
When using `spotlight.toggle` instead of `spotlight.open`, using down arrow key will jump the first item. 

## How to reproduce
```
import { spotlight } from '@mantine/spotlight';

...

spotlight.open() // Will have selected:-1

spotlight.close()

spotlight.toggle() // Will have selected: 0 instead of -1. Using Down Arrow Key will skip the first item
```

## Fix
Set the selected to `-1` when you open the spotlight using `toggle`


rtivital:
Please fix errors

Issue #6835, opened 2024-09-17
Calendar component aria labeling wrong and props to fix it don't work


astolman:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.12.2

### What package has an issue?

@mantine/dates

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

The Calendar component's forward and back buttons do not have proper aria labeling. Whether or not you use the nextLabel and previousLabel props, lighthouse is reporting that there is no accessible name for the buttons. The docs https://mantine.dev/dates/calendar/?t=props say that these nextLabel and previousLabel props should set these aria labels, but they aren't working.

Below is a screenshot of the lighthouse report. I have also included a code sandbox which i was able to run lighthouse over and replicate the issue.
![image](https://github.com/user-attachments/assets/fdd6fc4b-009f-40eb-9a10-bef0f75a0e8d)

### If possible, include a link to a codesandbox with a minimal reproduction

https://codesandbox.io/p/sandbox/mantine-react-template-forked-vstc7z

### Self-service

- [ ] I would be willing to implement a fix for this issue

Kenzo-Wada:
You can use `ariaLabels` props to set arialabels!

https://mantine.dev/dates/calendar/?t=props

astolman:
> You can use `ariaLabels` props to set arialabels!
> 
> https://mantine.dev/dates/calendar/?t=props

I couldnt find any docs that described what type this prop takes. I guess I'll go take a look in the repo. Will update here with the results.

astolman:
The ariaLabels prop does work, thank you! I think the broken props should probably just be removed or fixed and the docs updated.

Kenzo-Wada:
@astolman 

YEAH, you are so right! I just now created a PR for the correction! :ok:

PR #6847, opened 2024-09-18
[@mantine/date] fix: add missing aria-label props to `Calender`

Kenzo-Wada:
fixes: https://github.com/mantinedev/mantine/issues/6835

Added `nextLabel` and `previousLabel` which was written on docs


rtivital:
Thanks!

PR #6848, opened 2024-09-18
[mantine.dev] make sure to install embla-carousel@7

Kenzo-Wada:
modified to specify v7 since the latest version of embla-carousel-react is now v8


rtivital:
Thanks!

PR #6863, opened 2024-09-21
[@mantine/core] fix: apply flex-wrap in breadcrumbs

### LINKED ISSUE #6862: UI Breaks in Breadcrumbs Component on Mobile and Similar Devices
### Dependencies check up

7.12.2

@mantine/core

Next.js

Chrome

### Describe the bug

The Breadcrumbs component causes UI issues on mobile and similar devices. When viewed on smaller screens, the layout breaks.

Sample:
Screen Shot 2024-09-21 at 16 11 16 PM

https://codesandbox.io/p/sandbox/mantine-react-template-forked-5ngrmy

### Possible fix

Adding `flex-wrap: wrap;` to the **root** style could resolve the issue.

### Self-service

ozantekin:
Fixes #6862 

This fix addresses the issue of the Breadcrumbs component breaking on mobile and similar devices. By adding `flex-wrap: wrap;` to the **root** style, the layout will now adjust correctly on smaller screens, preventing UI breakage.

Before: 
Screen Shot 2024-09-21 at 16 11 16 PM

After:
Screen Shot 2024-09-21 at 16 11 49 PM

### DIFF
--- a/packages/@mantine/core/src/components/Breadcrumbs/Breadcrumbs.module.css
+++ b/packages/@mantine/core/src/components/Breadcrumbs/Breadcrumbs.module.css
@@ -1,6 +1,7 @@
 .root {
   display: flex;
   align-items: center;
+  flex-wrap: wrap;
 }
 
 .breadcrumb {


rtivital:
Thanks!

PR #6871, opened 2024-09-23
[@mantine/core] feat: switch theme variant for buttons

yusufbkr:
```javascript
<Button
  variant={{
    light: 'filled',
    dark: 'outline',
  }}
>
  Button
</Button>
``` 

actually this is just an idea, this can be applied to all other components. i think it is a good idea as a user experience.

### DIFF
--- a/packages/@docs/demos/src/demos/core/Button/Button.demo.switchThemeVariant.tsx
+++ b/packages/@docs/demos/src/demos/core/Button/Button.demo.switchThemeVariant.tsx
@@ -1,0 +1,49 @@
+import { Button, createTheme, Group, MantineThemeProvider } from '@mantine/core';
+import { MantineDemo } from '@mantinex/demo';
+
+const code = `
+import { Group, Button, MantineProvider, createTheme } from '@mantine/core';
+
+function Demo() {
+  return (
+    <MantineThemeProvider theme={theme}>
+      <Group>
+        <Button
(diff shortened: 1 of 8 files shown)


rtivital:
This will not work with SSR, it will produce a hydration mismatch as color scheme value is not known on the server (it is initialized before first mount on client).

yusufbkr:
> This will not work with SSR, it will produce a hydration mismatch as color scheme value is not known on the server (it is initialized before first mount on client).

we just need to add “use client”; for nextjs. 
```typescript
"use client";
import {
  Button as MantineButton,
  ButtonProps,
  useMantineColorScheme,
  ButtonVariant,
} from "@mantine/core";

const Button = ({
  children,
  ...props
}: Omit<ButtonProps, "variant"> & {
  variant: {
    dark: ButtonVariant;
    light: ButtonVariant;
  };
}) => {
  const { colorScheme } = useMantineColorScheme();
  const variant =
    colorScheme !== "auto"
      ? props.variant?.[colorScheme] || "filled"
      : "filled";

  return (
    <MantineButton {...props} variant={variant}>
      {children}
    </MantineButton>
  );
};

export default Button;

``` 

it works like this

rtivital:
'use client' components are rendered on the server and will produce hydration mismatch if the user color scheme does not match the value defined on the server.

Issue #6870, opened 2024-09-23
PinInput component populates old values when using Backspace


Vurmevska:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

"^7.12.2"

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

Steps to reproduce:
- Create a component that uses PinInput
- Add controlled state using useState or mantine form
- Paste value that is previously copied
- Try deleting the numbers using backspace
- When the last value is deleted, the inputs are populated with the old values

```
import { PinInput } from "@mantine/core";
import { useState } from "react";

const TestExample = (): JSX.Element => {
  const [value, setValue] = useState<string | undefined>(undefined);

  return <PinInput value={value} onChange={(value) => setValue(value)} />;
};

export default TestExample;
```

![Peek 2024-09-23 14-27](https://github.com/user-attachments/assets/64e9a634-b354-4340-8ee4-380c372490fb)

### If possible, include a link to a codesandbox with a minimal reproduction

https://codesandbox.io/p/sandbox/mantine-react-template-forked-jqdxwr?file=%2Fsrc%2FApp.tsx%3A13%2C1

### Self-service

- [ ] I would be willing to implement a fix for this issue

PR #6880, opened 2024-09-25
[@mantine/core] fix: disable `Backspace` if index of input is 0

Kenzo-Wada:
fixes: https://github.com/mantinedev/mantine/issues/6870


rtivital:
Thanks!

PR #6884, opened 2024-09-25
[mantine.dev] fix demos of Charts

Kenzo-Wada:
I've fixed some demo codes and demo component of charts in the docs :+1:


rtivital:
Thanks!

PR #6896, opened 2024-09-27
Clarify that hooks also count as Mantine components

domino14:
Both hooks and components need to be inside the MantineProvider.


rtivital:
Thanks for submitting a PR! I consider the current wording good enough for understanding how MantineProvider works. This change may lead to more confusion.

PR #6903, opened 2024-09-27
[@mantine/chart] remove unused css from `PieChart`

Kenzo-Wada:
Removed unused css from PieChart. Probably copied and pasted from DonutChart was left in place.


rtivital:
Thanks!

PR #6904, opened 2024-09-27
[@mantine/dates] fix styles api description of level groups

Kenzo-Wada:
just as the title. some description of styles api was incorrect.


rtivital:
Thanks!

PR #6906, opened 2024-09-27
fix: typo for onKeyDown method in MenuItem

Meierschlumpf:
Noticed that typo when I wanted to create an advanced interaction.


rtivital:
Thanks!

Issue #6873, opened 2024-09-24
Deleting element in TagsInput with allowDuplicates=True deletes all duplicate elements 


ctdunc:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

^7.6.2

### What package has an issue?

@mantine/core

### What framework do you use?

Other, I will specify in the bug description

### In which browsers you can reproduce the issue?

All

### Describe the bug

I am using mantine through dash-mantine-components, but have confirmed that this issue also appears on the mantine website.

For example, given a tags input containing `items=['a','b','a']`, clicking delete on `items[2]` also deletes `items[0]`.

### If possible, include a link to a codesandbox with a minimal reproduction

https://mantine.dev/core/tags-input/#allow-duplicates

### Possible fix

Clicking delete on a tag should delete only that tag. 

Possibly here https://github.com/mantinedev/mantine/blob/00a94ceefc22c80ad08af73acb943f8f0c01d662/packages/%40mantine/core/src/components/TagsInput/TagsInput.tsx#L339

Maybe something like this? If duplicates aren't allowed, the behavior will be identical. If duplicates are allowed, it will only remove the desired element.
```
 onRemove={() => {
        let _ = _value.splice(index, 1)
        setValue(_value);
        onRemove?.(item);
      }}
```

### Self-service

- [X] I would be willing to implement a fix for this issue

ctdunc:
If this is expected behavior, ill close this issue. Otherwise happy to submit a fix, just let me know. 

Thanks!

rtivital:
You are welcome to submit a PR with a fix

ctdunc:
BTW -- fixed an issue here surrounding controlled components not triggering on `x` click https://github.com/mantinedev/mantine/pull/7105

PR #6907, opened 2024-09-27
Fix TagsInput duplicate deletion behavior

ctdunc:
fixes https://github.com/mantinedev/mantine/issues/6873

### DIFF
--- a/packages/@mantine/core/src/components/TagsInput/TagsInput.tsx
+++ b/packages/@mantine/core/src/components/TagsInput/TagsInput.tsx
@@ -336,7 +336,7 @@ export const TagsInput = factory<TagsInputFactory>((_props, ref) => {
       key={`${item}-${index}`}
       withRemoveButton={!readOnly}
       onRemove={() => {
-        setValue(_value.filter((i) => item !== i));
+        setValue(_value.splice(index, 1));
         onRemove?.(item);
       }}
       unstyled={unstyled}


rtivital:
Thanks!

packages/@mantine/core/src/components/TagsInput/TagsInput.tsx
```diff
@@ -336,7 +336,7 @@ export const TagsInput = factory<TagsInputFactory>((_props, ref) => {
       key={`${item}-${index}`}
       withRemoveButton={!readOnly}
       onRemove={() => {
-        setValue(_value.filter((i) => item !== i));
+        setValue(_value.splice(index, 1));
```
adelamarAtGalileo, line 339: Is this correct? I'm seeing that clicking X to remove a tag now removes all other tags except the target

ctdunc: think I should've used `.toSpliced` here @adelamarAtGalileo can you let me know if this resolves it? 

Seems alright to me, but clearly I missed it yesterday. Storybook must not have refreshed :shrug:.

adelamarAtGalileo: `splice` modifies the array and returns the removed value

Screenshot 2024-09-30 at 1 55 04 PM

`toSpliced` doesn't modify the array and returns the remaining values

Screenshot 2024-09-30 at 1 57 00 PM

So yes, `.toSpliced` should work here

max-degterev: nice, hopefully released soon, already got complaints from users :D

adelamarAtGalileo: @max-degterev You can downgrade to  `"@mantine/core": "^7.11.1"` you can avoid the newly introduced bug while waiting for a new release with the fix

max-degterev: maybe @rtivital pushes this out soon, I keep my hopes up :)

PR #6922, opened 2024-09-30
Fix/tagsinput duplicate deletion

ctdunc:
This had bad behavior in the last PR that I've fixed here. Apologies!


rtivital:
toSpliced method does not fit into browser support.
1. https://caniuse.com/mdn-javascript_builtins_array_tospliced
2. https://mantine.dev/about/#browser-support

rtivital:
Ha
Ve you validated that it works correctly this time is Storybook?

rtivital:
Thanks!

PR #6923, opened 2024-09-30
feat: allow useFullscreen to update the event listener when ref changes

linspw:
Sometimes we need to change the ref value with another element


rtivital:
Thanks!

Issue #6883, opened 2024-09-25
AppShell causes hydration errors when using Suspense


RabeeAbuBaker:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.13.0

### What package has an issue?

@mantine/core

### What framework do you use?

Remix

### In which browsers you can reproduce the issue?

All

### Describe the bug

Using AppShell with Suspense inside causes the following error:

```
Uncaught Error: This Suspense boundary received an update before it finished hydrating. This caused the boundary to switch to client rendering. The usual way to fix this is to wrap the original update in startTransition.
```

I saw there was a related issue ([#5979](https://github.com/mantinedev/mantine/issues/5979)) that was fixed earlier, but it's still happening in the latest version (7.13.0). 

![image](https://github.com/user-attachments/assets/5a1b59ad-803d-459d-b814-41b33bbca6cd)

### If possible, include a link to a codesandbox with a minimal reproduction

https://github.com/RabeeAbuBaker/remix-defer/tree/simple-defer-app-shell

### Self-service

- [ ] I would be willing to implement a fix for this issue

RabeeAbuBaker:
Is there anything we can do locally to overcome this? @rtivital 
This is preventing us from using defer which has a huge impact on user experience. 
Thanks!

rtivital:
I do not know, I've never used Suspense

RabeeAbuBaker:
> I do not know, I've never used Suspense

Since this was looked into and fixed before, is it possible to look into fixing this again? Or providing the functionality to disabled what's breaking it.

rtivital:
Okay, I'll leave this issue with "help wanted" label, you and anyone else are welcome to research it and submit a PR with a fix

RabeeAbuBaker:
@rtivital I've opened a PR that fixes the issue.
I've investigated the issue locally and the change resolved the error. I have added a screen recording for before and after the fix. Please review.

rtivital:
Great, I will review and merge change with the next patch release

Issue #5979, opened 2024-03-25
AppShell Suspense error


uglyDrusLee:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.6.2

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

AppShell has useResizing, which leads to rerendering, due to setState inside useIsomorphicEffect. Because of this behavior - when hydrating, react throws an error "Uncaught Error: This Suspense boundary received an update before it finished hydrating".

### Possible fix

Consider wrapping setState inside startTransition, or at least provide an option to disable useResizing. This could be particularly useful when transitionDuration is set to 0.

### Self-service

- [ ] I would be willing to implement a fix for this issue

rtivital:
Please provide a repository with a minimal reproduction.

uglyDrusLee:
https://github.com/uglyDrusLee/drusMantineAppShellIssue/tree/main/
here's minimal reproduction. to start, just "npm run start". server will be started on 3000 port.

rtivital:
The issue is not related to use-resizing. If the code is chaged to 

```tsx
import { startTransition, useRef, useState } from 'react';
import { useIsomorphicEffect, useWindowEvent } from '@mantine/hooks';

interface UseResizingInput {
  transitionDuration: number | undefined;
  disabled: boolean | undefined;
}

export function useResizing({ transitionDuration, disabled }: UseResizingInput) {
  const [resizing, setResizing] = useState(true);
  const resizingTimeout = useRef<number>();
  const disabledTimeout = useRef<number>();

  useWindowEvent('resize', () => {
    setResizing(true);
    clearTimeout(resizingTimeout.current);
    resizingTimeout.current = window.setTimeout(() => setResizing(false), 200);
  });

  useIsomorphicEffect(() => {
    startTransition(() => {
      setResizing(true);
      clearTimeout(disabledTimeout.current);
      disabledTimeout.current = window.setTimeout(
        () => setResizing(false),
        transitionDuration || 0
      );
    });
  }, [disabled, transitionDuration]);

  return resizing;
}
```

Or resizing removed entirely, it does not fix the error message. 

You are welcome to research the issue and propose the fix.

uglyDrusLee:
[...]

PR #6927, opened 2024-10-02
fix(hooks): resolve hydration issues by deferring state updates with startTransition in useResizing

RabeeAbuBaker:
Fixes [#6883](https://github.com/mantinedev/mantine/issues/6883) and https://github.com/mantinedev/mantine/issues/5979

Here's a screen recording before and after the fix:

https://github.com/user-attachments/assets/ae9a799d-e586-4a27-80f7-a7368e0b68f3


rtivital:
Thanks!

PR #6938, opened 2024-10-03
[mantine.dev] Unify NumberFormatter and NumberInput prefixSuffix demo

mbarzda:
Hi,
  I changed `<NumberFormatter />` demo to use the same `prefix` and `suffix` values as in `<NumberInput />` demo.


rtivital:
Why do you think that these changes should be made?

mbarzda:
> Why do you think that these changes should be made?

Usually numeric input and output formatting should match in an application scope. I noticed the difference in Mantine's documentation and wanted to suggest a common look.

rtivital:
Thanks for submitting the PR, but I'd rather keep the current demos

mbarzda:
**Z**.

PR #6947, opened 2024-10-07
[mantine.dev] Add missing configuration for `virtualColors`

dxanh97:
Without this property, `virtualColors` won't work


rtivital:
The change is not correct, demo works as intended as `bg="primary"` is set on the box. Setting virtual color as primary color is not required.

PR #6959, opened 2024-10-10
[@mantine/dates] fix timeInputProps defaultValue not working

### LINKED ISSUE #6958: DateTimePicker timeInputProps defaultValue does not get applied
### Dependencies check up

7.13.2

@mantine/dates

Next.js

All

### Describe the bug

Passing a defaultValue through the timeInputProps of DateTimePicker does not get applied to the TimeInput due to formatTime is setting the default state for value to an empty string.

```tsx
const formatTime = (dateValue: Date) => dateValue ? dayjs(dateValue).format(withSeconds ? 'HH:mm:ss' : 'HH:mm') : '';
const [timeValue, setTimeValue] = useState(formatTime(_value!));
```

https://codesandbox.io/s/sleepy-dijkstra-xkwzdn?file=/src/App.tsx

### Possible fix

setting the fallback default value to undefined instead of an empty string
```tsx
const formatTime = (dateValue: Date) => dateValue ? dayjs(dateValue).format(withSeconds ? 'HH:mm:ss' : 'HH:mm') : undefined;
```

### Self-service

Cuzart:
Fixes #6958 and is self-explanatory

### DIFF
--- a/packages/@mantine/dates/src/components/DateTimePicker/DateTimePicker.tsx
+++ b/packages/@mantine/dates/src/components/DateTimePicker/DateTimePicker.tsx
@@ -142,7 +142,7 @@ export const DateTimePicker = factory<DateTimePickerFactory>((_props, ref) => {
   });
 
   const formatTime = (dateValue: Date) =>
-    dateValue ? dayjs(dateValue).format(withSeconds ? 'HH:mm:ss' : 'HH:mm') : '';
+    dateValue ? dayjs(dateValue).format(withSeconds ? 'HH:mm:ss' : 'HH:mm') : undefined;
 
   const [timeValue, setTimeValue] = useState(formatTime(_value!));
   const [currentLevel, setCurrentLevel] = useState(level || defaultLevel || 'month');


rtivital:
This won't work, if the default value is set on the TimeInput and then not modified by the user, it will not be treated as part of the date object passed to `onChange` handler.

Cuzart:
Yes you are right I missed that point. Out of interest I tested an approach in the handleDateChange function if we check if the formattedTime is different from the timeInputProps?.defaultValue we get the expected behaviour. But this could add a bit of complexity depending on the format of the defaultValue and as you mentioned you are not planing to support this any time soon I will try to fix this outside of the component via JavaScript. Anyhow it could be misleading that the type of timeInputProps is supporting defaultValue then.

```tsx
const handleDateChange = (date: DateValue) => {
    if (date) {
      if (timeInputProps?.defaultValue && formatTime(date) !== timeInputProps?.defaultValue) {
        date.setHours(parseInt(timeInputProps?.defaultValue?.split(':')[0]));
        date.setMinutes(parseInt(timeInputProps?.defaultValue?.split(':')[1]));
      }
      setValue(assignTime(_value, date));
    }
    timeInputRef.current?.focus();
  };

```

rtivital:
Okay, I'll remove defaultValue and value props from this type

Issue #6652, opened 2024-08-11
PasswordInput component Issue on mobile


mnakhli:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

Latest

### What package has an issue?

@mantine/core

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

None

### Describe the bug

On mobile, when I toggle the visibility after entering the password, the `cursorPosition` moves to the beginning of the entered value. (Video)
-- This issue doesn't exist on the desktop and works correctly.
-- Check it in [Mantine documentation](https://mantine.dev/core/password-input/) on mobile.

https://github.com/user-attachments/assets/d887da17-c831-42dc-82c2-a6ca5f26580b

### Possible fix

The `cursorPosition` should always be at the end of the characters in every situation:

- When the input is clicked for the first time on mobile.
- When the toggle button for showing or hiding the password is pressed.
- And in any other scenario, the `cursorPosition` should stay at the end of the characters and not move.

### Self-service

- [ ] I would be willing to implement a fix for this issue

mmounirf:
I can reproduce this as well on a desktop browser using mobile emulation, so I'm a bit confident that it might be due to a "ghost click" effect. Looking at the code, the input handles `mousedown` and `keydown` but not `touchend`
 https://github.com/mantinedev/mantine/blob/e0721af0aafc7acf2c32aa5275633b218d2fc6af/packages/%40mantine/core/src/components/PasswordInput/PasswordInput.tsx#L158-L169
 
 I believe something like this could help.
 
```javascript
	onTouchEnd={(event) => {
		event.preventDefault();
		visibilityToggleButtonProps?.onTouchEnd?.(event);
		toggleVisibility();
	}}
``` 

but I will need to dig deeper just to be 100% sure about it.

PR #6971, opened 2024-10-12
[@mantine/core] Handle touchend event on password visibility toggle

mmounirf:
Fixes https://github.com/mantinedev/mantine/issues/6652 by handling `touchend` event for password visibility toggle to avoid caret jump.


rtivital:
Thanks!

Issue #6960, opened 2024-10-10
Table with sticky header is incompatible with `Table.ScrollContainer`


nabby27:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.13.2

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

I've recently been making a comparison table of features for different pricing plans. With this case and to be responsive I wanted both the header and the first column to be sticky.

For that I put the parameters `stickyHeader` and `stickyHeaderOffset={0}` to the table and wrapped the table with the `Table.ScrollContainer` component as indicated in the documentation. But the combination of both makes the `stickyHeader` stop working.

Here is a minimal stackblitz where you can test it: https://stackblitz.com/edit/mantine-bug-table-sticky-header-with-scrollcontainer?file=src%2FTableDemo.tsx

### If possible, include a link to a codesandbox with a minimal reproduction

https://stackblitz.com/edit/mantine-bug-table-sticky-header-with-scrollcontainer?file=src%2FTableDemo.tsx

### Possible fix

I've been investigating and it seems that the `sticky` header doesn't get along well with having overflows above it. Ideally, this result could be achieved: https://codepen.io/chriscoyier/pen/yLVNErX

### Self-service

- [ ] I would be willing to implement a fix for this issue

RohanMishra315:
@nabby27 I'll do this

nabby27:
Hi @RohanMishra315, Thanks for the stackblitz, I'm sure it can be helpful to the Mantine team. I want this to be integrated into Mantine and have the team support it. Also, the horizontal scrolling in your example is for the entire page, not just the table content.

nicojav:
@nabby27 added a fix with a demo in the [PR](https://github.com/mantinedev/mantine/pull/6976)

nabby27:
Hi @nicojav! Thanks for your PR and the stackblitz to see how it works, it's very helpful.

Personally I think the solution doesn't seem the most appropriate, it seems more like a patch than a solution in itself. Why does the scroll container have to have a height of 95vh and not 90vh or 99vh? It seems random.

I would expect the solution to be in the table and not in the scroll container as this can be used in cases where that 95vh breaks with what is expected.

Anyway, I will wait for the Mantine team to decide if the solution is valid or not.

nicojav:
[...]

PR #6976, opened 2024-10-14
[@mantine/core] Added height to scrollContainer

nicojav:
Fix for the sticky header bug on the issue #6960.

- Bug: [Stackblitz-link](https://stackblitz.com/edit/mantine-bug-table-sticky-header-with-scrollcontainer?file=src%2FTableDemo.tsx)
- Fix: [Stackblitz-link](https://stackblitz.com/edit/mantine-bug-table-sticky-header-with-scrollcontai-u3wnif?file=vite.config.ts,package.json,src%2FApp.tsx,src%2Fdata.ts,src%2FTableDemo.tsx)

### DIFF
--- a/packages/@mantine/core/src/components/Table/Table.module.css
+++ b/packages/@mantine/core/src/components/Table/Table.module.css
@@ -100,6 +100,7 @@
 
 .scrollContainer {
   overflow-x: var(--table-overflow);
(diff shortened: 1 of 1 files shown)


rtivital:
That is not a valid fix, the scroll contained must not have a fixed height.

PR #6977, opened 2024-10-14
[@mantine/dropzone] Add image HEIF to IMAGE_MIME_TYPE in @mantine/dropzone

thaynam:
Fixes https://github.com/mantinedev/mantine/issues/5866.

In addition to HEIC, we also need to add HEIF as a MIME type, otherwise some files may not function properly.


rtivital:
Thanks!

PR #6985, opened 2024-10-16
Fix condition for isValidNumber in NumberInput

### LINKED ISSUE #6969: NumberInput with minimum broke since 7.13.2
### Dependencies check up

7.13.2

@mantine/core

Vite

Firefox

### Describe the bug

Our integration tests fail once we update to the current mantine version (from the last patch release 7.13.1). I looked into this and broke down the issue to a combination of min/defaultValue/value and the onChange handler. I pasted a reproduction link below.

How to reproduce:
- Focus the NumberInput
- Remove the 3
- Enter a minus sign
- Instead of the minus sign the 3 reappears

In the previous version the "-" stayed so that one could enter a negative number. I suspect this is due to b166592af44b6b064b5a914d88fe92a0bc29d1f0 which is the only change to NumberInput in this release. I'm not entirely sure about it but I guess due to the change to `isValidNumber` the value "-" is maybe considered as valid number?
```js
  if (typeof value === 'string') {
    return getDecimalPlaces(value) < 14 && value !== '';
  }
```

https://codesandbox.io/p/sandbox/mantine-react-template-forked-8vnf2m

### Possible fix

If I identified the problem correctly I think we could fix this by adding another condition (like `!Number.isNaN(Number(value))) `) to the "string" case or by merging both conditions somehow.

### Self-service

richardboehme:
Previously values that only contain a minus would have been considered valid which triggers things like min/max clamping even though it should not.

According to the types `value` should always be a string and never undefined which made it easier to collapse the conditions into one. This fixed my reproduction script that was posted in the issue.

Fixes #6969

### DIFF
--- a/packages/@mantine/core/src/components/NumberInput/NumberInput.tsx
+++ b/packages/@mantine/core/src/components/NumberInput/NumberInput.tsx
@@ -38,18 +38,14 @@ function getDecimalPlaces(inputValue: string | number): number {
   return inputValue.toString().replace('.', '').length;
 }
 
-function isValidNumber(
-  floatValue: number | undefined,
-  value: string | undefined
-): floatValue is number {
-  if (typeof value === 'string') {
-    return getDecimalPlaces(value) < 14 && value !== '';
-  }
-
+function isValidNumber(floatValue: number | undefined, value: string): floatValue is number {
   return (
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

Issue #6991, opened 2024-10-17
withBarValueLabel not displaying in correct positions in BarChart


namakshenas:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.13.2

### What package has an issue?

@mantine/charts

### What framework do you use?

Other, I will specify in the bug description

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

Hi,

I'm trying to use the `withBarValueLabel` property of BarChart. The value labels are not placed correctly. Here is the snapshot of the issue:

<img width="1151" alt="Screenshot 2024-10-17 at 16 08 52" src="https://github.com/user-attachments/assets/9a66280c-c8b2-494f-875f-e0b48151cc44">

Here is the working example in `@mantine/charts`:

```
<BarChart
          h={300}
          data={[
            { dmu: 'dmu1', 'Efficiency score': 100, Status: 'Efficient', color: '#32CD32' },
            { dmu: 'dmu2', 'Efficiency score': 78, Status: 'InEfficient', color: '#FFA500' },
            { dmu: 'dmu3', 'Efficiency score': 100, Status: 'Efficient', color: '#32CD32' },
            { dmu: 'dmu4', 'Efficiency score': 100, Status: 'Efficient', color: '#32CD32' },
            { dmu: 'dmu5', 'Efficiency score': 89, Status: 'InEfficient', color: '#FFA500' },
            { dmu: 'dmu6', 'Efficiency score': 95, Status: 'InEfficient', color: '#FFA500' },
          ]}
          withBarValueLabel
          orientation="vertical"
          gridAxis="y"
          barProps={{ radius: 5 }}
          dataKey="dmu"
          series={[{ name: 'Efficiency score' }]}
          tickLine="none"
/>
```

### Self-service

- [ ] I would be willing to implement a fix for this issue

fsd-niraj:
I'd like to work on this. Please assign if not picked up.

rtivital:
You are welcome to submit a PR with a fix

fsd-niraj:
![image](https://github.com/user-attachments/assets/852291df-7f2e-411c-8233-843c9c96a48f)

Please let me know if this looks good? I'll quickly raise my PR.

namakshenas:
> ![image](https://github.com/user-attachments/assets/852291df-7f2e-411c-8233-843c9c96a48f)
> 
> 
> 
[...]

PR #7011, opened 2024-10-22
withBarValueLabel not displaying in correct positions in BarChart - Issue #6991

fsd-niraj:
Added a function that determines "dy" and "dx" parameters of the label depending on height and width of the Bar Chart.

### DIFF
--- a/packages/@mantine/charts/src/BarChart/BarChart.tsx
+++ b/packages/@mantine/charts/src/BarChart/BarChart.tsx
@@ -55,9 +55,9 @@ export type BarChartCssVariables = {
 
 export interface BarChartProps
   extends BoxProps,
-    GridChartBaseProps,
-    StylesApiProps<BarChartFactory>,
-    ElementProps<'div'> {
+  GridChartBaseProps,
+  StylesApiProps<BarChartFactory>,
+  ElementProps<'div'> {
   /** Data used to display chart. */
   data: Record<string, any>[];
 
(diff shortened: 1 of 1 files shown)


Kenzo-Wada:
CI is failing because of prettier. you should run `npm run prettier:write`

fsd-niraj:
Yes, I ran that command already.

packages/@mantine/charts/src/BarChart/BarChart.tsx
```diff
@@ -130,10 +130,23 @@ export function BarLabel({
...
+  function labelCoordinates(): { dx: number, dy: number } {
+    let coord: { dx: number, dy: number } = { dx: 0, dy: 0 }
+    if (others.props.h < 300) {
+      coord = { ...coord, dx: Math.ceil(others?.width + 10) }
+    } else {
+      coord.dx = Math.ceil(others?.width - others?.viewBox?.x / 2)
+    }
+    coord = { ...coord, dy: Math.ceil(others.props.h / 15) }
+    return coord;
+  }
```
Kenzo-Wada, line 143: IMO: I think this way is much simple, WDYT? I prefer not using `let` in this case.

```suggestion
function labelCoordinates(): { dx: number, dy: number } {
  return {
    dx: others.props.h < 300
      ? Math.ceil(others?.width + 10)
      : Math.ceil(others?.width - others?.viewBox?.x / 2),
    dy: Math.ceil(others.props.h / 15)
  };
}
```

fsd-niraj: That looks good too.
I tried not to use ternary operators just in case we need to add more conditions in future.

Kenzo-Wada:
@fsd-niraj 

you're sure?? see this!
https://github.com/fsd-niraj/mantine/pull/1/files

packages/@mantine/charts/src/BarChart/BarChart.tsx
Kenzo-Wada: Since the code is not complicated yet, I think we should consider it only when additional conditions arise!

Issue #7006, opened 2024-10-21
Datepickerinput should close when disabled


MathRobin:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.13.3

### What package has an issue?

@mantine/dates

### What framework do you use?

Other, I will specify in the bug description

### In which browsers you can reproduce the issue?

All

### Describe the bug

I can reproduce this on Chrome and Firefox on Ubuntu, not able to try on Windows/Mac or with others browsers but think it's not related.

When a DatePickerInput popover is opened, if you disabled it without selecting any date or without closing it manually. The popover stays opened and continue to allow selecting of date.

Better is to show you a sample (sorry, text in the UI is in french). In this video I enabled the datepickerinput, see popover open at click. Juste after i selected another option by radio. And my UI needs to disable the datepicker (choosing a date relative option). But my popover stayed opened and I could change the date. Not desired behaviour.

[Capture vidéo du 2024-10-21 15-09-39.webm](https://github.com/user-attachments/assets/adccf720-336c-4fdf-bf7d-54a2aea2b8de)

### If possible, include a link to a codesandbox with a minimal reproduction

You can reproduce directly on the docs by disabling the picker when popover is opened
https://mantine.dev/dates/date-time-picker/#input-props

### Possible fix

When switching to disabled, popover should close. I'm not able to fix it alone, but what i was able to understand, the datepickerinput is based on a generic pickerinputbase in matine/dates package. Which can mean that the other pickers in popover, at least, have the same behaviour/bug.

### Self-service

- [ ] I would be willing to implement a fix for this issue

Many :heart: on you for this great project !

MathRobin:
Many thanks ! Have a nice day !

PR #7017, opened 2024-10-23
[@mantine/dates] fix: close DatePickerInput when disabled

Kenzo-Wada:
fixes: https://github.com/mantinedev/mantine/issues/7006


rtivital:
Thanks!

PR #7020, opened 2024-10-24
Issue #7010 key prop added to area chart

### LINKED ISSUE #7010: Issue with CompositeChart
### Dependencies check up

7.13.3

@mantine/charts

Vite

Chrome

### Describe the bug

I use CompositeChart with the following code:

```tsx
<CompositeChart
  h={185}
  data={chartData}
  dataKey="date"
  maxBarWidth={30}
  withRightYAxis
  series={[
    { name: 'views', label: 'Số lượt xem', color: 'yellow.8', type: 'area' },
    {
      name: 'edits',
      label: 'Số sửa đổi',
      color: 'rgba(18, 120, 255, 0.2)',
      yAxisId: 'right',
      type: 'bar',
    },
  ]}
  curveType="monotone"
  gridAxis="none"
/>
```

But the Chrome console shows this error:
Screenshot 2024-10-22 at 16 53 47

I think we should add `key` prop to this line: https://github.com/mantinedev/mantine/blob/6a067f1c90159b6c4b89586df3bcc98aea1aeecd/packages/%40mantine/charts/src/CompositeChart/CompositeChart.tsx#L289

Regards,
Plantaest.

_No response_

### Possible fix

_No response_

### Self-service

mariakax3:
Fixes #7010
The unique key property has been added to the area type of the composite chart to resolve the console error.

### DIFF
--- a/packages/@mantine/charts/src/CompositeChart/CompositeChart.tsx
+++ b/packages/@mantine/charts/src/CompositeChart/CompositeChart.tsx
@@ -286,6 +286,7 @@ export const CompositeChart = factory<CompositeChartFactory>((_props, ref) => {
       return (
         <Area
           {...getStyles('area')}
+          key={item.name}
           name={item.name}
           type={curveType}
           dataKey={item.name}


rtivital:
Thanks!

PR #7021, opened 2024-10-24
Fix number input NaN issue

### LINKED ISSUE #7005: Number input min/max can be bypassed by entering "0" first
### Dependencies check up

7.13.2

@mantine/core

Vite

Chrome

### Describe the bug

Go to a NumberInput and set a max of 40. Type in 072, then focus away. Notice how it sets the value to 72, which shouldn't be.

https://mantine.dev/core/number-input/
image

### Self-service

owenmoogk:
Closes #7005


rtivital:
Please fix errors

rtivital:
Thanks!

Issue #5338, opened 2023-11-27
Form hook causes infinite re-render when using as required effect dependency


lukasbash:
### What package has an issue?

@mantine/form

### Describe the bug

In the example of your docs (https://mantine.dev/form/recipes/#set-initial-values-with-async-request) you use the `useEffect` as a mount-only cycle.
However, considering that I can have the form mounted and the actual values can change via props (or any other hook), I might want to extend the `useEffect` hook with other properties that change.

Consider I want to change the values with an effect like this:

```jsx
useEffect(() => {
  const values = {
    a: prop_a,
    b: prop_b,
  };

  form.setValues(values);
  form.resetDirty(values);
}, [prop_a, prop_b]);
// React Hook useEffect has a missing dependency: 'form'. Either include it or remove the dependency array.
```

With the example above, eslint complains about the `form` object not being a dependency (which is technically correct, isn't it?). Now if I add the form object as a dependency (regardless of either using `form` directly or destructuring `setValues` and `resetDirty`), I get an infinite render cycle:

<img width="317" alt="image" src="https://github.com/mantinedev/mantine/assets/15958032/6b165227-f7da-4759-9d8e-4593ffe9637f">

How do you explain the correct approach, re-rendering on single prop changes and update the values accordingly while still being linting-compatible? Our project follows pretty strict linting rules, so exclusions are not really possible here. Also I assume re-setting initial values as a "one-time-re-render" is not an option. Further I guess that the form relies on all components being controlled and therefore always re-renders on a change, whereas a `setValue` call is a reason here.

Please give me some input on the technical correctness, and a working, linted code 🙏 

### What version of @mantine/* packages do you have in package.json? (Note that all @mantine/* packages must have the same version in order to work correctly)

7.1.5

### Do you know how to fix the issue

No

### Do you want to send a pull request with a fix?

No

rtivital:
Do not add `form` as a dependency in useEffect – there is no reason to do so. If you really need this feature, you can wrap `form` in useMemo, but it does not really make any sense.

lukasbash:
> Do not add `form` as a dependency in useEffect – there is no reason to do so. If you really need this feature, you can wrap `form` in useMemo, but it does not really make any sense.

While this the most obvious option, I am still confused why it does not make sense. As far as I understand the official documentation, every reactive value needs to be a declared dependency.

But `useMemo` is worth a try, thanks a lot!

twiddler:
[...]

PR #7029, opened 2024-10-24
Wrap resetDirty in a useCallback

anthony-dandrea:
I am using `useForm` and had the result in a dependency array. I noticed that it caused needless executions. I ended up destructuring the result of `useForm` and adding those items to the dependency array. I noticed it worked except for `resetDirty`. It appears it's the only action not wrapped in a `useCallback`. I don't see any reason why it couldn't be so I cut  a quick PR.

Semi-related issue: https://github.com/mantinedev/mantine/issues/5338


mr-pinzhang:
ping @rtivital, that'd be great if you could take a look at this issue.

rtivital:
Thanks!

PR #7031, opened 2024-10-25
[@mantine/core] Notification: Add closeButton ReactNode

aalirezamoradii:


rtivital:
Thanks for submitting a PR. This change will not work correctly. If the close button is replaced with the custom item, it will not close notification out of the box as it has no access to the onClose handler. If you want to change the icon, use `closeButtonProps` instead.

PR #7038, opened 2024-10-25
Docs - Tabs - added @tanstack/react-router example

jacobra19:
* added the most basic and naive example of how to use Tabs component with `@tanstack/react-router`
image


rtivital:
Thanks for submitting a PR! I'd rather not add this – the implementation is identical to react-router, there should not be any difficulties for anyone to figure this part out on their own.

PR #7046, opened 2024-10-28
[@mantine/charts] fix: Object literal may only specify known properties, and label does not exist in type RadarChartSeries

zaphire12:
add new properties in RadarChartSeries


rtivital:
Thanks!

PR #7062, opened 2024-11-01
Export type FormValidateInput from "@mantine/form"

### LINKED ISSUE #7061: Unable to import FormValidateInput from "@mantine/form/lib/types";
### Dependencies check up

^7.13.4

@mantine/form

Vite

All

### Describe the bug

I would like to be able to use the type `FormValidateInput` from `@mantine/form/lib/types` for my own custom hook.
However, I cannot import it, as it is not exported from the`@mantine/form` library.

```tsx
import { FormValidateInput } from "@mantine/form/lib/types";
```

Error:
`TS2307: Cannot find module @mantine/ form/ lib/ types or its corresponding type declarations`

_No response_

### Possible fix

Export the `FormValidateInput` type from the `@mantine/form` package.

### Self-service

pfo-omicsstudio:
Fixes #7061.


rtivital:
Thanks!

Issue #7055, opened 2024-10-31
Cannot Override "tabIndex" props on Tabs.Tab buttons


dev-banana:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.13.4

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

I'm facing an issue when trying to disable focus on the Tabs.Tab buttons by manually setting the tabIndex property to "-1" 
The active tab is still rendered with `tabindex="0"` and others with `tabindex="-1"`.

Here is a minimal code example to reproduce the issue:
```typescript
<Tabs defaultValue="gallery">
  <Tabs.List>
    <Tabs.Tab value="gallery" tabIndex={-1}>Gallery</Tabs.Tab>
    <Tabs.Tab value="messages" tabIndex={-1}>Messages</Tabs.Tab>
  </Tabs.List>

  <Tabs.Panel value="gallery">Gallery tab content</Tabs.Panel>
  <Tabs.Panel value="messages">Messages tab content</Tabs.Panel>
</Tabs>
```

### Possible fix

```typescript
export const TabsTab = factory<TabsTabFactory>((_props, ref) => {
  const props = useProps('TabsTab', defaultProps, _props);
  const {
    //...
      
    tabIndex, // <-- get "tabIndex" given by props
      
    ...others
  } = props;

  // ...
    
  return (
    <UnstyledButton
      {...others}
      // ...
      
      // If defined, use the given "tabIndex" props
      tabIndex={tabIndex !== undefined ? tabIndex : (active || ctx.value === null ? 0 : -1)}
      // ...
    >
       // ... 
    </UnstyledButton>
  );
});
```

### Self-service

- [ ] I would be willing to implement a fix for this issue

PR #7081, opened 2024-11-05
[@mantine/core] fix: enable overrriding `tabIndex`

Kenzo-Wada:
fixes: https://github.com/mantinedev/mantine/issues/7055


rtivital:
Thanks!

PR #7083, opened 2024-11-05
[mantine.dev] add example code of Nested Collapse Demo

Kenzo-Wada:
Added example code to https://mantine.dev/core/collapse/#nested-collapse-components


rtivital:
Thanks for submitting the PR, I'd rather keep this as demo only without code, it only demonstrates that nested collapses work correctly.

PR #7097, opened 2024-11-08
Narrow type of `NotificationData`

### LINKED ISSUE #5640: `showNotification` doesn't support data-* attributes
### Dependencies check up

7.4.2

@mantine/notifications

Vite

### Describe the bug

When passing data-* attributes to `showNotification` or `notifications.show` functions a TS error is thrown:
```
Argument of type '{ message: string; "data-<attribute>": string; }' is not assignable to parameter of type 'NotificationData'.
  Object literal may only specify known properties, and '"data-<attribute>"' does not exist in type 'NotificationData'.
```

Example:
data attribute used in show notification caused TS error

See link below for example code.

https://codesandbox.io/p/sandbox/focused-moon-m85ccq

### Possible fix

_No response_

### Self-service

twiddler:
e78a9bb9d23654349cf9413674dce08051322762 fixed #5640, but made the `NotificationData` type [overly broad](https://github.com/mantinedev/mantine/issues/5640#issuecomment-1936799846).

This change limits the additional keys to those that start with `data-`. This allows passing `data-` attributes _and_ narrowing the type.

### DIFF
--- a/packages/@mantine/notifications/src/notifications.store.ts
+++ b/packages/@mantine/notifications/src/notifications.store.ts
@@ -10,7 +10,9 @@ export type NotificationPosition =
   | 'bottom-right'
   | 'bottom-center';
 
-export interface NotificationData extends Omit<NotificationProps, 'onClose'>, Record<`data-${string}`, any> {
+export interface NotificationData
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #7103, opened 2024-11-08
Add transitionDuration prop to RingProgress

yorkeJohn:
Support the `transitionDuration` prop on `RingProgress` (similar to `SemiCircleProgress`).

Note: also added return types in the following files in order to fix the TypeScript Compiler error [The inferred type of "X" cannot be named without a reference to "Y". This is likely not portable. A type annotation is necessary.](https://github.com/microsoft/TypeScript/issues/42873)
`@mantine-tests/core/src/render.tsx`
`@mantine/core/src/components/Popover/use-popover.ts`
`@mantine/core/src/components/Tooltip/use-tooltip.ts`

Docs have also been updated.


rtivital:
Please keep changes that are only related to RingProgress, other components worked fine as is without modifications

yorkeJohn:
Type annotation changes were required in order to compile locally. These changes should be made regardless, but I can move them to another PR if you’d prefer?

rtivital:
I do not have any issue with local compilation, please remove these changes from the PR, if tests pass then these changes should not be made.

yorkeJohn:
OK, those changes have been reverted.

rtivital:
Thanks!

PR #7105, opened 2024-11-09
Change TagsInput onRemove to copy correctly

ctdunc:
Alright so my last commit introduced a new bug where controlled `TagsInput`s would not have any callbacks triggered when clicking the `x` button in a pill.

This commit fixes that issue. I tested it in dash-mantine-components as well, and the behavior is now correct.

### DIFF
--- a/packages/@mantine/core/src/components/TagsInput/TagsInput.tsx
+++ b/packages/@mantine/core/src/components/TagsInput/TagsInput.tsx
@@ -256,7 +256,6 @@ export const TagsInput = factory<TagsInputFactory>((_props, ref) => {
     if ((!isDuplicate || (isDuplicate && allowDuplicates)) && _value.length < maxTags!) {
       onOptionSubmit?.(val);
       setSearchValue('');
-
       if (val.length > 0) {
         setValue([..._value, val]);
       }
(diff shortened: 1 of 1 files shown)


omerchn:
We also encountered this issue. thanks for fixing it.
wouldn't this be more performant? (using only `filter` instead of `splice` and then `splice`)
```ts
setValue(_value.filter(_item => _item !== item));
```

ctdunc:
> We also encountered this issue. thanks for fixing it.
> wouldn't this be more performant? (using only `filter` instead of `splice` and then `splice`)
> ```ts
> setValue(_value.filter(_item => _item !== item));
> ```

No, as this does not work when `allowDuplicates` is enabled. 

The correct solution here is probably `toSpliced`, but hasnt caught on in terms of browser compatibility

omerchn:
> No, as this does not work when `allowDuplicates` is enabled.
> 
> The correct solution here is probably `toSpliced`, but hasnt caught on in terms of browser compatibility

Got you. What about:
```ts
setValue(_value.filter((_, _index) => _index !== index));
```
Using `splice` is prone to cause bugs when used in libraries as it mutates the original array, which is something that consumers of libraries don't expect.
I get that using `slice()` to create a copy of the array fixes this bug, but it seems more like a hack rather than a correct solutaion.

rtivital:
Thanks!

PR #7114, opened 2024-11-11
[@mantine/core] Fix combobox aria-controls attribute

### LINKED ISSUE #7085: Accessibility problem in ComboboxTarget component - @mantine/core
### Dependencies check up

7.11.0

@mantine/core

Vite

### Describe the bug

Path: src/mantine-core/src/components/ComboBox/use-combobox-target-props/use-combobox-target-props.ts

The combobox Target component has a attribute called `aria-controls`. This attribute has the id ( ` 'aria-controls': ctx.store.listId`) of the dropdown that it controls. 
The problem is when the dropdown is closed the `aria-controls` attribute is still existing on the combobox Target while the dropdown is not even visible in the DOM, so it points to nothing.
I get an accessibility warning in storybook: 

> Unable to determine if aria-controls referenced ID exists on the page while using aria-haspopup: aria-controls="mantine-hbj4u6pg6"

_No response_

### Possible fix

A solution is to check if the dropdown is open, if it's true, add "aria-controls" attribute, else, set as undefined:
```
'aria-controls': ctx.store.dropdownOpened
          ? ctx.store.listId
          : undefined,
```

### Self-service

Samy0412:
Resolves #7085

- Checks if the dropdown is open, if it's true, add "aria-controls" attribute, else, set as undefined.

### DIFF
--- a/packages/@mantine/core/src/components/Combobox/use-combobox-target-props/use-combobox-target-props.ts
+++ b/packages/@mantine/core/src/components/Combobox/use-combobox-target-props/use-combobox-target-props.ts
@@ -94,7 +94,7 @@ export function useComboboxTargetProps({
         'aria-haspopup': 'listbox',
         'aria-expanded':
           (withExpandedAttribute && !!(ctx.store.listId && ctx.store.dropdownOpened)) || undefined,
-        'aria-controls': ctx.store.listId,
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #7144, opened 2024-11-20
[@mantine/core] Add support for custom variant in PasswordInput visibility toggle button

pfo-omicsstudio:
The variant prop for the visibility toggle button is now taken from `visibilityToggleButtonProps` if provided.

Defaults to "subtle" when no variant is specified in `visibilityToggleButtonProps`.


rtivital:
Thanks!

PR #7147, opened 2024-11-20
[Bugfix] Gracefully handle environments without `window.matchMedia`

lsanwick:
I've been struggling with using Mantine within a test environment, and it seems like the dependency on `window.matchMedia` isn't really required in a test environment.  This change simply guards against the case where `window` and `window.matchMedia` are not defined, and avoids trying to use it in that case.

I've looked at doing mocking based on the docs [here](https://mantine.dev/guides/vitest/#configuration), but this creates conflicts with other libraries, like `framer-motion` that assume if `window.matchMedia` exists, that they are not in a test environment.


rtivital:
Thanks!

PR #7153, opened 2024-11-21
[DonutChart] Add render label function

me-gusta-v:
Use a similar label render function as the Pie Chart for Donut Chart in order to apply the _valueformatter_ to the label.

Also adds the _labelsType_: value or percent, similar to Pie Chart.

Omitted the inside label option since it doesn't make sense on the Donut Chart in my opinion.


rtivital:
Thanks!

PR #7178, opened 2024-11-27
[refactor] Replace global JSX with React.JSX

yesmeck:
The global JSX namespace has been removed in react 19. 

See https://react.dev/blog/2024/04/25/react-19-upgrade-guide#the-jsx-namespace-in-typescript


rtivital:
Thanks!

Issue #7176, opened 2024-11-27
Option validation in ComboBox is too strict


ustun-ed:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

Latest

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

All

### Describe the bug

If the Combobox has duplicate options, Mantine fails instead of showing a warning, unlike React. 

This might result in unexpected failures if the data coming from the backend has duplicate data all of a sudden and the frontend code was not written defensively to make the incoming data unique.

For example, if the Combobox appears in a Modal, the Modal crashes and is not shown at all.

It would be hard to make sure every input to Combobox is unique in all cases, so I think the validation should be relaxed and applied on-demand.

### Self-service

- [X] I would be willing to implement a fix for this issue

rtivital:
It is not planned to change this behavior. If options are not unique, Combobox component will not work correctly, all of its features are bound to option value, including keyboard navigation, option retrieving from the list and scrolling to a specific option. The function is implemented to throw an error to explicitly notify the developer that the data cannot be processed. 

To fix the issue, transform the data on your side before passing it to the component.

ustun-ed:
@rtivital OK, I see. I still think it is a bit too punishing as users might not be aware of this restriction.

In any case, I opened this PR that at least provides some warning in the docs about the uniqueness requirement: https://github.com/mantinedev/mantine/pull/7181

There might be a better place to communicate this in the docs as most may not look at the `props` tab of these components. Maybe we put that in the first paragraph of each component?

PR #7181, opened 2024-11-28
[@mantine/core] Update docs for Select, MultiSelect, Autocomplete to warn about data props consisting of unique values

ustun-ed:
Since we decided not to change the validation requirement for the options, we should add some docs about this constraint.

See https://github.com/mantinedev/mantine/issues/7176 for previous discussion


rtivital:
Thanks!

PR #7182, opened 2024-11-28
Fix typos in Tree code example

mariansimecek:
I was experimenting with tree examples and found that they don't work as shown in the demo. The values in `dataCode` are not unique.


rtivital:
Thanks!

PR #7189, opened 2024-11-29
[@mantine/modals] Expose Modals Manager props

pfo-omicsstudio:
I had a use case in my application where I needed to read from the Modals manager props, so I created this PR to allow them to be exposed to the user.

EDIT: I also discovered an error in my implementation of `updateContextModal`, which would throw the following error:
```hook.js:608 Warning: React does not recognize the `modalId` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `modalid` instead. If you accidentally passed it from a parent component, remove it from the DOM element. Error Component Stack```

This occurred because the modalId was unintentionally included in the props passed to the modal component. I have resolved this in the PR by ensuring that modalId is excluded from the props spread onto the modal.

### DIFF
--- a/packages/@mantine/modals/src/ModalsProvider.tsx
+++ b/packages/@mantine/modals/src/ModalsProvider.tsx
@@ -148,8 +148,8 @@ export function ModalsProvider({ children, modalProps, labels, modals }: ModalsP
   );
 
   const updateContextModal = useCallback(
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #7210, opened 2024-12-03
Donut chart legend modification

mariakax3:
Resolves https://github.com/orgs/mantinedev/discussions/6987
Added legendMode and legendOrientation properties. The legendMode allows to control how the legend behaves. When set to 'side', the legend is displayed on the side of the chart, and hovering over a row in the legend highlights the corresponding segment in the chart. When set to 'hover', the legend is hidden by default and only appears on hover. 
The legendOrientation property enables the customization of the legend's position.

### DIFF
--- a/packages/@mantine/charts/src/DonutChart/DonutChart.story.tsx
+++ b/packages/@mantine/charts/src/DonutChart/DonutChart.story.tsx
@@ -12,7 +12,7 @@ const data = [
 export function Usage() {
   return (
     <div style={{ padding: 40 }}>
-      <DonutChart data={data} strokeWidth={1} strokeColor="red" />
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks for the PR and sorry that it took so long to review! DonutChart now supports legend out of the box – set `withLegend` prop to display it, hovering legend items highlights corresponding segments. Legend position and orientation can be changed with `legendProps`, for example `legendProps={{ verticalAlign: 'middle', align: 'right', layout: 'vertical' }}`, which covers both side legend and orientation use cases from this PR.

This PR also changes defaults (`size`, tooltip is hidden unless `legendMode="hover"`) and now conflicts with master, so I'm closing it as already resolved.

Issue #7212, opened 2024-12-03
Property `withWeekNumbers` does not work in DatePickerInput


boristian:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.14.3

### What package has an issue?

@mantine/dates

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

All

### Describe the bug

Hi there, first of all, thanks for these great component libraries!

I was happy to see the new prop `withWeekNumbers` for calendar components in the 7.14.3 and decided to try it out in one of my projects with a DatePickerInput. Unfortunately it doesn't work.
The week numbers are not displayed and on the browser console i get this message:
 
> clients.js:xx Warning: React does not recognize the `withWeekNumbers` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `withweeknumbers` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

### Self-service

- [ ] I would be willing to implement a fix for this issue

PR #7213, opened 2024-12-03
Add withWeekNumbers to pickCalendarProps

me-gusta-v:
Fixes https://github.com/mantinedev/mantine/issues/7212

The withWeekNumbers prop is currently passed to the input base in DatePickerInput instead of as part of the calendar props

Add it to pickCalendarProps


rtivital:
Thanks!

PR #7226, opened 2024-12-06
TypographyStylesProvider CSS modification for Paragraphs inside lists

### LINKED ISSUE #7223: Paragraphs inside lists move to a different line from list bullets when using TypographyStylesProvider
### Dependencies check up

7.14.3

@mantine/core

Next.js

Chrome

### Describe the bug

Large paragraphs appear on a separate line from bullet points in lists wrapped in a TypographyStylesProvider - for instance in the RichTextEditor example page in the docs, adding extra texts to one of the list items puts the preceding bullet on the previous line. Single-line paragraphs are fine in the RichTextEditor but not in other lists inside the provider.

_No response_

### Possible fix

This seems to have been caused by the following commit https://github.com/mantinedev/mantine/commit/5e03fa3817484613eb0bf488ba0b0173d422478b

### Self-service

onkar-deshpande:
Fixes #7223 

Made changes in css which handles paragraphs text alignment.


rtivital:
Thanks!

PR #7230, opened 2024-12-07
Export date hooks

heybart:
Currently the date hooks are not exported, making it hard to create your own DatePickerInput. Since all other types/utils and components are exported, I think it makes sense to export the hooks as well.


rtivital:
These hooks are internal, they are not planned to be exported from the package

PR #7232, opened 2024-12-08
[@mantine/core] Slider: Fix thumb children color in dark theme

### LINKED ISSUE #7231: @mantine/core Slider thumb children have incorrect color when using dark theme
### Dependencies check up

7.14.3

@mantine/core

Vite

All

### Describe the bug

The thumb children of the Slider component blend in with the thumb background color when using mantine dark theme. To reproduce, navigate to https://mantine.dev/core/slider/#thumb-children and enable mantine dark theme in upper-right corner.

image

Working as expected in light theme.
image

_No response_

### Possible fix

_No response_

### Self-service

ZFudge:
Fixes #7231 

image

### DIFF
--- a/packages/@mantine/core/src/components/Slider/Slider.module.css
+++ b/packages/@mantine/core/src/components/Slider/Slider.module.css
@@ -97,6 +97,12 @@
     border-color: var(--mantine-color-white);
     background-color: var(--slider-color);
   }
+
+  & > svg {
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks for submitting a PR, unfortunately it is not a valid solution. I've resolved the issue on my side.

PR #7238, opened 2024-12-10
Replace backslashes by slashes so it works on Windows with git bash too

### LINKED ISSUE #7237: sassOptions's additionalData uses backslashes which git bash (Windows) cannot resolve, resulting in build failure

https://mantine.dev/styles/sass/

### What is wrong?

This config line results in an error if one wants to build in a Windows git bash environment:

In my case, we use Next.js 14, but I am sure for Vite it would be the same:

```
$ npm run build

....
./app/layout.module.scss
SassError: Can't find stylesheet to import.
  ╷
1 │ @use "C:\Users\jnachtigall\Devel\mh2-frontend\globals\theming\_mantine" as mantine;
  │ ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  ╵
  app\layout.module.scss 1:1  root stylesheet
```

The problem is that git bash cannot resolve paths with backslashes:
```sh
$ ls C:\Users\jnachtigall\Devel\mh2-frontend
ls: cannot access 'C:UsersjnachtigallDevelmh2-frontend': No such file or directory
```

but only with slashes:
```sh
$ ls C:/Users/jnachtigall/Devel/mh2-frontend
  ...this works...
```

But node's inbuild `path.join(process.cwd(), './globals/theming/_mantine')` always returns path with backslashes because it assumes a Windows environement. Adding a `.replace(/\\/g, '/')` fixes the problem.

PS See #7023 where the issue was introduced afaics

jnachtigall:
Not sure this fix also works for PowerShell, but is required for git bash to resolve the path the the _mantine sass file.

Fixes #7237

### DIFF
--- a/apps/mantine.dev/src/pages/styles/sass.mdx
+++ b/apps/mantine.dev/src/pages/styles/sass.mdx
@@ -35,7 +35,7 @@ export default defineConfig({
     preprocessorOptions: {
       scss: {
         api: 'modern-compiler',
-        additionalData: `@use "${path.join(process.cwd(), 'src/_mantine')}" as mantine;`,
+        additionalData: `@use "${path.join(process.cwd(), 'src/_mantine').replace(/\\/g, '/')}" as mantine;`,
       },
     },
   },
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #7255, opened 2024-12-13
Remove paragraph inline-block

### LINKED ISSUE #7254: Multi-line paragraphs in list items in RichTextEditor have the marker bottom aligned
### Dependencies check up

7.15.1

@mantine/tiptap

Next.js

Chrome

### Describe the bug

List item markers on multi-line text in the RichTextEditor are aligned to the bottom of the text block rather than the top as they should. This can be verified on the RichTextEditor demo in the docs.

_No response_

### Possible fix

It seems this is caused by a `display: inline-block` css rule applied to all `li > p` that are children of the rich text editor component. Removing this rule via inspect element fixes the problem. This was added in 7aac29f, not sure what the fix was intended to address.

### Self-service

QuinnStraus:
Fixes #7254

### DIFF
--- a/packages/@mantine/tiptap/src/RichTextEditor.module.css
+++ b/packages/@mantine/tiptap/src/RichTextEditor.module.css
@@ -18,7 +18,6 @@
 
   & li > p {
     margin: 0;
-    display: inline-block;
   }
 
   & ul li,


rtivital:
Thanks!

PR #7267, opened 2024-12-16
[@mantine/core] FloatingIndicator: fix translate calculations

lamualfa:
# Descriptions

The `borderTopWidth` & `borderLeftWidth` from `parent` & `target` should added to translate calculation because [`getBoundingClientRect`](https://developer.mozilla.org/en-US/docs/Web/API/Element/getBoundingClientRect) included that on calculation.

**Before**

image

**After**

Screenshot from 2024-12-16 16-24-48

It's a little bit difficult to see the difference unless you use wider border width.

# Notes

- I reuse [`toInt`](https://github.com/lamualfa/mantine/blob/e5faa0d2c4df522ea981787a7da8076e6e024ec4/packages/%40mantine/core/src/components/FloatingIndicator/use-floating-indicator.ts#L4) helper from `ScrollArea` utils to [convert computed style to int](https://github.com/lamualfa/mantine/blob/e5faa0d2c4df522ea981787a7da8076e6e024ec4/packages/%40mantine/core/src/components/FloatingIndicator/use-floating-indicator.ts#L55). That helper should be on more generic folder/file other than `ScrollArea`. I'm not sure where to put it. This refactor must be in different commit.

### DIFF
--- a/packages/@mantine/core/src/components/FloatingIndicator/use-floating-indicator.ts
+++ b/packages/@mantine/core/src/components/FloatingIndicator/use-floating-indicator.ts
@@ -1,6 +1,7 @@
-import { RefObject, useEffect, useRef, useState } from 'react';
+import { useEffect, useRef, useState, type RefObject } from 'react';
 import { useMutationObserver, useTimeout } from '@mantine/hooks';
 import { getEnv } from '../../core';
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #7270, opened 2024-12-16
(docs): updated docs, to reflect actual behaviour of allowSingleDateInRange

StyxOfDynamite:
`allowSingleDateInRange` allows for a single day to be selected, by clicking the same value in the date input for both start and end dates.

This was not reflected in the documentation.


rtivital:
Thanks!

PR #7272, opened 2024-12-16
useDebouncedCallback: add a flush method to returned callback

scamden:
As well as give an option to simply flush on unmount, which is desirable when the callback does persisting of values (but not when it's just updating state obviously)

### DIFF
--- a/packages/@mantine/hooks/src/use-debounced-callback/use-debounced-callback.test.ts
+++ b/packages/@mantine/hooks/src/use-debounced-callback/use-debounced-callback.test.ts
@@ -26,4 +26,59 @@ describe('@mantine/hooks/use-debounced-callback', () => {
     jest.advanceTimersByTime(100);
     expect(callback).toHaveBeenCalledWith(3);
   });
+
+  it('can be flushed immediately', () => {
+    const callback = jest.fn();
+    const { result } = renderHook(() => useDebouncedCallback(callback, 100));
+    result.current(1);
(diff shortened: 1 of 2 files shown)


packages/@mantine/hooks/src/use-debounced-callback/use-debounced-callback.test.ts
```diff
@@ -26,4 +26,52 @@ describe('@mantine/hooks/use-debounced-callback', () => {
...
+    const { result, unmount } = renderHook(() => useDebouncedCallback(callback, {delay : 100, flushOnUnmount : true}));
+    result.current(1);
+    result.current(2);
+    result.current(3);
+    unmount();
+    expect(callback).toHaveBeenCalledWith(3);
+  });
+
+
+  it('does not call after unmount if timer lapsed', () => {
```
scamden, line 51: this adds a test for the clearTimeout functionality in the useEffect, previously untested, not strictly related to this change :)

rtivital:
Please fix failing tests.

scamden:
> Please fix failing tests.

Will do. I think it’s just formatting

scamden:
> Please fix failing tests.

Fixed I believe, ran locally this time to check

scamden:
@rtivital looks like checks pass. Any other thoughts on this approach?

rtivital:
Thanks!

PR #7281, opened 2024-12-18
[@mantine/core] Tooltip: add middlewares prop

jvllmr:
This PR implements the middlewares prop for Tooltip which lets you control the Floating UI middlewares used with the Tooltip. The implementation is derived from the Popover component. One could unite these implementations, but then there would be coupling between the Tooltip and the Popover component which wouldn't be helpful for future adjustments.

I added only one simple test demonstration in the storybook to observe my changes. Let me know if you require more tests/demos.


jvllmr:
There is also room for a discussion if the inline prop should be deprecated in a new major version in favor of the middlewares prop.

rtivital:
Thanks!

PR #7290, opened 2024-12-20
[@mantine/core] Restrict unexpected margin removal in TypographyStylesProvider

### LINKED ISSUE #7289: Unexpected margin removal in TypographyStylesProvider
### Dependencies check up

7.15.1

@mantine/core

Next.js

Chrome

### Describe the bug

The TypographyStylesProvider contains the following css selectors 
```
& :first-child {
  margin-top: 0;
}

& :last-child {
  margin-bottom: 0;
}
```

These impact all elements that are either first or last within their container within the provider, and override any margin set on the top or bottom of any elements using a single class selector, since these selectors have higher specificity. This creates some unexpected behavior and also creates some issues with eg. the spoiler component, which adds margin-bottom to the container when the show or hide buttons are visible. 

https://codesandbox.io/p/sandbox/mantine-react-template-forked-c8hv6r

### Possible fix

Removing the selectors from the TypographyStylesProvider [css file](https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/components/TypographyStylesProvider/TypographyStylesProvider.module.css) would solve the problem. From looking through the css file, it seems like these styles were added to remove the margin added to some items by the provider if they are first or last (specifically ul, ol, headers, img, p, hr, pre, table, caption, and blockquote). Filtering by these tags therefore should solve the problem.

### Self-service

QuinnStraus:
Fixes #7289

### DIFF
--- a/packages/@mantine/core/src/components/TypographyStylesProvider/TypographyStylesProvider.module.css
+++ b/packages/@mantine/core/src/components/TypographyStylesProvider/TypographyStylesProvider.module.css
@@ -1,9 +1,13 @@
 .root {
-  & :where(h1, h2, h3, h4, h5, h6, img, p, ul, ol, hr, pre, table, caption, blockquote) :first-child {
+  &
+    :where(h1, h2, h3, h4, h5, h6, img, p, ul, ol, hr, pre, table, caption, blockquote)
+    :first-child {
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #7297, opened 2024-12-22
Fix unmet dependency on React 19

adrum:
react-textarea-autosize 8.5.6 adds React 19 as an allowed dependency version.


rtivital:
Thanks for submitting the PR. The fix is invalid, I've updated the dependency on my side.

PR #7303, opened 2024-12-24
[mantine.dev] update use-favicon hook example to use cookieless domain

angristan:
With the current example, when logged in to X, Chrome refuses to fetch the icon. I update the example to use another cookieless domain so that the example works for everyone.


rtivital:
Thanks!

PR #7322, opened 2024-12-31
Fix Avatar initials color hashing to use full name

YukiKoNeko:
**Problem:**
In the Avatar component, the `getInitialsColor` function was using only the initials of a name to generate a hash, which determined the color of the avatar. This led to the issue where different names with the same initials would receive the same color, even if the full names were different.

**Solution:**
The function now uses the full name to generate the hash. By removing the use of `getInitials()` inside the `hashCode()` call, each unique name will produce a unique hash, ensuring a unique color for each different name.

**Details:**
- The `hashCode()` call is updated to process the entire name, not just the initials.
- This change maintains the internal implementation without affecting the public API.
- The behavior of the `getInitialsColor` function remains consistent, mapping the hash to an index in the colors array.

**Impact:**
- Resolves the issue of same initials leading to same colors.
- Enhances the distinguishability of avatars in user interfaces.


rtivital:
Thanks!

PR #7358, opened 2025-01-14
[@mantine/core] Fix On Escape KeyDown handling to Combobox component

### LINKED ISSUE #7337: Escape to Close Combobox does not work on Safari
### Dependencies check up

7.15.1

@mantine/core

Vite

Safari

### Describe the bug

In Firefox, when I open a combobox I can close it using Escape. But on Safari, this is not true.

https://mantine.dev/core/combobox/

### Possible fix

https://stackoverflow.com/questions/719537/capture-an-escape-keypress-in-webkit-browsers

### Self-service

yuki0410-dev:
fixes #7337 

The method of judging escape key press events is now the same as for Modal.
https://github.com/mantinedev/mantine/blob/e8f7b01f85149e6f7d9776ef41703bab739f5d4d/packages/%40mantine/core/src/components/ModalBase/use-modal.ts#L37

### DIFF
--- a/packages/@mantine/core/src/components/Combobox/use-combobox-target-props/use-combobox-target-props.ts
+++ b/packages/@mantine/core/src/components/Combobox/use-combobox-target-props/use-combobox-target-props.ts
@@ -76,7 +76,7 @@ export function useComboboxTargetProps({
         }
       }
 
-      if (event.nativeEvent.code === 'Escape') {
+      if (event.key === 'Escape') {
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

elibroftw:
The issue is bigger than the escape key. It seems that onKeyDown itself does not run when the combobox is not an input. Will create a new issue.

PR #7359, opened 2025-01-14
[@mantine/hooks] Fix Handling of multiple Entries to useInViewport

### LINKED ISSUE #7340: inViewport using the `useInViewport` hook is `false` after drag and drop
### Dependencies check up

7.15.3

@mantine/core

Vite

Chrome

### Describe the bug

Reproduction scenario:

1. Starting from the drag and drop example in the documentation
2. Add a `useInViewport` to display the content of a card only when visible

Expected:

After using drag and drop, if a card is in the viewport the value of inViewport is `true`.

Observed:

After drag and drop, the inViewport value is `false` even if the card is in the viewport.

Can be reproduced using the sandbox code: 

https://stackblitz.com/edit/vitejs-vite-emcugsmi?file=src%2FDndList.tsx

view-port-dnd-issue-2

https://stackblitz.com/edit/vitejs-vite-emcugsmi?file=src%2FDndList.tsx

### Possible fix

_No response_

### Self-service

yuki0410-dev:
fixes #7340 

IntersectionObserver's Handler now handles cases where there is more than one Entry.
https://codesandbox.io/p/sandbox/mantine-7340-6w3krx?file=%2Fsrc%2FDndList.tsx%3A89%2C1-92%2C12

### DIFF
--- a/packages/@mantine/hooks/src/use-in-viewport/use-in-viewport.ts
+++ b/packages/@mantine/hooks/src/use-in-viewport/use-in-viewport.ts
@@ -7,8 +7,8 @@ export function useInViewport<T extends HTMLElement = any>() {
   const ref = useCallback((node: T | null) => {
     if (typeof IntersectionObserver !== 'undefined') {
       if (node && !observer.current) {
-        observer.current = new IntersectionObserver(([entry]) =>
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #7360, opened 2025-01-14
[@mantine/core] Handle undefined value in matchMedia

Benjaminsson:
[This PR](https://github.com/mantinedev/mantine/pull/7147) broke my testing setup. In some cases `window.matchMedia("(prefers-color-scheme: dark)")` returns undefined and breaks my tests. Optional chaining solves it.


rtivital:
Thanks!

PR #7364, opened 2025-01-15
[mantine.dev] Fix description in use-isomorphic-effect

kidonng:
According to implementation notes and meta data it's a replacement for `useLayoutEffect` instead of `useEffect`:

https://github.com/mantinedev/mantine/blob/e8f7b01f85149e6f7d9776ef41703bab739f5d4d/apps/mantine.dev/src/mdx/data/mdx-hooks-data.ts#L93

https://github.com/mantinedev/mantine/blob/e8f7b01f85149e6f7d9776ef41703bab739f5d4d/packages/%40mantine/hooks/src/use-isomorphic-effect/use-isomorphic-effect.ts#L3-L4


rtivital:
Thanks!

PR #7373, opened 2025-01-16
fix: changed spreading order in form.resetDirty to make it behave correctly

tomrehnstrom:
When using `form.resetDirty` and passing an object as argument, it's expected that the object being passed should act as the new baseline for the dirty checking. Currently, because of how the objects are being spread it does not behave like this and in fact basically does the same thing as not having passed any argument at all.

This PR should fix this :)


rtivital:
Thanks!

PR #7377, opened 2025-01-17
[@mantine/core] Menu.item: add data-disabled props

### LINKED ISSUE #7341: MenuItem not accepting data-disabled attribtue
### Dependencies check up

7.15.3

@mantine/core

Vite

Chrome

### Describe the bug

according to https://mantine.dev/core/button/#disabled-button-with-tooltip, should the menu item have the same behavior with button when displaying a tooltip/hovercard on disabled state?

currently the data-disabled attribtue is useless and will not be passed to the button, it only handles disable props internally.  https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/components/Menu/MenuItem/MenuItem.tsx#L106

_No response_

### Possible fix

_No response_

### Self-service

yuki0410-dev:
closes #7341

Added data-disabled attribute to Menu.Item component like Button component.


packages/@docs/demos/src/demos/core/Menu/Menu.demo.disabledTooltip.tsx
```diff
@@ -0,0 +1,79 @@
+import {
```
rtivital, line 1: Please remove this file, it is not needed.

yuki0410-dev: @rtivital 
Thank you for your review.
We have addressed your advice.

rtivital:
Thanks!

PR #7378, opened 2025-01-17
[@mantine/chart] AreaChart : fix console error

### LINKED ISSUE #7376: React does not recognize the `gridColor` prop on a DOM element
### Dependencies check up

7.16.0

@mantine/charts

Vite

Chrome

### Describe the bug

When using a AreaChart and passing in gridColor props I receive the following error from React:

Warning: React does not recognize the gridColor prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase gridcolor instead. If you accidentally passed it from a parent component, remove it from the DOM element.

_No response_

### Possible fix

> Assuming my intuition about the cause is correct, I expect a fix will follow the same pattern as the [PR](https://github.com/mantinedev/mantine/pull/6810) linked above on a similar issue.

### Self-service

yuki0410-dev:
closes #7376


rtivital:
Thanks!

PR #7380, opened 2025-01-17
Add missing 'withPages' prop to Pagination examples

cameronrdecker:
withPages prop is missing in the demo code for the Pagination component as well as the version 7.16 changelog


rtivital:
Thanks!

PR #7383, opened 2025-01-18
[@mantine/core]: NumberInput: fixes #7369 by making sure onChange is …

### LINKED ISSUE #7369: [NumberInput] validation gets cleared after second blur when input is empty
### Dependencies check up

7.16.0

@mantine/core

Next.js

Chrome

### Describe the bug

Heya,

When using a NumberInput component the validation error is never shown for empty values.

In the code sandbox the issue is a little different, there the validation error is vanishing after the second blur so I assume our issue might have something to do with next.js or React 19 as well.

I hope if we're able to fix it in the scenario of the code sandbox it might also be fixed in our project.
I'm not entirely sure why it happens but it has to do with the `trimLeadingZeroesOnBlur` property, as it doesn't happen when it is set to `false`.

Maybe a rerender timing could be the issue here since the value is internally updates on blur?
Should be tied to this code: https://github.com/mantinedev/mantine/blob/master/packages/@mantine/core/src/components/NumberInput/NumberInput.tsx#L496

The issue might already be solved if the trim is only done with the check `_value !== ""` or `_value.startsWith("0")`.
The problem only occurs for the case of an empty input. So validating something like value > 10 or something it works fine.

Best Regards,
Rico

https://codesandbox.io/p/sandbox/mantine-react-template-forked-5z4zsv

### Possible fix

```tsx
if (
  trimLeadingZeroesOnBlur &&
  typeof _value === 'string' &&
  _value !== "" &&
  getDecimalPlaces(_value) < 15
)
```

```tsx
if (
  trimLeadingZeroesOnBlur &&
  typeof _value === 'string' &&
  _value.startsWith("0") &&
  getDecimalPlaces(_value) < 15
)
```

### Self-service

btmnk:
…not triggered unnecessarily while onBlur

### DIFF
--- a/packages/@mantine/core/src/components/NumberInput/NumberInput.story.tsx
+++ b/packages/@mantine/core/src/components/NumberInput/NumberInput.story.tsx
@@ -232,19 +232,27 @@ export function FormValidateOnBlur() {
   const form = useForm({
     validateInputOnBlur: true,
     validate: {
-      age: (value) => (value < 18 ? 'Error' : null),
+      age: (value) => {
+        if (typeof value === 'string' && value === '') {
+          return 'Required';
+        }
+        if (typeof value === 'number' && value < 18) {
+          return 'Error';
+        }
+        return null;
+      },
       name: (value) => (value.length < 2 ? 'Error' : null),
     },
     initialValues: {
       name: '',
-      age: 2,
+      age: '' as string | number,
     },
   });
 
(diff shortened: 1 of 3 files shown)


btmnk:
closes https://github.com/mantinedev/mantine/issues/7369

rtivital:
Thanks!

PR #7389, opened 2025-01-21
[@mantine/dates] MonthPicker: improve set picked date update logic

### LINKED ISSUE #7336: MonthPicker in range mode in controlled form endlessly useEffect calls
### Dependencies check up

7.15.3 / 7.16.0

@mantine/dates

Vite

All

### Describe the bug

When using the `<MonthPicker>` in range mode within a controlled form `use-date-state.ts` endlessly calls `useEffect` at

https://github.com/mantinedev/mantine/blob/476b417cb126d4ac88a40ac4004d8b5c3c78cc0c/packages/%40mantine/dates/src/hooks/use-dates-state/use-dates-state.ts#L169-L178

The controlled form sets the value of the `<MonthPicker>` to `[new Date(), null]`.

https://codesandbox.io/p/sandbox/mantine-react-template-forked-qtxcvf

### Possible fix

In Version 7.15.1 the issue was not present.

### Self-service

openscript:
Fixes #7336


rtivital:
Thanks!

PR #7390, opened 2025-01-21
[@mantine/hooks] use-hotkey: Clarifying use of physical key assignments

danpeavey-classdojo:
We have multiple users of our apps using non-QWERTY keyboard layouts. Given the nature of the app, and the need for inputs to be collocated sensibly (thinking camera manipulation), this convenience hook caused issues for our non-QWERTY users. 

This change corrects that. Open to feedback & adjustments, and hoping this is useful.

- [KeyboardEvent: code property](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/code)
- [Code values for keyboard events](https://developer.mozilla.org/en-US/docs/Web/API/UI_Events/Keyboard_event_code_values)

### DIFF
--- a/apps/mantine.dev/src/pages/hooks/use-hotkeys.mdx
+++ b/apps/mantine.dev/src/pages/hooks/use-hotkeys.mdx
@@ -75,6 +75,7 @@ document.body.addEventListener(
 - `alt + shift + L` – you can use whitespace inside hotkey
 - `ArrowLeft` – you can use special keys using [this format](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/key/Key_Values)
 - `shift + [plus]` – you can use `[plus]` to detect `+` key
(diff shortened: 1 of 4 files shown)


rtivital:
Thanks!

Odonno:
The conditions here does not seem to be accurate. I still face an issue with French ISO layout. 
https://github.com/mantinedev/mantine/pull/7390

When pressing `CTRL+Z`, it currently executes both `CTRL+W` and `CTRL+Z`, even when setting `{ usePhysicalKeys: true }`. Here is what the "z" pressed event looks like for info:

image

We should find a way to remove the check using the `pressedCode` property.

danpeavey-classdojo:
I can look into this after-hours. Thanks for raising this.

Odonno:
Just checking where we are.

danpeavey-classdojo:
No traction lately; it shouldn't be a difficult or time-intensive fix but I haven't been in React land for several weeks. I'm anticipating being back in the domain in ~2 weeks. May be faster for you to toss over a PR, and I'm happy to test on multiple other keyboards (literal, physical, and logical configs).

Issue #7388, opened 2025-01-21
Datetime input field shows incorrect time while the dropdown/modal is accurate


intigratech:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.15.0

### What package has an issue?

@mantine/dates

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

All

### Describe the bug

The issue can be seen on https://mantine.dev/dates/dates-provider/#date-format-considerations
The input field shows a time - for example: 02/10/2000 12:40 which doesn't match the timezone set on the datesprovider. However, the modal/dropdown to input the time shows a different timezone, presumably based on the timezone provided in the datesprovider.

```
import { DatesProvider, DateTimePicker } from '@mantine/dates';

function Demo() {
  return (
    <DatesProvider settings={{ timezone: 'America/New_York' }}>
      <DateTimePicker label="Pick a Date" placeholder="Pick a Date" defaultValue={new Date('2000-10-03 02:10:00Z')} />
    </DatesProvider>
  );
}
```

### If possible, include a link to a codesandbox with a minimal reproduction

https://mantine.dev/dates/dates-provider/#date-format-considerations

### Possible fix

The input field and modal/dropdown must respect the timezone provided.

### Self-service

- [ ] I would be willing to implement a fix for this issue

mrbnclt:
@rtivital I would like to submit a PR that should fix this.

rtivital:
Okay, you are welcome to

PR #7400, opened 2025-01-22
[@mantine/dates] DateTimePicker: prevent converting value to DateProvider twice

mrbnclt:
`DateTimePicker` uses `useUncontrolledDates` hook which is already using the context from the `DateProvider` to convert the default value into the `DateProvider`s timezone. 

In the `DateTimePicker` we shouldn't do another `.tz()` that will convert the date & time into the `DateProvider`'s timezone, the result will be incorrect if that's the implementation. 

Fix was to pass `true` to the `.tz()`'s second parameter (`keepLocalTime`), which doesn't convert the value again to the `DateProvider`'s timezone.

### Before
before  

### After
 after

Closes https://github.com/mantinedev/mantine/issues/7388

### DIFF
--- a/packages/@mantine/dates/src/components/DateTimePicker/DateTimePicker.test.tsx
+++ b/packages/@mantine/dates/src/components/DateTimePicker/DateTimePicker.test.tsx
@@ -192,15 +192,15 @@ describe('@mantine/dates/DateTimePicker', () => {
         />
       </DatesProvider>
     );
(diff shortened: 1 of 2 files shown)


rtivital:
Please fix errors in tests

mrbnclt:
Hi @rtivital test are now passing, for your review please 🙇🏻

rtivital:
Thanks!

PR #7404, opened 2025-01-23
Fix potential dangerous access of ref value in useEffect cleanup

### LINKED ISSUE #7406: Inproper usage of refs throughout the code base
### Dependencies check up

7.16.0

@mantine/hooks

Vite

All

### Describe the bug

While working on https://github.com/mantinedev/mantine/pull/7404, I noticed that there are many places that depend on ref values inside `useEffect` hooks. Sometimes the ref value is included in the dependency array (for example in the `useHover` hook __[1]__) and sometimes it is used inside a `useEffect` hook but not declared as a dependency at all (for example in the `useMove` hook __[2]__).

Both are problematic and are a source of bugs. Updates to a ref value do **not** trigger re-renders and thus the code of the `useEffect` hook will never be executed in some scenarios.

Here are two sandbox examples, illustrating the bug:

- `useHover`: https://codesandbox.io/p/sandbox/mantine-react-template-forked-x745wn
- `useMove`: https://codesandbox.io/p/sandbox/mantine-react-template-forked-rz84sj

The `useHover` one does not work after clicking the "Show Div" button, but a triggering a re-rendering "fixes" it, because the value of `ref.current` is included in the dependency array:

https://github.com/user-attachments/assets/c094f3c7-b77c-4f83-a94f-16d64c12b65a

The `useMove` one never recovers, even after re-renders, because `ref.current` is not included in the dependency array:

https://github.com/user-attachments/assets/662f82f3-1664-4800-8250-4c4ad68b4e6b

These are just two examples, but there are many others, and since these hooks are also used in the Mantine components, they are most likely affected as well.

__[1]__ `ref.current` in dependency array of `useHover`
https://github.com/mantinedev/mantine/blob/85f6f0ac372172d0de8a72690bac39b9c7cfaa36/packages/%40mantine/hooks/src/use-hover/use-hover.ts#L9-L21

__[2]__ `ref.current` used in `useEffect` hook but not included in dependency array
https://github.com/mantinedev/mantine/blob/85f6f0ac372172d0de8a72690bac39b9c7cfaa36/packages/%40mantine/hooks/src/use-move/use-move.ts#L115-L124

_No response_

### Possible fix

Refs should not be used in these cases. Instead the html element should be stored inside a React state which triggers a re-render when the value changes.

```tsx
import { useCallback, useEffect, useState } from 'react';


export function useHover<T extends HTMLElement = any>() {
  const [hovered, setHovered] = useState(false);
  const [node, setNode] = useState<T>();
  const onMouseEnter = useCallback(() => setHovered(true), []);
  const onMouseLeave = useCallback(() => setHovered(false), []);

  useEffect(() => {
    if (node) {
      node.addEventListener('mouseenter', onMouseEnter);
      node.addEventListener('mouseleave', onMouseLeave);

      return () => {
        node?.removeEventListener('mouseenter', onMouseEnter);
        node?.removeEventListener('mouseleave', onMouseLeave);
      };
    }

    return undefined;
  }, [node]);

  return { ref: setNode, hovered };
}
```

The usage for the user remains the same, but strictly speaking this is a breaking change.

### Self-service

Bastian:
Fixes #7406

Directly accessing the value of a ref in a `useEffect` cleanup function can lead to problems, as refs “live” outside of React's rendering cycle and it can therefore happen that the value of the ref changes between the first execution and the cleanup.

See ESLint warning:
image

This is not just a theoretical problem. I encountered a bug after upgrading to React 19 while running in dev mode. My callback passed to the `onChange` prop of a `Slider` component (which uses `useMove` internally) was not updated and therefore used a stale variable that was captured by the closure. I believe this is related to React calling the useEffect hooks twice in development mode, although I'm not sure why this only happens with React 19 and not React 18, which introduced this behavior.

### DIFF
--- a/packages/@mantine/hooks/src/use-event-listener/use-event-listener.ts
+++ b/packages/@mantine/hooks/src/use-event-listener/use-event-listener.ts
@@ -8,9 +8,11 @@ export function useEventListener<K extends keyof HTMLElementEventMap, T extends 
   const ref = useRef<T>(null);
 
   useEffect(() => {
-    if (ref.current) {
-      ref.current.addEventListener(type, listener as any, options);
-      return () => ref.current?.removeEventListener(type, listener as any, options);
+    const node = ref.current;
+
+    if (node) {
+      node.addEventListener(type, listener as any, options);
+      return () => node?.removeEventListener(type, listener as any, options);
     }
     return undefined;
   }, [listener, options]);
--- a/packages/@mantine/hooks/src/use-focus-within/use-focus-within.ts
+++ b/packages/@mantine/hooks/src/use-focus-within/use-focus-within.ts
@@ -41,13 +41,15 @@ export function useFocusWithin<T extends HTMLElement = any>({
   };
 
   useEffect(() => {
-    if (ref.current) {
-      ref.current.addEventListener('focusin', handleFocusIn);
-      ref.current.addEventListener('focusout', handleFocusOut);
+    const node = ref.current;
+
+    if (node) {
+      node.addEventListener('focusin', handleFocusIn);
+      node.addEventListener('focusout', handleFocusOut);
 
       return () => {
-        ref.current?.removeEventListener('focusin', handleFocusIn);
-        ref.current?.removeEventListener('focusout', handleFocusOut);
+        node?.removeEventListener('focusin', handleFocusIn);
+        node?.removeEventListener('focusout', handleFocusOut);
       };
     }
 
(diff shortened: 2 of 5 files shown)


Bastian:
Follow up: https://github.com/mantinedev/mantine/issues/7406

tl;dr: Refs should probably not be used at all in this case. I would still appreciate it if this PR could be merged as it fixes an immediate bug I have in my code.

rtivital:
Thanks!

Issue #6681, opened 2024-08-16
ScrollArea doesn't always reset blocking pointer events on body tag


t-vince:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.12.0

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

In the ScrollArea component when scrolling, pointer-events are set to none and when scrolling stops this is reset back to auto. 
This behavior happens since https://github.com/mantinedev/mantine/commit/bd6fad374b88894c6615386b47f5899cdeab2929 with
```
onPointerDown={composeEventHandlers(props.onPointerUp, (event) => {
          [...]
          document.body.style.pointerEvents = 'none';
```
and gets reset with
```
onPointerUp={composeEventHandlers(props.onPointerUp, (event) => {
          [...]
          document.body.style.pointerEvents = 'auto';
```

However, this does not always work reliably. When clicking next to the scrollbar after scrolling, it can cause it to not reset the this pointer event, which makes it seem that the whole page freezes as it no longer takes pointer input. Only way to make the page function again is to remove the `pointer-events: none;` from the body tag or refreshing the page.

The last version where this is not reproducible is `7.10.1` (the version before the mentioned commit)
It is also reproducible on the mantine website, on the [ScrollArea page](https://mantine.dev/core/scroll-area/)

https://github.com/user-attachments/assets/e01dc3c6-109d-4ff6-afd4-d05dd2b7130b

### Possible fix

It seems that the `pointerUp` event is not always correctly triggered. I'd propose to just revert the pointer-events on body PR, make this optional so people can opt-out or scope it to the scrollArea instead of the body tag. Currently it's quite risky; when it fails, it disables the complete page and not just the ScrollArea component.

### Self-service

- [ ] I would be willing to implement a fix for this issue

rtivital:
Fixed in [7.12.2](https://github.com/mantinedev/mantine/releases/tag/7.12.2)

PR #7423, opened 2025-01-27
[@mantine/core] ScrollArea: Fix user-select being left as none after interaction with scrollbar

webkonstantin:
This should be a proper fix for the issue https://github.com/mantinedev/mantine/issues/6681

I think another option would be to also handle `pointercancel` event, but `lostpointercapture` should cover everything. Not sure that handling `pointerup` is even required in this case, but I left it just in case.

Related commit https://github.com/mantinedev/mantine/commit/ccc0f3c42284c6fdbe671732aabee0d9299eb98d

The issue is reproducible on the mantine website

https://github.com/user-attachments/assets/41f041c1-256f-4685-8756-6545987e7ff6


rtivital:
Thanks!

Issue #7441, opened 2025-01-31
Typing more than 4 numbers in DateInput component freezes the page on Safari


IvanRighthub:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.16.2

### What package has an issue?

@mantine/dates

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

Safari

### Describe the bug

The `DateInput` component freezes the page on Safari when you add more than 4 numbers. The issue is present on the component page in the official docs as well [DateInput](https://mantine.dev/dates/date-input/). The issue occurs in a similar way with different date formatting as well. Also, the freeze prevents any mouse click and opening the inspector tool. If you had the inspector tool already opened it is sometimes frozen as well, or the whole browser freezes and I have to restart it.

Here is a screenshot from Safari with the frozen screen:
![Image](https://github.com/user-attachments/assets/be88714c-a470-45b1-9496-141a097e666d)

Additionally, some users on the project that I'm working on reported that the same freeze occurs when typing a date that starts with a 0 but I do not have a screenshot from that. So, if you try to type `01/02/2025` as a date, the page freezes after the first 0.

### Self-service

- [ ] I would be willing to implement a fix for this issue

PR #7442, opened 2025-01-31
[@mantine/dates]:DateInput component freezes the page on Safari 

kingflamez:
fixes https://github.com/mantinedev/mantine/issues/7441

**Issue:**

The original code encountered an issue when creating a dayjs object using a formatted string (dayjs(day.format('YYYY-M-D'))). This resulted in Safari producing invalid dates, leading to NaN values. This caused the while loop to continuously execute the getEndOfWeek() function, ultimately freezing the component in Safari.

**Root Cause:**
The format() method in dayjs converts the date to a string (ISO format), which Safari does not parse correctly in certain edge cases, especially with extreme years (e.g., 4543 as described in https://github.com/mantinedev/mantine/issues/7441). When re-parsed back into a dayjs object, the resulting date was invalid, causing subsequent calculations and date manipulations to fail.

**Fix:**
- Removed the unnecessary string formatting step, which was responsible for the invalid date in Safari.
- Replaced the .format() method with direct usage of dayjs’s native date handling methods to ensure correct date parsing and to avoid compatibility issues with Safari.

### DIFF
--- a/packages/@mantine/dates/src/components/Month/get-month-days/get-month-days.ts
+++ b/packages/@mantine/dates/src/components/Month/get-month-days/get-month-days.ts
@@ -15,7 +15,7 @@ export function getMonthDays({
   consistentWeeks,
 }: GetMonthDaysInput): Date[][] {
   const day = dayjs(month).subtract(dayjs(month).date() - 1, 'day');
-  const start = dayjs(day);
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

Issue #7439, opened 2025-01-31
Text gradient degree defaults to 45 when deg = 0


FunctionDJ:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.16.2

### What package has an issue?

@mantine/core

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

Can be observed in the official docs: https://mantine.dev/core/text/#gradient-variant
Change "from" to black and "to" to white or vice-versa, then drag "Gradient degree" to "0" and see how it abruptly changes when it hits "0" vs. "1" or nearby values above "1".
DevTools shows that with value "0" the applied `linear-gradient` is defaulted to `45deg`.

### Possible fix

Since `deg` is an optional parameter, i assume the issue is an OR operator (`||`) instead of a nullish coalescing operator (`??`) for the fallback.
`0` is a falsy value and so `0 || 45` is `45` whereas `0 ?? 45` is `0` and will only use the fallback if lefthand value is `null` or `undefined`.

### Self-service

- [ ] I would be willing to implement a fix for this issue

PR #7444, opened 2025-02-02
fix: [@mantine/core] update get-gradient deg to use nullish coalescing operators

marcinlukanus:
Resolves https://github.com/mantinedev/mantine/issues/7439

The bug reporter identified the solution in the report. Because `0` is a valid value but considered falsy in Javascript, using the logical OR (`||`) was causing a `deg={0}` to default to the defaultTheme's `45` value.

Before:

https://github.com/user-attachments/assets/628daa5e-6a47-488d-a5dc-7172efa118e1

After:

https://github.com/user-attachments/assets/a32d452d-410e-453e-8184-189c26171244


rtivital:
Thanks!

PR #7487, opened 2025-02-15
Feature/fix refs usage

DanielSegal1:
fixing #7406 
In short, I changed most of hook relying on refs to use the ref callback instead.
this fixes issues described thoroughly at #7406 , such as hooks not working when the ref is changed.

Strictly speaking, this is a breaking change because it changes the return value of the hooks to RefCallback instead of RefObject.
Also this is my first time contributint to open source so I may have missed steps (?)
(I'm not sure if I'm suppposed to bump the version or not etc.)


rtivital:
Thanks for proposing a fix, I've decided to implement it a little different on my side. The fix will be available in 8.1.0 minor release.

PR #7488, opened 2025-02-15
Feature/fix use hover ref

DanielSegal1:
Fix of #7406 only to the use hover hook as poc
Added a ut for the failing use case of when the ref is not initialized in the first render (and there are no renders after it is initialized) which fails in master.

Strictly, it is breaking change turning the ref return value from `RefObject` to `RefCallback`


rtivital:
Thanks for proposing a fix, I've decided to implement it a little different on my side. The fix will be available in 8.1.0 minor release.

Issue #7482, opened 2025-02-14
Polymorphic component lacks focus state when rendered as <a />


amaury-hanser:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.16.3

### What package has an issue?

@mantine/core

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

All

### Describe the bug

When using a polymorphic component to render a  `<a>` element, the component does not receive the expected focus state, impacting accessibility and user experience.

### If possible, include a link to a codesandbox with a minimal reproduction

https://codesandbox.io/p/sandbox/mantine-react-template-q4dd5d

### Self-service

- [ ] I would be willing to implement a fix for this issue

amaury-hanser:
Upon further investigation, we can see that for this specific `<Paper />` component, there is `.root { outline: 0; }` by default.

[CSS file of the component](https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/components/Paper/Paper.module.css)

I'm not sure as why it is here, focus would only happen if the element is interacted with so...

Maybe we could replace:

``` 
.root {
  outline: 0;
}
```
with: 
```
.root:not(:focus) {
  outline: 0;
}

rtivital:
Styles on Paper are explicitly defined to disable focus styles. It is not planned to be added – in cases at which Paper is used in the library – focus styles are not needed. It is not planned to be changed.

PR #7489, opened 2025-02-16
Fix focus state for polymorphic <a> elements

timmy471:
This PR ensures that the focus state is properly handled when a polymorphic component renders an `<a>` tag.

Fixes: #7482

### DIFF
--- a/packages/@mantine/core/src/components/Paper/Paper.module.css
+++ b/packages/@mantine/core/src/components/Paper/Paper.module.css
@@ -1,7 +1,6 @@
 .root {
   --paper-radius: var(--mantine-radius-default);
 
-  outline: 0;
   -webkit-tap-highlight-color: transparent;
   display: block;
   touch-action: manipulation;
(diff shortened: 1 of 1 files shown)


amaury-hanser:
Hi there, 

Thanks for taking the time to solve the problem. 🙏 

Your change only updates the `<Paper />' component, but I believe the problem affects many (all?) polymorphic components.  

Any idea how to fix this for all components? I don't know why the `outline: 0;` is there at all.

timmy471:
Thanks for the feedback! 🙏 

I can update the CSS on global level to exclude interactive elements like `<a>`, `<button>`, and `<input>`, so only non-interactive ones (e.g., `<div>`, `<span>`) lose the outline.

This should do the trick:

`:where(:not(:is(a, button, input, textarea, select, [tabindex]))):focus {
  outline: none;
}
`

amaury-hanser:
I haven't delved into Mantine's code, so I can't say. Sorry.

rtivital:
This change is not planned to be implemented.

PR #7497, opened 2025-02-17
Update CodeHighlightTabs.tsx: Fallback unsupported code to plaintext

imcvampire:
Instead of throwing unsupported code error, it fallback to plaintext.

It's the current behaviour of CodeHighlight component.

### DIFF
--- a/mise.toml
+++ b/mise.toml
@@ -1,0 +1,2 @@
+[tools]
+yarn = "latest"
--- a/mise.toml
+++ b/mise.toml
@@ -1,2 +1,0 @@
-[tools]
-yarn = "latest"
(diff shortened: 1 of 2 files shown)


mise.toml
```diff
@@ -0,0 +1,2 @@
+[tools]
```
rtivital: Remove this file, it is not related to PR

imcvampire: It's already removed

imcvampire: Do you mean that you want to remove it from the git history? Or it's okay to have the final commit without it?

rtivital:
Thanks!

PR #7498, opened 2025-02-17
bug fix function names on document

mehdiraized:
we have problem with funcation names on document migrate v7 to v8


rtivital:
What is wrong with these names?

mehdiraized:
i saw this function name on document

i removed all 8x and 7x on function names like other sections

rtivital:
Thanks for reporting the issue, I've fixed it on my side.

PR #7508, opened 2025-02-18
Export isJSONString from index.ts

### LINKED ISSUE #7507: isJSONString function is not being exported from index.ts
**Dependencies check up**

I have verified that I use latest version of all @mantine/* packages

**What version of @mantine packages do you have in package.json?**
7.17.0

**What package has an issue?**
@mantine/form

**What framework do you use?**
Vite

**In which browsers you can reproduce the issue?**
All

**Describe the bug**
In the new version 7.17.0 two new functions were added to form validators, isJSONString and isNotEmptyHTML per [documentation](https://mantine.dev/changelog/7-17-0/#isjsonstring-and-isnotemptyhtml-form-validators). While the isNotEmptyHTML is being exported from @mantine/form/src/validators/index.ts the isJSONString function is not. 

**If possible, include a link to a codesandbox with a minimal reproduction**
No response

**Possible fix**
Export the function from index.ts

**Self-service**

I would be willing to implement a fix for this issue: yes

ysalo:
fixes #7507


packages/@mantine/form/src/validators/index.ts
```diff
@@ -5,3 +5,4 @@ export { hasLength } from './has-length/has-length';
 export { isInRange } from './is-in-range/is-in-range';
 export { matchesField } from './matches-field/matches-field';
 export { isNotEmptyHTML } from './is-not-empty-html/is-not-empty-html';
+export { isJSONString} from './is-json-string/is-json-string';
```
dannya, line 8: @ysalo can you fix the code formatting please: `{ isJSONString }`

ysalo: fixed

rtivital:
Thanks!

PR #7513, opened 2025-02-20
[mantine.dev] Add `not-hover` underline option to Anchor documentation and demo

kino90:
Saw in the changelog for the latest version that a `not-hover` value was added for the `underline` prop of the `Anchor` component, but the new value has not been added neither in the documentation page nor in the example. 

I added both of these changes.

Hope it's all good!


rtivital:
Thanks!

PR #7519, opened 2025-02-23
Add centered overlay demo

AlonMiz:
Introduce a new demo showcasing the use of the `center` prop in the `Overlay` component, allowing content to be centered within its container.

image

image


rtivital:
This demo is not needed

AlonMiz:
i disagree. actaully, i've made it because my colleague and i were debating wether the centered is related to the children or not.
only after we tested, we understood the meening of centered.
there's no example with overlay children and i think that this one is useful, and ambigous, so this can be helpful overall

Issue #7504, opened 2025-02-18
useDebouncedState incorrectly handles state update callback


KrystilizeNevaDies:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.17.0

### What package has an issue?

@mantine/hooks

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

All

### Describe the bug

Multiple partial state updates using the state update callback causes incomplete state.

In this case:
```ts
  const [search, setSearch] = useDebouncedState<string>('', 200, { leading: true })

  useEffect(() => {
    setSearch(prev => prev + 'a')
    setSearch(prev => prev + 'b')
    setSearch(prev => prev + 'c')
  }, [])
```

Only one of 'a', 'b', or 'c' are appended to the value, where all three should be.

### Possible fix

Adjust the behavior of `useDebouncedState` to sequentially apply the state updates correctly.

### Self-service

- [ ] I would be willing to implement a fix for this issue

rtivital:
It works as indended, use-debounced-state does not support state updated calback, it does not work the same value as `useState` hook. It is not planned to be changed.

KrystilizeNevaDies:
If "use-debounced-state does not support state updated callback", then isn't being able to call setState with the callback a bug?

PR #7523, opened 2025-02-25
Fixed debouncing incorectly handling the state from callback 

lndrj:
Adjusting behavior of useDeboucnedState - applying the changes sequentially by checking the ```latestValue``` 

Fixes: #7504


packages/@mantine/hooks/src/use-debounced-value/use-debounced-value.ts
```diff
@@ -5,19 +5,26 @@ export function useDebouncedValue<T = any>(value: T, wait: number, options = { l
...
 
   useEffect(() => {
     if (mountedRef.current) {
       if (!cooldownRef.current && options.leading) {
         cooldownRef.current = true;
-        setValue(value);
+        setValue((prev) => {
+          latestValue.current = typeof value === "function" ? value(prev) : value;
+          return latestValue.current;
+        })
```
rtivital, line 19: What is the difference from:

```tsx
setValue(value)
latestValue.current = value
```

PR #7526, opened 2025-02-26
fix: broken "go to definition" for `createStyles().classes` object

### LINKED ISSUE #7491: `createStyles`: Navigating to definition of `classes` ends up at TS definition instead of source code
### Dependencies check up

7.15.2

@mantine/core

Vite

All

### Describe the bug

This is actually a bug in the `@mantine/emotion` package, but the option was not available in the issue template.

Given this example:

```tsx
// dropdown.styles.ts
import { createStyles } from '@mantine/emotion';

export const useStyles = createStyles((theme) => ({
  root: {
    width: '100%',
  },
  dropdown: {
    boxShadow: theme.shadows.sm,
  },
}));

// dropdown.ts
import { useStyles } from './dropdown.styles'

export function Dropdown() {
  const { classes } = useStyles()

  return (
    <MultiSelect
      classNames={{
        root: classes.root,
        dropdown: classes.dropdown,
      }}
    />
  )
}
```

Attempting to "Go to Definition" for `classes.root` will take us to the TS definition, not `dropdown.styles.ts`. I believe this is due to a regression in the TS definitions for `createStyles`, specifically the use of `Record` over a mapped type: https://github.com/mantinedev/mantine/blob/3f1b82458a82d9f1e83ef9b59a3c9422e069ee00/packages/%40mantine/emotion/src/create-styles.ts#L44

_No response_

### Possible fix

Changing the type definition locally to a mapped type restores IDE functionality, confirming this bug:

```diff
// create-styles.d.ts

import { CSSObject } from '@emotion/serialize';
import { MantineBreakpoint, MantineTheme } from '@mantine/core';
export declare function getStylesRef(refName: string): string;
export declare const getHelpers: (theme: MantineTheme) => {
    light: string;
    dark: string;
    rtl: string;
    ltr: string;
    notRtl: string;
    notLtr: string;
    ref: typeof getStylesRef;
    smallerThan: (breakpoint: MantineBreakpoint | number) => string;
    largerThan: (breakpoint: MantineBreakpoint | number) => string;
};
export type EmotionHelpers = ReturnType<typeof getHelpers>;
export declare function createStyles<Key extends string = string, Params = void, Input extends Record<Key, CSSObject> = Record<Key, CSSObject>>(input: ((theme: MantineTheme, params: Params, helpers: EmotionHelpers) => Input) | Input): (params: Params) => {
-    classes: Record<keyof Input, string>;
+    classes: { [key in keyof Input]: string};
    cx: (...args: any) => string;
    theme: MantineTheme;
};

```

### Self-service

tranhl:
Fixes #7491

### DIFF
--- a/package.json
+++ b/package.json
@@ -20,7 +20,7 @@
     "hc:exports": "tsx scripts/docs/hc-exports",
     "hc:deploy": "npm run hc:build && wrangler pages deploy apps/help.mantine.dev/out --branch master --project-name help-mantine-dev",
     "chmod": "sh scripts/utils/chmod.sh",
-    "setup": "npm run docs:docgen && tsx scripts/docs/generate-css-exports",
+    "setup": "npm run docs:docgen && npm run docs:sponsors && tsx scripts/docs/generate-css-exports",
     "plop": "tsx scripts/plop",
     "format-changelogs": "tsx scripts/docs/format-changelogs",
     "changelog": "tsx scripts/docs/changelog",
--- a/packages/@mantine/emotion/src/create-styles.ts
+++ b/packages/@mantine/emotion/src/create-styles.ts
@@ -44,7 +44,7 @@ export function createStyles<
     const classes = Object.keys(cssObject).reduce<Record<string, string>>((acc, key) => {
       acc[key] = css(cssObject[key]);
       return acc;
-    }, {}) as Record<keyof Input, string>;
+    }, {}) as { [key in keyof Input]: string };
 
     return { classes, cx, theme };
   };


package.json
```diff
@@ -20,7 +20,7 @@
     "hc:exports": "tsx scripts/docs/hc-exports",
     "hc:deploy": "npm run hc:build && wrangler pages deploy apps/help.mantine.dev/out --branch master --project-name help-mantine-dev",
     "chmod": "sh scripts/utils/chmod.sh",
-    "setup": "npm run docs:docgen && tsx scripts/docs/generate-css-exports",
+    "setup": "npm run docs:docgen && npm run docs:sponsors && tsx scripts/docs/generate-css-exports",
```
tranhl, line 23: @rtivital This change fixes CI by generating the `sponsors.json` file that was missing. It's a minor enough step that I think we can include it as part of the setup script. Happy to change it if you have a better suggestion.

rtivital:
Thanks!

Issue #6364, opened 2024-06-10
Stepper component error: DOMException: Failed to execute 'removeChild' on 'Node': The node to be removed is not a child of this node.


Uniiq:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.10.1

### What package has an issue?

@mantine/core

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

The App crashes when the user uses google translate and clicks on the "Next step" button in the Stepper component.

https://github.com/mantinedev/mantine/assets/30861773/c0ad0259-af71-4a6a-a44c-325073f49089

### Self-service

- [ ] I would be willing to implement a fix for this issue

rtivital:
Issue with React, not Mantine – https://github.com/facebook/react/issues/11538
You are welcome to research it and submit a PR with a fix if you are interested.

yhranik-vp:
I managed to fix this issue in the app code by manually assigning a step icon (number) to each step. The icon needs to be wrapped in a `<span>` element

```
<Stepper.Step icon={<span>1</span>}>
```

PR #7548, opened 2025-03-03
Fix Stepper removeChild react issue when using browser's translate feature

zieu:
Fixes [mantinedev/mantine#6364](https://github.com/mantinedev/mantine/issues/6364)


rtivital:
Thanks for proposing a fix! Unfortunately, it cannot be merged – the change is potentially breaking and it also introduces an unwanted element with inline styles (high specificity).

Issue #7562, opened 2025-03-06
Recharts ReferenceArea component not working with mantine charts


Algirdyz:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.17.1

### What package has an issue?

@mantine/charts

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

I am trying to use the Recharts feature called ReferenceArea by importing it from there:
`import {  ReferenceArea } from "recharts";`

I expect that if I add it as a child to any Mantine chart component it would render like in recharts.

I looked at the source code and I see that a child is added in the right place. I'm not sure what makes it not show up.

Here is an example in codesandbox with it being added to mantine vs recharts directly.

https://codesandbox.io/p/sandbox/mantine-react-template-forked-4kk89v?file=%2Fsrc%2FApp.tsx%3A14%2C15

### Self-service

- [ ] I would be willing to implement a fix for this issue

rtivital:
Please provide a sandbox with a minimal reproduction. The provided link is the Mantine template link.

Algirdyz:
Oh, right, sorry about that. Here it is:
https://codesandbox.io/p/sandbox/mantine-react-template-forked-4kk89v?file=%2Fsrc%2FApp.tsx%3A14%2C15

rtivital:
I was unable to find the source of the issue, help wanted with this issue.

rtivital:
You can find a fix for and a working example here – https://github.com/mantinedev/mantine/pull/7592/files
It is not planned to change anything in the library

PR #7590, opened 2025-03-14
add reference area to chart components

Ernesto-tha-great:
## Overview (fixes issue: #7562 )

This PR adds support for reference areas in the BarChart component, allowing users to highlight specific regions on the chart. Reference areas are useful for visualizing target zones, thresholds, or important data ranges.

## Changes

- Added `ReferenceArea` import from recharts
- Added referenceAreas prop to the BarChart component interface
- Implemented rendering of reference areas with customizable styling
- Added support for labels within reference areas
- Added story examples demonstrating reference area functionality
- Updated styles API to include reference area styling

## Examples

The PR includes two new story examples:
- ReferenceAreas - Demonstrates basic reference areas with both x-axis and y-axis based highlighting
- VerticalWithReferenceAreas - Shows how reference areas work with vertical orientation

## Testing
The implementation has been tested with:
- Different chart orientations (vertical/horizontal)
- Various label positions
- Different color schemes
- Multiple reference areas on the same chart

### DIFF
--- a/packages/@docs/styles-api/src/data/AreaChart.styles-api.ts
+++ b/packages/@docs/styles-api/src/data/AreaChart.styles-api.ts
@@ -21,6 +21,7 @@ export const AreaChartStylesApi: StylesApiData<AreaChartFactory> = {
     tooltipItemData: 'Tooltip item data',
     tooltipLabel: 'Label of the tooltip',
     referenceLine: 'Reference line',
+    referenceArea: 'Reference area',
     axisLabel: 'X and Y axis labels',
   },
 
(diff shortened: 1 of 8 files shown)


rtivital:
Thanks for submitting a PR, it is not a valid fix. The issue states that BarChart does not display ReferenceArea passed as children. It is not planned to add new props to the component.

Ernesto-tha-great:
> Thanks for submitting a PR, it is not a valid fix. The issue states that BarChart does not display ReferenceArea passed as children. It is not planned to add new props to the component.

Ah, noted. sending a new PR in a couple minutes. thank you for spotting this!

PR #7592, opened 2025-03-14
chore: create examples on how to use reference area in charts

Ernesto-tha-great:
Regarding (issue: #7562 )

After investigating, Reference area does work when imported as child components. I have added some examples to the storybook component to demonstrate how to use reference area(from rechart) with mantine. also left comments in the story file that will guide users on how to successfully use reference area in their charts.

Screenshot 2025-03-14 at 14 51 47


Ernesto-tha-great:
@rtivital 
check this out

rtivital:
Thanks!

PR #7599, opened 2025-03-17
fix: Scrollbars remain interactive with offsetScrollbars="present" when thumb is hidden

pfo-omicsstudio:
When using `offsetScrollbars="present"`, it was possible to hover over and reveal the scrollbar background even when the thumb was hidden. This behavior was incorrect, as the scrollbar should be entirely hidden when the thumb isn't visible.

This is how it looked when hovering over a vertical scrollbar that is hidden using `offsetScrollbars=present`:

image

With this PR, the scrollbar background no longer appears when the thumb is hidden, as expected.


rtivital:
Thanks!

PR #7601, opened 2025-03-18
[mantine.dev] Add `variantColorResolver`documentation

prosenjit-manna:
variantColorResolver documentation for custom HTML tag theming is sometimes needed for custom HTML auto-contrast features.


rtivital:
Thanks for submitting the PR, but it is not the intended use of the function.

PR #7602, opened 2025-03-18
fix: add pointer-events: none to disabled state

ItaiAxelrad:
The pagination controls can still be clicked when the disabled prop is present. Adding `pointer-events: none` to the CSS module fixes this bug

### DIFF
--- a/packages/@mantine/core/src/components/Pagination/Pagination.module.css
+++ b/packages/@mantine/core/src/components/Pagination/Pagination.module.css
@@ -27,6 +27,7 @@
   }
 
   &:where(:disabled, [data-disabled]) {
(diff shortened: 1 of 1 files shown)


rtivital:
- How can pagination controls be clicked when disabled prop is set?
- This is a potentially breaking change that breaks usage several scenarios. For example, when a tooltip added to disabled elements.

ItaiAxelrad:
@rtivital I am running into this issue when using controls as links [https://mantine.dev/core/pagination/#controls-as-links](https://mantine.dev/core/pagination/#controls-as-links). No worries, I can work around this if it causes breaking changes

PR #7603, opened 2025-03-18
fix: modal exports for node16

connorads:
I have a type checking issue in my repository, similar to the one described in https://github.com/mantinedev/mantine/issues/4991#issuecomment-1784254633 with `@mantine/modals/lib/index.d.mts`
Which was then fixed in https://github.com/mantinedev/mantine/pull/5697

image

I couldn't get the [paths workaround](https://github.com/mantinedev/mantine/issues/4991#issuecomment-1784456011) working, so to solve this locally I can
1. Delete`@mantine/modals/lib/index.d.mts` in `postinstall` script and it will rely on the non-module file `@mantine/modals/lib/index.d.ts`
2. Or, patch the file to add `.js` to exports
image

This PR makes the change to do `2.`
Tbh I don't really know what I'm doing, I suspect we don't want to do this because I believe it will also add `.js` to the non-module `@mantine/modals/lib/index.d.ts` file which we probs don't want.

Was hoping that @waweber might be able to help 🙏

### DIFF
--- a/packages/@mantine/modals/src/index.ts
+++ b/packages/@mantine/modals/src/index.ts
@@ -9,7 +9,7 @@ export {
   updateModal,
   updateContextModal,
   modals,
-} from './events';
+} from './events.js';
 
 export type { ModalsProviderProps } from './ModalsProvider';
 export type {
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

connorads:
No problem!

I was hoping that @waweber might validate the solution 😅 But if you're happy with it then great 👍

PR #7605, opened 2025-03-19
Update slider.mdx - args for slider label

ansonkao:
Updated "Documentation" tab to match the "Props" tab


rtivital:
The statement is incorrect and does not logically match the type of the label `ReactNode | ((value: number) => ReactNode)`.

PR #7609, opened 2025-03-20
fix: `AppShell[disabled]` does not cancel aside and footer margins

ofk:
[AppShell's disabled prop](https://mantine.dev/core/app-shell/#disabled-prop) doesn't disable aside and footer margins, which means that currently AppShell create margins at bottom and right (right margin is 316, bottom margin is 76):
スクリーンショット 2025-03-20 16 40 26

This PR will remove margins.


rtivital:
Thanks!

PR #7636, opened 2025-03-30
[@mantine/hook] useHotkeys: Correctly handle key matcher for FR keyboard layout

Odonno:
Fix issue with `useHotkeys` hook with a French ISO layout. `Q` and `A` as long as `Z` and `W` keys are interverted with a FR keyboard layout. Using both the `key` and the `code` property currently triggers both events (on `Z` and `W` for example). This fix will prevent using the `code` property that lead to this issue.

See more details here https://github.com/mantinedev/mantine/pull/7390


rtivital:
Thanks!

rtivital:
Sorry, had to revert, there were failing tests, you are welcome to submit a new PR with a better fix

PR #7644, opened 2025-04-02
[@mantine/core] better type support for size related theme overrides

### LINKED ISSUE #7642: Theme overrides are not fully recognized in code intelligence (auto completion)
### Dependencies check up

7.17.2

@mantine/core

Vite

All

### Describe the bug

This may not be an issue at all, as I'm not sure if it's intended.

Type definition for theme seems to assume `fontSizes`, `radius`, `spacing`, `lineHeights`, `shadows` and `breakpoints` would all share the same "size variants" https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/core/MantineProvider/theme.types.ts#L157-L162
That is obviously not always the case, user might need certain `radius` (a full circle roundness regardless height for example) that does not exist in other sizes.
Documentation suggest different approaches for that need, like having custom cssVariableResolver or something similar, like mentioned here https://arc.net/l/quote/rajuwokg
But that is not as intuitive as how you can "extend" `colors`, like explained here https://arc.net/l/quote/eijkmjxb

Yes I understand user can use `theme.other` to achieve something similar, but again, that's not as intuitive and will not be honored in auto completion. For example, if I have custom radius defined as `theme.radius.max` for full circle roundness, then I'd expect to have `max` be an option to `radius` besides `xs`, `sm`, etc..

I'm thinking maybe something could be done in theme types to expose interfaces for each "size", so user would have chance to augment module by adding their customized "size"s into theme definition, like how custom colors can be picked up via augmenting `MantineThemeColorsOverride`.

_No response_

### Possible fix

Explained above, basically, export [these types ](https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/core/MantineProvider/theme.types.ts#L164-L169) as interfaces so they can be augmented.

Refactoring their corresponding values types would be necessary, but something similar to how color override is handled should suffice.

### Self-service

mark-night:
This PR is to address issue #7642
Fixes #7642 

What's changed:
- Export `MantineThemeSizesOverride`, which can help to enhance types for size related theme overrides, like `MantineThemeColorsOverride` is to `theme.colors`.
- Updated docs on how to augment sizes override.
- Updated `size` prop types in some components (from referencing `MantineSize` to referencing more dedicated size, like `MantineFontSize`, `MantineSpacing`, etc..)
  - only those have clear clues that the prop expects a certain kind of size, were updated (e.g. `Text.size`)
  - some components use size prop for multiple purposes, like `Button.size`, and should be kept as they were (referencing `MantineSize`)

All changes should be backward compatible, not breaking anything (to the best of my knowledge).

### DIFF
--- a/apps/mantine.dev/src/pages/guides/typescript.mdx
+++ b/apps/mantine.dev/src/pages/guides/typescript.mdx
@@ -207,6 +207,34 @@ declare module '@mantine/core' {
 }
 ```
 
+You can also customize size related types for `theme.spacing`, `theme.radius`,
+`theme.breakpoints`, `theme.fontSizes`, `theme.lineHeights`, and `theme.shadows` similarly.
+
+To override `theme.spacing` and `theme.radius`
+
+```tsx
+import {
+  DefaultMantineSize,
+  MantineThemeSizesOverride,
+} from '@mantine/core';
+
+type ExtendedCustomSpacing =
+  | 'myCustomSpacing'
+  | 'anotherCustomSpacing'
+  | DefaultMantineSize;
+
+type ExtendedCustomRadius =
+  | 'myCustomRadius'
+  | DefaultMantineSize;
+
+declare module '@mantine/core' {
+  export interface MantineThemeSizesOverride {
+    spacing: Record<ExtendedCustomSpacing, string>;
+    radius: Record<ExtendedCustomRadius, string>;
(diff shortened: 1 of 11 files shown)


mark-night:
@rtivital let me know if there are anything I missed or need further attention.

Issue #7574, opened 2025-03-11
useIdle: State stays idle when horizontally scrolling


kevduc:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.17.1

### What package has an issue?

@mantine/hooks

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

Firefox

### Describe the bug

When using the `useIdle` example on mantine.dev, once the state is idle, if I horizontally scroll (I'm using the track pad on my mac), the state doesn't change to "not idle". It looks like it's using the defaults so it includes the `scroll` events, and I'd expect that horizontal scroll to be detected.

My user agent: `Mozilla/5.0 (Macintosh; Intel Mac OS X 10.15; rv:136.0) Gecko/20100101 Firefox/136.0`

https://github.com/user-attachments/assets/4ef938ff-495e-4a48-a341-c8caf6d33263

### If possible, include a link to a codesandbox with a minimal reproduction

https://mantine.dev/

### Self-service

- [ ] I would be willing to implement a fix for this issue

AliAkrem:
@rtivital, can you assign this to me?

rtivital:
You are welcome to submit a PR with a fix

PR #7669, opened 2025-04-09
fix: add 'wheel' event to use-idle hook

AliAkrem:
# changes 

- Add 'wheel' event to the default events of the use-idle hook.

# after fix 

[Screencast from 2025-04-09 15-35-02.webm](https://github.com/user-attachments/assets/79ebdb08-5f9a-47d3-8538-4d41af11b8af)

fix: #7574

### DIFF
--- a/apps/mantine.dev/src/components/MdxProvider/MdxTemplatesList/community-data.ts
+++ b/apps/mantine.dev/src/components/MdxProvider/MdxTemplatesList/community-data.ts
@@ -62,4 +62,10 @@ export const COMMUNITY_TEMPLATES_DATA: Template[] = [
     link: 'https://github.com/doiska/waku-mantine-starter',
     description: 'Vite + Waku + Mantine template',
   },
+  {
+    type: 'vite',
+    name: 'tanstack-start-cloudflare-template',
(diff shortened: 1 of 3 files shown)


rootboard:
@AliAkrem sorry for commenting here, but just wanted to mention, that [keypress](https://github.com/mantinedev/mantine/blob/75a52cb0859b4b2070547caef8d43286bfe618e4/packages/%40mantine/hooks/src/use-idle/use-idle.ts#L4) is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event)

When user clicks Tab it stays idle.

It would be great if you change it to `keydown` in this PR instead of creating a new one and it will be fixed all at once.

AliAkrem:
[keypress](https://github.com/mantinedev/mantine/blob/75a52cb0859b4b2070547caef8d43286bfe618e4/packages/%40mantine/hooks/src/use-idle/use-idle.ts#L4) is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/Element/keypress_event)
> 
> When user clicks Tab it stays idle.
> 
> It would be great if you change it to `keydown` in this PR instead of creating a new one and it will be fixed all at once.
 
@rtivital, can you check this?

rtivital:
Thanks!

Issue #7687, opened 2025-04-13
Add llms.txt


sandcastle:
Add `llms.txt` a file that helps LLMs and Agents like `Cursor` and `GitHub Colpilot Agent` use Mantine efficiently.

## Example

- [Daisy usage example](https://x.com/inkko44/status/1910700408957215148)
- [Daisy Docs on LLM.txt](https://daisyui.com/docs/editor/)
- [llms.txt](https://daisyui.com/llms.txt)

PR #7688, opened 2025-04-13
llms.txt v1

sandcastle:
See: https://github.com/mantinedev/mantine/issues/7687


rtivital:
This is a bad implementation. I do not think that llms.txt should be generated with llms.

PR #7708, opened 2025-04-18
Fix useHotKeys matching physical key codes when usePhysicalKeys unset

### LINKED ISSUE #7707: Bug in useHotkeys() hook: hotkeys match BOTH physical and layout keys
### Dependencies check up

7.17.4

@mantine/core

Vite

All

### Describe the bug

When using a keyboard layout other than QWERTY (in my case Colemak, which I'll use for my example) and not setting `usePhysicalKey`, the hotkey accepts BOTH the qwerty and colemak versions (that is, both the physical key despite `usePhysicalKey` not being set, and the virtual keyboard-layout specific key trigger the action). So if I set `ctrl-k`, in colemak `k` is physical `n`: when using colemak, I can trigger this action with both of the following physical keys: `ctrl-n` and `ctrl-k`. I expect only physical keys `ctrl-n` to trigger the action.

The bug is in this condition:

https://github.com/mantinedev/mantine/blob/87639c2b706884f9809fe1f68dc7a6cbe74abfe3/packages/%40mantine/hooks/src/use-hotkeys/parse-hotkey.ts#L65-L69

The issue is: `(!usePhysicalKeys && virtualCondition) || physicalCondition`, here `physicalCondition` will get checked both if `usePhysicalKeys` is set (which is what we want), but ALSO when `virtualCondition` is false, which is *not* what we want.

I've confirmed that this bug still exists in the `v8` branch too.

_No response_

### Possible fix

The fixed version is:

```typescript
  if (
    key &&
    ((!usePhysicalKeys && pressedKey.toLowerCase() === key.toLowerCase()) ||
      (usePhysicalKeys && pressedCode.replace('Key', '').toLowerCase() === key.toLowerCase()))
  ) {
```

Alternatively:
```typescript
key && (
    usePhysicalKeys ? 
        pressedCode.replace('Key', '').toLowerCase() === key.toLowerCase() :
        pressedKey.toLowerCase() === key.toLowerCase())
    )
``` 

Essentially, looking at `(!usePhysicalKeys && virtualCondition) || physicalCondition`, again, in the fixed version, we make sure to check the `physicalCondition` *only if* `usePhysicalKeys` is actually true, otherwise we only check the `virtualConditoin`.

### Self-service

danielytics:
Fixes #7707

### DIFF
--- a/packages/@mantine/hooks/src/use-hotkeys/parse-hotkey.test.ts
+++ b/packages/@mantine/hooks/src/use-hotkeys/parse-hotkey.test.ts
@@ -78,6 +78,32 @@ describe('@mantine/hooks/use-hot-key/parse-hotkey', () => {
         })
       )
     ).toBe(false);
+
+    expect(
+      getHotkeyMatcher(
+        'mod+k',
+        true
+      )(
+        new KeyboardEvent('keydown', {
+          ctrlKey: true,
+          code: 'KeyK',
+          key: 'n',
+        })
+      )
+    ).toBe(true);
+
+    expect(
+      getHotkeyMatcher(
+        'mod+k',
+        false
+      )(
+        new KeyboardEvent('keydown', {
+          ctrlKey: true,
+          code: 'KeyK',
+          key: 'n',
+        })
+      )
+    ).toBe(false);
   });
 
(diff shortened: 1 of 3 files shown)


rtivital:
Thanks!

PR #7302, opened 2024-12-24
fixed dayjs another local and calendar

baravak:
fixed #7276

PR #7712, opened 2025-04-19
Fix: Revert isSame usage due to Day.js calendar bug

baravak:
In [pull request #7302](https://github.com/mantinedev/mantine/pull/7302), two main changes were introduced:

### Support for non-Gregorian calendars:
We patched parts of the codebase to allow the library to work correctly with alternative calendars (e.g., Jalali). This part of the change was successful — we’ve integrated it into our own project and it's now functioning properly within a non-Gregorian calendar setup.

### Replacement of isSameMonth with dayjs().isSame(...):
The assumption was that Day.js’s isSame function would correctly handle month comparisons across all calendar systems. However, we later discovered that isSame does not work reliably for non-Gregorian calendars — it fails to compare months correctly. This PR reverts that change and restores the original internal isSameMonth function to maintain expected behavior.

This fix ensures consistent support for both Gregorian and non-Gregorian calendars and avoids breaking functionality due to incorrect assumptions about Day.js.


rtivital:
Thanks!

PR #7739, opened 2025-04-24
fix UseFieldInput.onValueChange is not being updated #7735

### LINKED ISSUE #7735: Bug: UseFieldInput.onValueChange is not being updated
Hello, good people! Congrats on making a great component library! 

I seem to have found a bug in the `useField` function. `onValueChange` is not listed as a dependency of `setValue` which results in the initial version of `onValueChange` being called even after it was updated.

The following is the line `onValueChange` (I think) should be listed as a dependency.

https://github.com/mantinedev/mantine/blob/87639c2b706884f9809fe1f68dc7a6cbe74abfe3/packages/@mantine/form/src/use-field.ts#L190

I hope reproduction is not needed as the problem is clearly seen in the code.

Cheers!

scato3:
Fixes #7735

I fixed the bug I discovered in #7735 by adding onValueChange to the dependency array.

### DIFF
--- a/packages/@mantine/form/src/use-field.ts
+++ b/packages/@mantine/form/src/use-field.ts
@@ -187,7 +187,7 @@ export function useField<
         _validate();
       }
     },
-    [error, clearErrorOnChange]
+    [error, clearErrorOnChange, onValueChange]
   );
 
   const reset = useCallback(() => {


rtivital:
Thanks!

PR #7740, opened 2025-04-24
Update inject-props.tsx

gfazioli:
Use a regex with the global flag to replace all occurrences when multiline is false


rtivital:
Thanks!

gfazioli:
🤝

PR #7745, opened 2025-04-25
[@mantine/hooks] useDebouncedCallback flushOnUnmount does not work #7684

### LINKED ISSUE #7684: useDebouncedCallback flushOnUnmount does not work
### Dependencies check up

7.17.4

@mantine/core

Next.js

None

### Describe the bug

flushOnUnmount appears not to work.
Looks like `lastCallback.flush` is always `noop ` in the clean up scope

https://codesandbox.io/p/devbox/mantine-react-template-forked-wmqkcg

was implemented from https://github.com/orgs/mantinedev/discussions/7271 so tagging @scamden in case can help :)

_No response_

### Possible fix

_No response_

### Self-service

scato3:
Fixes #7684 

Fixed the hook by adding isPendingRef to track pending callbacks. Now it only executes on unmount when truly necessary, passing the failing test. Added a Storybook demo to show the flushOnUnmount functionality in action

Issue
In the previous implementation of useDebouncedCallback, the flushOnUnmount option didn't work correctly in certain cases. When a component unmounted, even with flushOnUnmount: true, the debounced value wasn't properly applied.
The issue was that lastCallback.flush was initially set to noop and dynamically reassigned during each invocation, but the cleanup function in useEffect didn't always have access to the latest version of this function.
Fix
Modified the implementation to
Add isPendingRef to explicitly track whether there's a pending callback awaiting execution
Use lastArgsRef to store the latest arguments passed to the callback
Call handleCallback directly during unmount when appropriate, instead of relying on the potentially stale lastCallback.flush function

### DIFF
--- a/packages/@docs/demos/src/demos/hooks/Hooks.demos.story.tsx
+++ b/packages/@docs/demos/src/demos/hooks/Hooks.demos.story.tsx
@@ -303,6 +303,11 @@ export const Demo_useDebouncedCallbackUsage = {
   render: renderDemo(demos.useDebouncedCallbackUsage),
 };
 
+export const Demo_useDebouncedCallbackUnmount = {
+  name: '⭐ Demo: useDebouncedCallbackUnmount',
+  render: renderDemo(demos.useDebouncedCallbackUnmount),
+};
+
 export const Demo_useThrottledStateUsage = {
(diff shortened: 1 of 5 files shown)


rtivital:
Thanks!

PR #7746, opened 2025-04-25
Fix 7684 make a failing test for the theorized reason and fix it

### LINKED ISSUE #7684: useDebouncedCallback flushOnUnmount does not work

scamden:
Fix #7684

### DIFF
--- a/packages/@docs/demos/src/demos/hooks/Hooks.demos.story.tsx
+++ b/packages/@docs/demos/src/demos/hooks/Hooks.demos.story.tsx
@@ -303,6 +303,11 @@ export const Demo_useDebouncedCallbackUsage = {
   render: renderDemo(demos.useDebouncedCallbackUsage),
 };
 
+export const Demo_useDebouncedCallbackUnmount = {
+  name: '⭐ Demo: useDebouncedCallbackUnmount',
+  render: renderDemo(demos.useDebouncedCallbackUnmount),
+};
+
 export const Demo_useThrottledStateUsage = {
   name: '⭐ Demo: useThrottledStateUsage',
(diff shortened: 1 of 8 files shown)


scamden:
@rtivital i believe this is ready for your review. Pretty simple fix with a few additional tests. 

I verified the OP's codesandbox works with the fix: https://codesandbox.io/p/sandbox/mantine-react-template-forked-lqrk5c

scamden:
@rtivital i think the flush ref is a little bit simpler than the fix you merged (although I love the storybook demo) would it make sense to rebase against that but keep the fix I put up here? (There’s really no need to track pending and args etc since the flush was already idempotent)

rtivital:
Sure, you are welcome to change the implementation, I was planning to merge both PRs for tests and stories

scamden:
> Sure, you are welcome to change the implementation, I was planning to merge both PRs for tests and stories

Awesome, ya i think it's a good bit simpler now, thanks. Should be good to merge if checks pass.

scamden:
> Please fix errors in tests

Whoops merge conflict got weird, should be good now

rtivital:
Thanks!

PR #7749, opened 2025-04-25
[@mantine/charts] RadarChart: Add Tooltips and Dots

LilyStilson:
In the latest Mantine version `7.17.5` RadarChart does not have any tooltips. This PR implements them with a Storybook demo of custom tooltip.

Tooltip example:
image

Custom tooltip demo:
image

### DIFF
--- a/packages/@docs/demos/src/demos/charts/RadarChart/RadarChart.demo.customTooltip.tsx
+++ b/packages/@docs/demos/src/demos/charts/RadarChart/RadarChart.demo.customTooltip.tsx
@@ -1,7 +1,7 @@
 import { RadarChart } from '@mantine/charts';
+import { Paper, Text } from '@mantine/core';
 import { MantineDemo } from '@mantinex/demo';
 import { multiData, multiDataCode } from './_data';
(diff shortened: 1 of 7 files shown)


LilyStilson:
Added dots to match `LineChart`
image

LilyStilson:
Btw, I couldn't figure out how to put cursor under dots. Looks like its default behavior in Recharts...

rtivital:
Please fix the following:
1. By default, the tooltip should be disabled and dots should not be visible. It is needed to avoid unexpected breaking changes. Please also add a separate demo to showcase tooltip feature.
2. When dot is active, it is being overlaid by the axis, see screenshot. It must render above the axis the same way as an inactive dot.

Снимок экрана 2025-04-28 в 09 13 52

LilyStilson:
@rtivital Should dots be enabled automatically when `withTooltip` is specified?

LilyStilson:
Fixed cursor line being above dots, apparently the only way to do this - put tooltip above `{radars}`.

image

Also, added demo for default tooltip.

rtivital:
Thanks!

PR #7752, opened 2025-04-27
[@mantine/modals] ModalsProvider supports rendering multiple modals and add closeOnEscapeTopOnly prop

lizhong-magic-eden:
### This PR introduces two new features to the `ModalsProvider` component:

1. Support rendering multiple modals at once
    - Added the `showAllModals` prop to ModalsProvider, `false` by default.
    - When set to `true`, all opened modals are rendered in the order they were opened, with the most recently opened modal on top.

2. Fine-grained control over Escape key behavior
    - Added the `closeOnEscapeTopOnly` prop to ModalsProvider.
    - When set to `true` (default), only the topmost modal responds to the Escape key.
    - When set to `false`, pressing Escape will close all opened modals at once.

Additional changes:
- Refactored internal modal stack logic for clarity and maintainability.
- Added tests for the new `ModalsProvider` features and behaviors.
- Updated documentation to clearly explain the new props and behaviors.
- Added interactive demos for `showAllModals` to the documentation.

---

### Motivation

Our team encountered a real-world need for rendering and managing multiple modals simultaneously in our project. The current implementation of @mantine/modals only allows the most recently opened modal to be visible, which limits more advanced modal flows and user experiences.

Additionally, this feature request has been raised by other community members as well, as seen in [mantinedev discussions #4945. ](https://github.com/orgs/mantinedev/discussions/4945).

---

### Checklist
[x] Feature implementation
[x] Documentation updates
[x] Interactive demos

---

### Example

https://github.com/user-attachments/assets/ae3a90be-a388-44ea-968d-1474371976f1

### DIFF
--- a/apps/mantine.dev/src/pages/x/modals.mdx
+++ b/apps/mantine.dev/src/pages/x/modals.mdx
@@ -160,10 +160,33 @@ With `modals.open` function you can open a modal with any content:
 
 ## Multiple opened modals
 
-You can open multiple layers of modals. Every opened modal is added as first element in modals queue.
+You can open multiple modals one after another.
+Each modal you open is added to the modals queue, but by default, only the most recently opened modal is visible.
 To close all opened modals call `modals.closeAll()` function:
 
 <Demo data={ModalsDemos.multipleSteps} />
+
+To display all opened modals at once, set the `showAllModals` prop to `true` on `ModalsProvider`.
+All modals will be rendered in the order they were opened, with the most recently opened modal on top:
+
+<Demo data={ModalsDemos.showAllModals} />
+
+By default, when multiple modals are open, only the topmost modal responds to the Escape key press.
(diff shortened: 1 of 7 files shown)


lizhong-magic-eden:
I Removed the current property from the internal modal state in my commit [55e8732](https://github.com/mantinedev/mantine/pull/7752/commits/55e873254a8a74507b634b8e7a6d3ec0e3ec2ba6).

After careful analysis, I found that `current` was only potentially different from modals[modals.length - 1] in the `CLOSE` and `CLOSE_ALL` actions, but in both cases, the modals array becomes empty. Since the UI only renders a modal when modals.length > 0, the value of current is irrelevant in these cases. Therefore, `current` is redundant and can be safely removed.

rtivital:
Thanks for showing interest in contributing to Mantine! Unfortunately, this contribution cannot be accepted as it does not align with the project roadmap. Be sure to discuss changes that you are planning to implement with the maintainer before submitting a PR like it is described in the contributing guideline – https://mantine.dev/contribute/#contributing-workflow

PR #7757, opened 2025-04-29
center text within container

ayovev:


rtivital:
I think it looks better without centering.

PR #7785, opened 2025-05-06
[@mantine/dates] TimePicker: allow definition of value range for each column

### LINKED ISSUE #7784: TimePicker: Confusing value clamp when using arrow keys with steps enabled
### Dependencies check up

8.0.0

@mantine/dates

Vite

All

### Describe the bug

First of all, thanks a lot for implementing the TimePicker. Now I don't have to patch around the confusing behaviors of native time inputs in different browsers anymore and can just use TimePicker 🚀 

When testing out the TimePicker in the docs, I found it kind of confusing that the maximum value is not always related to the step size.
I expect that when I have a minute step size of 10 and my value is at 50, that the value stays at 50 or that the value jumps back to 00 and maybe even bumps/reduces the hour value, but the value jumps to the max value instread (59 in the docs). Especially after the value went to 59 it becomes kind of frustrating to use the mouse when the whole point of using the arrow keys was to not use the mouse.

Related demo in the docs: https://mantine.dev/dates/time-picker/#hoursminutesseconds-step

https://github.com/user-attachments/assets/329898ec-dddc-4709-a08c-18889b1cd7f8

_No response_

### Possible fix

Simplest fix for now would be to not step if the next step is outside of the defined value range.

### Self-service

jvllmr:
https://github.com/user-attachments/assets/41dcfcb3-4e12-4367-ab9c-f8db9f3e13f1

closes #7784

### DIFF
--- a/apps/mantine.dev/src/pages/dates/time-picker.mdx
+++ b/apps/mantine.dev/src/pages/dates/time-picker.mdx
@@ -65,11 +65,13 @@ By default, the dropdown is visible when one of the inputs is focused.
 
 <Demo data={TimePickerDemos.withDropdown} />
 
-## Hours/minutes/seconds step
+## Hours/minutes/seconds step with max/min Hours/minutes/seconds
 
 Use `hoursStep`, `minutesStep` and `secondsStep` props to control step for each input.
 These props are used to control value by which the input is incremented or decremented when
 up/down arrow keys are pressed and to generate corresponding values range in the dropdown.
+
+`maxHours`, `minHours`, `maxMinutes`, `minMinutes`, `maxSeconds` and `minSeconds` props can be used to control the value range of each column.
 
 <Demo data={TimePickerDemos.steps} />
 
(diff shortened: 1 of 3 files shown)


rtivital:
Thanks for submitting the PR! I do not think that this implementation is good. 
TimePicker has already min and max props that limit the value that can be picked by the user. Adding more props that kind of do the same thing but much wordy is a bad API.

jvllmr:
Yeah, I thought so too, but I couldn't think of another way to achieve this without messing with the default expected behavior of `min` and `max` of the time picker... I will think about it again.

mikkurogue:
Could use that `min` and `max` accept either a string or some identifier that also inherits it to be a part of the time portion?

some really badly thought of idea way too early with little coffee;
```ts
min?: string | { hours: number, minutes: number }
```

Means the initial value *could* stay the same, but that checking if the shape of `min` is `{[K: string]: number }` could allow for a slightly more flexible api, whilst also having the fallback.

Though, I dont think this is the best solution, it might help on the thought process

rtivital:
I do not see any difference between specifying `10:15` and `{ hours: 10, minutes: 15 }`.

rtivital:
I've resolved the issue on my side, the fix will be available in 8.0.1 release

PR #7788, opened 2025-05-06
[@mantine/dates] TimePicker: 00 in dropdown is reachable with arrow keys when it was outside of scroll

### LINKED ISSUE #7787: TimePicker: 00 not reachable with arrow keys in dropdown after it was outside of view
### Dependencies check up

8.0.0

@mantine/core

Vite

All

### Describe the bug

When using the dropdown with arrow keys, 00 cannot be reached inside dropdown after it was outside of view.

https://github.com/user-attachments/assets/eaaf9550-5822-413d-8524-c4b1976a0a4a

_No response_

### Possible fix

null value is not correctly checked.

### Self-service

jvllmr:
fixes #7787


rtivital:
Thanks!

PR #7795, opened 2025-05-07
[@mantine/core] Stepper: fix separator color

gdostie:
Hello 👋 I noticed that the Stepper separator had a different color between horizontal and vertical orientations. For instance you can see this on the docs website:

| Vertical | Horizontal |
| --- | --- |
|image|image|
| `mantine-color-gray-1` | `mantine-color-gray-2` |

While looking at the code I saw some dead code so I figured I would open a PR with a small cleanup. Here's a summary of what I did:

- `getStyles('separator')` is only every applied when `orientation === 'horizontal'` so the whole rule for `.separator &:where([data-orientation='vertical'])` can be deleted.
- defined `--stepper-outline-color` variable to make sure both orientations use the same color for the separator. 
  - I chose the one that was previously applied for horizontal orientation (gray 2), let me know if that works for you. 
  - I also apply this color to the non completed step background since it was the same for vertical, but I am unsure if it is ok (looks okay to me, but can revert if needed).
- defined a `--stepper-outline-thickness` variable for ease of customization

Here is a before and after
| Before | After |
| --- | --- | 
|image|image|
|image|image|
|image|image|
|image|image|

### DIFF
--- a/packages/@mantine/core/src/components/Stepper/Stepper.module.css
+++ b/packages/@mantine/core/src/components/Stepper/Stepper.module.css
@@ -11,6 +11,15 @@
   --stepper-spacing: var(--mantine-spacing-md);
   --stepper-radius: 1000px;
   --stepper-fz: var(--mantine-font-size-md);
+  --stepper-outline-thickness: 2px;
+
+  @mixin light {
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #7807, opened 2025-05-09
[mantine.dev] Make the demo and actual code consistent

kawarimidoll:
I noticed that the code attached as a demo on this page ( https://github.com/mantinedev/mantine/blob/39360296221bced9adaf098135ff676f214c5022/packages/%40docs/demos/src/demos/core/MultiSelect/MultiSelect.demo.renderOption.tsx ) differs from the actual working implementation. To avoid confusion, the demo should reflect the exact same code as the one that is actually being executed.


rtivital:
Thanks!

PR #7810, opened 2025-05-09
[@mantine/core] Fix: add boolean to data attribute types

KurtGokhan:
Allows data-* attributes to be boolean type.

For us, it's common to write data attributes like:

```tsx
  <Button
    data-disabled
    ...
  />
```

which just serializes like `data-disabled="true"`. I think there is no need to prevent that.


rtivital:
Thanks!

Issue #6247, opened 2024-05-16
Passing step via timeInputProps has no effect


souhagaa:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.9.1

### What package has an issue?

@mantine/dates

### What framework do you use?

create-react-app (CRA)

### In which browsers you can reproduce the issue?

All

### Describe the bug

When trying to pass `step` prop via `timeInputProps` to `DateTimePicker` but it is not working and if I change it manually in the HTML it works. 
It seems like Mantine `TimeInput` component ignores this prop here https://github.com/mantinedev/mantine/blob/0976f8aa5e0bc214a6bfd76fd773bd9742ff71bb/packages/%40mantine/dates/src/components/TimeInput/TimeInput.tsx#L124

a possible solution would be  to change that line to `step={props.step || withSeconds ? 1 : 60}`

### If possible, include a link to a codesandbox with a minimal reproduction

 https://rb.gy/jmrgpd

### Self-service

- [ ] I would be willing to implement a fix for this issue

rtivital:
Fixed in [7.9.2](https://github.com/mantinedev/mantine/releases/tag/7.9.2)

tianhuil:
Hi @rtivital: I'm not quite sure how https://github.com/mantinedev/mantine/pull/6204 helps solve the problem.  Ideally, we would have
```ts
step={props.step || withSeconds ? 1 : 60}
```
as suggested by @souhagaa and I would be happy to implement this and submit a PR.

jannikbuschke:
Also running into this issue

PR #7811, opened 2025-05-09
Allow to set step in TimeInput component

jrozbicki:
<input type="time"> allows setting step to increment by defined amount of seconds.

Current order of props basically disregards any `step` setting passed to TimeInput `<TimeInput step={60 * 10} />` (10min increments example)

Relates to https://github.com/mantinedev/mantine/issues/6247


rtivital:
Thanks!

PR #7813, opened 2025-05-09
Deep merge default props

me-gusta-v:


rtivital:
1. It is intended behavior that should not be changed.
2. Mantine must not depend on lodash.

Issue #7801, opened 2025-05-08
Select.onSearchChange is triggered every time the data prop is modified


kaareal:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

8.0.0

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

In the Select component the onSearchChange is triggered every time the data attribute is modified. Causing the below code to trigger a `Maximum update depth exceeded.` Error. 

```
import { useState } from "react";
import { Select } from "@mantine/core";

export default function App() {
  const [data, setData] = useState([]);

  return (
    <div
      className="App"
      style={{
        padding: "24px",
      }}
    >
      <h1>Hello CodeSandbox</h1>
      <Select
        searchable
        data={data}
        onChange={(value) => {
          console.log("onChange", value);
        }}
        onSearchChange={(value) => {
          console.log("onSearchChange", value);
          setData([...value]);
        }}
      />
    </div>
  );
}
```

I’m trying to build an async autocomplete using the Select component, where onSearchChange and searchValue let the user type a query, trigger a fetch for matching results, and display those results as selectable options.

### If possible, include a link to a codesandbox with a minimal reproduction

https://codesandbox.io/p/sandbox/wild-hill-j8lg4c

### Self-service

- [x] I would be willing to implement a fix for this issue

rtivital:
You are welcome to submit a PR with a fix

meslzy:
Thank you for v8 its awesome @rtivital
but this is not working as well. 

```ts
const [searchValue, setSearchValue] = useState("");

<Select
  searchValue={searchValue}
  data={[]}
  searchable
  onSearchChange={setSearchValue}
/>
```

kaareal:
@rtivital unsure if I got this right https://github.com/mantinedev/mantine/pull/7814 let me know what you think.

PR #7814, opened 2025-05-09
Select checking for searchControlled

kaareal:
If the value is not controlled, and you mutate the data prop, the onSearchChange will call causing problems.
From what I can tell the issue was introduced with https://github.com/mantinedev/mantine/commit/b22de41c6dfe68720ef031e3cd98b3efa2de0ab9 

Relates to https://github.com/mantinedev/mantine/issues/7801


rtivital:
Thanks!

PR #7821, opened 2025-05-10
[@mantine/core] stop scaling explicit rem units in rem function

jvdsande:
As discussed on Discord, this MR changes the logic of the `rem` function to only scale px values, not rem values.

This impacts a few test files, I tried to change things to relevant values for what each test was testing.


rtivital:
Thanks!

PR #7841, opened 2025-05-14
Allow useDebouncedCallback to be fired immediately

jonjomckay:
This follows the same pattern as `useDebouncedState` to allow the caller to choose whether the callback should be fired immediately, or whether to wait until the delay duration passes.


rtivital:
Thanks!

PR #7869, opened 2025-05-21
Add `defaultValue` to `form.getInputProps()` docs

gentunian:
When form mode is `"uncontrolled"` no `value` is present. Instead, `defaultValue` is being used.


rtivital:
Thanks!

PR #7893, opened 2025-05-26
Refactor Date Utilities for Enhanced Type Precision to avoid awkward Null Checks

YukiKoNeko:
I've refactored `toDateString` and `toDateTimeString` for enhanced type precision using generics.

This approach uses the conditional helper type `ExactOptionalReturn<T, MainType>` (defined as `T extends undefined ? undefined : T extends null ? null : MainType`). This helper maps inputs as follows:
  - `undefined` inputs result in an `undefined` return type.
  - `null` inputs result in a `null` return type.
  - Other valid inputs result in `DateStringValue` or `DateTimeStringValue`.

Runtime behavior is preserved: an empty string (`''`) input still returns `''`, which is type-compatible with `DateStringValue` (as `string`).

The primary benefit is more accurate type inference. For instance, an input of type `Date | null` now correctly infers a return type of `DateStringValue | null`, rather than `DateStringValue | null | undefined`. This significantly reduces the need for unnecessary null or undefined checks in consuming code, directly addressing the goal of avoiding awkward null checks and leading to cleaner, more type-safe development.


YukiKoNeko:
@rtivital I saw you added the `?` tag to this.

TL;DR: when calling your `toDateTimeString()` function with a defined (non-null, non-undefined) input, TypeScript would still suggest that the output might both either null or undefined, when it really can't be, leading to having to to another useless null & undefined check.

Now: When input might be undefined/null, output might be undefined/null. When input defined, output defined.

rtivital:
Thanks!

PR #7906, opened 2025-05-27
Fix toRgba function

ylvaselling:
I encountered a behavior which I believe is unintentional with the `toRgba` function. The behavior was that if you input a rgba string to the function with an alpha value of 0, the alpha value in the RGBA object would be `1`. 

Before fix: 
```
toRgba("rgba(24, 21, 42, 0)") -> { R: 24, G: 21, B: 42, A: 1 }
```

After fix: 
```
toRgba("rgba(24, 21, 42, 0)") -> { R: 24, G: 21, B: 42, A: 0 }
```
The fix is simply to create a check for if alpha is `undefined` rather than falsy.


rtivital:
Thanks!

PR #7907, opened 2025-05-27
Fix ordering of vendor-prefixed backdrop-filter rules in Overlay component 

### LINKED ISSUE #7898: incorrect ordering of vendor-prefixed CSS rule backdrop-filter in Overlay component
### Dependencies check up

8.0.1

@mantine/core

Next.js

Chrome

### Describe the bug

Using Mantine 8.0.1 in a Next 15 project using tailwind 4 and postcss

Using the drawer component with overlayProps > blur

There is a known issue with lightningcss (used by postcss) where a non-prefixed rule is ignored if a vendor-prefixed rule appears after it in the style declaration. This manifests as a blur filter not being applied in Chrome because the `-webkit-backdrop-filter` prefixed rule is used instead of the expected `backdrop-filter` that Chrome accepts.

The correct ordering of CSS rules should have the vendor-prefixed versions first.

I appreciate that this isn't really a Mantine issue, but it's a quick fix and aligns to the conventional ordering of vendor-prefixed css rules.

References:
https://github.com/tailwindlabs/tailwindcss/discussions/16443
https://github.com/parcel-bundler/lightningcss/issues/785
https://github.com/parcel-bundler/lightningcss/issues/695

_No response_

### Possible fix

Update the order of the backdrop-filter rules so the non-prefixed version is last.

I'm not hugely familiar with the codebase, but this seems to be where the rules are applied:

https://github.com/mantinedev/mantine/compare/master...VirtuoUK:mantine:master

### Self-service

VirtuoUK:
Fixes #7898 

Where PostCSS is used in a project, the lightningcss dependency will exclude the non-vendor-prefixed rule if it appears _before_ any vendor prefixed rules, this PR simply reorders the declarations so that the non-prefixed variant appears last.

This allows overlays with blurred backgrounds to render correctly on browsers that don't require vendor prefixing for the `backdrop-filter` CSS rule - e.g. Chrome

### DIFF
--- a/apps/mantine.dev/src/components/HomePage/HomePageComponents/HomePageComponents.tsx
+++ b/apps/mantine.dev/src/components/HomePage/HomePageComponents/HomePageComponents.tsx
@@ -15,7 +15,7 @@ export function HomePageComponents() {
 
         <HomePageDescription className={classes.description}>
           Build your next app faster with high-quality, well-tested components. Mantine includes
-          everything you need to create complex we applications with ease: custom select, date
+          everything you need to create complex web applications with ease: custom select, date
           pickers, notifications, modals, and more.
         </HomePageDescription>
 
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mantine-a91763c0e2",
-  "version": "8.0.1",
+  "version": "8.0.2",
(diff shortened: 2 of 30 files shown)


rtivital:
Thanks!

PR #7911, opened 2025-05-28
fix(Collapse): Fix inert attribute handling for React 18 compatibility

jonavila:
## What
Fixes incorrect `inert` attribute handling in the Collapse component for React 18 compatibility.

## Problem
The current implementation has backwards logic for React version detection and always sets the `inert` attribute as a boolean, which causes warnings in React 18:

`Warning: Received false for a non-boolean attribute inert.`

## Solution
- **React 18**: Use empty string `''` for `inert=true`, `undefined` for `inert=false`
- **React 19+**: Use boolean values (`true`/`false`)
- Fixed the backwards conditional logic that was checking `!React.version.startsWith('18')`

## Changes
- Updated `getCollapseProps` function in `use-collapse.ts` to properly handle `inert` attribute based on React version
- Added proper TypeScript handling with `any` type for props object to allow dynamic property assignment

## Testing
- All existing Collapse component tests pass
- Verified TypeScript compilation succeeds for the core component
- Manual testing confirms proper `inert` attribute behavior in both React versions

### DIFF
--- a/packages/@mantine/core/src/components/Collapse/use-collapse.ts
+++ b/packages/@mantine/core/src/components/Collapse/use-collapse.ts
@@ -107,16 +107,23 @@ export function useCollapse({
 
   function getCollapseProps({ style = {}, refKey = 'ref', ...rest }: GetCollapseProps = {}) {
     const theirRef: any = rest[refKey];
-    const props = {
+    const props: any = {
       'aria-hidden': !opened,
-      inert: !opened,
       ...rest,
       [refKey]: mergeRefs(el, theirRef),
       onTransitionEnd: handleTransitionEnd,
       style: { boxSizing: 'border-box', ...style, ...styles },
     };
 
-    if (!React.version.startsWith('18')) {
+    // Handle inert attribute based on React version
+    if (React.version.startsWith('18')) {
+      // React 18: Use empty string for inert=true, undefined for inert=false
+      if (!opened) {
+        props.inert = '';
+      }
+      // Don't set inert property when opened=true (leave it undefined)
+    } else {
+      // React 19+: Use boolean values
       props.inert = !opened;
     }
 


nicolasff:
I can confirmed this resolved an issue for me with the `Collapse` component. On React 18, upgrading from Mantine 7.17.1 to 8.0.2 caused the following to be logged:
```none
Warning: Received `true` for a non-boolean attribute `inert`. If you want to write it to the DOM,
pass a string instead: inert="true" or inert={value.toString()}.
```

### Integrating this patch

Here's how I integrated this patch while I wait for the fix to be merged and integrated in a new release.

I updated my local `node_modules/@mantine/core/esm/components/Collapse/use-collapse.mjs` to make the same changes as the ones from this PR. There are only two changes to port: removing the line `inert: !opened,` from the `props` object declaration, and updating the `if` block.

Note: this PR updates a TypeScript file, but when you install Mantine as a dependency you'll get `use-collapse.mjs` (or maybe `use-collapse.js` with CommonJS? I'm not sure). This is why we have to re-apply the patch manually.

After making these changes, I installed `patch-package` and pointed it to `@mantine/core`:
```shell
npm install --save-dev patch-package
npx patch-package @mantine/core
```
This created `patches/@mantine+core+8.0.2.patch`.

<details>
<summary>Here's the full `.mjs` patch, if anyone needs it</summary>

```diff
diff --git a/node_modules/@mantine/core/esm/components/Collapse/use-collapse.mjs b/node_modules/@mantine/core/esm/components/Collapse/use-collapse.mjs
index 92a756a..7f1b6c8 100644
--- a/node_modules/@mantine/core/esm/components/Collapse/use-collapse.mjs
+++ b/node_modules/@mantine/core/esm/components/Collapse/use-collapse.mjs
@@ -80,13 +80,20 @@ function useCollapse({
     const theirRef = rest[refKey];
     const props = {
       "aria-hidden": !opened,
-      inert: !opened,
       ...rest,
       [refKey]: mergeRefs(el, theirRef),
       onTransitionEnd: handleTransitionEnd,
       style: { boxSizing: "border-box", ...style, ...styles }
     };
-    if (!React.version.startsWith("18")) {
+    // Handle inert attribute based on React version
+    if (React.version.startsWith('18')) {
+      // React 18: Use empty string for inert=true, undefined for inert=false
+      if (!opened) {
+        props.inert = '';
+      }
+      // Don't set inert property when opened=true (leave it undefined)
+    } else {
+      // React 19+: Use boolean values
       props.inert = !opened;
     }
     return props;
```
</details>

Finally, I added a post-install step to my `package.json`:
```json
"postinstall": "patch-package",
```

To validate this setup, I removed `node_modules` entirely, then re-installed dependencies with `npm install` and observed the patch being applied:

```none
patch-package 8.0.0
Applying patches...
@mantine/core@8.0.2 ✔
```

After rebuilding, the warning mentioned above was no longer being logged. I also dumped `node_modules/@mantine/core/esm/components/Collapse/use-collapse.mjs` and verified that the local patch was being applied as expected.

Caveat: I am not associated with the Mantine project and cannot comment on the change itself. It just took me a while to track down what was causing a new warning after upgrading to 8.0, so I figured documenting this process could help someone in a similar situation.

And thanks @jonavila for the fix!

rtivital:
Thanks!

PR #7914, opened 2025-05-29
Cancel idle timer when hook is being unmounted in use-idle.ts

deedw:
Idle timer needs cancelling when hook is being unmounted


rtivital:
Thanks!

PR #7929, opened 2025-06-03
[@mantine/cone] Fix focus loss on tab navigation when Combobox has withinPortal=false

wancup:
When `withPortal` is set to false on components using Combobox (e.g., Select), tab navigation causes focus to be lost. This happens because, during tabbing, focus moves to the scroll area of the Combobox options only for a moment, but the options immediately close, resulting in focus loss.
In my opinion, since Combobox already supports keyboard navigation with arrow keys, tab focus on the options is unnecessary.
To address this, the default `viewportProps` for options now includes `tabIndex: -1`, preventing the scroll area from receiving focus during tab navigation.

## Before

1. Set `comboboxProps={{ withinPortal: false }}` on the Select component.
2. Focus the Select component, then press the Tab key.
3. Ideally, focus should move to the Button next to the Select, but instead, focus is lost.

https://github.com/user-attachments/assets/e3b3d6f0-e358-4deb-9114-159a6f7472c2

## After

https://github.com/user-attachments/assets/063868be-3d55-4d2a-b6cb-457a9e1f8d83

### DIFF
--- a/packages/@mantine/core/src/components/Combobox/OptionsDropdown/OptionsDropdown.tsx
+++ b/packages/@mantine/core/src/components/Combobox/OptionsDropdown/OptionsDropdown.tsx
@@ -1,4 +1,5 @@
 import cx from 'clsx';
+import { deepMerge } from '../../../core';
 import { CheckIcon } from '../../Checkbox';
 import { ScrollArea, ScrollAreaProps } from '../../ScrollArea/ScrollArea';
 import { Combobox } from '../Combobox';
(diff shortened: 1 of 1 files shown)


wancup:
Oh... My understanding of listbox behavior was incorrect. According to MDN:

> When a list is tabbed to, the first item in the list will be selected if nothing else already is.
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Roles/listbox_role

So, it seems that the expected behavior is for the options not to close when pressing Tab. At the very least, the approach in this pull request is incorrect, so I will close it.

PR #7934, opened 2025-06-05
[@mantine/cores] [Tree] fix: Remove handle hovering cause rerender whole tree

trananhvan1102vn:


rtivital:
It is not a valid fix:
1. this just removes hover functionality 
2. It is a breaking change – `data-hovered` attribute can be used by the library consumers

PR #7940, opened 2025-06-06
[@Mantine/core] Fix: detached DOM causing memory leak in Tree component

### LINKED ISSUE #7677: Tree component retain references to detached DOM,  leading to memory leak
### Dependencies check up

7.17.4

@mantine/core

Vite

Chrome

### Describe the bug

- Open https://mantine.dev/core/tree/
- Open Devtools -> Memory
  - Select Profiling "Detached Elements"
- Expand/Collapse any node
- Make snapshot

Image

In my case, when tree contains 500 nodes, it leads to +1.5Mb memory leak after each expand/collapse

_No response_

### Possible fix

_No response_

### Self-service

sungpaks:
fixed #7677 

A few changes in `TreeNode` nested tree nodes, it worked
as-is : `<ul role="group" ... >`
to-be: `<Box component="ul" role="group">`

https://github.com/user-attachments/assets/78853a04-6c7c-4f51-a8c2-d234dd94a507

But I'm still not sure exactly *why* it works, and what differences between them making these results?
If anyone has any insight on this, I'd love to hear it.

### DIFF
--- a/packages/@mantine/core/src/components/Tree/TreeNode.tsx
+++ b/packages/@mantine/core/src/components/Tree/TreeNode.tsx
@@ -1,5 +1,5 @@
 import { useRef } from 'react';
-import { findElementAncestor, GetStylesApi } from '../../core';
+import { Box, findElementAncestor, GetStylesApi } from '../../core';
 import type { RenderNode, TreeFactory, TreeNodeData } from './Tree';
 import type { TreeController } from './use-tree';
 
@@ -196,9 +196,9 @@ export function TreeNode({
       )}
 
       {controller.expandedState[node.value] && nested.length > 0 && (
-        <ul role="group" {...getStyles('subtree')} data-level={level}>
+        <Box component="ul" role="group" {...getStyles('subtree')} data-level={level}>
           {nested}
-        </ul>
+        </Box>
       )}
     </li>
   );


sungpaks:
I realized that that the reason why *working with `Box`* does not caused detached DOM elements: 
The point was *"wrapped in seperated component"*, I think

ex:

```tsx
export function App() {
  const [expanded, setExpanded] = useState(false);

  return (
    <>
      <button
        onClick={() => {
          setExpanded((prev) => !prev);
        }}
      >
        {expanded ? "close" : "open"}
      </button>
      {expanded && <div>fooo</div>}
    </>
  );
}
```

In this case, when we click the button twice (`open` -> `close`), it remains detached DOM
But when we make this `<div>fooo</div>` as seperated component:

```tsx
function Foo() {
  return <div>fooo</div>
}

export function App() {
  return (
    <>
      <button
        onClick={() => {
          setExpanded((prev) => !prev);
        }}
      >
        {expanded ? "close" : "open"}
      </button>
      {expanded && <Foo />}
    </>
  );
}
```

In this case, no detached DOM remains

sungpaks:
Then there seem to be two ways to solve this problem in `Tree` component: 
1. using `<Box>`, just like what I did in this PR
2. break the old `<ul role="group" ...>` into a seperated component (e.g. with the name `TreeSubItem` ?)

Which one would be better?

rtivital:
Thanks!

sungpaks:
@rtivital I'm sorry but.. I think this work was not necessary.
I realized that, essentially, it was issue that **occurred when React DevTools was on.**
Should this PR be reverted?

PR #7965, opened 2025-06-11
feat(hooks): add cancel functionality to useDebouncedCallback

scamden:
- Add cancel method alongside existing flush method

- Cancel clears pending timeouts and resets leading state without executing callback

- Maintains backward compatibility with existing API

- Add comprehensive test coverage (8 new tests) covering:

  - Basic cancellation preventing execution

  - Multiple cancel calls safety

  - Cancel and re-call behavior

  - Interaction with already executed callbacks

  - Leading mode reset behavior

  - Cancel/flush interaction patterns

Resolves the need to cancel pending debounced executions in scenarios like component unmounting or input direction changes.


packages/@mantine/hooks/src/use-debounced-callback/use-debounced-callback.ts
```diff
@@ -21,12 +21,11 @@ export function useDebouncedCallback<T extends (...args: any[]) => any>(
...
   const handleCallback = useCallbackRef(callback);
   const debounceTimerRef = useRef(0);
-  const flushRef = useRef(() => {});
   const leadingRef = useRef(leading);
 
-  const lastCallback = Object.assign(
-    useCallback(
-      (...args: Parameters<T>) => {
+  const lastCallback = 
+    useMemo(() =>
```
scamden: using useMemo instead of useCallback lets us treat the callback itself as the ref since it's not created every render. Ties the leading and flush state to the callback itself, which means things will reset when options like leading, delay, etc change (which i think is desired)

scamden:
@rtivital think this is ready for your review if you have a moment. Adds cancel functionality and cleans up the code a bit to track state in fewer places as well.

packages/@mantine/hooks/src/use-debounced-callback/use-debounced-callback.test.ts
```diff
@@ -81,6 +91,29 @@ describe('@mantine/hooks/use-debounced-callback', () => {
     expect(callback).toHaveBeenCalledWith(3);
   });
 
+  // Note: this might not be desired but this is what happens
```
scamden, line 94: realized this was the behavior for both of these cases so i thought I'd document it with a test. We could consider doing something differently if we don't like this but it's non-regression

packages/@mantine/hooks/src/use-debounced-callback/use-debounced-callback.test.ts
```diff
@@ -170,4 +203,135 @@ describe('@mantine/hooks/use-debounced-callback', () => {
...
+    );
+    rerender({ leading: false });
+    result.current(1);
+    result.current(2);
+    expect(callback).not.toHaveBeenCalled();
+    jest.advanceTimersByTime(100);
+    expect(callback).toHaveBeenCalledTimes(1);
+  });
+
+  it('does call again on leading edge if options change and it was already called before the change', () => {
```
scamden, line 221: the was behaving the opposite way before. If i change any option i don't get a fresh leading edge. My instinct is that it should reset when the options reset but again open to changing it

rtivital:
Thanks!

PR #7972, opened 2025-06-13
Add `onOverflowChange` callback to `ScrollArea.Autosize`

pfo-omicsstudio:
This PR adds a new `onOverflowChange?: (overflowing: boolean) => void` prop to ScrollArea.Autosize.

It fires when content starts or stops exceeding max-height, enabling consumers to react when the container becomes scrollable.

https://github.com/user-attachments/assets/e43d5e79-934f-4907-9326-40c816c5bb46

### DIFF
--- a/packages/@mantine/core/src/components/ScrollArea/ScrollArea.tsx
+++ b/packages/@mantine/core/src/components/ScrollArea/ScrollArea.tsx
@@ -290,8 +290,8 @@ export const ScrollAreaAutosize = factory<ScrollAreaAutosizeFactory>((props, ref
   // Overflow detection (Autosize-only)
   const [resizeObserverRef, rect] = useResizeObserver<HTMLDivElement>();
   const combinedViewportRef = useMergeRefs([viewportRef, resizeObserverRef]);
-  const didMountRef = useRef(false);
-  const lastOverflowRef = useRef(false);
+  const [overflowing, setOverflowing] = useState(false);
+  const didMount = useRef(false);
 
   useEffect(() => {
     if (!onOverflowChange) {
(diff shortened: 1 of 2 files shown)


packages/@mantine/core/src/components/ScrollArea/ScrollArea.tsx
```diff
@@ -272,8 +282,45 @@ export const ScrollAreaAutosize = factory<ScrollAreaFactory>((props, ref) => {
...
+  // Overflow detection (Autosize-only)
+  const viewportObserverRef = useRef<HTMLDivElement>(null);
+  const combinedViewportRef = useMergeRefs([viewportRef, viewportObserverRef]);
+
+  const [overflowing, setOverflowing] = useState(false);
+  const didMount = useRef(false);
+
+  useEffect(() => {
+    const el = viewportObserverRef.current;
+    if (!el) {
```
rtivital, line 303: This logic might have a significant impact on performance, add early return if onOverflowChange is not a function

pfo-omicsstudio: Done, I also refactored it to use the `useResizeObserver` hook.

packages/@mantine/core/src/components/ScrollArea/ScrollArea.tsx
```diff
@@ -272,8 +283,40 @@ export const ScrollAreaAutosize = factory<ScrollAreaFactory>((props, ref) => {
...
     onBottomReached,
     onTopReached,
+    onOverflowChange,
     ...others
-  } = useProps('ScrollAreaAutosize', defaultProps, props);
+  } = useProps('ScrollAreaAutosize', defaultProps, props as ScrollAreaAutosizeProps);
+
+  // Overflow detection (Autosize-only)
+  const [resizeObserverRef, rect] = useResizeObserver<HTMLDivElement>();
+  const combinedViewportRef = useMergeRefs([viewportRef, resizeObserverRef]);
```
rtivital: Not a good solution: adds significant overhead with resize observer tracking when `onOverflowChange` is not passed. ResizeObserver must only listen for changes when it is required.

pfo-omicsstudio: Good point - I reverted the last two commits and added the early return guard.

rtivital:
Thanks!

Issue #7954, opened 2025-06-10
`onToggle` function's type inference about `&&` operator


JeonJaewon:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

8.1.0

### What package has an issue?

@mantine/core

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

Not applicable – issue is not related to the browser

### Describe the bug

In `@mantine/core/src/components/Popover/use-popover.ts`,

```ts
const onToggle = () => !options.disabled && setOpened(!_opened);
// const onToggle: () => false | void
```
The `onToggle` from `usePopover` is using the && operator, which caused TypeScript to infer its return type as () => false | void. I guess that this was not the intended behavior. This type inference could lead to a little confusion.

### Possible fix

```ts
const onToggle = () => {
  if (!options.disabled) {
    setOpened(!_opened);
  }
};
// const onToggle: () => void
```
Replace the previous usage of the && operator with a simple if statement

### Self-service

- [x] I would be willing to implement a fix for this issue

PR #7978, opened 2025-06-15
[@mantine/core] Improve type inference of onToggle

JeonJaewon:
Fixes https://github.com/mantinedev/mantine/issues/7954

This pull request includes a really simple change to the `usePopover` function in the `Popover` component. 
The change improves type inference of the `onToggle` function to be properly inferred as `() => void`.


rtivital:
Thanks!

PR #7983, opened 2025-06-17
[@mantine/core] optimize Popover autoUpdate for keepMounted elements

jonavila:
- Conditionally apply whileElementsMounted based on keepMounted prop
- Use manual autoUpdate control when keepMounted=true to avoid unnecessary updates on CSS-hidden elements
- Add keepMounted prop to usePopover options interface

Fixes performance issue where autoUpdate ran continuously on hidden popover dropdowns

### DIFF
--- a/packages/@mantine/core/src/components/Popover/Popover.tsx
+++ b/packages/@mantine/core/src/components/Popover/Popover.tsx
@@ -300,6 +300,7 @@ export function Popover(_props: PopoverProps) {
     positionRef,
     disabled,
     preventPositionChangeWhenVisible,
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

liber-gong:
Issue still in Chrome device simulator (choose iphone XR), infinite re-render and page get stuck. @jonavila @rtivital

rbozan:
> Issue still in Chrome device simulator (choose iphone XR), infinite re-render and page get stuck. @jonavila @rtivital

Do you know how to reproduce it?

Issue #7836, opened 2025-05-13
Can't access Select menu or its options by role in my Vitest tests (7.17.0 to 8.0.0)


chumba-wamba:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

8.0.0

### What package has an issue?

@mantine/core

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

None

### Describe the bug

Hello,
My repository wraps Mantine components with our custom styles. We have some Vitest unit tests that check the Select component's behavior by clicking on the Input field and selecting the options. These tests access the dropdown items by the role 'option' and the container by the role 'listbox'.

These tests run successfully in Mantine v7.17.0 but have started failing since we upgraded the package to 8.0.0 in our repo.

I can see the roles in the accessibility tree on the browser, but unfortunately, the tests still fail.

Any help would be appreciated.
Thanks!

### Self-service

- [ ] I would be willing to implement a fix for this issue

rtivital:
Can you provide a repository with a minimal reproduction?

chumba-wamba:
> Can you provide a repository with a minimal reproduction?

Couldn't get it to work on codesandbox because of all the config required. Created an (awful) GitHub repo with the reproduction that you can find here - https://github.com/chumba-wamba/mantine-select-test.

`npm ci` followed by `npx vitest` should do the trick.

Please let me know if you have any questions.

Thank you!

cmoesel:
I've run into this as well with `getbyRole('option', ...)` as well as `getByRole('menuitem', ...)`. These worked fine in 7.17.7 but stopped working in 8.0.0 and 8.0.1.

chumba-wamba:
Hello @rtivital,
Any update on this issue? Is there anything I can do to help?

Thank you!  = )

hjelmeir:
I have run into this problem also with failing tests. I think my problem is the new Popover `hideDetached` prop being set to true (see migration guide) where the menu items still have "display: none" set. My fix for now, pass in comboxProps to Mantine select to set hideDetached to false (you will also probably need to do this with Popover, or any component that uses popover/combobox)

```
<Select
    comboboxProps={{ hideDetached: false }}
    {...props}
/>
```

Note: upon doing this I get a console warning: `React does not recognize the 'hideDetached' prop on a DOM element`

jonavila:
Thanks for the tip @hjelmeir. I've added 

```
[...]

PR #7992, opened 2025-06-18
[mantine.dev] Add env option to HeadlessMantineProvider  (#7836)

chumba-wamba:
Adding an env prop to the `HeadlessMantineProvider` since it can be used instead of `MantineProvider` for a custom render function.

This is my first PR for the mantine project. Please let me know if I am doing something incorrectly.


rtivital:
Thanks!

PR #8002, opened 2025-06-23
breaking change: make it clearer when a disabled switch is already checked

jgravois:
no offense taken if you don't consider this an improvement, but in our own app we differentiate the two components below👇 by swapping the default color back in and applying a partial opacity.

```tsx
<Switch disabled />
<Switch disabled checked />
```
before:
Screenshot 2025-06-23 at 3 24 19 PM

after:
Screenshot 2025-06-23 at 3 12 13 PM


rtivital:
I consider this to be a breaking change. Cannot merge for 8.x releases. I will consider it with 9.0 release.

jgravois:
since this didn't make it into 9.x i'll go ahead and close this PR.

feel free to incorporate the change at your leisure though.

PR #8023, opened 2025-06-29
Implement defaultDate prop on DateTimePicker

deedw:
This update implements defaultDate prop functionality in DateTimePicker to match that in DatePicker.


packages/@mantine/dates/src/components/DateTimePicker/DateTimePicker.tsx
```diff
@@ -154,6 +155,8 @@ export const DateTimePicker = factory<DateTimePickerFactory>((_props, ref) => {
     withTime: true,
   });
 
+  const _defaultDate = Array.isArray(_value) ? _value[0] || defaultDate : _value || defaultDate;
```
rtivital: DateTimePicker value cannot be Array

deedw: _defaultDate adjusted accordingly

rtivital:
Thanks!

PR #8026, opened 2025-06-30
[@mantine/form] Add Support for Cascading Form Updates

### LINKED ISSUE #8024: `form.watch` Does Not Fire for Changes to Parent Objects
### Dependencies check up

8.1.2

@mantine/core

Vite

All

### Describe the bug

When calling `form.setFieldValue('person',  { name: "john doe" })`, watchers of the key `person.name` aren't notified. They are only notified when the keypath is set directly (i.e. `form.setFieldValue('person.name', "john doe")`)

https://codesandbox.io/p/devbox/mantine-react-template-forked-tyc94v

### Possible fix

the implementation currently seems to only notify subscribers of the *exact* keypath. this could be changed to a *startsWith*, perhaps with an opt-in flag for more granular control 

### Self-service

erwijet:
## Description

Fixes an issue where changes to parent objects did not trigger updates for subscribers of nested keys, and where subscribers of parent keys were not notified when a nested key of that parent object changed.

To avoid introducing unexpected behavior for existing users, this new behavior is gated behind an opt-in flag: `cascadeUpdates?: boolean`.

---

closes #8024

### DIFF
--- a/apps/mantine.dev/src/pages/form/values.mdx
+++ b/apps/mantine.dev/src/pages/form/values.mdx
@@ -203,6 +203,15 @@ function Demo() {
 }
 ```
 
+## form.watch cascade
+
+To loosely subscribe to changes, you can set `cascadeUpdates: true`.
+This allows for parent objects to be written to directly, while still having
+subscribers to nested keys updated. Additionally, writes to nested keys
+will bubble up triggering parent key subscriptions as well.
+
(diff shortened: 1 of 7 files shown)


erwijet:
added some tests! lmk if anything else needs to be addressed :D

rtivital:
Please fix errors in tests

rtivital:
Thanks!

PR #8032, opened 2025-07-01
Update schema-validation.mdx on zod4 validation

Retwix:
In Zod v4, the `message` key has been deprecated in favor of `error`.

Reference: https://zod.dev/v4/changelog?id=deprecates-message


rtivital:
Thanks!

PR #8051, opened 2025-07-07
fix: Resolved the mql compatibility issue

Ives7:
image

In some browser versions, such as Safari, only the addListener method of MQL is supported, but this method is deprecated.

The current W3C standard only supports addEventListener of EventTarget.

Then, in some versions that don't support the new standard, the following error will be reported:

image

So in order to support some intermediate versions, compatibility issues need to be addressed.


rtivital:
Safari 13 is not supported by Mantine. You can find a list of the supported browsers here – https://mantine.dev/browser-support/#supported-browsers

The supported browsers do not require this logic. If you need to support older browsers, you can implement a polyfill on your side instead.

PR #8063, opened 2025-07-08
fix Select cursor display even when not searchable on Firefox only

### LINKED ISSUE #8057: Select cursor display even when not searchable on Firefox only
### Dependencies check up

7.17.8

@mantine/core

Vite

Firefox

### Describe the bug

When using `<Select />`  component in Firefox, the cursor is displayed even when the component don't have `searchable` prop.

The bug is reproducible on latest v7 and latest v8.
[v7](https://stackblitz.com/edit/react-y7f86d?file=src%2FApp.js)
[v8](https://stackblitz.com/edit/react-qukrgmmb?file=src%2FApp.js)

https://stackblitz.com/edit/react-y7f86d?file=src%2FApp.js

### Possible fix

_No response_

### Self-service

ihsanberkozcan:
fixed #8057

### DIFF
--- a/packages/@mantine/core/src/components/Input/Input.module.css
+++ b/packages/@mantine/core/src/components/Input/Input.module.css
@@ -212,6 +212,7 @@
     background-color: var(--input-disabled-bg);
     color: var(--input-disabled-color);
   }
+
   &:read-only {
     caret-color: transparent;
   }


rtivital:
Thanks!

ikapta:
[PasswordInput](https://mantine.dev/core/password-input/) component caret invisble. This occurs because the `PasswordInput` input component is rendered as a `div`, which is a `:not(:read-write)` element will selected by `:read-only`

[MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/:read-only)
> [!IMPORTANT]
> The `:read-only` pseudo-class matches any element that does not match `:read-write`, effectively equivalent to `:not(:read-write)`.

image

### Proposed Solution
update the CSS to target only `input` elements with `:read-only`. 
```
 &:is(input):read-only {
    caret-color: transparent;
  }
```

PR #8068, opened 2025-07-10
[@mantine/core] Add clearSearchOnChange back on MultiSelect

jobrk:
Adding back this prop that used to be part of v6
See https://v6.mantine.dev/core/multi-select/?t=props for reference


rtivital:
Thanks!

Issue #8066, opened 2025-07-10
TagsInput: Backspace can remove tags even when readOnly is set


Redbird10:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

8.1.3

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

All

### Describe the bug

When using the `TagsInput` component with the `readOnly` prop, users should not be able to add or remove tags. However, when the input is focused, pressing backspace will still remove the last tag, even though the component is set to `readOnly`.

### If possible, include a link to a codesandbox with a minimal reproduction

https://mantine.dev/core/tags-input/#read-only

### Self-service

- [ ] I would be willing to implement a fix for this issue

rtivital:
Fixed in [8.2.4](https://github.com/mantinedev/mantine/releases/tag/8.2.4)

PR #8069, opened 2025-07-10
fix: prevent tag removal via backspace when readOnly is enabled

naaa760:
#8066

### Problem
The `TagsInput` component with `readOnly={true}` was incorrectly allowing users to remove tags by pressing backspace when the input field was focused. This violated the expected read-only behavior where no modifications should be possible.

### Root Cause
Multiple input event handlers were missing `readOnly` property checks, allowing tag modifications even when the component was supposed to be read-only.

### Solution
Added `readOnly` checks to all input handlers that can modify tags:
- ✅ Backspace key (tag removal)
- ✅ Enter key (tag addition) 
- ✅ Split characters like comma (tag addition)
- ✅ Paste events (tag addition)
- ✅ Blur with `acceptValueOnBlur` (tag addition)
- ✅ Dropdown option selection (tag addition)

### Testing
- Added 6 comprehensive test cases covering all input scenarios
- All existing tests continue to pass (62/62)
- Verified behavior matches other Mantine read-only components


package.json
```diff
@@ -210,6 +210,10 @@
     "@babel/core": "7.26.0",
     "@types/react": "19.0.1",
     "@types/react-dom": "19.0.1",
-    "react-is": "^19.0.0"
+    "react-is": "^19.0.0",
```
rtivital: Please remove changes made to the package.json dependencies

rtivital:
Please fix errors

nehaaprasad:
What error is still persisting?
I fixed already.

rtivital:
You can view the failed CI run here – https://github.com/mantinedev/mantine/actions/runs/16505698640/job/46678522814

rtivital:
Thanks for submitting the PR, I've resolved issue on my side.

PR #8092, opened 2025-07-21
Update TipTap.demo.bubbleMenu.tsx

icflorescu:
BubbleMenu is exported from `@tiptap/react/menus` (as of tiptap v3, I believe).


icflorescu:
Apologies for the hasty PR, I just saw that "@tiptap/react" is still at "2.9.1" in the demo project.
@rtivital I had to upgrade tiptap to the latest `3.0.7` in my current project, everything works just fine with lots of extensions, the only issue being that `BubbleMenu` must now be imported from `@tiptap/react/menus` instead of `@tiptap/react`.

Since the docs don't specifically mention locking tiptap to `2.9`, people will assume it works with the latest version.
Which is true, except for the BubbleMenu example.

rtivital:
Thanks for submitting a PR. Due to security reasons, PRs with dependencies updates are not accepted as contributions to the project. The new tiptap major version will be evaluated in one of the next releases.

Issue #8094, opened 2025-07-21
[New Feature] Implement new color theme toggle with 3 states for mantine.dev using segmented control


mantine-contrib:
Currently, the [Mantine website](https://mantine.dev/) uses a single switch to toggle between light and dark color schemes:

<img width="2816" height="112" alt="Image" src="https://github.com/user-attachments/assets/5291da70-a4be-456a-90d8-dc9f66dc1c9b" />

<img width="2816" height="112" alt="Image" src="https://github.com/user-attachments/assets/ec9a6a38-dcdf-4c82-b9dd-9fd66d6ecc1a" />

This setup can be somewhat confusing, as the color schemes remains fixed once the user interacts with the switch, even if their system preferences change (as indicated by the prefer-color-scheme media query in the browser, as explained [here](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme)).

<img width="2816" height="112" alt="Image" src="https://github.com/user-attachments/assets/e72c893b-ea37-400f-bab8-f61de215bcd6" />
<img width="2816" height="112" alt="Image" src="https://github.com/user-attachments/assets/20cebbdc-9b70-4f15-b2d8-0370abe88b4f" />
<img width="2816" height="112" alt="Image" src="https://github.com/user-attachments/assets/e8c14bd3-ffed-4cea-8484-91db19c7ecdd" />
<img width="2816" height="112" alt="Image" src="https://github.com/user-attachments/assets/3ae2e498-d683-4d3f-accf-e1b0b64367b9" />

To address this, I implemented a three-state toggle, the left most option is `Light`, the rightmost is `Dark`, and the middle one is `Auto`. With this change, users can easily revert their selection and let the color scheme follow their system's preferences, improving the overall user experience.

PR #8095, opened 2025-07-21
[mantine.dev] Implement new color theme toggle with 3 states for mantine.dev using segmented control

mantine-contrib:
closes: https://github.com/mantinedev/mantine/issues/8094


rtivital:
It is not planned to implement this kind of switch on mantine.dev website, I'm happy with the current implementation.
If you want, you can add a demo with this component to the [color-scheme](https://mantine.dev/theming/color-schemes/) documentation. Otherwise you can close the PR.

PR #8107, opened 2025-07-24
Fix for #8078: TimePicker min and max props not enforced

### LINKED ISSUE #8078: TimePicker min and max props not enforced — invalid times can be typed and selected
### Dependencies check up

8.1.3

@mantine/dates

Vite

Chrome

### Describe the bug

The min and max props in the TimePicker component do not seem to be enforced correctly:

- Users can manually type a time value outside the specified range.
- The dropdown list still allows selecting values outside the min/max bounds.
- Invalid times are not disabled in the dropdown, unlike how DatePicker disables out-of-range dates.

This issue can be reproduced on the official documentation site:
🔗 https://mantine.dev/dates/time-picker/#min-and-max-values

**Expected behavior:**

- Times outside the valid range should be disabled in the dropdown.
- Manually entered times should be validated and restricted to the defined range.

https://github.com/user-attachments/assets/c78e3c4d-026e-4e57-ab39-d378f1d7718c

https://codesandbox.io/p/sandbox/mantine-react-template-forked-6lmhv6

### Possible fix

_No response_

### Self-service

idealconceptz:
Fixes #8078
Invalid times can be typed and selected.

The time is clamped between min and max when the TimePicker element loses focus.

### DIFF
--- a/packages/@mantine/core/src/components/HoverCard/use-hover-card.ts
+++ b/packages/@mantine/core/src/components/HoverCard/use-hover-card.ts
@@ -61,7 +61,7 @@ export function useHoverCard(settings: UseHoverCard) {
       delay: withinGroup ? groupDelay : { open: settings.openDelay, close: settings.closeDelay },
     }),
     useRole(context, { role: 'dialog' }),
-    useDismiss(context, { enabled: typeof settings.opened === 'undefined' }),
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #8109, opened 2025-07-24
fix: 1px gap border table issue #8080

### LINKED ISSUE #8080: Table withTableBorder prop 1 pixel gap
### Dependencies check up

8.1.3

@mantine/core

Vite

Chrome

### Describe the bug

Table component with specified `withTableBorder` prop results in a 1-pixel gap at the top of the table header.

https://github.com/user-attachments/assets/a0fb9b01-6301-4f10-bb8b-ed59687bc1bd

https://codesandbox.io/p/sandbox/mantine-react-template-forked-kzgn9p

### Possible fix

_No response_

### Self-service

themm-s:
Fixes #8080

Fixed the 1px gap issue that appeared at the top of the table header when both withTableBorder and sticky header (stickyHeader) were enabled in the Mantine Table component.
How it works:
When the table has an outer border and the header (thead) is sticky, browsers render the sticky header above the table border, causing a visible gap.
To fix this, a ::before pseudo-element is added to the sticky thead that draws a 1px line with the same color as the table border, perfectly covering the gap.
The sticky thead is set to position: sticky and z-index: 4 to ensure the pseudo-element is rendered above the table border.
Any unnecessary border-top on the sticky thead and the first th is removed to avoid double borders.


rtivital:
Thanks!

PR #8116, opened 2025-07-28
V8 DateTimePicker and DatePickerInput lose responsive layout on narrow screens Bug Fix

moulichan99:
Added multiline prop to wrap into two lines without overlapping the border


rtivital:
Not a full valid fix. It was removed to make DatePickerInput with xs size the same as other inputs (e.g. TextInput). You need to update styles for this as well.

moulichan99:
Okay. Will try different approach then.

moulichan99:
> Not a full valid fix. It was removed to make DatePickerInput with xs size the same as other inputs (e.g. TextInput). You need to update styles for this as well.

Adding `white-space: no-wrap ` css also fixes the issue, can i proceed with this change ?
image

rtivital:
Not a good solution. Keep multiline, update styles for xs size to match other inputs.

rtivital:
I've resolved the issue on my side.

PR #7304, opened 2024-12-24
[@mantine/hooks] Support ref cleanup function in React 19 with useMergedRef

elecdeer:
fixes #7269 

useMergedRef now properly handles ref callbacks that return cleanup functions in React 19. 
For backward compatibility with React 18, it maintains the previous behavior of calling refs with null during unmount.

I have verified the functionality through a separate test repository, as direct testing within Mantine is not possible due to its React 18 dependency.
https://github.com/elecdeer/react-19-use-merge-ref/blob/main/react19/src/use-merged.ref.test.tsx

PR #8118, opened 2025-07-29
fix: improve use-merged-ref hook cleanup implementation

seansps:
**Dependencies check up**

I have verified that I use latest version of all @mantine/* packages
What version of @mantine/* packages do you have in package.json?
7.17.8 and also tested on 8.1.3

**What package has an issue?**
@mantine/hooks

**What framework do you use?**
Vite

**In which browsers you can reproduce the issue?**
Chrome, Safari, Firefox

**Describe the bug**
It seems that when merging mantine refs to ReactDND refs, on unmount, we get errors about this cleanup function implemented from a prior PR: https://github.com/mantinedev/mantine/pull/7304/files#diff-7aaef3b012f983195f83037e3474a958b02ad5da696aaf2348b36bb529ce3b41

**If possible, include a link to a codesandbox with a minimal reproduction**
[No response
](https://codesandbox.io/p/devbox/mantine-forked-cwq8ss?workspaceId=ws_HT9CUathF2Ud8Se39j2Wwt)

**Possible fix**
This PR attempts to fix the issue by checking if the cleanup function returned is actually a function

**Self-service**

### DIFF
--- a/packages/@mantine/hooks/src/use-merged-ref/use-merged-ref.test.tsx
+++ b/packages/@mantine/hooks/src/use-merged-ref/use-merged-ref.test.tsx
@@ -33,4 +33,35 @@ describe('@mantine/hook/use-merged-ref', () => {
     unmount();
     expect(refCalled).toEqual([expect.any(HTMLButtonElement), null]);
   });
+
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #8123, opened 2025-07-30
fix: SubMenu overflows on small screen widths

ikapta:
# Submenu Positioning with Floating UI

## Overview of Floating UI
**flip()**: Adjusts placement (e.g., `top` to `bottom`) to prevent overflow.
**shift()**: Moves the floating element along the x or y axis to stay within the viewport. By default, only [`mainAxis=true`.](https://floating-ui.com/docs/shift#crossaxis)

## Problem Description
Since the submenu is set to `position=right-start`, the `mainAxis` changes to the y-axis in this horizontal placement. By default, `shift()` only adjusts the `mainAxis`, which may lead to inconsistent behavior due to Floating UI's handling.

## Solution
Enable `crossAxis: true` in `shift()` to allow x-axis adjustments, aligning the submenu to the viewport's right boundary. As `shift()` precedes `flip()` in the [`usePopover` middleware order](https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/components/Popover/use-popover.ts#L74-L90), `flip()` does not trigger horizontal placement changes. The current middleware order in `usePopover` is sufficient, ensuring a visually consistent right-aligned submenu.

## Visual Results
image

image

image

[Per Floating UI docs: "Enabling this can lead to the floating element **overlapping** the reference element, which may not be desired and is often replaced by the flip() middleware."](<https://floating-ui.com/docs/shift#crossaxis:~:text=Enabling%20this%20can%20lead%20to%20the%20floating%20element%20overlapping%20the%20reference%20element%2C%20which%20may%20not%20be%20desired%20and%20is%20often%20replaced%20by%20the%20flip()%20middleware.>)
However, this approach is effective for our use case.

## Alternative Approach
Updating the [middleware order](https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/components/Popover/use-popover.ts#L74-L90) to `[flip(), shift()]` would flip the submenu to `left-start` on overflow before shifting. The results would appear as follows:
image

image

## Conclusion
Both solutions are effective. I opted for the minimal code change by enabling `crossAxis: true` in `shift()` for the align-right approach. Users can modify position and middleware but should be familiar with Floating UI middleware to understand the implications of their changes.

### DIFF
--- a/packages/@mantine/core/src/components/Menu/MenuSub/MenuSub.tsx
+++ b/packages/@mantine/core/src/components/Menu/MenuSub/MenuSub.tsx
@@ -32,6 +32,12 @@ const defaultProps = {
   offset: 0,
   position: 'right-start',
   transitionProps: { duration: 0 },
+  middlewares: {
+    shift: {
+      // Enable crossAxis shift to keep submenu dropdown within viewport bounds when positioned horizontally
+      crossAxis: true,
+    },
+  },
 } satisfies Partial<MenuSubProps>;
 
 export function MenuSub(_props: MenuSubProps) {


rtivital:
Thanks!

clicktodev:
Thanks! However the second behavior seems like a much better default because the merged behavior completely hides the parent menu if the child has a multiple word label and the menu is on the edge of the screen.

ikapta:
Yes, it’s a thoughtful approach. 
Pros: In most cases, with only a second-level submenu, if it hits the right edge, it can flip to the left.
Cons: This way not solve the long label, smaller screens or beyond second submenus, the submenu may still hit an edge and overlap the parent menu. This may rare cases.

@rtivital What's your advice? If it's okay, could adjust the middleware order by default.

rtivital:
I like the merged behavior more. Let's leave it as is. It should still be possible to customize on the user side.

Issue #8088, opened 2025-07-20
sub-row for rowSpan > 1 in Table component does not have column border


jvllmr:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

8.1.3

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Firefox

### Describe the bug

When having cells with rowSpan > 1 and column borders, the resulting "sub-rows" do not have a column border.

### If possible, include a link to a codesandbox with a minimal reproduction

https://codesandbox.io/p/devbox/mantine-react-template-forked-dw6lqw?file=%2Fsrc%2FApp.tsx%3A29%2C1&workspaceId=ws_QxeLZv7KFshBEKowzyTfNg

### Self-service

- [x] I would be willing to implement a fix for this issue

TheMineWay:
I see. You refer to the missing border between the second row of "Heading 2" and the "Heading 3" cell.
This is also happening to me in 8.2.2

<img width="1774" height="191" alt="Image" src="https://github.com/user-attachments/assets/85133234-b6fa-4395-ae62-a90fb83e921a" />

TheMineWay:
I have just created a PR fixing this issue :)

PR #8137, opened 2025-08-03
[@mantine/core] Table: Fix missing border when rowspan is present

TheMineWay:
# Description
When using `withColumnBorders` along with rowSpan cells, some borders do not appear (#8088).
imagen

# Why
Table styles define that borders are given at the end pof each cell that is not last child. The issue is that when rowspan is used, even it seems like a cell is not the last one, for the DOM it is.

# Solution
In order to fix this via CSS I decided to add another rule that places a border at the start of every cell if it is not the first one. This fixes the cases where rowspan is used.
imagen

I choose this solution as I did not found any other way to fix this without requiring the developer to provide manually a "lastCell" indicator.


jvllmr:
Thank you for creating this PR. I initially wanted to provide a solution myself, but failed to do so because of limited time.

rtivital:
Thanks!

PR #8149, opened 2025-08-06
Fix carret no longer displaying on password inputs

EdwardEB:
PasswordInput carret was invisivle because of Pull Request #8063.

Tested both Select component and PasswordInput on Chrome, Edge and FireFox as the original pull request said it was broken in FireFox.


rtivital:
Thanks!

PR #8156, opened 2025-08-07
feat(input): set default size to sm

Jannchie:
image

It should be able to fix the issue where the textarea padding falls back to 0, rather than being set explicitly.


rtivital:
Thanks!

PR #8157, opened 2025-08-08
Fix use-list-state type

ofk:
`useState` allows you to pass an initializer value. ([Reference](https://react.dev/reference/react/useState#avoiding-recreating-the-initial-state))
It would be convenient if useListState also supported initializers. Technically, we were passing values directly to useState, so we modified the type definitions accordingly.


rtivital:
Thanks!

PR #8160, opened 2025-08-09
Fix Horizontal Scrollbars not updating when content  overflows

### LINKED ISSUE #8075: ScrollArea doesn't update its horizontal scrollbar when children are resized
### Dependencies check up

8.1.3

@mantine/core

Vite

### Describe the bug

`ScrollArea` doesn't seem to recognize when its children grow and doesn't show a horizontal scrollbar as a result when the container overflows. Vice versa, the scrollbar won't get removed when the children shrink.

The vertical scrollbar seems to work as intended. Using `ScrollArea.AutoSize` kinda works, but only because the native scrollbars take over when the container overflows.

I tested it in Opera GX (Chromium) and Firefox.

https://codesandbox.io/p/sandbox/mantine-react-template-forked-lr9v9y

### Possible fix

_No response_

### Self-service

pgrones:
Fixes #8075

The problem is that `display: block` by default lets its content overflow vertically, but only grows to 100% horizontally. This could be fixed in two ways

1. Make the content have `display: inline-block`
2. Make the content have a fixed width

I fixed the issue by giving the content container of `ScrollArea.Autosize` `width: min-content`. I think that's the least intrusive way of handling this issue that hopefully doesn't break anything else. This way the container resizes whenever the content changes its width.

I deliberately only applied the fix to `ScrollArea.Autosize` as applying it to `ScrollArea` messes up other components like `MultiSelect`.

If anyone has a better solution to this problem feel free to let me know. I kinda worry that applying these styles messes up other components relying on `ScrollArea.Autosize` down the line.


rtivital:
Thanks!

Issue #8164, opened 2025-08-12
Aside on home page cannot be seen


zrll12:
### Link to the page where something is not right

mantine.dev

### What is wrong?

On mobile phone, the aside section of the home page cannot be seen, only the burger has animation. On any other page this is fine.

If this is intended, then the burger icon should be hidden.

### Self-service

- [ ] I would be willing to implement a fix for this issue

mrdjohnson:
I've noticed this a few times as well! Seems the cause is that the `Shell` file hides the mobileNavbar if `withNav` is set to false. I'm willing to take this on

rtivital:
Thanks for reporting, the issue is fixed and will be released with the next patch

PR #8169, opened 2025-08-14
[mantine.dev] Homepage: Fix navbar not showing on mobile (#8164)

mrdjohnson:
The home page does not display the navbar on mobile because the subnav is not displayed on mobile. 

Fix: subnav and navbar should be separated

image

This does not affect the Combobox page (which also uses Shell) because that one has its own sidebar implementation (strange)


mrdjohnson:
I noticed that if the navbar is open when using spotlight to navigate, the navbar stays open. I can add fix to this PR (close the navbar if the pathname changes useEffect) or in another PR

rtivital:
Thanks!

PR #8181, opened 2025-08-18
fix(hooks): make cancel a stable reference

chyzwar:
When using useDebouncedValue cancel in useEffect we need to either biome-ignore or eslint-ignore to avoid accidental cancels.


rtivital:
Thanks!

PR #8189, opened 2025-08-20
fix(hooks): Safari iframe matchMedia crash

lorem--ipsum:
This fixes an issue with Mantine in an iframe context in Safari. Original issue can be found [here](https://github.com/metabase/metabase/issues/62302).

The original bug happens when switching tabs in an interactive embedded Metabase. It actually comes from the way Safari implements MediaQueryList event handling. It seems that events are triggered while an iframe is being removed, leading to an RTE. The crash only happening in Safari and also crashing the dev tools prevents from having a look at the error, but a git bisect shows that the issue was introduced in https://github.com/metabase/metabase/pull/53447.

Unfortunately there's no way of deterministically knowing when the error will be thrown, hence the `try {...} catch()`.

Also removed the unnecessary `useRef` usage.

Temporary patch in our repo is [here](https://github.com/metabase/metabase/pull/62407).

### DIFF
--- a/packages/@mantine/hooks/src/use-media-query/use-media-query.ts
+++ b/packages/@mantine/hooks/src/use-media-query/use-media-query.ts
@@ -46,14 +46,14 @@ export function useMediaQuery(
   );
   useEffect(() => {
     try {
-        const mediaQuery = window.matchMedia(query);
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #8198, opened 2025-08-24
[@mantine/form] make Values and TransformedValues covariant

jvllmr:
This allows type-safely passing a form to a func which accepts a form with only selected fields of the original form.

Here an example:
```typescript
import { UseFormReturnType, useForm } from '@mantine/form';



function PasswordSection<TForm extends UseFormReturnType<{password: string}>(
  {form}: {form: TForm}
) {
  // add buttons such as password generator button
  // ...
}

interface SignupFormValues {
  username: string;
  password: string;
}

export function SignupForm() {
  const form = useForm<SignupFormValues>({ initialValues: { username: '', password: '' } })
   
  return <form>
  {
  // ...
  }
  <PasswordSection form={form} />
  {
  // ...
  }
 </form>;
}

```

Related discussions:
  - https://github.com/orgs/mantinedev/discussions/7171
  - https://github.com/orgs/mantinedev/discussions/7195

This is just a proposal. Feel free to close the PR if you'd like.
I needed to create a new UseFormReturnType, but kept the old one around so that no breaking changes are made.
Edit: It is possible without an extra type


jvllmr:
I also looked into improved inference and value passing for the TransformValues / TransformedValues type, but I think for that TransformValues must be removed in favor of using only TransformedValues. But that would be a breaking change. I could start to create a draft PR for v9 though if there is interest.

jvllmr:
@rtivital could you already take a look at this PR? Would be happy about any feedback.

rtivital:
Yes, I've reviewed it, but I do not fully understand the new TS syntax (`out` part). I need more time for a deep dive.

jvllmr:
Okay, take your time 👍🏻  Thanks for letting me know.

jvllmr:
Now that it is right around the corner, is there interest in including this change in mantine v9?

rtivital:
There are many conflicts, I've added these changes in this commit https://github.com/mantinedev/mantine/pull/8748/changes/7aeb1ad848830004789d92f04508b9661d9b5151, closing the PR, thanks for proposals!

Issue #8188, opened 2025-08-19
TextInput - div with multiple paragraph elements passed to the error prop results in a dirty HTML


chumba-wamba:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

8.1.2

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

All

### Describe the bug

The error prop in the TextInput component accepts any values of type ReactNode. The value passed to it is wrapped in a p tag.

I have a case where I want to render multiple error values associated with a field. I achieve this by passing a component that renders multiple p tags within a div (as in the screenshot below). This results in a dirty HTML structure, and causes Vitest to log the following message when writing tests -

"In HTML, <p> cannot be a descendant of <p>. This will cause a hydration error."

<img width="1027" height="166" alt="Image" src="https://github.com/user-attachments/assets/80974d54-0960-4525-a916-a2b70107b657" />

### Possible fix

Ideally, the element should be wrapped in a generic element like a div or span. Another potential solution (that does not break existing users) would be to render the error in a p tag if it is of type string and render it in a generic element otherwise. 

### Self-service

- [x] I would be willing to implement a fix for this issue

s0h1s2:
same issue here

s0h1s2:
This is a temporary solution that works.
It has a type issue, though, but the browser applies it.
```
const theme = createTheme({
  components: {
    InputError: InputError.extend({ defaultProps: { component: "div" } }),
  },
});
```

ricaxvi:
TextInput error prop wraps everything in a <p> tag
When you pass a component with multiple <p> tags inside, you get nested <p> tags
This creates invalid HTML (<p> can't contain other <p> tags) and causes hydration errors

So Override the default component globally

Seongkyun-Yu:
Can I fix this problem?

chumba-wamba:
I discovered that this error is also reproducible in the DateTimePicker component.

rtivital:
Fixed in [8.3.1](https://github.com/mantinedev/mantine/releases/tag/8.3.1)

rtivital:
```tsx
const theme = createTheme({
  components: {
    InputError: InputError.extend({ defaultProps: { component: "div" } }),
  },
});
```

Now works without errors.

acsbendi:
[...]

PR #8200, opened 2025-08-24
fix: apply conditional p/div rendering and add aria-live

Seongkyun-Yu:
fix https://github.com/mantinedev/mantine/issues/8188

- Render p when content is a plain string; use div for other types.
- Add aria-live to enhance accessibility for dynamic updates. [MDN Link](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Reference/Attributes/aria-live)

Result
- string input
스크린샷 2025-08-25 오전 1 35 59

- React node input
스크린샷 2025-08-25 오전 1 40 13

- Related content
https://github.com/facebook/react/issues/24519
https://github.com/vercel/next.js/pull/46677


rtivital:
Thanks for submitting the PR! The proposed change is breaking. I'd prefer to resolve the issue differently, and I've added a fix on my side.

PR #8206, opened 2025-08-27
🩹 Fix font-size calculation in Input.module.css

k-utsumi:
CSS does not allow circular references.

- [demo @ JSFiddle](https://jsfiddle.net/32q04gfe/)


rtivital:
Thanks!

PR #8209, opened 2025-08-28
Add 'offset' option to useScrollSpy hook

andreaspersson-sciber:
This PR adds an 'offset' option to the useScrollSpy hook, used when calculating the active heading with respect to the top of the viewport.

- `offset` - offset from the top of the viewport to use when determining the active heading, by default `0` is used

This allows the hook to accommodate pages with floating headers or other elements that obscures the top of the viewport.

Example usage:
```tsx
function Demo() {
  return (
    <TableOfContents
      scrollSpyOptions={{
        offset: 70 // Offsets the "top of the viewport" used when calculating the active element
      }}
    />
  );
}
```

This fixes an issue where the Table of Contents may display an unexpected heading as active when using the AppShell header.
https://codesandbox.io/p/sandbox/mantine-react-template-forked-wlrnk8

### Without offset:
image

("Title 3" is the topmost visible title in the user's window, but TableOfContents shows "Title 2" as active)

### With offset:
image

TableOfContents shows the expected Title 3 as active, accommodating for the header.


rtivital:
Thanks!

PR #8211, opened 2025-08-28
[@mantine/hooks] support nodenext module resolution

seanwu1105:
This PR fix the similar issue mentioned in https://github.com/mantinedev/mantine/issues/4991 for @mantine/hooks.


rtivital:
Can you provide steps to reproduce the issue?

shwu-jt:
Steps:

1. Install `@mantine/hooks` and `typescript-eslint`
2. Configure the `tsconfig.json` with the following settings

```jsonc
{
  "$schema": "https://json.schemastore.org/tsconfig",
  "compilerOptions": {
    "module": "nodenext",
    "moduleResolution": "nodenext",
    "verbatimModuleSyntax": true,
    // ...
  }
}
```

3. Creating the following file using `@mantine/hooks`.

```tsx
import { useDocumentTitle } from "@mantine/hooks";

export function MyComponent() {
  useDocumentTitle("My Component");

  return <div>My Component</div>;
}
```

4. Running `eslint .` with `typescript-eslint` yields the following error.

```txt
> eslint .


/home/shwu/ui/apps/web/app/routes/txnrepo/temp.tsx
  4:3  error  Unsafe call of a(n) `error` type typed value  @typescript-eslint/no-unsafe-call

✖ 1 problem (1 error, 0 warnings)
```

`@mantine/dates` and `@mantine/core` do not have such error because the exports in their [`index.ts` files explicitly specify `.js` in the path](https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/index.ts), which is the correct way for `nodenext` module resolution.

rtivital:
Thanks!

PR #8215, opened 2025-08-30
[@mantine/hooks] use-document-visibility: Set initial document visibility state

sky0014:
The initial document visibility state maybe 'hidden', should set it when useEffect runs.


rtivital:
Thanks!

PR #8224, opened 2025-09-03
Add `.js` for nodenext module resolution

JoannaWang1216:
This PR applies the same thing in https://github.com/mantinedev/mantine/pull/8211 to other packages. The way to reproduce the issue is identical to [the comment](https://github.com/mantinedev/mantine/pull/8211#issuecomment-3237159318).


rtivital:
Thanks!

PR #8268, opened 2025-09-15
FIX: Nested list overflow issue

ondrejkonec:
I've added safe fix for issue with nested items and text overflow. It can be improved with redesigning component, but this would be more complicated and it's not part of my intention.

### before
Screenshot 2025-09-15 at 9 50 41

### after
Screenshot 2025-09-15 at 9 51 01


rtivital:
Thanks for submitting a PR, the issue has already been resolved in other PR.

Issue #8165, opened 2025-08-12
Nested List causes layout break and overflow


Su-Vic:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

8.2.4

### What package has an issue?

@mantine/core

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

When using multiple levels of nested ‎`<List>` components, the content sometimes exceeds the parent container’s width, causing unwanted horizontal scrolling (overflow-x). This happens even when the parent container has a fixed or limited width.

<img width="1061" height="548" alt="Image" src="https://github.com/user-attachments/assets/34baf014-f6ba-4274-9129-bf91458b3c55" />

### Self-service

- [ ] I would be willing to implement a fix for this issue

shaadcode:
As a workaround, `text-wrap` didn't solve the problem?

Su-Vic:
Hi @shaadcode,

I tested adding `text-wrap: wrap;` to the `li`, but this causes the marker and the content to wrap together. When the content is long, the marker gets pushed to the next line, which makes the list layout look incorrect.

<img width="460" height="753" alt="Image" src="https://github.com/user-attachments/assets/54679c82-78e5-4d46-a8a5-ac93eb03b69f" />

rootboard:
@Su-Vic `list-style-position: inside;` causes the problem, as a workaround just change style for root class to `outside` and add some paddings if needed.

https://github.com/mantinedev/mantine/blob/6fba1a12b9db5a048bb2af0f94ffa7ddc6da4225/packages/%40mantine/core/src/components/List/List.module.css#L5

shaadcode:
Hi @Su-Vic 
As a new solution, give `li`s `text-wrap-style: balance;`

AyushTyagi2:
hello, I would like to work on this issue.

rtivital:
You are welcome to submit a PR with a fix

terrydkim:
Hello! I'd like to solve this issue :>

rtivital:
You are welcome to submit a PR with a fix

ondrejkonec:
> You are welcome to submit a PR with a fix

I've suggest safe and quick fix and [created PR](https://github.com/mantinedev/mantine/pull/8268).

terrydkim:
> You are welcome to submit a PR with a fix

[...]

Issue #2778, opened 2022-10-23
List.Item overflows List 


edymusajev:
### What package has an issue

@mantine/core

### Describe the bug

The itemWrapper div inside of `List.Item` overflows the width of the List. 
<img width="612" alt="Screenshot 2022-10-24 at 00 04 54" src="https://user-images.githubusercontent.com/9220737/197405465-cca9d96f-da84-4dfc-ba4b-b4d861738507.png">

### What version of @mantine/hooks page do you have in package.json?

5.6.1

### If possible, please include a link to a codesandbox with the reproduced problem

https://codesandbox.io/s/relaxed-matan-upqijw

### Are you willing to participate in fixing this issue and create a pull request with the fix

Yes

koralarts:
@rtivital I'd like to help take a look at this.

koralarts:
Did a little research about why this is happening. From what I researched, `list-style-position: inside` is handled differently by different browsers according to ([MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/list-style-position)). After checking different browsers (Chrome, Safari, Firefox, and Edge) they seem to act the same.

**Quick Solution**

A quick solution is to pass a custom bullet icon.

```javascript
<List icon="•"> ... </List>
```

---

**Caveat(s)**

The `::marker` element is controlled by having an element have `display: list-item`.

**Possible Solutions**

1. Calculate the width of the item wrapper to take into account the bullet. This by default isn't a good solution since sizing won't be consistent between browsers.
2. Use a custom bullet. This can be done in multiple ways -- via CSS or custom element.

**2.1 CSS**

A custom bullet can be created using a `::before` element. I used `display: list-item` instead of passing a bullet in `content` because it will render the bullet using the browser's preferred bullet.

```css
li {
  display: flex;
  
  &::before {
    content: ' ';
    display: list-item;
  }
}
```

https://codesandbox.io/s/mantine-list-item-overflow-4n19br?file=/src/App.tsx

**2.2 Custom Element**

A custom element rendering a bullet could be the default element rendered if `icon` is not passed.

---

I think either solution would be the least invasive as both won't require breaking structural changes. If we feel one is a viable solution, I can make a PR with the selected solution; otherwise, we'll brainstorm another solution.

Manas-Nagelia:
A more straightforward solution is to add <span> tags inside the <List.Item> tag. This is because this commit with removed the original <span> tags causing this overflow to happen: #2146

[...]

PR #8269, opened 2025-09-15
🐛Fix: nestedList overflow issue

terrydkim:
Hello👋,
As a solution to issue #8165 , I changed `list-style-position: inside` to `list-style-position: outside`.

Issue #2778  also suggested using `outside`, and I believe this change is beneficial for preventing potential future bugs. Should I instead maintain `inside` and make modifications in other areas?

Additionally, after changing to `outside`, there's now a spacing difference between the marker and text, and for the "With Icons" case, padding-inline-start has been applied to secure the marker's position. To maintain the same appearance as before, additional adjustments would be needed.

If you could suggest an alternative approach, I'd be happy to implement it accordingly. Thank you.

# Before
스크린샷 2025-09-15 오후 10 56 05
스크린샷 2025-09-15 오후 10 57 28
스크린샷 2025-09-15 오후 10 57 44

# After
스크린샷 2025-09-15 오후 10 56 29
스크린샷 2025-09-15 오후 10 56 55
스크린샷 2025-09-15 오후 11 21 56

### DIFF
--- a/packages/@mantine/core/src/components/List/List.module.css
+++ b/packages/@mantine/core/src/components/List/List.module.css
@@ -1,20 +1,22 @@
 .root {
   --list-fz: var(--mantine-font-size-md);
   --list-lh: var(--mantine-line-height-md);
+  --list-marker-gap: var(--mantine-spacing-lg);
 
-  list-style-position: inside;
+  list-style-position: outside;
   font-size: var(--list-fz);
   line-height: var(--list-lh);
   margin: 0;
   padding: 0;
+  padding-inline-start: var(--list-marker-gap);
 
   &:where([data-with-padding]) {
-    padding-inline-start: var(--mantine-spacing-md);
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

packages/@mantine/core/src/components/List/List.module.css
```diff
@@ -1,20 +1,22 @@
...
   --list-lh: var(--mantine-line-height-md);
+  --list-marker-gap: var(--mantine-spacing-lg);
 
-  list-style-position: inside;
+  list-style-position: outside;
   font-size: var(--list-fz);
   line-height: var(--list-lh);
   margin: 0;
   padding: 0;
+  padding-inline-start: var(--list-marker-gap);
```
sirianni, line 11: This change introduced a regression where `<<List listStyleType="none"> are now improperly left-padded

PR #8276, opened 2025-09-17
[@mantine/hooks] Improve type safety and cleanup handling in mergeRefs and useMergedRef hooks

wo-o29:
- Refined `assignRef` to use `StrictRef` type and handle nullable refs correctly  
- Filtered out `null` and `undefined` refs early in `mergeRefs` for better type safety  
- Updated `mergeRefs` to always return a cleanup function for consistent return type  
- Simplified cleanup logic by always returning a cleanup callback, removing conditional branching  
- Overall improvements to typings and code readability for React ref merging utilities  

This change enhances the reliability and maintainability of ref merging utilities by enforcing stricter type guarantees and consistent cleanup behavior.

### DIFF
--- a/packages/@mantine/hooks/src/use-merged-ref/use-merged-ref.ts
+++ b/packages/@mantine/hooks/src/use-merged-ref/use-merged-ref.ts
@@ -1,41 +1,40 @@
-import { Ref, useCallback, type RefCallback } from 'react';
+import { useCallback, type Ref, type RefCallback } from 'react';
 
 type PossibleRef<T> = Ref<T> | undefined;
-
+type StrictRef<T> = NonNullable<Ref<T>>;
 type RefCleanup<T> = ReturnType<RefCallback<T>>;
 
-export function assignRef<T>(ref: PossibleRef<T>, value: T): RefCleanup<T> {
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks for submitting the PR. These changes make the hook incompatible with React 18, which is still supported by Mantine 8.

wo-o29:
> Thanks for submitting the PR. These changes make the hook incompatible with React 18, which is still supported by Mantine 8.

@rtivital Thank you for your reply!

My PR changes mergeRefs so that the returned callback always provides a cleanup function. In React 18, this return value isn't actually used, so I believe there are no major runtime issues (in other words, it does not cause errors).

On the other hand, in React 19, the cleanup function returned by the ref callback is now explicitly utilized, so having mergeRefs consistently return a cleanup improves type consistency and developer experience.

That is why the callback always returns a cleanup function—if there is nothing to clean up, the function simply does nothing, otherwise it performs the necessary cleanup.

I’m curious if there are any specific cases in the Mantine codebase where this pattern might cause unexpected issues in React 18, or if there are additional considerations I should be aware of. I’d appreciate hearing your thoughts!

PR #8278, opened 2025-09-18
[@mantine/core] CheckboxIndicator: Update icon color to use CSS variable  #8271

### LINKED ISSUE #8271: Checkbox iconColor prop ignored due to usage of wrong css variable
### Dependencies check up

8.3.1

@mantine/core

Vite

### Describe the bug

The checkbox component offers an "icon color" prop that does not affect the output, because it sets the `--checkbox-icon-color` css variable, but the icon sets its color based on `--checkbox-color` - see screenshots below. This behavior does not line up with the respective prop descriptions in the docs (https://mantine.dev/core/checkbox/?t=props) which makes me assume this to be unintended.

Image
Image

We are using a custom icon component in the screenshots, but confirmed the bug to exist for both the standard and the custom icons.

_No response_

### Possible fix

Use the `--checkbox-icon-color` variable with `--checkbox-color` as fallback to avoid breaking changes to existing code bases.

### Self-service

sarathkumarsasi:
Fixes #8271
Will fix the Checkbox iconColor prop ignored due to the usage of the wrong CSS variable #8271

### DIFF
--- a/packages/@mantine/core/src/components/Checkbox/CheckboxIndicator/CheckboxIndicator.module.css
+++ b/packages/@mantine/core/src/components/Checkbox/CheckboxIndicator/CheckboxIndicator.module.css
@@ -81,7 +81,7 @@
 .icon {
   display: block;
   width: 60%;
-  color: transparent;
+  color: var(--checkbox-icon-color);
   pointer-events: none;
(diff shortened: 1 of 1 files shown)


rtivital:
There is no issue to fix, associated reported issue is invalid.

PR #8282, opened 2025-09-19
[@mantine/hooks] Enhance useClickOutside to propagate event to handler

ThoDon:
This PR extends the use-click-outside hook so that the callback receives the triggering event object as its first argument. This gives developers more flexibility when handling outside clicks without introducing breaking changes.

**Details**

- Updated the internal event listener to forward the event to the callback.
- Type definitions updated to reflect the new callback signature.
- Existing behavior remains unchanged for callbacks that don’t use arguments.

Linked feature request : https://github.com/orgs/mantinedev/discussions/8281


rtivital:
Thanks!

PR #8287, opened 2025-09-21
[@mantine/hooks] Optimize useWindowScroll hook by scroll updates with requestAnimationFrame

wo-o29:
### Problem
The original hook updated the scroll position state on every scroll and resize event. Given that these events can fire many times per frame, this caused excessive React state updates and component re-renders, resulting in degraded performance, especially on high-frequency scrolls.

### Solution
- Introduced a `ticking` ref flag to avoid multiple requestAnimationFrame calls per frame.
- Wrapped the scroll position update inside a requestAnimationFrame callback, ensuring updates happen once per animation frame.
- Cleaned up event listeners and cancelled any pending animation frames on unmount to prevent memory leaks.
- Kept the initial position set on mount.

### Benefits
- **Reduced unnecessary React renders:** Updates are batched per animation frame instead of per raw scroll event.
- **Smoother scrolling experience:** Less UI jank and better frame rates.


rtivital:
Thanks!

PR #8290, opened 2025-09-23
[@mantine/dates] prevent endless loop in getEndOfWeek

sjroesink:
Calling `getEndOfWeek` with an invalid date now causes an endless loop (`getEndOfWeek("", "")` for instance).
See https://codesandbox.io/p/sandbox/priceless-flower-8v9mhl?file=%2Fsrc%2FApp.js%3A7%2C5-7%2C17

This function is at least also called from `DateInput`. Simply setting an invalid value there also causes the loop after clicking the input:
```
<DateInput value={'invalid'} />
```


rtivital:
Can you give an example of code which causes an issue for you?

sjroesink:
> Can you give an example of code which causes an issue for you?

@rtivital I updated the description.

rtivital:
Thanks!

PR #8294, opened 2025-09-23
[@mantine/core] NavLink: fix onClick/onKeyDown blocking expand/collapse

### LINKED ISSUE #8293: NavLink: Uncontrolled parent NavLink with children fails to toggle when onClick or onKeyDown is provided
### Dependencies check up

8.3.1

@mantine/core

Vite

None

### Describe the bug

In an uncontrolled parent NavLink with children, passing a custom onClick or onKeyDown prevents the link from expanding/collapsing.
- Clicking the parent does not toggle expansion when onClick is supplied.
- Pressing Space does not toggle expansion when onKeyDown is supplied.

This happens because {...others} is spread after onClick={handleClick} (and onKeyDown), overriding the internal toggle handlers.

_No response_

### Possible fix

Extract onClick/onKeyDown from props and compose them with the internal handlers instead of letting them be overridden by {...others}.

### Self-service

Redbird10:
### Issue
In an uncontrolled parent NavLink with children, passing a custom `onClick` or `onKeyDown` prevents the link from expanding/collapsing.
- Clicking the parent does not toggle expansion when `onClick` is supplied.
- Pressing Space does not toggle expansion when `onKeyDown` is supplied.

This regression was introduced at dbf319e, where `{...others}` began overriding the internal handlers.
### Fix
- This fix extracts `onClick`/`onKeyDown` from props and composes them with internal handlers to ensure they are not overridden by `{...others}`. 
- The extracted handlers are typed against `<HTMLElement>` rather than `<HTMLAnchorElement>` to align with the polymorphic design of NavLink.
- A test was also added to confirm that expansion works when a custom `onClick` is provided.
- Fixes #8293

### DIFF
--- a/packages/@mantine/core/src/components/NavLink/NavLink.test.tsx
+++ b/packages/@mantine/core/src/components/NavLink/NavLink.test.tsx
@@ -51,6 +51,17 @@ describe('@mantine/core/NavLink', () => {
     expect(screen.getByRole('link')).toHaveAttribute('data-expanded');
   });
 
+  it('sets correct data-expanded attribute on link (uncontrolled) when onClick is provided', async () => {
+    render(
+      <NavLink {...defaultProps} onClick={() => {}}>
+        <div>test-dropdown</div>
+      </NavLink>
+    );
+    expect(screen.getByRole('link')).not.toHaveAttribute('data-expanded');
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #8306, opened 2025-09-26
[@mantine/hooks]Fix useDidUpdate running on mount in React Strict Mode during development

barkinkaradeniz-tr:


rtivital:
Thanks!

Issue #8253, opened 2025-09-11
Absolute Bottom Left/Right Combobox second click.


TrSherlock:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

8.3.0

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

```js
return <MantineProvider forceColorScheme="dark">
      <div className="absolute right-0 bottom-0">
        <MantineCombobox />
      </div>
  </MantineProvider>  
}
```
[https://codesandbox.io/p/sandbox/mantine-react-template-forked-vvv39q](url)
absolute or fixed
first click up, second click down

Note: I don't speak English. I apologize for my spelling mistakes.

### If possible, include a link to a codesandbox with a minimal reproduction

https://codesandbox.io/p/sandbox/mantine-react-template-forked-vvv39q

### Self-service

- [ ] I would be willing to implement a fix for this issue

pipepasta:
Hi, I’ve opened a PR https://github.com/mantinedev/mantine/pull/8308 that should fix this issue.

TrSherlock:
Thank you 👏👏👏

PR #8308, opened 2025-09-27
fix: Prevent combobox popover position flip when opened twice

macaroni10y:
fixes https://github.com/mantinedev/mantine/issues/8253

## Cause
The issue was caused by an interaction between the `preventPositionChangeWhenVisible` property and position reset logic in the Popover component.

1. First click: Floating UI detects insufficient space at bottom and flips to top, storing the flipped position
2. On close: The `onExited` callback unconditionally resets position back to original (`bottom`)
3. Second click: Uses the reset position (`bottom`) with flip middleware disabled, causing overflow

## Changes
- In `onExited` in `Popover.tsx`, keep previous position when `preventPositionChangeWhenVisible` is true
- Added two Combobox stories to verify behavior when positioned at the bottom-left and bottom-right corners

## issue
https://github.com/mantinedev/mantine/issues/8253

### DIFF
--- a/packages/@mantine/core/src/components/Combobox/Combobox.story.tsx
+++ b/packages/@mantine/core/src/components/Combobox/Combobox.story.tsx
@@ -1,6 +1,8 @@
 import { useState } from 'react';
 import { Anchor } from '../Anchor';
 import { Button } from '../Button';
+import { Input } from '../Input';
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #8309, opened 2025-09-28
[@mantine/hooks] Fix focus capture timing to useFocusReturn hook 

wo-o29:
## Description  
  
This PR fixes the focus capture timing issue in the `useFocusReturn` hook and adds SSR safety.  
  
### Problem  
  
The current implementation captures `document.activeElement` after the modal opens using `useDidUpdate`, which means it often captures the modal's internal elements instead of the button that opened the modal. This breaks the intended behavior of returning focus to the original trigger element.  
  
### Solution  
  
1. **Fixed timing issue**: Capture `document.activeElement` during hook initialization instead of in `useDidUpdate` when `opened` becomes `true`  
2. **Added SSR safety**: Added `typeof document !== "undefined"` check to prevent errors during server-side rendering  
3. **Simplified logic**: Removed the `opened` condition from `useDidUpdate` since we only need to handle the closing case  

### Changes  
  
- Move focus capture from `useDidUpdate` to hook initialization  
- Add SSR safety check for `document.activeElement`  
- Remove unnecessary `opened` condition in `useDidUpdate`  
- **Remove redundant optional chaining in `returnFocus` function**: Changed `lastActiveElement.current?.focus()` to `lastActiveElement.current.focus()` since the if condition already guarantees the element exists and has a focus method  
- Maintain backward compatibility with existing API

### DIFF
--- a/packages/@mantine/hooks/src/use-focus-return/use-focus-return.ts
+++ b/packages/@mantine/hooks/src/use-focus-return/use-focus-return.ts
@@ -12,14 +12,17 @@ export function useFocusReturn({
   opened,
   shouldReturnFocus = true,
 }: UseFocusReturnOptions): UseFocusReturnReturnValue {
-  const lastActiveElement = useRef<HTMLElement>(null);
+  const lastActiveElement = useRef<HTMLElement>(
+    typeof document !== 'undefined' ? (document.activeElement as HTMLElement) : null
+  );
+
   const returnFocus = () => {
     if (
       lastActiveElement.current &&
       'focus' in lastActiveElement.current &&
       typeof lastActiveElement.current.focus === 'function'
     ) {
-      lastActiveElement.current?.focus({ preventScroll: true });
+      lastActiveElement.current.focus({ preventScroll: true });
     }
   };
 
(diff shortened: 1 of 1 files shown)


rtivital:
Can you provide an example sandbox of issue that this fixes?

wo-o29:
> Can you provide an example sandbox of issue that this fixes?

example: https://codesandbox.io/p/devbox/usefocusreturn-mantine-forked-4t6mq4
main code: `App.tsx`, `components/Modal`, `hooks/use-focus-return`

### **[Context]**
  Blue button – The Default useFocusReturn has the same implementation as the original mantine useFocusReturn.
  Red button – The Custom useFocusReturn is an example using the useFocusReturn that I modified.

### **[Example]**
  When opening the modal and either clicking the close button or closing it using the ESC key, I expected the focus to return to the button that triggered the modal. 
  
  However, with the blue button (Default useFocusReturn), lastActiveElement.current is captured as an input inside the modal, so it does not work as expected. 
  
  With the red button (Custom useFocusReturn), lastActiveElement.current is captured as the modal trigger button, so you can see that when the modal is closed, the focus correctly returns to the modal trigger button.
  
  The original useFocusReturn updates lastActiveElement.current inside useDidUpdate after the focus has moved inside the modal, so I believe this kind of timing issue occurs.

PR #8341, opened 2025-10-08
[@mantine/dates] DatePicker: add missing `multiple` type for `DatePickerPreset`

### LINKED ISSUE #8338: DatePickerInput on type multiple accepts the the wrong type
### Dependencies check up

8.3.3

@mantine/dates

Vite

### Describe the bug

When using preset on a DatePickerInput with `type="multiple"`, TypeScript wants a value of `string | null` when it should be `string[] | null`.

This is only a TypeScript issue. It works normally when provided the correct value of `string[]`.

_No response_

### Possible fix

Replace [DatePicker.tsx](https://github.com/mantinedev/mantine/blob/6694c53674b1a2a4f4678771e614584a3bada414/packages/%40mantine/dates/src/components/DatePicker/DatePicker.tsx#L41)

```
export interface DatePickerPreset<Type extends DatePickerType> {
  value: Type extends 'range'
    ? [DateStringValue | null, DateStringValue | null]
    : DateStringValue | null;
  label: React.ReactNode;
```

With

```
export interface DatePickerPreset<Type extends DatePickerType> {
  value: Type extends "range"
    ? [DateStringValue | null, DateStringValue | null]
    : Type extends "multiple"
      ? DateStringValue[]
      : DateStringValue | null;
  label: React.ReactNode;
}
```

### Self-service

ahmedsemih:
Fixes #8338 

This PR resolves a type issue in the `DatePickerPreset` interface.

The `value` property was incorrectly typed for the `'multiple'` selection mode. I've updated the conditional type to ensure that when `Type` is `'multiple'`, the `value` is correctly defined as `DateStringValue[]`. This fixes the type mismatch and ensures proper type safety.

### DIFF
--- a/packages/@mantine/dates/src/components/DatePicker/DatePicker.tsx
+++ b/packages/@mantine/dates/src/components/DatePicker/DatePicker.tsx
@@ -40,7 +40,9 @@ import classes from './DatePicker.module.css';
 export interface DatePickerPreset<Type extends DatePickerType> {
   value: Type extends 'range'
     ? [DateStringValue | null, DateStringValue | null]
-    : DateStringValue | null;
+    : Type extends 'multiple'
+      ? DateStringValue[]
+      : DateStringValue | null;
   label: React.ReactNode;
 }
 


rtivital:
Thanks!

PR #8350, opened 2025-10-10
🐛fix(popover): prevent horizontal ghost scrolling

### LINKED ISSUE #8321: Select can scroll to infinity
### Dependencies check up

8.3.2

@mantine/core

Next.js

Chrome

### Describe the bug

You can scroll to the right in a select dropdown to infinity — see the scroll bar in the image below.

Image

https://codesandbox.io/p/sandbox/mantine-react-template-q4dd5d

### Possible fix

_No response_

### Self-service

terrydkim:
Fixes #8321 

Hello :D ,
I've resolved the infinite scrolling issue in the Select component.

### Changes:
Modified the shift padding value from `5` to `0` in `use-popover.ts`

### Issue Details:
The Combobox renders through a portal, which made this issue a bit tricky to track down. 😅
To reproduce this issue, you can easily observe it by removing the padding from the Combobox's parent div in the Demo.

### Root Cause:
The problem occurred when the Combobox had left padding. Since the shift middleware is enabled by default in the popover's defaultProps, the component would get pushed by `5px` padding and then pulled back by the shift, creating a loop that resulted in infinite scrolling.

### Solution:
I've set the default padding value of the shift middleware to `0` in `use-popover.ts.`

### Side Effects:
This change may affect the Popover Arrow component. If you'd like to modify only the Combobox without impacting other components, you can configure the Combobox props with `shift: { padding: 0 }` instead.

```jsx
<Combobox 
  middlewares={{ 
    shift: { padding: 0 }
  }}
>
  {/* ... */}
</Combobox>
```

If you could suggest an alternative approach, I'd be happy to implement it accordingly. Thank you.

### Normal
스크린샷 2025-10-11 오전 1 10 02 
 
### Bug
스크린샷 2025-10-11 오전 1 09 55

https://github.com/user-attachments/assets/94baf595-4362-44b1-8ef8-3f3fa8ae2437

## After
스크린샷 2025-10-11 오전 2 10 14

https://github.com/user-attachments/assets/23ca0190-35f3-40c7-878a-b0dc8583d99f

### DIFF
--- a/packages/@mantine/core/src/components/Popover/use-popover.ts
+++ b/packages/@mantine/core/src/components/Popover/use-popover.ts
@@ -75,8 +75,8 @@ function getPopoverMiddlewares(
     middlewares.push(
       shift(
         typeof middlewaresOptions.shift === 'boolean'
-          ? { limiter: limitShift(), padding: 5 }
-          : { limiter: limitShift(), padding: 5, ...middlewaresOptions.shift }
+          ? { limiter: limitShift(), padding: 0 }
+          : { limiter: limitShift(), padding: 0, ...middlewaresOptions.shift }
       )
     );
   }


rtivital:
I'd rather not change those values in the library. Instead, it is better to update the documentation and include a section in select component that mentions this behavior.

terrydkim:
Thank you for the feedback!
I'll close this PR and create a new one that adds documentation instead. 😆

PR #8355, opened 2025-10-11
📜docs(select): add horizontal infinite scroll workaround

### LINKED ISSUE #8321: Select can scroll to infinity

terrydkim:
Fixes #8321 

Hello😆!

Following your feedback from PR #8350 , I've updated the Select component documentation instead of changing the library code.

I added the workaround section after "Dropdown offset" – I hope the placement makes sense! 
Please let me know if you'd prefer it in a different location or if any other changes are needed.

I've tested the workaround and confirmed it resolves the horizontal infinite scrolling issue.

Thanks :>

## After
스크린샷 2025-10-11 오후 6 41 24

### DIFF
--- a/apps/mantine.dev/src/pages/core/select.mdx
+++ b/apps/mantine.dev/src/pages/core/select.mdx
@@ -201,6 +201,27 @@ To change dropdown offset, set `offset` prop in `comboboxProps`:
 
 <Demo data={SelectDemos.dropdownOffset} />
 
+## Prevent horizontal infinite scrolling
+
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #7613, opened 2025-03-20
[@mantine/core] Checkbox - Set indeterminate prop (#7608)

andrewborstein:
Closes https://github.com/mantinedev/mantine/issues/7608

Set `indeterminate` property in native `input` element when `indeterminate` component prop is true. Prefer `indeterminate` over `aria-checked` per spec (https://www.w3.org/TR/html-aria/#att-checked).

References
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/indeterminate
- https://github.com/testing-library/jest-dom/blob/main/src/to-be-partially-checked.js
- https://www.w3.org/WAI/ARIA/apg/patterns/checkbox/examples/checkbox-mixed/
- https://medium.com/indigoag-eng/indeterminate-checkboxes-are-weird-704b246c0f19

Issue #7608, opened 2025-03-20
[core] <Checkbox> indeterminate attribute not set on input element


andrewborstein:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.17.2

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

I noticed when unit testing checkboxes locally with testing library that an indeterminate checkbox would fail an assertion like

```
expect(getByRole('checkbox', { name: 'Select all' })).toBePartiallyChecked()
```

[The `toBePartiallyChecked` matcher](https://github.com/testing-library/jest-dom/blob/918b6fbcde10d4409ee8f05c6e4eecbe96a72b7a/src/to-be-partially-checked.js#L24-L32) checks for either the presence of an `indeterminate` attribute or `aria-check="mixed"`, which makes sense based on [the spec](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/indeterminate).

When I inspect the DOM, [the indeterminate example in the docs](https://mantine.dev/core/checkbox/#indeterminate-state) seems to only set the `data-indeterminate` attribute, not the native `indeterminate` attribute. This matches what I'm seeing locally and in [this sandbox](https://stackblitz.com/edit/vitejs-vite-dacozs2a?file=src%2FApp.tsx).

### If possible, include a link to a codesandbox with a minimal reproduction

https://stackblitz.com/edit/vitejs-vite-dacozs2a?file=src%2FApp.tsx,package.json

### Possible fix

Could [the fix](https://github.com/mantinedev/mantine/blob/0b7ccbfe061b4e07c83c14747e612c89507db3b8/packages/%40mantine/core/src/components/Checkbox/Checkbox.tsx#L215) be as simple as this?

```diff
        <Box
          component="input"
          id={uuid}
          ref={ref}
          checked={checked}
          disabled={disabled}
+         indeterminate={indeterminate}
          mod={{ error: !!error, indeterminate }}
          {...getStyles('input', { focusable: true, variant })}
          onChange={onChange}
          {...rest}
          {...contextProps}
          type="checkbox"
        />
```

### Self-service

- [x] I would be willing to implement a fix for this issue

rtivital:
You are welcome to research the issue and provide a PR with a fix

andrewborstein:
[...]

PR #8363, opened 2025-10-14
[@mantine/core] Checkbox: Fix data-indeterminate attribute removal on state change

chlee1001:
## Description

  Fixes an issue where the `data-indeterminate` attribute is not removed from the DOM when the `indeterminate` prop changes
  from `true` to `false`, causing the checkbox to visually remain in the indeterminate state even when all items are checked.

  ## Problem

  When using the "select all" pattern with indeterminate checkboxes:
  1. User manually checks all individual items (email, sms, push)
  2. State correctly shows `allChecked: true, indeterminate: false`
  3. However, the checkbox still displays the indeterminate icon
  4. DevTools shows `data-indeterminate` attribute still present in DOM

  **Root cause**: The `mod` prop system in Box component cannot reliably remove data attributes when they transition from
  truthy to falsy values due to React's reconciliation behavior.

  ## Solution

  Extended the existing `useEffect` hook (added in #7613) to:
  1.  Set native `indeterminate` property (original purpose - accessibility/testing)
  2.  Manually manage `data-indeterminate` attribute removal (new fix)

  Changes:
  - Added explicit `setAttribute`/`removeAttribute` calls in useEffect
  - Removed `indeterminate` from `mod` prop to prevent conflicts
  - Ensures synchronization between native property and data attribute

## Demo

  **Before:** Select all checkbox stuck in indeterminate state

  https://github.com/user-attachments/assets/802358cb-35ed-4040-a745-be0c3e2dde6b

  **After:** Select all checkbox correctly shows checked state

https://github.com/user-attachments/assets/fd54f3fd-9d5b-40fa-b0fb-a2f0ad638bac

  ## Testing

  Tested with the indeterminate demo:
  - Indeterminate state displays correctly
  -  "Select all" checkbox shows checked when all items are manually checked
  -  `toBePartiallyChecked()` test still passes (maintains #7613 functionality)
  -  CSS styling works correctly with `[data-indeterminate]` selector

  ## Related Issues

  Related to #7613 and #7608 - maintains the native `indeterminate` property functionality while fixing the data attribute
  synchronization issue.

### DIFF
--- a/packages/@mantine/core/src/components/Checkbox/Checkbox.tsx
+++ b/packages/@mantine/core/src/components/Checkbox/Checkbox.tsx
@@ -192,11 +192,8 @@ export const Checkbox = factory<CheckboxFactory>((_props, forwardedRef) => {
 
   useEffect(() => {
     if (ref && 'current' in ref && ref.current) {
-      // Set native indeterminate property for accessibility and testing
       ref.current.indeterminate = indeterminate || false;
 
-      // Manually sync data-indeterminate attribute with indeterminate state
-      // This is necessary because mod prop system cannot reliably remove attributes
       if (indeterminate) {
         ref.current.setAttribute('data-indeterminate', '');
       } else {


rtivital:
Please remove comments

chlee1001:
Done! Comments removed. 👍

rtivital:
Thanks!

packages/@mantine/core/src/components/Checkbox/Checkbox.tsx
```diff
@@ -193,6 +193,12 @@ export const Checkbox = factory<CheckboxFactory>((_props, forwardedRef) => {
   useEffect(() => {
     if (ref && 'current' in ref && ref.current) {
       ref.current.indeterminate = indeterminate || false;
+
+      if (indeterminate) {
+        ref.current.setAttribute('data-indeterminate', '');
```
predrag-nikolic-kombinat-dev, line 198:
```suggestion
        ref.current.setAttribute('data-indeterminate', 'true');
```

Some small suggestion
To be 100% compatible with the old behavior, I would suggest adding "true"

Before the PR, mantine added data-indeterminate='true'
Image

But after this PR, only `data-indeterminate` is added

Image

While this is nothing big, 
it can cause subtle issues if someone used `data-indeterminate='true'` as a selector in CSS.

chlee1001: Good point! I'll submit a follow-up PR to use `setAttribute('data-indeterminate', 'true')`
  instead of an empty string to maintain backward compatibility. Thanks!

PR #8387, opened 2025-10-24
[@mantine/dates] Allow for custom placeholders in <TimePicker />

LaExploradora:
Resolves https://github.com/orgs/mantinedev/discussions/8386.

Added three new optional props to `<TimePicker />`:

- `hoursPlaceholder`
- `minutesPlaceholder`
- `secondsPlaceholder`

Added one new optional prop to `<SpinInput />`:

- `placeholder`

Now the users can set up custom input placeholders for every type of input separately.
Did not change any of the styling due to the concern of implementing a breaking change.


rtivital:
Thanks!

PR #8391, opened 2025-10-25
[@mantine/core] Align checked and unchecked combobox dropdown options

dmitri-gb:
I'd like to propose this change which modifies the way unchecked combobox options are displayed so that they are aligned with checked options.

Here's what it looks like **before** this change:
image

And here's what it looks like **after**:
image

I find that aligning the options like this creates a better, visually more pleasing and consistent look. It also prevents the label from jumping when selecting or deselecting it. For what it's worth, this is also how native `<select>` elements are rendered on macOS:
image


rtivital:
Thanks for submitting the PR! THe new behavior must be controlled by a new prop to avoid breaking changes.

rtivital:
@dmitri-gb please complete the PR for it to get merged

dmitri-gb:
Added a new prop. I wasn't exactly sure what to call it, so I settled on `withAlignedLabels`.

rtivital:
Thanks!

PR #8396, opened 2025-10-27
[@mantine/core] Add disable prop for inline input groups

Spichkenss:
The problem is that I can't pass the disabled prop to Checkbox.Group, Radio.Group, or Switch.Group to disable all elements in the group. This PR fixes this issue.

### DIFF
--- a/apps/mantine.dev/src/pages/core/checkbox.mdx
+++ b/apps/mantine.dev/src/pages/core/checkbox.mdx
@@ -81,6 +81,10 @@ You can add any number of custom sizes with [data-size](/styles/data-attributes/
 
 <Demo data={CheckboxDemos.groupConfigurator} />
 
+## Checkbox.Group disabled
+
+<Demo data={CheckboxDemos.groupDisabled} />
+
 ## Controlled Checkbox.Group
 
 ```tsx
--- a/apps/mantine.dev/src/pages/core/switch.mdx
+++ b/apps/mantine.dev/src/pages/core/switch.mdx
@@ -25,6 +25,10 @@ function Demo() {
 }
 ```
 
+## States
+
+<Demo data={SwitchDemos.states} />
+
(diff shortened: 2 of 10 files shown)


rtivital:
Please add demos for this feature to the documentation

rtivital:
Thanks!

packages/@mantine/core/src/components/Radio/Radio.tsx
```diff
@@ -173,16 +174,15 @@ export const Radio = factory<RadioFactory>((_props, ref) => {
...
-        checked: ctx.value === rest.value,
-        name: rest.name ?? ctx.name,
-        onChange: (event: React.ChangeEvent<HTMLInputElement>) => {
-          ctx.onChange(event);
-          onChange?.(event);
-        },
-      }
-    : {};
+  const withContextProps = {
+    checked: ctx?.value === rest.value || checked || undefined,
```
rozrowe, line 178: This commit causes issues -- in the previous version, `contextProps` would have always contained `true` or `false` for `checked` when a `Radio.Group` is present, and only the `InlineInput` `data-checked` attribute would have ended up `undefined`

In the new version, `withContextProps` will end up passing `undefined` for any currently unchecked `Radio`, causing the "A component is changing an uncontrolled input to be controlled..." error when choosing a different option.

packages/@mantine/core/src/components/Radio/Radio.tsx
```diff
@@ -173,16 +174,15 @@ export const Radio = factory<RadioFactory>((_props, ref) => {
...
+    name: ctx?.name ?? rest.name,
```
dj-rabel, line 179: This line as well causes issues. It's not possible to overwrite the name anymore. `ctx.name` always gets preference over `rest.name`.

PR #8398, opened 2025-10-27
[@mantine/core]  Add Modal.Footer and Drawer.Footer components

Spichkenss:
This component is useful when we want to make the modal footer sticky so that it can be seen regardless of the scroll position, when there is a lot of content in the ModalBody


rtivital:
Please add a demo with the new component to Drawer and Modal documentation

travistruett:
Anything I can do to help here? This would be great to have.

Spichkenss:
> Anything I can do to help here? This would be great to have.

No, I think this is the final solution.

slargat:
any updates please ?

Spichkenss:
@rtivital branch updated. pls review

rtivital:
Thanks for the PR! `Modal.Footer` and `Drawer.Footer` are already implemented for the upcoming Mantine 10.0, together with `stickyHeader`/`stickyFooter` props and documentation. Adding a separate footer API in 9.x would conflict with that implementation, so I'm closing this PR in favor of 10.0.

Until 10.0 is released, a footer can be composed with `Modal.Root`/`Drawer.Root` by placing an element with `position: sticky; bottom: 0` after `Modal.Body`/`Drawer.Body`.

PR #8403, opened 2025-10-29
[@mantine/charts]: Add options to set curveType per series

drewhoener:
I ran into a situation recently where it was helpful to have a different curve type on the `rightAxis`, and I wasn't able to set it through normal props.  

This PR adds an extra prop `curveType` to the series types on `LineChart` and `AreaChart`. If `undefined`, the `curveType` prop from the Chart component will be used.


rtivital:
Thanks!

PR #8404, opened 2025-10-29
[@mantine/charts]: Pass labelFormatter prop provided by Recharts into default ChartTooltip

drewhoener:
This PR proxies the `labelFormatter` prop provided by the Recharts `Tooltip` component into the default `ChartTooltip` component.

The formatter function can be passed in via the `tooltipProps` on applicable charts


rtivital:
Thanks!

Issue #8415, opened 2025-11-02
[@mantine/tiptap] Sticky toolbar hides bubble menu


Alf-Melmac:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

8.3.6

### What package has an issue?

@mantine/tiptap

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

All

### Describe the bug

From the update to Tiptap 3 in Mantine 8.3 onwards, sticky RTE toolbars hide bubble menus. This is probably related to the [switch from Tippy.js to Floating UI](https://tiptap.dev/docs/guides/upgrade-tiptap-v2#menu-systems).
<img width="153" height="119" alt="Image" src="https://github.com/user-attachments/assets/a22f6c58-1500-497d-b046-9980ffa75120" />
For reference the previous behaviour with Mantine 8.2 and Tiptap 2: https://codesandbox.io/p/sandbox/tiptap-bubble-menu-hidden-by-toolbar-older-working-version-g6z29y

### If possible, include a link to a codesandbox with a minimal reproduction

https://codesandbox.io/p/sandbox/tiptap-bubble-menu-hidden-by-toolbar-vlh5r9

### Possible fix

The introduction of the tiptap package introduced a `z-index: 1` on the toolbar https://github.com/mantinedev/mantine/commit/a65a5c119b7eb64447c9dfb0f41d63b47aae8342#diff-e4ca2f1802c85c71520530dd7efebdbb2cafcf74c724760c423d06626b19dfe2R14. I can't see any reason for it to be there. If that is the case, simply deleting [this](
https://github.com/Alf-Melmac/mantine/blob/master/packages/%40mantine/tiptap/src/RichTextEditor.module.css#L312) is sufficient. Another option would be to give the tooltip a higher z-index.
I've created a pull request for the first idea.

### Self-service

- [x] I would be willing to implement a fix for this issue

rtivital:
Fixed in [8.3.7](https://github.com/mantinedev/mantine/releases/tag/8.3.7)

PR #8416, opened 2025-11-02
[@mantine/tiptap] Remove unused z-index to display bubble menu above toolbar

Alf-Melmac:
As proposed in #8415, simply remove the z-index definition to make the bubble menu appear above the toolbar. I don't think it's worth creating a new story for this, but please let me know if you think otherwise


rtivital:
Thanks!

Issue #8418, opened 2025-11-06
AppShell Header and Footer bounce when modals open/close


schummar:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

8.3.6

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

When a modal opens or closes - or rather whenever the body scroll is blocked - the header and footer elements bounce quickly to the side and back. I have attached a video and a repo link.

https://github.com/user-attachments/assets/55dad2ae-a838-435b-8e62-9ee899aead37

### If possible, include a link to a codesandbox with a minimal reproduction

https://codesandbox.io/p/sandbox/mantine-react-template-forked-vg5j55

### Possible fix

The issue comes from the `right` property of header and footer being animated. As can be tested in the sandbox, changing `transitionProperty` to `transform` fixes the issue. I'm not sure though why `left` and `right` were animated in the first place, so I don't know what the implications are. Maybe only changing `transitionProperty` while `data-scroll-locked` is in place?

### Self-service

- [x] I would be willing to implement a fix for this issue

rtivital:
Fixed in [8.3.7](https://github.com/mantinedev/mantine/releases/tag/8.3.7)

PR #8420, opened 2025-11-06
fix: prevent AppShell Header and Footer from bouncing when scrollbar …

schummar:
…is hidden/unhidden

Possible fix for #8418. As I said there, I'm not sure why left and right were animated in the first place, so I don't know what the implications are. Feedback welcome.

Maybe only changing transitionProperty while data-scroll-locked is in place could be an alternative.

### DIFF
--- a/packages/@mantine/core/src/components/AppShell/AppShellMediaStyles/assign-aside-variables/assign-aside-variables.test.ts
+++ b/packages/@mantine/core/src/components/AppShell/AppShellMediaStyles/assign-aside-variables/assign-aside-variables.test.ts
@@ -58,6 +58,7 @@ describe('@mantine/core/AppShell/assign-aside-variables', () => {
           '--app-shell-aside-width': '100%',
           '--app-shell-aside-transform-rtl': 'translateX(calc(var(--app-shell-aside-width) * -1))',
           '--app-shell-aside-transform': 'translateX(var(--app-shell-aside-width))',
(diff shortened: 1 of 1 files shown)


rtivital:
left and right properties are animated to allow smooth transitions when navbar/aside becomes hidden/visible, removing these values from transition would result in broken animations.

schummar:
When I tried to reproduce this - by collapsing `NavBar` and `Aside` in alt layout - I found no `left` or `right` property that was set. It seems `inset-inline-start` and `inset-inline-end` are used to position the header then. Is there another scenario that I am not considering?

Edit: Alright, seems in this context that is kind of the same as `left` and `right`. I'll look into an alternative approach then...

schummar:
I was able to make it work by using and animating `margin-inline-start`/`margin-inline-end` instead. That way it doesn't collide with `right` that react-remove-scroll sets.

While testing it, I noticed that the problem in alt layout is even worse, because there the aside element jumps as well. So I included that in the fix as well. I had to hide the aside while scroll-locked is on - otherwise part of it would be visible under the scroll bar.

Stories are updated to test the modal behavior in both layouts:

https://github.com/user-attachments/assets/a2b914a6-bb45-46da-b3ac-043ab4c29ca9

https://github.com/user-attachments/assets/d25a9a04-8745-4a80-9670-18270656e473

rtivital:
Thanks!

PR #8421, opened 2025-11-06
[@mantine/tiptap] Prevent custom onClick handler overriding editor control

Alf-Melmac:
Partially reverts #8148. The onClick handler of a rich text editor control shouldn't be overridden by a custom implementation.
Fixes Controls no longer working in menus https://codesandbox.io/p/sandbox/rte-control-menu-w3p8h3

Ref: https://discordapp.com/channels/854810300876062770/1434681144913170595


rtivital:
Thanks!

Issue #8427, opened 2025-11-11
`rightSection` in `Button` component is not working properly


shaadcode:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

8.3.7

### What package has an issue?

@mantine/core

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

None

### Describe the bug

for RTL ❌:

<img width="193" height="146" alt="Image" src="https://github.com/user-attachments/assets/9db4cec0-8967-4697-a5bc-3eb90b74edbf" />

for LTR ✔:

<img width="229" height="167" alt="Image" src="https://github.com/user-attachments/assets/72ff9475-efa0-475d-80ef-f29aa1167113" />

good for know: I am also using `DirectionProvider`.

### Self-service

- [ ] I would be willing to implement a fix for this issue

shaadcode:
The code that caused this problem:

```tsx
'use client';
import { IconX } from '@tabler/icons-react';
import { Button, Group, useDirection } from '@mantine/core';

const Home = () => {
  const dir = useDirection();

  return (
    <Group h={500} wrap="nowrap" justify="center">
      <Button lang="fa" rightSection={<IconX />} >
        {dir.dir}
      </Button>
    </Group>
  );
};

export default Home;
```

rtivital:
This issue usually happens when you do not follow rtl instructions. You must set `dir` attribute on the root element of your application. See documentation – https://mantine.dev/styles/rtl/#dir-attribute

shaadcode:
<img width="500" height="auto" alt="Image" src="https://github.com/user-attachments/assets/fb869dd4-9e21-4743-af59-9082b3444f3d" />

I checked and did everything according to the documentation but the problem still exists!
This only happens in RTL mode!
I've seen this in previous versions

rtivital:
I cannot reproduce it on mantine.dev website, provide a sandbox with a minimal reproduction.

<img width="819" height="232" alt="Image" src="https://github.com/user-attachments/assets/ddda0a0b-c4e1-4432-9ba2-c912999bc652" />

shaadcode:
Problem solved!
It was due to one of the postCSS libraries

PR #8429, opened 2025-11-11
fix(Button): RTL directionality in a Button by swapping left and right sections

hasegawa-101:
Issue #8427

  ## 📝 Description

  Fixed RTL (Right-to-Left) directionality support in the Button component. Previously, `leftSection` and `rightSection` did not swap their visual positions in RTL layouts, causing icons and sections to appear on the incorrect
  side for Arabic, Hebrew, Persian, and other RTL languages.

  ## ⛳️ Current behavior (updates)

  In RTL layouts, when using `rightSection` prop on a Button component:
  - The section appears on the **left** side visually (incorrect)
  - `leftSection` appears on the **right** side visually (incorrect)
  - The DOM order remains `[leftSection][label][rightSection]` regardless of direction

  This breaks the user experience for RTL languages where visual reading direction is right-to-left.

  ## 🚀 New behavior

  With this fix:
  - In **LTR** mode: `rightSection` appears on the right (unchanged)
  - In **RTL** mode: `rightSection` appears on the right side *visually* (DOM order swapped to `[rightSection][label][leftSection]`)
  - The component automatically detects direction using `useDirection()` hook and swaps section rendering order
  - Maintains semantic meaning: `rightSection` always means "right side" regardless of text direction

  **Changes:**
  - Added `useDirection()` hook to detect current direction context
  - Implemented conditional section swapping: `startSection` and `endSection` variables swap based on `dir`

スクリーンショット 2025-11-11 17 11 27

  ## 💣 Is this a breaking change (Yes/No):

  **No**

  This is a bug fix that improves existing functionality. The API remains unchanged - developers still use `leftSection` and `rightSection` props as before. The fix only affects visual rendering in RTL contexts, making the
  component work as expected.

  ## 📝 Additional Information

  - Tested with `DirectionProvider` in both LTR and RTL modes
  - The fix follows the same pattern used in other Mantine components that support RTL (e.g., `TabsTab`)
  - No performance impact as direction detection happens via React context

### DIFF
--- a/packages/@mantine/core/src/components/Button/Button.tsx
+++ b/packages/@mantine/core/src/components/Button/Button.tsx
@@ -13,6 +13,7 @@ import {
   PolymorphicFactory,
   rem,
   StylesApiProps,
+  useDirection,
   useProps,
   useStyles,
 } from '../../core';
@@ -167,6 +168,7 @@ export const Button = polymorphicFactory<ButtonFactory>((_props, ref) => {
     ...others
   } = props;
 
+  const { dir } = useDirection();
   const getStyles = useStyles<ButtonFactory>({
     name: 'Button',
     props,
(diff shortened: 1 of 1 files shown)


rtivital:
The PR is invalid because the reported issue is invalid.

hasegawa-101:
Got it. Thanks for checking!

PR #8439, opened 2025-11-12
[@mantine/modals] Enhance contextModal functions

AzzouQ:
This PR aim to provide a better UX using context modal.

- contextModal's functions now correctly infer modalKeys, providing autocompletion and strong typescript support.
```ts
openContextModal({ modalKey: 'myCustomModal' });
updateContextModal({ modalKey: 'myCustomModal' });
closeContextModal('myCustomModal');
```

- Fix closeContextModal to correctly work with modalKey

- Make innerProps optionnal when not needed
```ts
// Without InnerProps
openContextModal({ modalKey: 'customWithNoInner' });
updateContextModal({ modalKey: 'customWithNoInner' });

// With InnerProps
openContextModal({ innerProps: { text: 'example' }, modalKey: 'customWithInner' });
updateContextModal({ innerProps: { text: 'example' }, modalKey: 'customWithInner' });
```

- Strong typing for `ModalsProvider`
```tsx
const signInModals = { signInStepOne: SignInStepOneModal, signInStepTwo: SignInStepTwoModal };
export type SignInModals = typeof signInModals;

const registerModals = { registerStepOne: RegisterStepOneModal, registerStepTwo: RegisterStepTwoModal };
export type RegisterModals = typeof registerModals;

declare module '@mantine/modals' {
  export interface MantineModalsOverride {
    modals: SignInModals & RegisterModals;
  }
}

<ModalsProvider modals={{ ...signInModals }}> /* Typescript error, missing registerModals */
```

- 1:1 parity between function from `useModals` hook and events exported by `@mantine/modals` to avoid confusion
```ts
// Before
openContextModal('myCustomModal', { innerProps: { text: '' } }); // Hook
openContextModal({ modal: 'myCustomModal', innerProps: { text: '' } }); // Event

closeContextModal(); // Only by hook, no re-exported event

closeAll(); // Hook
closeAllModals(); // Event

// After
openContextModal({ modalKey: 'myCustomModal', innerProps: { text: '' } }); // Hook & Event
closeContextModal('myCustomModal'); // Hook & Event
closeAllModals(); // Hook & Event
```

contextModal's function will use `modalKey` to properly differentiate with `modalId`. Before, `openContextModal` required a prop named `modal` or `modalId` depending on how it was used (hook vs event). Also, contextModals can still use the proper `modalId` prop along `modalKey`, so that differentiation make sense to me.

This PR include some breaking change:

1. `closeContextModal` now works with `modalKey`, previsouly it was just a duplicate of `closeModal` and only worked with `modalId` even tho it was typed to works with `modalKey`.
2. `openContextModal` has a different signature when used with `useModals` or when imported. Now it share a single signature.
3. `closeAllModals` has different names when used with `useModals` or when imported. Now it share a single name.
4. If there's no `innerProps` for a contextModal, `innerProps` will become optional, and this PR typed it as `never` so the old fashion will produce an error. This can be modified to either allow settling a "wrong" innerProps, or keep it like this to avoid unecessary code.
```ts
openContextModal({ modal: 'myCustomModal', innerProps: {} }); // Type '{}' is not assignable to type 'undefined'.
```

I originally only wanted to enforce correct typing and provide autocompletion to contextModal, but then I realize closeContextModal was wrongly typed and the fix could be a breaking change, so I end up proposing all this.

I'll be happy to modify things if needed, and if you accept the following modification, I can also work on some documentation.

Thanks !

### DIFF
--- a/packages/@docs/demos/src/demos/modals/Modals.demo.context.tsx
+++ b/packages/@docs/demos/src/demos/modals/Modals.demo.context.tsx
@@ -31,7 +31,7 @@ function Demo() {
     <Button
       onClick={() =>
         modals.openContextModal({
-          modal: 'demonstration',
+          modalKey: 'demonstration',
           title: 'Test modal from context',
           innerProps: {
             modalBody:
--- a/packages/@docs/demos/src/demos/modals/Modals.demo.updateContextModal.tsx
+++ b/packages/@docs/demos/src/demos/modals/Modals.demo.updateContextModal.tsx
@@ -77,7 +77,7 @@ function Demo() {
     <Button
       onClick={() => {
         const modalId = modals.openContextModal({
-          modal: 'asyncDemonstration',
+          modalKey: 'asyncDemonstration',
           title: 'Processing...',
           closeOnEscape: false,
           closeOnClickOutside: false,
@@ -90,6 +90,7 @@ function Demo() {
 
         setTimeout(() => {
           modals.updateContextModal({
+            modalKey: 'asyncDemonstration',
             modalId,
             title: 'Processing Complete!',
             closeOnEscape: true,
--- a/packages/@mantine/modals/src/Modals.story.tsx
+++ b/packages/@mantine/modals/src/Modals.story.tsx
@@ -27,7 +27,7 @@ export function Usage() {
   const showContextModal = () =>
     openContextModal({
       modalId: 'context-modal',
-      modal: 'hello',
+      modalKey: 'hello',
       title: 'Context modal',
       centered: true,
       onClose: () => console.log('context modal closed'),
@@ -160,7 +160,7 @@ function CloseAllApp() {
       centered: true,
       onClose: () => {
         console.log('Close modal', modalId);
-        modals.closeAll();
+        modals.closeAllModals();
       },
     });
   };
@@ -195,7 +195,8 @@ function UpdateContextModal() {
   const modals = useModals();
 
   const handleOpenAsyncConfirmModal = () => {
-    const modalId = modals.openContextModal('asyncProcessing', {
+    const modalId = modals.openContextModal({
+      modalKey: 'asyncProcessing',
       title: 'Processing...',
       innerProps: {
         modalBody: 'You cannot close the modal during this operation.',
@@ -209,6 +210,7 @@ function UpdateContextModal() {
 
     setTimeout(() => {
       modals.updateContextModal({
+        modalKey: 'asyncProcessing',
         modalId,
         title: 'Processing Complete!',
         closeButtonProps: { disabled: false },


packages/@mantine/modals/src/events.ts
```diff
@@ -40,44 +19,41 @@ export const openConfirmModal: ModalsEvents['openConfirmModal'] = (payload) => {
...
 
-export const updateModal = (payload: { modalId: string } & Partial<ModalSettings>) =>
+export const updateModal: ModalsEvents['updateModal'] = (payload) =>
   createEvent('updateModal')(payload);
 
-export const updateContextModal = (payload: { modalId: string } & Partial<OpenContextModal<any>>) =>
+export const updateContextModal: ModalsEvents['updateContextModal'] = (payload) =>
   createEvent('updateContextModal')(payload);
 
 export const modals: {
```
AzzouQ: This could also be unified to either remove or keep all mention of `Modal` at the end of each function. I left it like it because it would have required to modified some other files in demo, which was unecessary for the core of this PR. Also since every function is already exported, I'm not sure exporting `modals` is useful at all, as it could bring some confusion about the `modals` returned by `useModal`.

packages/@mantine/modals/src/ModalsProvider.tsx
```diff
@@ -112,55 +103,72 @@ export function ModalsProvider({ children, modalProps, labels, modals }: ModalsP
...
     },
     [dispatch]
   );
 
+  const updateContextModal: ModalsContextProps['updateContextModal'] = useCallback(
+    ({ modalKey, modalId, ...newProps }) => {
+      const id =
+        modalId ??
+        stateRef.current.modals.find((m) => m.type === 'context' && m.ctx === modalKey)?.id ??
+        modalKey;
```
AzzouQ, line 159: I'm unsure if you want to have that kind of logic inside the Provider of if you prefer to have it inside the reducer. But doing it in the reducer may require a new dispatch action for updateContext like 'UPDATE_CONTEXT'

packages/@mantine/modals/src/context.ts
```diff
@@ -1,16 +1,20 @@
...
 export interface OpenConfirmModal extends ModalSettings, ConfirmModalProps {}
-export interface OpenContextModal<CustomProps extends Record<string, any> = {}>
-  extends ModalSettings {
-  innerProps: CustomProps;
-}
+
+export type ContextModalInnerProps<
+  TKey extends MantineModal,
+  P = Parameters<MantineModals[TKey]>[0]['innerProps'],
+> = keyof NonNullable<P> extends never ? { innerProps?: never } : { innerProps: P };
```
AzzouQ, line 14: `never` can be switched to `any` or `P` depending on the behavior you want.

With `innerProps?: never`: 
```ts
openContextModal({ modal: 'myCustomModal', innerProps: {} }); // Type '{}' is not assignable to type 'undefined'.
```

With `innerProps?: P`: 
```ts
openContextModal({ modal: 'myCustomModal', innerProps: {} }); // No error.
```

packages/@mantine/modals/src/events.ts
```diff
@@ -1,29 +1,8 @@
...
-  ) => string;
-  closeModal: (id: string) => void;
-  closeContextModal: <TKey extends MantineModal>(id: TKey) => void;
-  closeAllModals: () => void;
-  updateModal: (
-    payload: { modalId: string } & Partial<ModalSettings> & Partial<OpenConfirmModal>
-  ) => void;
-  updateContextModal: (payload: { modalId: string } & Partial<OpenContextModal<any>>) => void;
-};
+type ModalsEvents = Omit<ModalsContextProps, 'modals' | 'modalProps'>;
```
AzzouQ, line 5: I took the liberty of doing it like this but I can also revert it back to defining each function. Doing it like that is less error prone in my opinion as everything type related is actually concentrated in `Context.ts`.

packages/@mantine/modals/src/context.ts
```diff
@@ -1,16 +1,20 @@
...
+
+export type OpenContextModal<TKey extends MantineModal = string> = ModalSettings &
+  ContextModalInnerProps<TKey>;
```
AzzouQ, line 17: `interface` doesn't support conditionnal, so I had to switch it to `type`.

packages/@mantine/modals/src/events.ts
```diff
@@ -40,44 +19,41 @@ export const openConfirmModal: ModalsEvents['openConfirmModal'] = (payload) => {
...
   open: openModal,
-  close: closeModal,
-  closeAll: closeAllModals,
   openConfirmModal,
   openContextModal,
+  close: closeModal,
+  closeContext: closeContextModal,
+  closeAll: closeAllModals,
   updateModal,
   updateContextModal,
```
AzzouQ, line 58: This could also be unified to either remove or keep all mention of `Modal` at the end of each function. I left it like it because it would have require modifying files in demo, which was unecessary for the scope of this PR. Also since every function is already exported, I'm not sure exporting `modals` is useful at all, as it could bring some confusion about the `modals` returned by `useModal`.

rtivital:
Please ignore the copilot review, I use it only for reference, if anything needs changing, I'll let you know.

rtivital:
Thanks!

PR #8444, opened 2025-11-13
[@mantine/charts] Add cellProps to PieChart and DonutChart

ScotchAndSoda:
Similar to `barProps` in `BarChart`, allow passing `cellProps` for more granular control over Pie segments


rtivital:
Thanks!

PR #8447, opened 2025-11-13
[@mantine/core] Alert: Allow custom role prop to override default alert role

wancup:
Added support for overriding the default `role` attribute in the `Alert` component, similar to the `Notification` component (https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/components/Notification/Notification.tsx#L137-L138).
This allows setting less critical messages to `role="status"` or other roles to improve a11y.


rtivital:
Thanks!

PR #8459, opened 2025-11-17
[@mantine/core] Checkbox: Remove redundant useEffect causing indeterminate state conflicts

chlee1001:
## What
   Removes the redundant `useEffect` hook that manages the `data-indeterminate` attribute in the Checkbox component.

   ## Why
   After commit 608df1d24a added `indeterminate` to the `mod` prop, the `useEffect` hook became redundant and started causing conflicts with checkbox group behavior.

   ### Problem
   The checkbox component had two mechanisms managing the same `data-indeterminate` attribute:

   1. **Declarative (mod prop)**: `mod={{ error: !!error, indeterminate }}`
   2. **Imperative (useEffect)**: Manual DOM manipulation via `setAttribute`/`removeAttribute`

   This duplication caused checkbox group issues where:
   - Selecting individual checkboxes in a group would show `indeterminate` state incorrectly
   - The "select all" checkbox would lose its checked state when all items were selected

   ### Root Cause
   The conflict occurs because:
   - React's `mod` prop sets `data-indeterminate` during render cycle
   - `useEffect` runs after render and manipulates the same attribute
   - Timing issues between these two mechanisms led to incorrect state display

   ## Solution
   Remove the `useEffect` hook entirely. 

  **Before:** Select all checkbox stuck in indeterminate state

  https://github.com/user-attachments/assets/802358cb-35ed-4040-a745-be0c3e2dde6b

  **After:** Select all checkbox correctly shows checked state

https://github.com/user-attachments/assets/fd54f3fd-9d5b-40fa-b0fb-a2f0ad638bac

   ## Related
   - Introduced in: 608df1d24a (Add `disabled` prop for Radio.Group, Checkbox.Group and Switch.Group)
   - https://github.com/mantinedev/mantine/commit/608df1d24ae8a0052c5ff4a5a2fc34a0e0c7f636#diff-0201478ccc3ddf52937be3a9319cb3aba7e4f3afb6b5b96922a7543a9216d448

### DIFF
--- a/packages/@mantine/core/src/components/Checkbox/Checkbox.test.tsx
+++ b/packages/@mantine/core/src/components/Checkbox/Checkbox.test.tsx
@@ -95,21 +95,12 @@ describe('@mantine/core/Checkbox', () => {
     expect(screen.getByRole('checkbox')).toBeChecked();
   });
 
-  it('sets data-indeterminate and indeterminate property based on indeterminate prop', () => {
-    const ref = createRef<HTMLInputElement>();
-
-    const { rerender } = render(<Checkbox indeterminate ref={ref} />);
+  it('sets data-indeterminate attribute based on indeterminate prop', () => {
+    const { rerender } = render(<Checkbox indeterminate />);
     expect(screen.getByRole('checkbox')).toHaveAttribute('data-indeterminate');
-    expect(screen.getByRole('checkbox')).toBePartiallyChecked();
-    expect(ref.current).toBePartiallyChecked();
-
-    rerender(<Checkbox indeterminate />);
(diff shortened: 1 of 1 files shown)


rtivital:
It is not a valid fix 

```tsx
ref.current.indeterminate = indeterminate || false;
```

Is required to keep checkbox accessible.

chlee1001:
@rtivital 

  Hi! I saw the feedback the accessibility requirement for `ref.current.indeterminate = indeterminate || false;`

  I've taken a different approach in #8460 that preserves accessibility while fixing the visual transition bug.

  The key difference is that #8460 keeps the `useEffect` intact (maintaining the accessibility property) but removes only the redundant `indeterminate` from the `mod` prop to eliminate the conflict.

  Please review #8460 . Thanks!

PR #8466, opened 2025-11-19
[@mantine/core] Badge: Fix circle prop with defaultProps.radius

### LINKED ISSUE #8465: Badge: not round when using the "circle" prop with defaultProps.radius
### Dependencies check up

8.3.8

@mantine/core

Vite

All

### Describe the bug

If the theme sets a small radius as the default prop for Badge, then when using the circle prop, it does not look round.

Image

```
    <MantineProvider
      theme={{
        components: {
          Badge: Badge.extend({
            defaultProps: {
              radius: "xs",
            },
          }),
        },
      }}
    >
      <Center mih={300}>
        <Badge circle>1</Badge>
        <Badge circle radius="sm">
          2
        </Badge>
        <Badge circle radius="md">
          3
        </Badge>
        <Badge circle radius="md" size="lg">
          3
        </Badge>
      </Center>
    </MantineProvider>
```

https://codesandbox.io/p/sandbox/mantine-react-template-badge-circle-problem-2rrhmk

### Possible fix

When using circle you need to redefine border-radius as is done for the default radius "xl"

### Self-service

semimikoh:
## What does this PR do?

Fixes #8465

## Problem

When the Badge component has `defaultProps.radius` set to a small value (e.g., "xs"), the `circle` prop doesn't create a perfect circle. Instead, it shows a rounded square.

## Solution

Modified `varsResolver` to check for the `circle` prop. When `circle` is true, `--badge-radius` returns `undefined`, allowing the CSS default value (1000px) to be used.

## Before & After

**Before:** Circle badge with defaultProps radius "xs" shows as rounded square
**After:** Circle badge displays as perfect circle regardless of defaultProps radius

## Testing

Manually tested with the reproduction code from the issue - all badges now display as perfect circles.

### DIFF
--- a/packages/@mantine/core/src/components/Badge/Badge.tsx
+++ b/packages/@mantine/core/src/components/Badge/Badge.tsx
@@ -82,7 +82,7 @@ export type BadgeFactory = PolymorphicFactory<{
 }>;
 
 const varsResolver = createVarsResolver<BadgeFactory>(
-  (theme, { radius, color, gradient, variant, size, autoContrast }) => {
+  (theme, { radius, color, gradient, variant, size, autoContrast, circle }) => {
     const colors = theme.variantColorResolver({
       color: color || theme.primaryColor,
       theme,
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #8471, opened 2025-11-21
[@mantine/emotion] Add utility function to merge sx props

oguzgelal:
I've added a utility function that merges `sx` props. Main use-case would be to make it easier to pass `sx` props into custom components and merge them with the existing `sx`, like so:

```ts
import { Box } from '@mantine/core'
import { EmotionSx, mergeSx } from '@mantine/emotion'

interface MyCustomBoxProps {
  sx?: EmotionSx
}

function MyCustomBox({ sx }: MyCustomBoxProps) {
  return (
    <Box sx={mergeSx(theme => ({ ... }), sx)}>...</Box>
  )
}

function App() {
  return (
    <MyCustomBox sx={(theme) => ({ ... })} />
  )
}


```

### DIFF
--- a/packages/@mantine/emotion/src/index.ts
+++ b/packages/@mantine/emotion/src/index.ts
@@ -3,6 +3,7 @@ export { keyframes } from '@emotion/react';
 export * from './create-styles';
 export * from './emotion-transform';
 export * from './Global';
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #8486, opened 2025-11-24
[@mantine/form] export FormArrayElement type

jvllmr:
I have custom components which insert and remove list items. They need `FormArrayElement` to be typed correctly.


rtivital:
Thanks!

PR #8489, opened 2025-11-25
optional-multi-modal: Adding `shouldReplaceExistingModal` prop to mod…

Huiet:
…al manager.

Adding a property (`shouldReplaceExistingMOdal`) to be used by modal manager that allow modals to be opened without closing other modals that may be opened. This works in a way similar to nesting modals in the DOM, but can be done programatically.

The underlying change was to change the `current` modal in the modal manager to be an array, that way multiple modals can be rendered at a time instead of just one.


rtivital:
It works incorrectly with escape key. When multiple modals are opened and escape key is pressed, the first one that closes is the underlying modal. See example of incorrect behavior on the video.

https://github.com/user-attachments/assets/0c706c26-c182-4eb4-bbc4-cbdbbfeecb12

rtivital:
Thanks for the PR and for iterating on it! Stacked modals support is already implemented for the upcoming Mantine 10.0 with a shared overlay stack, so I'm closing this PR in favor of that implementation. Note that the current version of this PR also removes the close transition for all modals opened with the modals manager (closed modals are removed from `current` and unmounted immediately).

PR #8490, opened 2025-11-25
ix-notifications-dom-pollution-8470

saurabh24thakur:


packages/@mantine/notifications/src/Notifications.tsx
```diff
@@ -217,37 +217,27 @@ export const Notifications = factory<NotificationsFactory>((_props, ref) => {
...
-      <Box {...getStyles('root')} data-position="bottom-left" {...others}>
-        <TransitionGroup>{groupedComponents['bottom-left']}</TransitionGroup>
-      </Box>
-
-      <Box {...getStyles('root')} data-position="bottom-center" {...others}>
-        <TransitionGroup>{groupedComponents['bottom-center']}</TransitionGroup>
-      </Box>
+      {positions.map((pos, index) => {
+        const hasNotifications = grouped[pos].length > 0;
+        if (!hasNotifications) return null;
```
rtivital, line 222: This does not work correctly with react-transition-group – when the last notification is removed the animation does not work.

saurabh24thakur: Thank you for the feedback! You're absolutely right about the animation issue.
After investigating, I realize that properly fixing this while maintaining TransitionGroup's exit animations is more complex than initially expected. The container needs to stay mounted during the exit transition, which conflicts with the goal of removing empty containers from the DOM.
Given that:
- The 6 empty divs have minimal performance impact
- Fixing this properly requires significant complexity with react-transition-group
- The main issue was test diff noise, which can be addressed with better test selectors
I think it might be best to close this issue as "won't fix" for now. What do you think?
Alternatively, if you have a preferred approach for handling this with TransitionGroup, I'm happy to implement it!

Issue #8499, opened 2025-11-27
Inconsistent CSSObject import in create-styles.ts causing potential type mismatches


zoubingwu:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

8.3.9

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

None

## Description

In `packages/@mantine/emotion/src/create-styles.ts`, there's an import of `CSSObject` from `@emotion/serialize`:

```typescript
import { CSSObject } from '@emotion/serialize';
```

This imported `CSSObject` is then used in the generic constraint of the `createStyles` function:

```typescript
export function createStyles<
  Key extends string = string,
  Params = void,
  Input extends Record<Key, CSSObject> = Record<Key, CSSObject>,
>(input: ((theme: MantineTheme, params: Params, helpers: EmotionHelpers) => Input) | Input) {
  // ...
}
```

However, the package also defines its own `CSSObject` interface in `packages/@mantine/emotion/src/types.ts`:

```typescript
export interface CSSObject
  extends CSSPropertiesWithMultiValues,
    CSSPseudos,
    CSSOthersObject,
    CSSTssSpecials {}
```

## Problem

When users import and use the locally defined `CSSObject` and related types from `@mantine/emotion` to annotate return types in their `createStyles` calls, it can lead to type mismatches because:

1. The generic constraint uses `@emotion/serialize`'s `CSSObject`
2. Users might use the locally exported `CSSObject` from `@mantine/emotion`
3. These two interfaces are not identical. The local one has Mantine-specific extensions:
   - `ref` property (via `CSSTssSpecials`)
   - `ArrayCSSInterpolation` uses `Array<CSSInterpolation>` instead of Emotion's `ReadonlyArray<CSSInterpolation>`
   - `CSSPropertiesWithMultiValues` uses `Array<string>` instead of Emotion's `ReadonlyArray<string>`

```typescript
// Mantine's version (packages/@mantine/emotion/src/types.ts):
export type CSSTssSpecials = {
  ref?: string;
};

export interface ArrayCSSInterpolation extends Array<CSSInterpolation> {}

export type CSSPropertiesWithMultiValues = {
  [K in keyof CSSProperties]: CSSProperties[K] | Array<Extract<CSSProperties[K], string>>;
};

// vs Emotion's version (@emotion/serialize):
[...]

PR #8500, opened 2025-11-27
[@mantine/emotion] Update import path for CSSObject in createStyles

zoubingwu:
fix https://github.com/mantinedev/mantine/issues/8499


rtivital:
Thanks!

Issue #8502, opened 2025-11-27
Checkbox description should not display the pointer cursor


radui1398:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

8.3.9

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

All

### Describe the bug

Since clicking on the description element is not checking the checkbox component, the cursor should not be pointer when it's set to pointer in the mantine theme.

### If possible, include a link to a codesandbox with a minimal reproduction

https://codesandbox.io/p/devbox/mantine-react-template-forked-r93mgp

### Self-service

- [x] I would be willing to implement a fix for this issue

PR #8503, opened 2025-11-28
[@mantine/core] Set description on default cursor

radui1398:
Fixes: https://github.com/mantinedev/mantine/issues/8502


rtivital:
Thanks!

PR #8504, opened 2025-11-28
[@mantine/core] Apply default styles also to shadow DOM hosts

dmitri-gb:
This change applies default Mantine styles also to the `:host` element (in addition to `:root` and `body`). This makes it a little easier to use Mantine inside the shadow DOM.

Previously one could _almost_ achieve this with `<MantineProvider cssVariablesSelector=":host">`, however this didn't fully apply all styles (e.g. global font-family, line-height, color-scheme etc were not applied). It also didn't benefit from variable deduplication like `:root` does.


dmitri-gb:
For comparison, baseline styles [in Tailwind](https://github.com/tailwindlabs/tailwindcss/blob/563a016f963795f99bae651451d92a93919f4777/packages/tailwindcss/preflight.css#L28-L29) are similarly applied to `html, :host`.

rtivital:
Thanks!

PR #8505, opened 2025-11-29
fix: Checkbox description should not display pointer cursor

### LINKED ISSUE #8502: Checkbox description should not display the pointer cursor

hlongc:
## 📋 Description

Fixes #8502

When the Mantine theme's `cursorType` is set to `'pointer'`, the Checkbox component's description and error elements incorrectly inherit the pointer cursor. This creates a misleading UX, as clicking these elements does not toggle the checkbox.

## 🔧 Changes

- Set `cursor: default` on `.description` and `.error` classes in [InlineInput.module.css](cci:7://file:///Users/hulongchao/Documents/code/source-code/mantine/packages/@mantine/core/src/utils/InlineInput/InlineInput.module.css:0:0-0:0)
- Enhanced [CursorPointer](cci:1://file:///Users/hulongchao/Documents/code/source-code/mantine/packages/@mantine/core/src/components/Checkbox/Checkbox.story.tsx:108:0-126:1) story with description and error examples for visual verification

## ✅ Testing

- ✅ All stylelint checks pass
- ✅ Code formatted with prettier
- ✅ Added visual test story to verify the fix

## 📸 Behavior

**Before:** Description and error show pointer cursor (misleading)
**After:** Description and error show default cursor (correct)

The label still correctly shows the pointer cursor and remains clickable.

### DIFF
--- a/packages/@mantine/core/src/components/Checkbox/Checkbox.story.tsx
+++ b/packages/@mantine/core/src/components/Checkbox/Checkbox.story.tsx
@@ -109,7 +109,19 @@ export function CheckboxGroup() {
 export function CursorPointer() {
   return (
     <MantineThemeProvider theme={{ cursorType: 'pointer' }}>
-      <Checkbox label="Hello" value="1" />
+      <Stack style={{ padding: 40 }}>
+        <Checkbox label="Without description" value="1" />
+        <Checkbox
+          label="With description"
+          description="Description should have default cursor, not pointer"
+          value="2"
+        />
+        <Checkbox
+          label="With error"
+          error="Error should have default cursor, not pointer"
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks for submitting the PR! It is already resolved by other contributor – https://github.com/mantinedev/mantine/pull/8503

PR #8508, opened 2025-12-01
fix: Change modal property from 'modal' to 'modalKey' in docs

Tuck1297:
When migrating from an older version to v8.3.8 my linter caught this change. Requesting this change to keep the docs consistent. Thanks


rtivital:
This change was a mistake and was reverted in 8.3.9. The documentation update is not needed.

Issue #7591, opened 2025-03-14
Cannot read properties of undefined (reading 'onClick') on Popover and Menu


thomaslecoeur:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.17.2

### What package has an issue?

@mantine/core

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

Not applicable – issue is not related to the browser

### Describe the bug

I get this error sometimes on Menu and Popover. I'm using Next.JS 15 and React 19. Error was not happening with Next.JS 14 and previous React version.

```
 ⨯ TypeError: Cannot read properties of undefined (reading 'onClick')
    at @mantine/core/MenuTarget (../../../../src/components/Menu/MenuTarget/MenuTarget.tsx:30:52)
  28 |   const _childrenProps = children.props as any;
  29 |
> 30 |   const onClick = createEventHandler(_childrenProps.onClick, () => {
     |                                                    ^
  31 |     if (ctx.trigger === 'click') {
  32 |       ctx.toggleDropdown();
  33 |     } else if (ctx.trigger === 'click-hover') { {
  digest: '3306055790'
}
```

Here is the JSX for this error :

```
<Menu shadow="md" width={200}>
  <MenuTarget>
    <ActionIcon size="lg" variant="transparent" aria-label="Settings">
      <IconDotsVertical size={16} stroke={1.5} />
    </ActionIcon>
  </MenuTarget>
  <MenuDropdown>
    <ABomDeleteArticleButton abom={node.abom} abomArticleId={node.articleId}>
      <MenuItem leftSection={<IconTrash size={16} />}>
        Supprimer l&apos;article
      </MenuItem>
    </ABomDeleteArticleButton>
  </MenuDropdown>
</Menu>
```

### Self-service

- [ ] I would be willing to implement a fix for this issue

rtivital:
Thanks for submitting the issue! In order to proceed, we need a codesandbox with a minimal reproduction. You can use this template to create it – https://codesandbox.io/s/mantine-react-template-q4dd5d?file=/src/App.js.

[...]

PR #8522, opened 2025-12-08
[@mantine/core] Fix SSR-related errors in Tooltip, Popover.Target etc

dmitri-gb:
Occasionally, components like `Tooltip`, `Popover.Target`, `HoverCard.Target` and others, when used in a Next.js App Router based application, will cause an error similar to one of the following:

```
Uncaught TypeError: Cannot read properties of undefined (reading 'onClick')
    at @mantine/core/MenuTarget (MenuTarget.tsx:30:52)
```
```
Uncaught TypeError: Cannot read properties of undefined (reading 'className')
    at @mantine/core/Tooltip (Tooltip.tsx:352:14)
```
```
Uncaught Error: Element type is invalid: expected a string (for built-in components) or a class/function (for composite components) but got: undefined. You likely forgot to export your component from the file it's defined in, or you might have mixed up default and named imports.
```

(Some have already been reported in #7591.)

The reason for this is that the `children` prop received by these components can have an unexpected shape that these components cannot handle. Namely, sometimes `children` can be the [`Lazy`](https://github.com/facebook/react/blob/ec9cc003d232b5a3eed735311df152463a883b32/packages/shared/ReactSymbols.js#L33) type, which doesn't have a `props` field and cannot be passed directly to `cloneElement`.

More generally speaking, APIs like `cloneElement` and `children.props` seem to be ["soft deprecated"](https://github.com/facebook/react/issues/32392#issuecomment-3133110279), so they don't work well with newer React features. Nevertheless, React team [_did_ add support](https://github.com/facebook/react/pull/28284) for the Lazy type in `Children.map` and `Children.toArray` APIs. `Tooltip` and others can use them to handle these scenarios.

In the long term, `Children` APIs might also become deprecated though, so this is only a stopgap solution.

FYI, a similar workaround has also been applied to [radix-ui](https://github.com/radix-ui/primitives/pull/3680), although they use a different method. I find the `Children.toArray` approach to be nicer, as it delegates all the special handling to React itself.

### DIFF
--- a/packages/@mantine/core/src/components/Combobox/ComboboxEventsTarget/ComboboxEventsTarget.tsx
+++ b/packages/@mantine/core/src/components/Combobox/ComboboxEventsTarget/ComboboxEventsTarget.tsx
@@ -1,6 +1,6 @@
-import { Children, cloneElement } from 'react';
+import { cloneElement } from 'react';
 import { useMergedRef } from '@mantine/hooks';
-import { factory, Factory, getRefProp, isElement, useProps } from '../../../core';
+import { factory, Factory, getRefProp, getSingleElementChild, useProps } from '../../../core';
 import { useComboboxContext } from '../Combobox.context';
 import { useComboboxTargetProps } from '../use-combobox-target-props/use-combobox-target-props';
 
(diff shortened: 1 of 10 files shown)


rtivital:
There appears to be a lot of repeating logic. Please create a utility function for this.

dmitri-gb:
> There appears to be a lot of repeating logic. Please create a utility function for this.

Added one.

rtivital:
Thanks!

PR #8526, opened 2025-12-09
[@mantine/core] Checkbox: Add aria-describedby support for error and description (#8340)

### LINKED ISSUE #8340: Error message of an erroneous checkbox component is not accessible to voice readers
### Dependencies check up

8.3.1

@mantine/core

Vite

All

### Describe the bug

Image

Image

_No response_

### Possible fix

### Self-service

hlongc:
## What

Fixes #8340

Checkbox error messages are now accessible to screen readers like VoiceOver. Previously, the error and description text were not announced when focusing on the checkbox input.

## Changes

- Added `aria-describedby` attribute to checkbox input element
- The attribute references the IDs of description and error elements
- Updated InlineInput to accept optional `descriptionId` and `errorId` props
- Added test cases to verify the accessibility improvements

## How it works

When a Checkbox has an error or description:
1. Generate unique IDs for description/error elements
2. Set these IDs on the actual description/error elements
3. Link the checkbox input to these elements via `aria-describedby`

This ensures screen readers properly announce the description and error messages when users focus on the checkbox.

## Testing

All tests pass (54/54):
- ✅ aria-describedby set correctly when only error is present
- ✅ aria-describedby set correctly when only description is present  
- ✅ aria-describedby references both IDs when both are present
- ✅ No aria-describedby when neither is present

Tested with Mac VoiceOver - error messages are now properly announced. 🎉

### DIFF
--- a/packages/@mantine/core/src/components/Checkbox/Checkbox.test.tsx
+++ b/packages/@mantine/core/src/components/Checkbox/Checkbox.test.tsx
@@ -133,4 +133,45 @@ describe('@mantine/core/Checkbox', () => {
     render(<Checkbox {...defaultProps} rootRef={ref} />);
     expect(ref.current).toBeInstanceOf(HTMLDivElement);
   });
+
+  it('sets aria-describedby on input when error is present', () => {
+    render(<Checkbox label="test-label" error="test-error" />);
+    const input = screen.getByRole('checkbox');
+    const ariaDescribedBy = input.getAttribute('aria-describedby');
+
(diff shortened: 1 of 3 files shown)


robert-j-webb:
I have tested this PR and it does in fact fix the problem, without the PR, the error messages are not read by mac voice over.

rtivital:
Thanks for validation!

rtivital:
The issue was resolved in other PR

PR #8528, opened 2025-12-09
[@mantine/hooks] use-focus-within: Fix stale closure in onFocus/onBlur callbacks

### LINKED ISSUE #8507: `useFocusWithin` calls outdated `onFocus` and `onBlur` handlers
### Dependencies check up

8.3.9

@mantine/hooks

create-react-app (CRA)

All

### Describe the bug

The `useFocusWithin` hook calls outdated callback handlers that capture stale state values from the initial render.

#### Reproduction

In the following code, the ref returned by `useFocusWithin` is attached to a form element:

```tsx
import { useFocusWithin } from "@mantine/hooks";
import { useState } from "react";

function MyComponent() {
  const [focusCount, setFocusCount] = useState(0);
  const [blurCount, setBlurCount] = useState(0);

  const { focused, ref } = useFocusWithin({
    onFocus: () => {
      console.log("focused!");
      setFocusCount(focusCount + 1);
    },
    onBlur: () => {
      console.log("blurred!");
      setBlurCount(blurCount + 1);
    },
  });
  return (
    <div>
      <p>focusCount = {focusCount}</p>
      <p>blurCount = {blurCount}</p>
      <p>focused = {focused ? "true" : "false"}</p>
      <form ref={ref}>
        <input type="text" placeholder="inside form" />
      </form>
      <button>outside form</button>
    </div>
  );
}
```

#### Expected behavior

When repeatedly moving focus between the input (inside the form) and the button (outside the form), the corresponding value of `focusCount` or `blurCount` should increment each time the focus changes.

#### Actual behavior

`focusCount` and `blurCount` never exceed 1, even though "focused!" and "blurred!" are logged to the console every time the focus changes.
This suggests that `useFocusWithin` continues to call the `onFocus` and `onBlur` handlers from the initial render, which captured the initial values of `focusCount` and `blurCount` (both 0).

https://codesandbox.io/p/devbox/mantine-react-template-forked-jccfsr

### Possible fix

In [the source code of `useFocusWithin`](https://github.com/mantinedev/mantine/blob/e682a19602bcdb51dc5202bd61bfd262c4b26cf2/packages/%40mantine/hooks/src/use-focus-within/use-focus-within.ts), there are places where `previousNode.current` is read, but it is never updated. Assigning `previousNode.current = node` within `callbackRef` may resolve this issue (though I haven't verified this yet).

### Self-service

muhammedaksam:
The `useFocusWithin` hook was updated to prevent stale closures in the `onFocus` and `onBlur` callbacks. This change ensures that the latest state values are used when these callbacks are invoked, allowing the focus and blur counts to increment correctly. Additionally, tests were added to verify that the callbacks are called with the latest state. Fixes #8507

### DIFF
--- a/packages/@mantine/hooks/src/use-focus-within/use-focus-within.story.tsx
+++ b/packages/@mantine/hooks/src/use-focus-within/use-focus-within.story.tsx
@@ -19,3 +19,34 @@ export function Usage() {
     </div>
   );
 }
+
+export function StaleClosureFix() {
+  const [focusCount, setFocusCount] = useState(0);
+  const [blurCount, setBlurCount] = useState(0);
+
+  const { focused, ref } = useFocusWithin({
+    onFocus: () => {
+      console.log('focused!');
+      setFocusCount(focusCount + 1);
+    },
+    onBlur: () => {
+      console.log('blurred!');
+      setBlurCount(blurCount + 1);
+    },
+  });
+
+  return (
+    <div style={{ padding: 40 }}>
+      <p>focusCount = {focusCount}</p>
+      <p>blurCount = {blurCount}</p>
+      <p>focused = {focused ? 'true' : 'false'}</p>
+      <form ref={ref} style={{ padding: 20, border: '2px solid gray' }}>
+        <input type="text" placeholder="Click here (inside form)" />
+      </form>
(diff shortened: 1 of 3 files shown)


rtivital:
Thanks!

PR #8532, opened 2025-12-10
[@mantine/dates] TimePicker: Fix AM/PM input accepting arbitrary text

### LINKED ISSUE #8530: Bug: TimePicker with presets & format="12h" parses values incorrectly for AM/PM
### Dependencies check up

 "@mantine/dates": "^8.3.10"

@mantine/dates

Vite

Chrome

### Describe the bug

Image

Image

The AM/PM Input seems to accept any value (when used in combination with presets and 12h format).

The expected behaviour would be the same as the default example shown [here](https://mantine.dev/dates/time-picker/#12-hour-format) - i.e. user should not be allowed to type arbitrarily and should only select from valid AM/PM values.

The issue is two-fold because after accepting incorrect values, it also reverts back to the default AM value - regardless of whether or not the input ends up saying `PMm` (or similar)

https://codesandbox.io/p/devbox/mantine-react-template-forked-wx2m4z

### Possible fix

_No response_

### Self-service

muhammedaksam:
The AM/PM input in the TimePicker component previously allowed users to enter arbitrary text, which caused incorrect parsing of values. This update enforces read-only behavior on the AM/PM input, preventing any invalid input and ensuring that users can only select from valid AM/PM values.

Fixes #8530

### DIFF
--- a/packages/@mantine/dates/src/components/TimePicker/AmPmInput/AmPmInput.tsx
+++ b/packages/@mantine/dates/src/components/TimePicker/AmPmInput/AmPmInput.tsx
@@ -85,7 +85,8 @@ export const AmPmInput = forwardRef<HTMLSelectElement, AmPmInputProps>(
           ref={ref as any}
           value={displayValue}
           size={inputSize}
-          onChange={(event) => !readOnly && onChange(event.target.value || null)}
+          readOnly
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

Issue #8535, opened 2025-12-11
NotificationProps variant type is hardcoded


jakub-gaik:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

8.3.5

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Not applicable – issue is not related to the browser

### Describe the bug

NotificationProps defines the `variant` type instead of using the one defined in the StylesApiProps like other comonents. This prevents the user from overwriting the type and adding specific variants to type suggestion.

https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/components/Notification/Notification.tsx#L36

```ts
export interface NotificationProps
  extends BoxProps,
    StylesApiProps<NotificationFactory>,
    ElementProps<'div', 'title'> {
  variant?: string;

  /** Called when the close button is clicked */
  onClose?: () => void;

  /** Controls notification line or icon color, key of `theme.colors` or any valid CSS color @default `theme.primaryColor` */
  color?: MantineColor;
...
}
```

### Possible fix

Remove the hardcoded type from NotificationProps

```ts
export interface NotificationProps
  extends BoxProps,
    StylesApiProps<NotificationFactory>,
    ElementProps<'div', 'title'> {

  /** Called when the close button is clicked */
  onClose?: () => void;

  /** Controls notification line or icon color, key of `theme.colors` or any valid CSS color @default `theme.primaryColor` */
  color?: MantineColor;
...
}
```

### Self-service

- [x] I would be willing to implement a fix for this issue

PR #8537, opened 2025-12-11
[@mantine/core] Notification: Remove variant prop from NotificationProps

jakub-gaik:
# What
Fixes https://github.com/mantinedev/mantine/issues/8535

Removes `variant` property type definition from `NotificationProps` to enable overwriting similar to other components.

# Changes

- Removed `variant` property from `NotificationProps` interface
- Removed `variant` property extraction inside the `Notification` component render (it is now passed via `other`

# How it works

- Extending `NotificationProps` in the mantine.d.ts (ex) file with custom `variant` definition will now allow the type overwriting instead of hardcoded `string` type.


rtivital:
Thanks!

PR #8546, opened 2025-12-16
feat: Introduce Popover.Context and onOpen handler

Newbie012:
Adds `Popover.Context`, a render-prop component providing access to the popover context, including new `onOpen`. This allows custom state control while still using built-in components. I also Included documentation and a test case for hover behavior.

Example:

```ts
<Popover>
  <Popover.Target>
    <Popover.Context>
      {({ onOpen, onClose }) => (
        <button type="button" onMouseEnter={onOpen} onMouseLeave={onClose}>
          test-target
        </button>
      )}
    </Popover.Context>
  </Popover.Target>

  <Popover.Dropdown>
    <div>test-dropdown</div>
  </Popover.Dropdown>
</Popover>
```

### DIFF
--- a/apps/mantine.dev/src/pages/core/popover.mdx
+++ b/apps/mantine.dev/src/pages/core/popover.mdx
@@ -191,27 +191,6 @@ events, for example, `mouseup` and `touchend`:
 
 <Demo data={PopoverDemos.clickOutsideEvents} />
 
-## onOpen and onClose
-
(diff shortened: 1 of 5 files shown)


shwu-jt:
I think we can make `<Popover>` to take `ReactNode | (args: {onOpen, onClose}) => ReactNode` instead so we don't need another layer of `<Popover.Context>`.

PR #8573, opened 2025-12-30
docs: Update /colors-generator example to set myColor as primaryColor

iamdevlinph:
I generated a project using the [vite template](https://github.com/mantinedev/vite-template). I followed the usage example in the [colors generator](https://mantine.dev/colors-generator/) and it doesn't update the default blue color.

Turns out we have to set the `myColor` as `primaryColor`.


rtivital:
Thanks!

PR #8575, opened 2025-12-30
docs: Add mantineHtmlProps to html element for SSR

corydeppen:
Updates the getting started page to mention adding `mantineHtmlProps` to avoid a hydration warning if using SSR.


rtivital:
Thanks!

Issue #8578, opened 2026-01-02
Bug: Documentation issues in `CodeHighlighting` usage examples


Masterquiz:
### Link to the page where something is not right

https://mantine.dev/x/code-highlight/#create-custom-adapter

### What is wrong?

While this is the interface of `CodeHighlightAdapter`:
```typescript
interface CodeHighlightAdapter {
  loadContext?: () => Promise<any>;
  getHighlighter: (ctx: any) => Highlighter;
}
```

In the example the adapter has `createHighlighter` key:
```typescript
export const customShikiAdapter: CodeHighlightAdapter = {
  loadContext: loadShiki,
  createHighlighter: (ctx) => {
    ...
  },
}
```

Also the first line is wrong:
```typescript
import type { CodeHighlightAdapter, stripShikiCodeBlocks } from '@mantine/code-highlight';
```
1. `CodeHighlightAdapter` isn't exported by the package.
2. `stripShikiCodeBlocks` isn't a type, so it cannot be imported like so.

NOTE: I think that `CodeHighlightAdapter` should be exported, to avoid hardcoding three types/interfaces (`HighlighterInput`, `Highlighter`, `CodeHighlightAdapter`).

### Self-service

- [ ] I would be willing to implement a fix for this issue

PR #8581, opened 2026-01-04
[@mantine/code-highlight] fix missing export and incorrect docs

rilrom:
This PR:

- Exports the `CodeHighlightAdapter` type.
  - This type is used in existing documentation so I've used that as an assumption that it should be exported.
- Updates incorrect documentation.

Fixes: #8578.


rtivital:
Thanks!

PR #8584, opened 2026-01-05
[@mantine/core] Select: prevent iOS autofill bubble on readonly

rilrom:
Fixes [#6103](https://github.com/mantinedev/mantine/issues/6103).

If the `Select` component is `readonly` or not searchable, we prevent the pointer event from continuing, which is the event that causes the autofill bubble to appear (at least in modern iOS, I did not test earlier versions).


rilrom:
Closing this as it prevents text selection of read only inputs, something I would consider a breaking change.

PR #8591, opened 2026-01-07
Fix issue 8539

### LINKED ISSUE #8539: ScrollArea with horizontal scroll should not prevent parent vertical scroll
## What version of @mantine/* packages do you have in package.json?
7.17.x

## What package has an issue?
@mantine/core

## What framework are you using?
React Router

## In which browsers did you experience the issue?
All

## Describe the bug
When a horizontal `ScrollArea` (using `scrollbars="x"`) is nested inside a vertically scrollable container, vertical wheel events are captured and do not propagate to the parent.

**Expected:** A horizontal-only ScrollArea should never consume vertical scroll events. They should always propagate to the parent.

**Actual:** Vertical wheel events are captured by the horizontal ScrollArea, making it impossible to vertically scroll the page while the cursor is over it.

## CodeSandbox link
_Will add reproduction_

## Possible fix
Only capture wheel events where `deltaX !== 0` when `scrollbars="x"`.

Hosung99:
# Problem
This PR fixes issue  #8539 

# Solution
Added auto for x,y axis.

**before**

https://github.com/user-attachments/assets/d5837edf-6ab1-42b1-8fb0-bb16be4f9751

**after**

https://github.com/user-attachments/assets/9fb58175-843a-42cc-a4f4-995d0a1cc80f

Fixes #8539

### DIFF
--- a/packages/@mantine/core/src/components/ScrollArea/ScrollArea.tsx
+++ b/packages/@mantine/core/src/components/ScrollArea/ScrollArea.tsx
@@ -105,12 +105,24 @@ const defaultProps = {
 } satisfies Partial<ScrollAreaProps>;
 
 const varsResolver = createVarsResolver<ScrollAreaFactory>(
-  (_, { scrollbarSize, overscrollBehavior }) => ({
-    root: {
-      '--scrollarea-scrollbar-size': rem(scrollbarSize),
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #8595, opened 2026-01-09
[@mantine/core] Fix popover aria-controls attribute

### LINKED ISSUE #8594: Accessibility problem in Menu component - @mantine/core
### Dependencies check up

8.3.11

@mantine/core

Vite

### Describe the bug

The Menu Target component has a attribute called aria-controls. This attribute has the id ( 'aria-controls': "dropdown-id") of the dropdown that it controls.
The problem is when the dropdown is closed the aria-controls attribute is still existing on the Menu Target while the dropdown is not even visible in the DOM, so it points to nothing.

Currently, I am getting an accessibility warning in storybook:

Unable to determine if aria-controls referenced ID exists on the page while using aria-haspopup: aria-controls="mantine-xxxxxxxxx-dropdown"

This is very similar to this issue: https://github.com/mantinedev/mantine/issues/7085. I believe it would be a similarly simple fix. 

_No response_

### Possible fix

I think there is a possible fix in https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/components/Popover/PopoverTarget/PopoverTarget.tsx#L51.  
**Proposed change:** 'aria-controls': ctx.opened ? ctx.getDropdownId() : undefined. 

### Self-service

jonathanhuynh70:
Resolves #8594 
Checks if the dropdown has been opened, if it's true, set the "aria-controls" attribute, else, set as undefined.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "mantine-a91763c0e2",
-  "version": "8.3.11",
+  "version": "8.3.12",
   "description": "Mantine Components Monorepo",
   "packageManager": "yarn@4.10.3",
   "license": "MIT",
--- a/packages/@mantine/carousel/package.json
+++ b/packages/@mantine/carousel/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@mantine/carousel",
-  "version": "8.3.11",
+  "version": "8.3.12",
   "description": "Embla based carousel",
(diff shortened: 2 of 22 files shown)


jonathanhuynh70:
@rtivital Hey just wait on review for this MR, whenever you're available!

rtivital:
Thanks!

PR #8596, opened 2026-01-09
fix(modals): improve type inference for openContextModal

### LINKED ISSUE #8588: openContextModal modal name not type-checked despite MantineModalsOverride augmentation
## Description

When using module augmentation to add type safety to context modals as documented, the `modal` property in `openContextModal()` accepts any string instead of being restricted to the defined modal keys. However, `innerProps` IS correctly type-checked.

## Steps to reproduce

1. Define modals and augment `MantineModalsOverride` as per docs:

```typescript
// modals/index.ts
export const modals = {
  createDepartment: AddDepartmentModal,
  createUser: CreateUserModal,
} as const;

// registry.d.ts
import type { modals } from './modals';

declare module '@mantine/modals' {
  export interface MantineModalsOverride {
    modals: typeof modals;
  }
}

export {};
```

2. Call `openContextModal` with an invalid modal name:

```typescript
openContextModal({
  modal: 'invalidModalName', // ❌ No TypeScript error, but should error
  title: 'Test',
  innerProps: { orgId: 123 },
});
```

3. Note that `innerProps` IS type-checked correctly:

```typescript
openContextModal({
  modal: 'createDepartment',
  title: 'Test',
  innerProps: { wrongProp: 123 }, // ✅ TypeScript error as expected
});
```

## Expected behavior

`modal: 'invalidModalName'` should produce a TypeScript error like:
```
Type '"invalidModalName"' is not assignable to type '"createDepartment" | "createUser"'
```

## Actual behavior

No error. Any string is accepted for `modal`.

## Root cause analysis

The issue is in `context.d.ts`:

```typescript
export type MantineModalsOverwritten = MantineModalsOverride extends {
  modals: Record<string, React.FC<ContextModalProps<any>>>;
} ? MantineModalsOverride : {
  modals: Record<string, React.FC<ContextModalProps<any>>>;
};

export type MantineModals = MantineModalsOverwritten['modals'];
export type MantineModal = keyof MantineModals;
```

When TypeScript evaluates `MantineModalsOverwritten['modals']`, it intersects the type with the conditional's constraint (`Record<string, ...>`), causing `keyof` to return `string` instead of the literal union.

**Proof:**
```typescript
// Direct access works correctly:
type ModalsDir = MantineModalsOverride['modals'];
type ModalDir = keyof ModalsDir;  // "createDepartment" | "createUser" ✓

// Through conditional type doesn't:
type Modals = MantineModalsOverwritten['modals'];
type Modal = keyof Modals;  // string ✗
```

## Suggested fix

Replace the conditional type pattern with direct property access, or use a different conditional structure that doesn't cause type widening:

```typescript
// Option 1: Direct access when modals property exists
export type MantineModals = MantineModalsOverride extends { modals: infer M }
  ? M
  : Record<string, React.FC<ContextModalProps<any>>>;

// Option 2: Use a mapped type or other pattern
```

## Environment

- `@mantine/modals`: 8.3.11
- TypeScript: 5.9.3
- tsconfig: `moduleResolution: "bundler"`, `strict: false`

alecsis77:
## Summary
Fixes the type inference issue where `openContextModal` accepts any string for the `modal` parameter instead of being restricted to the defined modal keys from module augmentation.

The fix uses a simpler `infer` pattern that preserves the literal types from the user's module augmentation.

## Changes
- Simplified `MantineModals` type to use direct `infer` pattern instead of intermediate `MantineModalsOverwritten` type

Fixes #8588


rtivital:
Please fix errors in tests

rtivital:
Closing as stale PR

Issue #8600, opened 2026-01-10
RangeSlider don't respect theme 'vars' and 'styles' fields


AgentRBY:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

8.3.12

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

All

### Describe the bug

If you change `vars` or `styles` for `RangeSlider` in the theme, they will not work.

### If possible, include a link to a codesandbox with a minimal reproduction

https://codesandbox.io/p/sandbox/mantine-react-template-forked-ctr2j4

### Possible fix

RangeSlider.tsx, line 218

```diff
-    name: 'Slider',
+    name: 'RangeSlider',
```

### Self-service

- [x] I would be willing to implement a fix for this issue

PR #8601, opened 2026-01-10
fix: update RangeSlider name in useStyles

AgentRBY:
closes https://github.com/mantinedev/mantine/issues/8600


rtivital:
Thanks!

PR #8611, opened 2026-01-14
[@mantine/core] ScrollArea: Fix `overscrollBehaviour` prop not being available in `ScrollArea.Autosize` component

yverby:
Fixed: `overscrollBehaviour` prop for `ScrollArea.Autosize`


rtivital:
Thanks!

Issue #8608, opened 2026-01-14
ScrollArea onBottomReached not work when zoom in/out browser


tinhtinh95:
### Dependencies check up

- [ ] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.17.7

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

All

### Describe the bug

ScrollArea onBottomReached not work when zoom in/out browser

### Self-service

- [ ] I would be willing to implement a fix for this issue

PR #8616, opened 2026-01-15
Increase subpixel tolerance to fix `ScrollArea.onBottomReached()` while zoomed out.

robert-j-webb:
Fixes https://github.com/mantinedev/mantine/issues/8608

I could only reproduce this issue when zoomed out to 33%, where I would see the difference between `(scrollTop - (scrollHeight - clientHeight) == 0.77`.   I found I could no longer reproduce the issue, even at extreme zooms with the tolerance set to 0.8. It does pass tests.


rtivital:
Thanks!

Issue #7050, opened 2024-10-29
Bug with horizontal scrolling in ScrollArea inside the Drawer


AlexeyEsin:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.13.4

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

`ScrollArea` with horizontal scrolling works with errors inside the `Drawer`.

One of the reproduction options that you can try out in CodeSandbox:
1. Scroll all the way to the right using `Shift` and the mouse wheel.
2. Scroll down and then back up to the top.
3. Try to scroll through `Shift` (horizontal scrolling does not work).

I know that there is a [Usage with scrollArea](https://mantine.dev/core/drawer/#usage-with-scrollarea) section in the `Drawer` documentation, but it is not suitable for my situation. I need a specific element inside the `Drawer` to be scrollable, not the entire content. Besides this problem can potentially occur not only inside the `Drawer`. I just gave an example that I came across.

I am using Windows 11 and Chrome v130.0.6723.70. There is no problem in Safari 18.0.

### If possible, include a link to a codesandbox with a minimal reproduction

https://codesandbox.io/p/sandbox/mantine-react-template-forked-cnc5ff

### Self-service

- [ ] I would be willing to implement a fix for this issue

FurquanAnwer:
I would like to take this issue. Could you assign me this ?

rtivital:
You are welcome to submit a PR with a fix

descent0:
Hey , I would like to solve this issue

rtivital:
You are welcome to submit a PR with a fix

AyushTyagi2:
Hi, I'd like to work on this issue.

AlexeyEsin:
@rtivital I recently discovered that horizontal scrolling in the `Select` component also does not work inside `Drawer`. I found out that disabling `lockScroll` solves the problem. At the same time, `lockScroll` is also set to `true` by default in the `Modal` component, but this bug does not occur there.

I didn't see any difference between enabling and disabling `lockScroll` in `Drawer`, so the question arises: why is it needed, and will there be other problems if I disable it via `createTheme` and `defaultProps`?

robert-j-webb:
[...]

Issue #8037, opened 2025-07-02
Horizontal scroll Shift+Wheel in modal window unstable work


GabbereX:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

"@mantine/core": "^8.1.2", "@mantine/hooks": "^8.1.2", "@mantine/modals": "^8.1.2",

### What package has an issue?

@mantine/modals

### What framework do you use?

React Router

### In which browsers you can reproduce the issue?

All

### Describe the bug

I have a component that creates a horizontal scroll. There are no problems outside the modal window. The following scenarios caused problems inside the modal window:
* Hold down Shift and scroll with the wheel - It is expected that scrolling will be left or right.
1. If there is no vertical scrollbar, then there are no horizontal scroll movements at all.
2. If the vertical scroll is scrolled to 0%, then the horizontal does not move to the left, but moves to the right.
3. If the vertical scroll is scrolled to 100%, then the horizontal one does not move to the right, but it moves to the left.
4. If the vertical scroll is scrolled by 1-99%, then there are no problems with the horizontal scroll.

```
modals.open({
  id: item.id,
  title: (
    <Title component='p' order={4}>
      {item.name}
    </Title>
   ),
      size: '100%',
      centered: true,
      overlayProps: { blur: 3 },
      children: <CustomComponent {...{ ...record, columns }} />
   })
```
Please note that the horizontal scroll can be located outside the component or inside the component, the result is always the same.

### Possible fix

I don't know how to solve this, I tried to hack it by handling onWhell, disabled scroll styles. All this did not give results. Perhaps there is some error in the scroll bar handler. It is unlikely that I will be able to solve this problem.

### Self-service

- [ ] I would be willing to implement a fix for this issue

rtivital:
Thanks for submitting the issue! In order to proceed, we need a codesandbox with a minimal reproduction. You can use this template to create it – https://codesandbox.io/s/mantine-react-template-q4dd5d?file=/src/App.js.

To learn what is a minimal, reproducible example, please see the article – https://stackoverflow.com/help/minimal-reproducible-example. In short: Minimal reproducible example (MRE) includes only the essential code and data needed to reproduce the problem, excluding unrelated details or complexities. An effective MRE ensures that others can quickly replicate the issue in their own environment without requiring additional context or assumptions.

[...]

PR #8619, opened 2026-01-18
When at the bottom or top in Windows Edge, the browser will try to scroll the parent when using the scroll wheel.

robert-j-webb:
We detect this case by looking for the shift key (unfortunately the `deltaX` value seems to always be -0)  being down during scroll and then stop propagation to allow the scroll area to move horizontally.

This would also prevent scrolling horizontally outside of the current container, we could add an additional check to make sure the scroll is not at a horizontal edge, but I am not sure if that is worth it.

Fixes https://github.com/mantinedev/mantine/issues/7050 and https://github.com/mantinedev/mantine/issues/8037


rtivital:
Did you have a chance to validate that it works on Windows?

robert-j-webb:
>  Did you have a chance to validate that it works on Windows?

Yes I used a windows VM on my macbook, with a USB mouse. The USB is passthrough to the VM OS so it should not matter. However, I don't have a mouse with a horizontal scroll wheel, so I am not sure about that case.

rtivital:
Thanks!

PR #8622, opened 2026-01-19
[@mantine/dates] TimePicker: Fix TimePicker clearing in uncontrolled mode not propagating empty value

### LINKED ISSUE #8606: TimePicker empty value is not submitted in form when used with useForm in uncontrolled mode.
### Dependencies check up

8.3.12

@mantine/dates

Vite

Chrome

### Describe the bug

When using the TimePicker component with useForm in mode "uncontrolled", the TimePicker value is not reflected in the form submission when deleting the value with the Backspace key from the keyboard.

Image

**Steps to reproduce:**

1. Open the codesandbox
2. Click on the time picker field, and delete the values using the keyboard Backspace key
4. Click on Submit

**Expected:**
Submit is called with empty string value for the time picker field value.

**Actual:**
Submit method is called with the previous value, even though the input appears empty.

Other observations:

- When clicking on the clear button (that is visible when clearable=true) the input behaves as expected
- When adding isNotEmpty validation predicates on the field, the error is not shown when using the backspace method to clear the input and clicking on the submit button.

Lastly, thanks for the amazing library.

 https://codesandbox.io/p/devbox/mantine-react-template-forked-vnvxnj?workspaceId=ws_GfVLU128bkkHvL28UxiqcY

### Possible fix

_No response_

### Self-service

pggggggggh:
When using `TimePicker` in uncontrolled mode (without `value`), clearing the time with `Backspace` or `Delete` does not propagate an empty string. Visually the inputs are empty, but the external form value still holds the previous time.

image

In the screenshot above, the minutes field is cleared, but the submitted value is still 10:00.

Previously, `useTimePicker` only called `onChange` when:

```ts
if (typeof value === 'string' && value !== '') {
  onChange?.('');
}
```

The intent was to prevent redundant `onChange` calls. This worked in controlled mode, where `value` is given, but failed in uncontrolled mode.

Now we track the validity with a ref `wasInvalidBefore` instead of relying on the `value` prop.

I also added a regression test in `TimePicker.test.tsx` 

Fixes #8606

### DIFF
--- a/packages/@mantine/dates/src/components/TimePicker/use-time-picker.ts
+++ b/packages/@mantine/dates/src/components/TimePicker/use-time-picker.ts
@@ -115,7 +115,7 @@ export function useTimePicker({
       onChange?.(timeString.value);
     } else {
       acceptChange.current = false;
-      if (typeof value === 'string' && value !== '') {
+      if (!wasInvalidBefore.current) {
         onChange?.('');
         wasInvalidBefore.current = true;
       }


pggggggggh:
I'm sorry, there was a problem with my git. That `if (!wasInvalidBefore.current)` copilot suggested is exactly what I wrote and passed the test, but in some reason it was not applied. I fixed it.

rtivital:
Thanks!

PR #8624, opened 2026-01-20
✨ Add offset to InputsGroupFieldset

k-utsumi:
Adds offset to `InputsGroupFieldset`
for proper spacing between description/error text and input controls.

Reference (Input) |  Before | After
-|-|-
<kbd> localhost_2356__path=_story_input--demo-wrapper | <kbd> Without offset - tight spacing between description and controls | <kbd> With offset - proper spacing between description and controls

## Changes
- Apply the same offset as Input component ([ref](https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/components/Input/Input.tsx#L172-L173)) by utilizing InputWrapper context (`offsetTop`/`offsetBottom`)
- Adjust demos accordingly

### DIFF
--- a/packages/@docs/demos/src/demos/core/Checkbox/Checkbox.demo.groupConfigurator.tsx
+++ b/packages/@docs/demos/src/demos/core/Checkbox/Checkbox.demo.groupConfigurator.tsx
@@ -11,7 +11,7 @@ function Demo() {
       defaultValue={['react']}
       {{props}}
     >
-      <Group mt="xs">
+      <Group>
         <Checkbox value="react" label="React" />
         <Checkbox value="svelte" label="Svelte" />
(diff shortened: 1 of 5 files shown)


rtivital:
This is a breaking change, it is not planned to be implemented. The current approach is designed this way deliberately to provide flexibility.

k-utsumi:
Thanks for the clarification.
Would it be possible to add an optional prop instead, for example `withOffset` (opt-in, default false)?

In our team, hard-coding values or adding custom spacing tokens is discouraged,
so we would prefer a built-in solution if possible.

---

- [回避策メモ @ X](https://x.com/kaba_box/status/2032439186503217302?s=20)

Issue #8626, opened 2026-01-20
Select dropdown options not being read with screen reader inside Modal


harry-ignite:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

8.3.13

### What package has an issue?

@mantine/core

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

Repro:
- place a Select inside Modal
- open the modal window
- focus Select
- observe/listen the Voiceover's output
- use arrows up/down to navigate through dropdown options
- observe/listen the Voiceover's output

Observed: 
Voiceover recognises and announces the listbox, but remains silent when going through options

Expected: 
Voiceover recognises and announces the listbox, then announces each visited listbox option

https://github.com/user-attachments/assets/49fcc734-e5d1-419e-b0bf-bfaef76918cf

### If possible, include a link to a codesandbox with a minimal reproduction

https://codesandbox.io/p/sandbox/mantinemodalwithselect-forked-mwsl3n

### Possible fix

When using Select component inside a Modal, the dropdown content is portaled to the document body by default. This breaks listbox accessibility because screen readers expect all content to be within the modal's DOM subtree. Currently the Voiceover doesn't announce listbox options.

Dropdown options rendered inside a Modal should be part of the modal's DOM tree to be read by assistiove technology.

### Self-service

- [x] I would be willing to implement a fix for this issue

harry-ignite:
Proposed solution
https://github.com/mantinedev/mantine/pull/8627

PR #8627, opened 2026-01-20
[@mantine/core] Fix Select dropdown options not being read with screen reader inside Modal (#8626)

harry-ignite:
- Added FloatingPortalProvider context that allows Modal to provide a portal target for popovers
- Popovers now automatically render inside the Modal's DOM tree when used within a Modal
- This preserves listbox accessibility without requiring any changes from users

### DIFF
--- a/packages/@mantine/core/src/components/Popover/Popover.tsx
+++ b/packages/@mantine/core/src/components/Popover/Popover.tsx
@@ -287,7 +287,7 @@ export function Popover(_props: PopoverProps) {
     ? { ...portalProps, target: floatingPortalTarget }
     : portalProps;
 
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks for the PR! Unfortunately, I cannot merge it in 9.x – it changes default behavior of all floating components inside Modal and Drawer (portal target and `floatingStrategy`), moves dialog role to a different element and adds new public exports, which may break existing applications.

As a workaround, you can render the dropdown inside the modal with `comboboxProps={{ withinPortal: false }}` (or set it for all components with theme default props), so the options are inside the `aria-modal` dialog and can be read by screen readers.

PR #8630, opened 2026-01-23
[@mantine/core] Chip: Fix --chip-checked-padding to use padding-inline

chlee1001:
## Summary
- Fixed `--chip-checked-padding` CSS variable to use `padding-inline` instead of `padding`
- The documentation states this variable "Controls horizontal padding when chip is checked", but the CSS was applying padding to all sides

image

## Changes
```diff
&:where([data-checked]) {
-    padding: var(--chip-checked-padding);
+    padding-inline: var(--chip-checked-padding);
}
```


rtivital:
Thanks!

PR #8634, opened 2026-01-24
[@mantine/core] MultiSelect: Fix clearable text overlap

### LINKED ISSUE #8632: In the MultiSelect clearable component, the width of the clear icon is not taken into account for the padding
### Dependencies check up

8.3.12

@mantine/core

Vite

Chrome

### Describe the bug

In the MultiSelect clearable component, the padding-inline-end of the mantine-MultiSelect-input element does not account for the width of the clear icon (mantine-InputClearButton-root). This issue occurs when a long value is selected.

https://mantine.dev/core/multi-select/#clearable

Image

_No response_

### Possible fix

_No response_

### Self-service

ckhe1215:
## Description
Fixes #8632 - MultiSelect clearable: text overlaps with clear icon
                                                          
### Changes
- Added `paddingInlineEnd` to `Pill.Group` when clearable is enabled                                    
- Uses same padding values as `InputClearSection` for consistency (xs: 41px, sm: 50px, md: 60px, lg: 72px, xl: 89px)                                                  
- Padding only applied when clearable is true AND has selected values
                                                             
**As-is:** Text overlaps clear button
SCR-20260124-idmw

**To-be:** Proper spacing between text and clear button
SCR-20260124-iudh

### DIFF
--- a/packages/@mantine/core/src/components/MultiSelect/MultiSelect.tsx
+++ b/packages/@mantine/core/src/components/MultiSelect/MultiSelect.tsx
@@ -343,7 +343,9 @@ export const MultiSelect = factory<MultiSelectFactory>((_props, ref) => {
 
   const filteredData = filterPickedValues({ data: parsedData, value: _value });
   const _clearable = clearable && _value.length > 0 && !disabled && !readOnly;
-  const pillsListStyle = _clearable ? { paddingInlineEnd: clearSectionOffset[size] } : undefined;
+  const pillsListStyle = _clearable
+    ? { paddingInlineEnd: clearSectionOffset[size] ?? clearSectionOffset.sm }
+    : undefined;
 
   return (
     <>


jmcgavin:
Does this also (coincidentally) fix an issue where the multiselect's up/down chevron icon is not clickable when the component is clearable?

ckhe1215:
@jmcgavin 
No, this PR doesn't address the chevron click issue - they're separate problems.

However, while investigating, I found the root cause of the chevron issue:
```tsx
rightSectionPointerEvents: rightSectionPointerEvents || (_clearable ? 'all' : 'none'),
```

When `clearable` is true, the right section gets `pointer-events: all`, causing it to capture click events and preventing the input’s onClick handler from firing.

I traced this back to commit e2ffcd2a. Interestingly, `Select` had the same code but it was later removed, which is why `Select` doesn't have this problem anymore.

@rtivital Would you like me to:
1. Open a separate issue for the chevron click problem and submit a fix there, or
2. Include the fix in this PR as an additional commit?

I'm happy to work on either approach.

rtivital:
Thanks!

PR #8645, opened 2026-01-30
[@mantine/core] Switch - Fix test visibility (#8370)

### LINKED ISSUE #8370: Cannot interact with Switch component using playwright
### Dependencies check up

8.2.1

@mantine/core

Vite

All

### Describe the bug

👋

I've tried to use these playwright locators, which are basically the same:

```typescript
page.getByLabel("my-switch-label");

page.getByRole("switch", {
  name: "my-switch-label"
});
```

I can read the current value but cannot change it:

```typescript
// ok
expect(locator).toBeChecked()

// doesn't work
await locator.click();
await locator.uncheck();
await locator.check();

// this one also fails due to timeout
expect(locator).toBeVisible()
```

I guess the problem is that mantine puts the input off  screen and makes it invisible using this:

```css
height: 0;
width: 0;
opacity: 0;
position: absolute;
```

by using this locator it's possible to interact with the switch, but not to read the current value:

```typescript
// ok
const locator = page.locator(".mantine-Switch-label");
await locator.click();

// obviously not
await expect(locator).toBeChecked();
await locator.check();
```

_No response_

### Possible fix

_No response_

### Self-service

andrewborstein:
Closes #8370

As noted in https://github.com/orgs/mantinedev/discussions/8372 playwright is unable to interact with the switch because it interprets it as obscured / not visible in the viewport due to  its opacity, width and height.

This fix ensures the native checkbox is still visually hidden but now accessible to playwright and still to the DOM and a11y tree. See also: https://scottaohara.github.io/a11y_styled_form_controls/src/checkbox--switch/

Tested with latest version of Playwright 1.58.0.

Screenshot 2026-01-30 at 1 15 13 PM

https://github.com/user-attachments/assets/06eb9168-6d26-4e10-93bb-b531b0064cd4

### DIFF
--- a/packages/@mantine/core/src/components/Switch/Switch.module.css
+++ b/packages/@mantine/core/src/components/Switch/Switch.module.css
@@ -45,7 +45,7 @@
   /* Hide the input visually without removing it from the accessibility tree or obscuring it in the DOM */
   height: 100%;
   width: 100%;
-  opacity: 0.0001;
+  opacity: 0;
 
   margin: 0;
   padding: 0;


packages/@mantine/core/src/components/Switch/Switch.module.css
```diff
@@ -42,9 +42,11 @@
 }
 
 .input {
-  height: 0;
-  width: 0;
-  opacity: 0;
+  /* Hide the input visually without removing it from the accessibility tree or obscuring it in the DOM */
+  height: 100%;
+  width: 100%;
+  opacity: 0.0001;
```
andrewborstein: After attempting to find the source code in playwright that determines whether an element can be clicked, I found [some docs on "actionability"](https://playwright.dev/docs/actionability#introduction) that assert elements can have `opacity: 0` and still be [considered visible](https://playwright.dev/docs/actionability#visible), but can''t have "zero size". 

I did some more local testing: tests are unaffected by opacity but fail as soon as `height` or `width` are left at `0`. So now I think leaving opacity as-is would be better, and feel a tad less hacky/magic. 

```suggestion
  opacity: 0;
```

So the input really is clickable just not visible (i.e. it's actionable).

I was torn on if I should leave a code comment, and if it should be specific to A11Y or playwright. I would have liked to unit test this change, but it's not easy to test CSS module styles and the bug is specific to playwright anyway.

rtivital:
I'm not sure about this fix, I'll merge and return if there are any complaints.

PR #8650, opened 2026-02-01
update ChartTooltip valueFormatter type

tvandoren:
Minor issue: the usage accepts a `ReactNode`, but the type specifies a `string`.


rtivital:
Thanks!

PR #8674, opened 2026-02-11
[@mantine/core] implements selectFirstOptionOnDataChange

Alan-Gomes:
Adds a `selectFirstOptionOnDataChange` option for combobox-like components. When enabled, the combobox automatically selects the first option whenever its data changes. This is useful when options are dynamically updated (e.g., from a backend search), allowing the first result to be preselected and improving the experience for keyboard users.


Alan-Gomes:
Just to clarify: This option is different from `selectFirstOptionOnChange`.
* `selectFirstOptionOnChange` selects the first option when the input value changes (i.e., on user typing or any value change).
* `selectFirstOptionOnDataChange` selects the first option when the options list itself changes (e.g., after async loading, filtering from a backend, or any external data update), even if the input value did not change.

rtivital:
Thanks for submitting the PR, but I'd rather not implement this feature for Select component. If you need this functionality, it is better to create a custom component with Combobox.

PR #8688, opened 2026-02-16
fix(docs): add autocomplete=off to homepage input demos to prevent Firefox password autofill

### LINKED ISSUE #8682: Stray password pop-up shows up when going from mantine.dev to "Getting Started" page

https://mantine.dev/ -> https://mantine.dev/getting-started/

### What is wrong?

Image

Attached above is the password pop-up I get. This has been happening from since I've been using Mantine (more than 2 years now), and only now decided to report it.

Browser: Firefox
Description: Happens when I type in the URL mantine.dev and then click "Getting Started" button. Does not happen when I'm going back and forth between these pages once loaded. Does not happen in my Chrome browser.

### Self-service

pierreeurope:
## Summary

Fixes #8682

Firefox identifies `PasswordInput` and `ColorInput` in the homepage component demos as login fields, triggering the password manager popup when navigating from `mantine.dev` to the Getting Started page.

## Changes

Added `autoComplete: 'off'` to the shared `inputProps` object in `HomePageInputsDemo.tsx`. This is a one-line change that applies to all demo inputs on the homepage, preventing Firefox from offering to save/fill passwords for these decorative demo inputs.

## Root Cause

Firefox password manager heuristics (see [Mozilla Bug 956906](https://bugzilla.mozilla.org/show_bug.cgi?id=956906)) detect `<input type="password">` elements on the page and assume they are part of a login form, even when they are just component demos. The `autocomplete="off"` attribute tells the browser these inputs are not login-related.

## Testing

- Verified the change only affects the docs site homepage demo, not the library components themselves
- The fix is minimal: 1 file changed, 1 line added

### DIFF
--- a/apps/mantine.dev/src/components/HomePage/HomePageComponents/demos/HomePageInputsDemo.tsx
+++ b/apps/mantine.dev/src/components/HomePage/HomePageComponents/demos/HomePageInputsDemo.tsx
@@ -15,6 +15,7 @@ import { DatePickerInput } from '@mantine/dates';
 const inputProps = {
   size: 'lg' as const,
   radius: 'md',
+  autoComplete: 'off',
 };
 
 export function HomePageInputsDemo() {


rtivital:
Thanks for submitting the PR! The issue was fixed differently in another commit.

PR #8703, opened 2026-02-23
[@mantine/hooks] feat: Allow `interval` parameter on `useInterval` to dynamically change after mount

BreakingBeakers:
## Overview
This pull request updates the `useInterval` hook to add the `interval` parameter a one of the dependencies of the memoized `start` function.

While working with the `useInterval` hook I ran into an issue where I wanted the delay time of the `useInterval` hook to dynamically change based on the data I retrieved on my page. After some debugging I realized the root of the issue was due to the `interval`'s time being memoized in the start function and not getting updated since there was an empty dependency array on the `start` function the comes back from `useInterval`.

To create this solution I used TDD to first create a test the error from the behavior I expected to see, and then made updates to `useInterval` to get the test to pass.

### Test
```
npm run jest packages\@mantine\hooks\src\use-interval\use-interval.test.ts
```
#### Before
image

#### After
image


BreakingBeakers:
Happy to run prettier to fix the `useSpotlight` linting issue. My assumption is that will be fixed in another PR as of right now.

Pirulax:
Just a side-note: This hook also doesn't seem to handle changes to the fnref, does it? Or I'm not understanding the code.

BreakingBeakers:
> Just a side-note: This hook also doesn't seem to handle changes to the fnref, does it? Or I'm not understanding the code.

I believe you are correct, I noticed that too. I mainly just did the minimal change in hopes to get this PR in sooner.

rtivital:
Thanks for the PR! I took a look but I don't think this is actually fixing a bug. The `interval` value is already mirrored into `intervalValueRef` on every render (line 30–31), and the `useEffect` on line 70 already has `interval` in its dependency array — so when `interval` changes, the effect's cleanup stops the old timer and re-runs `start()` with the new value.

I verified this by running your new test against `master` (without the source change) and it passes:

```
✓ should change interval timing when interval parameter changes
```

Adding `interval` to `start`'s deps only changes the function's identity on each interval change, but `start` still reads from `intervalValueRef.current` either way, so behavior is unchanged.

The dynamic-interval case was previously addressed in 1dfef7442d. Going to close this one, but happy to revisit if you have a repro where the current code actually fails.

Issue #8712, opened 2026-02-28
Textarea autosize renders hidden mirror <textarea> that triggers accessibility tool false positives (WAVE)


dsuzener:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

8.3.15

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

All

### Describe the bug

Hi Mantine team,

When using Textarea with the autosize prop enabled, Mantine renders a hidden mirror <textarea> element in the DOM for measurement purposes.

Example of the generated element:
```html
<textarea
  tabindex="-1"
  aria-hidden="true"
  style="height: 0px; visibility: hidden; position: absolute; z-index: -1000;"
></textarea>
```
While this element is correctly marked with aria-hidden="true" and is not focusable, automated accessibility tools (specifically WAVE) flag it as form control missing label error.

This creates accessibility audit noise and makes it difficult to pass automated compliance checks, even though the element is not exposed to assistive technologies.

### Possible fix

Recommended approach:

1. Keep aria-hidden="true" and tabindex="-1" to ensure the mirror is ignored by assistive technologies.
2. Add aria-label="autosize measurement mirror" so automated accessibility tools do not report false positives.
3. Optionally, maintain role="presentation" to further clarify its decorative/technical nature.

When directly added into the HTML, the above resolves the WAVE error.

Thank you for the great work on Mantine. This is a small issue, but it affects strict accessibility compliance workflows.

### Self-service

- [x] I would be willing to implement a fix for this issue

rtivital:
You are welcome to submit a PR with a fix

dsuzener:
PR is now awaiting review.

rtivital:
Fixed in [9.0](https://github.com/mantinedev/mantine/releases/tag/9.0.0)

PR #8713, opened 2026-02-28
[@mantine/core] Textarea: Fix accessibility false positives for autosize (#8712)

dsuzener:
The hidden mirror <textarea> used internally by react-textarea-autosize (when autosize is enabled) triggers the error “Form control does not have a corresponding label”

Although the mirror element has aria-hidden="true" and tabindex="-1" and is not interactive or visible to users, tools like WAVE still flag it because it lacks an accessible name.

### Solution

This PR adds a runtime patch that detects the hidden mirror <textarea> based on aria-hidden="true, tabindex="-1", and some styles. Then, it adds role="presentation" and aria-label="autosize measurement mirror textarea".

The patch runs only when autosize is enabled and does not affect other textareas.

### Notes
- This is a workaround until react-textarea-autosize includes proper accessibility attributes on the mirror element.
- Screen readers and real user accessibility are not impacted. 
- The change eliminates WAVE false positives on the Textarea documentation page.

### DIFF
--- a/packages/@mantine/core/src/components/Textarea/Textarea.tsx
+++ b/packages/@mantine/core/src/components/Textarea/Textarea.tsx
@@ -1,4 +1,5 @@
 import TextareaAutosize from 'react-textarea-autosize';
+import { useEffect } from 'react';
 import {
   BoxProps,
   ElementProps,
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks for submitting the PR! The issue with textarea was resolved differently in another commit

PR #8720, opened 2026-03-02
[@mantine/notifications] Add callbacks to distinguish Auto Close vs Manual Close

drewhoener:
When a notification handled by the notification extension system closes, there's no indication if it was closed by the user (via the close button), or hidden by the auto-hide system.  
In the auto-hide case, I want to push it into a "notification queue" that the user can read in a separate drawer, but listening to onClose doesn't provide information about if it's already been acknowledged.

This PR fixes that by adding in additional callbacks for when each behavior is used. The overall API surface hasn't been changed, `onClose` still exists and functions as before.

### DIFF
--- a/apps/mantine.dev/src/pages/x/notifications.mdx
+++ b/apps/mantine.dev/src/pages/x/notifications.mdx
@@ -91,6 +91,8 @@ import {
 - `withBorder` – determines whether notification should have a border
 - `withCloseButton` – determines whether the close button should be visible
 - `onClose` – calls when notification is unmounted
+- `onAutoClose` – calls when notification is automatically closed via timeout
+- `onManualClose` – calls when the close button is clicked
 - `onOpen` – calls when notification is mounted
 - `autoClose` – defines timeout in ms on which notification will be automatically closed, use `false` to disable auto close
 - `message` – required notification body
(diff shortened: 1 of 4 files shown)


Pirulax:
Just noting that the exact same would be very nice for modals as well.
Thank you.

Pirulax:
Also, perhaps it would be beneficial to just provide this in `onClose` directly rather than having 3 `onClose`  type callbacks?
Eg.:
```ts
type OnCloseArgs = {
    closedBy: 'user' | 'automatically';
};
```
And then `onClose` would just be:
```ts
onClose(args: OnCloseArgs) => void;
```
or
```ts
onClose(args: {
    closedBy: 'user' | 'automatically';
}) => void;
```
But I feel like having a separate type for the arguments may be beneficial (Especially with the overall trend I see in Mantine, with exposing as many types as possible)

drewhoener:
> Also, perhaps it would be beneficial to just provide this in `onClose` directly rather than having 3 `onClose` type callbacks?

Good point, I'll wait on @rtivital to weigh in I suppose? I can also open another PR for the modals if that's something he wants

drewhoener:
@rtivital Ping on this PR? Waiting for a comment about the above suggestion

rtivital:
There were some recent changes to `@mantine/notifications` package that are not compatible with this PR. I'd rather not add this feature now.

PR #8725, opened 2026-03-03
fix(tiptap): use useEditorState in createControl for React Compiler compatibility

ReemX:
## Summary

Toolbar controls created via `createControl` call `editor.isActive()` and `isDisabled?.(editor)` directly during render. Because TipTap's editor is a mutable object (same reference across transactions), React Compiler memoizes the JSX and never updates the `data-active` / `disabled` attributes — toolbar buttons never show their active (toggled) state.

This PR replaces the direct calls with `useEditorState` from `@tiptap/react`, which subscribes to editor transactions via `useSyncExternalStore` under the hood. This gives the compiler a genuinely changing value to track, fixing active-state indicators without requiring consumers to add workarounds.

**Rebased on Mantine 9 (v9.0.2) — compatible with `@tiptap/react >=3.3.0`.**

### Before (broken with React Compiler)
\`\`\`tsx
active={isActive?.name ? editor?.isActive(isActive.name, isActive.attributes) : false}
disabled={isDisabled?.(editor) || false}
\`\`\`

### After (works with React Compiler)
\`\`\`tsx
const editorState = useEditorState({
  editor: editor ?? null,
  selector: (ctx) => ({
    active: isActive?.name
      ? ctx.editor?.isActive(isActive.name, isActive.attributes) ?? false
      : false,
    disabled: isDisabled?.(ctx.editor) ?? false,
  }),
});

const active = editorState?.active ?? false;
const disabled = editorState?.disabled ?? false;
\`\`\`

### Why this matters
- React 19 ships with React Compiler, which aggressively memoizes render output
- TipTap's `editor` is a mutable singleton — the compiler sees the same reference and skips re-renders
- `useEditorState` uses `useSyncExternalStore` internally, making state changes visible to both React and the compiler
- Without this fix, consumers must use workarounds like passing `data-rev={useEditorState({ editor, selector: s => s.transactionNumber })}` on every `<RichTextEditor>` to force child re-renders

### Related issues
- [tiptap#6566](https://github.com/ueberdosis/tiptap/issues/6566) — TipTap toolbar active state broken with React Compiler
- [mui-tiptap#425](https://github.com/sjdemartini/mui-tiptap/issues/425) — Same issue in mui-tiptap

### Changes
- **`RichTextEditorControl.tsx`**: Import `useEditorState` from `@tiptap/react` and use it in `createControl` to compute `active` and `disabled` reactively. Handles `editor: null` safely via optional chaining.


rtivital:
Thanks!

PR #8729, opened 2026-03-04
[mantine.dev] Fix definition of use-element-size hook

maltekiessling:


rtivital:
Thanks for submitting the PR! The issue was fixed in another commit

PR #8732, opened 2026-03-05
refactor(@mantine/hooks): update useViewportSize to useSyncExternalStore

mocutasorin:
This PR refactors the `useViewportSize` hook to utilize React 18's `useSyncExternalStore` instead of the legacy `useState` + `useEffect` pattern. 

### Motivation & Benefits:
1. **Prevents Double Renders:** The previous implementation initialized with `{ width: 0, height: 0 }` and then updated via `useEffect` after the component mounted. `useSyncExternalStore` reads the actual window dimensions synchronously during the initial client render, preventing layout shifts or visual flashes.
2. **Concurrent Rendering Safe:** Fully compatible with React 18's concurrent features, preventing "tearing" if the window resizes during a paused render transition.
3. **Global Subscription:** The subscription logic and cache have been hoisted outside the hook. If multiple components use `useViewportSize` simultaneously, they now share a single cache and `window` event listener, rather than attaching multiple redundant listeners via `useWindowEvent`.
4. **Hydration Safe:** Safely falls back to `{ width: 0, height: 0 }` on the server via `getServerSnapshot` to prevent hydration mismatches, matching the existing API behavior.

### Changes:
- Replaced `useState` and `useEffect` with `useSyncExternalStore`.
- Hoisted event listeners outside the component tree.
- Retained `{ passive: true }` for scroll/resize performance.
- Retained identical return types and API contract.

### DIFF
--- a/packages/@mantine/hooks/src/use-viewport-size/use-viewport-size.test.ts
+++ b/packages/@mantine/hooks/src/use-viewport-size/use-viewport-size.test.ts
@@ -1,0 +1,76 @@
+import { act, renderHook } from '@testing-library/react';
+import { useViewportSize } from './use-viewport-size';
+
+function setWindowSize(width: number, height: number) {
+  Object.defineProperty(window, 'innerWidth', { writable: true, configurable: true, value: width });
+  Object.defineProperty(window, 'innerHeight', {
+    writable: true,
+    configurable: true,
+    value: height,
+  });
+}
+
+describe('@mantine/hooks/use-viewport-size', () => {
+  beforeEach(() => {
+    setWindowSize(1024, 768);
+  });
+
+  it('returns initial viewport dimensions from window', () => {
+    const { result } = renderHook(() => useViewportSize());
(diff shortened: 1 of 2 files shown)


mocutasorin:
@rtivital I was doing some deeper testing on this PR and dug into React 18's hydration rules, and I actually found a significant edge case that makes me rethink this approach for Mantine.

Because the server snapshot ({ width: 0, height: 0 }) will inherently mismatch the client snapshot ({ width: 1920, height: 1080 }) on mount, useSyncExternalStore forces a synchronous re-render during hydration to prevent tearing.

The issue is that if a Mantine user wraps a component using useViewportSize inside a <Suspense> boundary, this synchronous hydration mismatch will instantly trigger the Suspense fallback (flashing a loading spinner/state). The current useState + useEffect approach safely avoids this because the effect runs after the initial paint.

Given this, the current useEffect implementation is actually much safer for a component library where we don't know how users are structuring their Suspense boundaries. If you agree with this assessment, let's close this PR.

PR #8736, opened 2026-03-05
fix: Avoid re-opening modal with the same id

Pirulax:
I'm not exactly sure if this is backwards compatible - But I believe there's no reason to open a modal with the same `id`.

I'm using it exclusively to void opening the  same modal twice from a `useEffect` callback (which is called twice in strict mode)


rtivital:
Thanks!

PR #8745, opened 2026-03-11
Document returnFocus default for Menu

pfo-omicsstudio:
Fixed incorrect returnFocus default value documentation for `Menu`. 

Currently, it says the default value is false, which is wrong. It's true by default for `Menu`, false by default for `Popover` (which is where this error comes from - `Menu` inherits docs from `Popover`).


rtivital:
Thanks!

PR #8752, opened 2026-03-15
[DOCS] [@mantine/dates] add missing information to getting started

slk333:
Added section on date values as strings for @mantine/dates.

Most of it comes from the 7->8 changelog, and **was missing from the docs** (and those formats are not obvious either)


rtivital:
Thanks!

PR #8756, opened 2026-03-18
fix #8755 radio name property overwrite

### LINKED ISSUE #8755: name property cannot be set in Radio component
### Dependencies check up

8.3.18

@mantine/core

Vite

All

### Describe the bug

The default behavior has been changed in #8396 ( https://github.com/mantinedev/mantine/pull/8396/changes#diff-ab5686646b1db846c1de7f0a5504207b16362a83cbc4aae83735323a4ee73b68R179 )

User cannot apply name property for Radio component. It will always be overwritten

_No response_

### Possible fix

change back to old behavior or edit L181 in https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/components/Radio/Radio.tsx

```
  const withContextProps = {
    checked: contextChecked ?? checked,
    name: rest.name ?? ctx?.name, // <-- correct version
    onChange: (event: React.ChangeEvent<HTMLInputElement>) => {
      ctx?.onChange(event);
      onChange?.(event);
    },
    disabled: ctx?.disabled ?? disabled,
  };
```

### Self-service

euronics-qaddir:
fixes #8755

### DIFF
--- a/packages/@mantine/core/src/components/Radio/Radio.tsx
+++ b/packages/@mantine/core/src/components/Radio/Radio.tsx
@@ -178,7 +178,7 @@ export const Radio = factory<RadioFactory>((_props, ref) => {
 
   const withContextProps = {
     checked: contextChecked ?? checked,
-    name: ctx?.name ?? rest.name,
+    name: rest.name ?? ctx?.name,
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #8763, opened 2026-03-24
fix: add missing role and aria-expanded to Combobox target

imkrishn:
### Solved

This PR adds missing role and aria-expanded attributes to the Combobox trigger.

### Issue

Improves accessibility by making the component more compliant with ARIA practices and better for screen readers.

**Implementation**

Added appropriate role to the Combobox target
Added aria-expanded to reflect open/closed state


rtivital:
Thanks for submitting the PR! The fix was already implemented in #8767

PR #8766, opened 2026-03-25
fix(FunnelChart): strokeColor and labelColor forwarded to DOM

### LINKED ISSUE #8759: FunnelChart forwards strokeColor prop to Box causing React DOM warning
### Dependencies check up

8.3.18

@mantine/charts

Next.js

All

### Describe the bug

When using the strokeColor prop in FunnelChart, the prop is incorrectly forwarded to the underlying Box component, which results in a React warning because strokeColor is not a valid DOM attribute.

https://codesandbox.io/p/devbox/mantine-react-template-forked-gmh5gp

### Possible fix

_No response_

### Self-service

restareaByWeezy:
Fixes #8759

\`strokeColor\` and \`labelColor\` were missing from the destructuring list,
so both props ended up in \`...others\` and got forwarded to the inner \`Box\`,
triggering React DOM warnings.

Both are already handled by \`varsResolver\` to set CSS custom properties,
so they don't need to reach the DOM. Adding them to the destructuring is enough.

### DIFF
--- a/packages/@mantine/charts/src/FunnelChart/FunnelChart.tsx
+++ b/packages/@mantine/charts/src/FunnelChart/FunnelChart.tsx
@@ -137,6 +137,8 @@ export const FunnelChart = factory<FunnelChartFactory>((_props, ref) => {
     funnelProps,
     labelsPosition,
     tooltipDataSource,
+    strokeColor,
+    labelColor,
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #8767, opened 2026-03-25
fix: add role="combobox" and aria-expanded to Select, MultiSelect, Autocomplete, TagsInput

### LINKED ISSUE #8758: Bug Report: Missing aria-expanded and role="combobox" on Combobox-based Components
# Bug Report: Missing `aria-expanded` and `role="combobox"` on Combobox-based Components
 
## Dependencies Check-Up
 
- [x] I have verified that I use the latest version of all `@mantine/*` packages
 
## Environment
 
| Field                          | Value            |
| ------------------------------ | ---------------- |
| **`@mantine/*` version**       | 8.3.9            |
| **Package with issue**         | `@mantine/core`  |
| **Framework**                  | Next.js          |
| **Reproducible in browsers**   | Chrome             |
 
### Current Behavior
 
The `Select`, `MultiSelect`, `Autocomplete`, and `TagsInput` components do not set `aria-expanded` on their input elements. This is because `Combobox.Target` and `Combobox.EventsTarget` both default `withExpandedAttribute` to `false`, and none of these components override it to `true`. This violates the ARIA combobox pattern, which requires `aria-expanded` on the triggering element.
 
Additionally, the input element is missing `role="combobox"`. Without this role, `aria-expanded` is not a valid ARIA attribute on a plain `<input readonly>` element — axe flags it as an `aria-allowed-attr` violation. Per the [WAI-ARIA combobox pattern](https://www.w3.org/WAI/ARIA/apg/patterns/combobox/), the triggering element must have `role="combobox"` along with `aria-expanded`.
 
### Expected Behavior
 
The input element should have:
- `role="combobox"`
- `aria-expanded="true"` when the dropdown is open
- `aria-expanded="false"` when it is closed
 
### Steps to Reproduce
 
1. Render a `<Select>` (or `MultiSelect`, `Autocomplete`, `TagsInput`) with any data.
2. Inspect the input element in DevTools.
3. Open the dropdown.
4. Notice `aria-expanded` is never present on the input.
5. Notice `role="combobox"` is missing from the input.
 
## Minimal Reproduction
 
_No response_
 
## Possible Fix
 
The fix requires two changes:
 
### 1. Add `role="combobox"` in `useComboboxTargetProps`
 
In `use-combobox-target-props.ts`, add `role: 'combobox'` to the `ariaAttributes` object:
 
```typescript
const ariaAttributes = withAriaAttributes
  ? {
      role: 'combobox' as const,
      'aria-haspopup': 'listbox' as const,
      'aria-expanded': withExpandedAttribute
        ? !!(ctx.store.listId && ctx.store.dropdownOpened)
        : undefined,
      // ... rest unchanged
    }
  : {};
```
 
### 2. Enable `withExpandedAttribute` on each component
 
In `Select` and `Autocomplete`, pass `withExpandedAttribute` to `Combobox.Target`:
 
```jsx
<Combobox.Target targetType={searchable ? 'input' : 'button'} autoComplete={autoComplete} withExpandedAttribute>
```
 
In `MultiSelect` and `TagsInput`, pass `withExpandedAttribute` to `Combobox.EventsTarget`:
 
```jsx
<Combobox.EventsTarget autoComplete={autoComplete} withExpandedAttribute>
```
 
**Note:** Adding `role="combobox"` affects all components that use `Combobox.Target` and `Combobox.EventsTarget`, so this may need careful review.
 
## Affected Components
 
`Select`, `MultiSelect`, `Autocomplete`, `TagsInput`
 
## Self-Service
 
restareaByWeezy:
Fixes #8758

\`Select\`, \`MultiSelect\`, \`Autocomplete\`, and \`TagsInput\` were missing
\`role="combobox"\` and \`aria-expanded\` on their inputs, which violates
the WAI-ARIA combobox pattern.

Two things needed fixing:
- \`useComboboxTargetProps\` wasn't including \`role: 'combobox'\` in its aria attributes
- The four components weren't passing \`withExpandedAttribute\`, so \`aria-expanded\` was never set

\`role: 'combobox'\` is gated on \`withExpandedAttribute\` so the two always appear
together — having role without aria-expanded would cause an axe \`aria-required-attr\` violation.

Updated \`getByRole('textbox')\` queries to \`getByRole('combobox')\` in Select and
MultiSelect tests to reflect the new role.

### DIFF
--- a/apps/mantine.dev/src/pages/x/extensions.mdx
+++ b/apps/mantine.dev/src/pages/x/extensions.mdx
@@ -40,16 +40,26 @@ Community extensions list:
 - [DataTable](https://icflorescu.github.io/mantine-datatable/) – data table component without dependencies
 - [MantineReactTable](https://v2.mantine-react-table.com/) – data table component based on TanStack table package
 - [BorderAnimate](https://gfazioli.github.io/mantine-border-animate/) – border animation styles (beam, glow, more...)
-- [Clock](https://gfazioli.github.io/mantine-clock/) - analog clock component
-- [Compare](https://gfazioli.github.io/mantine-compare/) - image comparison slider component
+- [Clock](https://gfazioli.github.io/mantine-clock/) – analog clock component
+- [Compare](https://gfazioli.github.io/mantine-compare/) – image comparison slider component
 - [Flip](https://gfazioli.github.io/mantine-flip/) – flip animation component
-- [SplitPane](https://gfazioli.github.io/mantine-split-pane/) – resizable split pane component
+- [JsonTree](https://gfazioli.github.io/mantine-json-tree/) – interactive JSON tree viewer with syntax highlighting
+- [Led](https://gfazioli.github.io/mantine-led/) – LED indicator component for status feedback
+- [ListViewTable](https://gfazioli.github.io/mantine-list-view-table/) – Finder-style list view table with column reordering and resizing
 - [Marquee](https://gfazioli.github.io/mantine-marquee/) – marquee component
+- [Mask](https://gfazioli.github.io/mantine-mask/) – cursor-follow spotlight mask component
 - [Onboarding](https://gfazioli.github.io/mantine-onboarding-tour/) – onboarding / tour component
 - [Parallax](https://gfazioli.github.io/mantine-parallax/) – parallax component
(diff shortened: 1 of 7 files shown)


rtivital:
Thanks!

PR #8768, opened 2026-03-25
fix(ChartLegend): highlight gets stuck when moving cursor quickly over legend items

### LINKED ISSUE #8750: Chart Legend Highlight Stuck During Rapid Cursor Movement
### Dependencies check up

8.3.17

@mantine/charts

Next.js

### Describe the bug

When hovering over legend items in @mantine/charts `CompositeChart` (or `AreaChart`, `LineChart`, etc.) and moving the cursor rapidly, the highlight state can become "stuck" — causing chart series to remain dimmed/highlighted incorrectly until the cursor is moved back out or into the chart component or is hovered over another legend item.

#### Expected Behavior

1. Hover over a legend item → that series is highlighted, others are dimmed
2. Move cursor off the legend item → all series return to normal opacity

#### Actual Behavior

During rapid cursor movements:

* Swiping through a legend item quickly can leave that series stuck as highlighted. This usually occurs when swiping the cursor up or down across the legend and chart area.
    Image

* Moving quickly from one legend item to another can cause the highlight to be cleared entirely (no series highlighted) even though the cursor is hovering over a legend item. This is more difficult to reproduce, but occurs when the cursor moves from one legend item, off of it, and onto another legend item within a couple frames.
    Image

#### 1. CSS `:not(:hover)` Override
```css
.chartBox:not(:hover) :global(.recharts-line-curve) {
  stroke-opacity: 1;
}
```
**Limitation:** Only helps when cursor leaves component entirely, not legend→chart area transitions.

#### 2. Native `pointerleave` Listener
```javascript
el.addEventListener('pointerleave', () => {
  // Dispatch pointerout to trigger Mantine's cleanup
  el.firstElementChild?.dispatchEvent(
    new PointerEvent('pointerout', { bubbles: true, relatedTarget: el })
  )
})
```
**Limitation:** Only fires when leaving the component entirely.

https://mantine.dev/charts/composite-chart/#legend

### Possible fix

_No response_

### Self-service

restareaByWeezy:
Fixes #8750

## Root Cause

Each legend item called \`onHighlight(null)\` in its \`onMouseLeave\` handler.
This triggers a state update in the parent chart, causing a re-render that
replaces the legend item DOM nodes. If the cursor has already moved on,
the new node never receives the mouseleave event — leaving the highlight stuck.

There's also a race: when moving quickly from A to B, A's \`onMouseLeave\`
(which sets null) could fire after B's \`onMouseEnter\` (which sets B),
clearing the highlight even though the cursor is on B.

## Fix

Moved \`onHighlight(null)\` from individual item \`onMouseLeave\` handlers to
the container Box's \`onMouseLeave\`.

Since \`mouseleave\` is non-bubbling, attaching it to the container doesn't
interfere with item-level enter events. It only fires when the cursor actually
leaves the entire legend area.

## Tests

Added \`ChartLegend.test.tsx\` with 6 tests:
- mouseenter calls onHighlight with the correct dataKey
- moving between items does NOT call onHighlight(null) (regression test)
- container mouseleave calls onHighlight(null)
- rapid A→B transition produces no intermediate null
- sequential hover through all items produces no null between transitions


restareaByWeezy:
Thanks for the review, @HardikDewra! Also just pushed a merge commit to resolve the conflict with upstream master — all tests passing now.

rtivital:
Please remove files with personal setup from the PR (`.omc` folder)

restareaByWeezy:
@rtivital oops, sorry about that — those slipped in from my local tooling. removed in 84595d4.

rtivital:
Thanks!

Issue #8769, opened 2026-03-25
`useLocalStorage` has incorrect return type when no `defaultValue` is specified


mifi:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

8.3.18

### What package has an issue?

@mantine/hooks

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Not applicable – issue is not related to the browser

### Describe the bug

Hi! I believe the returned typescript type of the returned value of `useLocalStorage` is wrong. Example:
```ts
const [test] = useLocalStorage({ key: 'testkey' });
console.log(typeof test)
```
In the above example vscode reports `test` as having the type `string`, however runtime it logs `typeof test` as `undefined`.

The same applies to if I use a generic type like this:
```ts
const [test] = useLocalStorage<number>({ key: 'testkey' });
console.log(typeof test)
```
vscode reports `test` as being `number`, however runtime it logs `typeof test` as `undefined`.

I believe this is a typescript-only issue, because I do indeed expect `undefined` to be returned when there is no such key in local storage.

### If possible, include a link to a codesandbox with a minimal reproduction

need to sign up to use this service :/

### Possible fix

I believe the fix is to always add `| undefined` to the return type of `useLocalStorage` **unless** there is a `defaultValue` prop specified, in that case return that type.

Let me know what you think. Thanks!

### Self-service

- [x] I would be willing to implement a fix for this issue

dkpark10:
I'd like to work on this issue.

  The fix would be adding function overloads:
  - With `defaultValue`: returns `T`
  - Without `defaultValue`: returns `T | undefined`

  This is technically a breaking change for TypeScript users
  who didn't provide `defaultValue` (their code may now show type errors).

  Should I proceed with a PR?

mifi:
Im not a maintainer but sounds good to me! I dont think it’s a breaking change, because it is a bugfix that will change the types from incorrect (currently) to correct

rtivital:
Fixed in [9.0](https://github.com/mantinedev/mantine/releases/tag/9.0.0)

PR #8770, opened 2026-03-26
Fix/local storage defaultvalue type

dkpark10:
Fixes: #8769 

Adds function overloads to useLocalStorage and useSessionStorage for correct return type inference based on
  defaultValue presence.

## Changes
- When defaultValue is provided: return type is T
- When defaultValue is not provided: return type is T | undefined

## Before
```javascript
// Returns T even without defaultValue (incorrect type)
const [value, setValue] = useLocalStorage<string>({ key: 'my-key' });
// value: string (but could actually be undefined)
```

## After
```javascript
// Without defaultValue → T | undefined
const [value, setValue] = useLocalStorage<string>({ key: 'my-key' });
// value: string | undefined ✅

// With defaultValue → T
const [value, setValue] = useLocalStorage({ key: 'my-key', defaultValue: 'hello' });
// value: string ✅
```

### DIFF
--- a/packages/@mantine/hooks/src/use-session-storage/use-session-storage.ts
+++ b/packages/@mantine/hooks/src/use-session-storage/use-session-storage.ts
@@ -1,4 +1,17 @@
-import { createStorage, readValue, UseStorageOptions } from '../use-local-storage/create-storage';
+import {
+  createStorage,
+  readValue,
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks for submitting the PR! The fix was implemented in another PR

Issue #8757, opened 2026-03-19
Uncontrolled Popover.Target children onClick doesn't receive event object


kamyku12:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

8.3.1

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Firefox

### Describe the bug

When the Popover is uncontrolled the event object is not passed down to the Popover.Target children onClick method. 

But in a controlled Popover, the Popover.Target children onClick method receives that event object.

### If possible, include a link to a codesandbox with a minimal reproduction

https://codesandbox.io/p/sandbox/mantine-react-template-forked-cqt7wq

### Self-service

- [ ] I would be willing to implement a fix for this issue

PR #8771, opened 2026-03-26
fix(PopoverTarget): pass event to child onClick handler

kamyku12:
Fixes: #8757 

Passes down event object down to the child onClick function when PopoverTarget component was in "uncontrolled" state.


rtivital:
Thanks!

PR #8772, opened 2026-03-26
fix(hooks): correct return type for useLocalStorage when no defaultValue

### LINKED ISSUE #8769: `useLocalStorage` has incorrect return type when no `defaultValue` is specified

JivinSardine:
When useLocalStorage is called without a defaultValue prop, the return type now correctly includes '| undefined' since the actual value can be undefined at runtime when the key doesn't exist in localStorage.

Fixes #8769

### DIFF
--- a/packages/@mantine/hooks/src/use-local-storage/use-local-storage.ts
+++ b/packages/@mantine/hooks/src/use-local-storage/use-local-storage.ts
@@ -1,5 +1,7 @@
-import { createStorage, readValue, UseStorageOptions } from './create-storage';
+import { createStorage, readValue, UseStorageOptions, UseStorageReturnValue } from './create-storage';
 
+export function useLocalStorage<T>(props: UseStorageOptions<T> & { defaultValue: T }): UseStorageReturnValue<T>;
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks for submitting the PR! The fix was implemented in another PR

PR #8777, opened 2026-03-27
fix(charts): handle equal values in HeatmapChart color scale

### LINKED ISSUE #8754: Haetmap only display values if there is more than one different values
### Dependencies check up

8.3.18

@mantine/charts

Next.js

Chrome

### Describe the bug

Mantine  Heatmap does not show any values when data= {"2025-01-01": 1, "2025-02-01":1} but shows both values when data= {"2025-01-01": 1, "2025-02-01": 2}. The criteria seems to be that you must have different values. 

_No response_

### Possible fix

_No response_

### Self-service

maether44:
## Root Cause
When all values are equal, `max - min === 0` produces `NaN`, 
making all cells invisible.

## Fix
- const percent = (value - min) / (max - min);
+ const percent = max === min ? 1 : (value - min) / (max - min);

Closes #8754


rtivital:
Thanks!

PR #8792, opened 2026-04-01
[@mantine/mcp-server] Fix stdio transport to comply with MCP spec

### LINKED ISSUE #8789: MCP server protocol handshake is not compatible with MCP clients (VS Code etc): waits forever for `initialize` request
### Dependencies check up

9.0.0-alpha.1

@mantine/core

### Describe the bug

When running @mantine/mcp-server with Visual Studio Code as an MCP client, the server never responds to the initial `initialize` request and the handshake fails.

**Root cause:**
The MCP server implementation expects messages with LSP-style HTTP headers (i.e., `Content-Length: ...` and double carriage returns `\r\n\r\n`), which is not compatible with the Model Context Protocol standard. MCP clients (such as VS Code) send newline-delimited JSON (NDJSON) without headers. As a result, MCP clients wait forever for a response because the server never reads the incoming requests.

**Steps to reproduce:**
1. Install and configure the Mantine MCP server with VS Code MCP extension support (or any other compliant MCP client)
2. Observe that the server never responds to the `initialize` request

**Expected behavior:**
The MCP server should understand both standard NDJSON and (optionally) LSP framed requests—or at minimum, NDJSON as required by the MCP spec.

**Suggested fix:**
Update the stdin transport handling in `packages/@mantine/mcp-server/src/server.ts` to support newline-delimited JSON

Feel free to reach out for a code snippet or PoC if needed.

_No response_

### Possible fix

The relevant code to update is inside `startServer()` in mcp-server/src/server.ts. Replace the LSP-style header & buffer approach with a `readline`-based NDJSON stream parser on stdin. This would enable compatibility with all MCP-compliant clients including VS Code.

### Self-service

semimikoh:
## What does this PR do?
  Fixes #8789

  ## Problem
  The MCP server uses LSP-style `Content-Length` header framing for stdio transport, which is incompatible with the MCP specification. This causes MCP clients (VS Code, etc.) to hang indefinitely during the
  `initialize` handshake.

  ## Solution
  - Replaced `Content-Length` header framing with NDJSON (newline-delimited JSON) as required by the MCP spec
  - Switched stdin parsing from buffer-based header extraction to `readline`-based line processing
  - Removed the now-unnecessary `readContentLength()` function

  ## Testing
  Manually verified that the server correctly parses NDJSON input and responds without headers, matching the MCP stdio transport specification.

### DIFF
--- a/packages/@mantine/mcp-server/src/server.ts
+++ b/packages/@mantine/mcp-server/src/server.ts
@@ -1,3 +1,4 @@
+import readline from 'readline';
 import { MantineMcpDataClient } from './data-client';
 import { GetItemArgs, ListItemsArgs, SearchDocsArgs } from './types';
 
@@ -118,7 +119,7 @@ function parseSearchDocsArgs(args: Record<string, unknown>): SearchDocsArgs | nu
 }
 
 function writeMessage(payload: unknown) {
-  process.stdout.write(JSON.stringify(payload) + '\n');
+  process.stdout.write(`${JSON.stringify(payload)}\n`);
 }
 
 function writeResponse(id: string | number | null, result: unknown) {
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #8810, opened 2026-04-05
[@mantine/form] Fix missing TS suggestions for path in clearFieldError, resetField, clearFieldDirty

### LINKED ISSUE #8808: mantine/form some field functions don't provide TS suggestions for `path` parameter
### Dependencies check up

9.0.0

@mantine/form

Vite

All

### Describe the bug

`restField` and `clearFieldError` not providing TS suggestions for `path` parameter

`setFieldValue`, `validateField` and `setFieldError` working fine

Image
Image
Image
Image
Image

https://stackblitz.com/edit/vitejs-vite-fdyygz1w?file=src%2FApp.tsx

### Possible fix

_No response_

### Self-service

semimikoh:
## Problem

`clearFieldError` and `resetField` have `path` parameters typed as `unknown` and `PropertyKey`, which disables TypeScript autocomplete for form values. This is inconsistent with `setFieldValue`, `setFieldError`, and `validateField`, which use `LooseKeys<Values>` and provide proper suggestions.

## Fix

Aligned the `path` parameter type across the form API by using the `LooseKeys<Values>` generic:

| Function          | Before                         | After                                                    |
| ----------------- | ------------------------------ | -------------------------------------------------------- |
| `clearFieldError` | `(path: unknown) => void`      | `<Field extends LooseKeys<Values>>(path: Field) => void` |
| `resetField`      | `(path: PropertyKey) => void`  | `<Field extends LooseKeys<Values>>(path: Field) => void` |
| `clearFieldDirty` | `(path: unknown) => void`      | `<Field extends LooseKeys<Values>>(path: Field) => void` |

`clearFieldDirty` wasn't mentioned in the original issue but had the same problem, so I included it for consistency with the rest of the form API. Its runtime implementation (`getPath`-based, same as `setFieldValue`) already supports nested paths, so the type change reflects actual behavior.

## Testing

- Type check passes (`tsc --noEmit`)
- All 389 `@mantine/form` tests pass
- Lint/format clean
- Verified the bug reproduces in the Stackblitz from the issue (no autocomplete on `clearFieldError` / `resetField`)
- Fix verified via local type check against the same usage pattern


rtivital:
Thanks!

PR #8811, opened 2026-04-06
[@mantine/core] fix: darkHidden missing in light LoadingOverlay

ScotchAndSoda:
In [this commit](https://github.com/mantinedev/mantine/commit/17acca212ebdcb107498f05071b73b6c718ce957) darkHidden has been removed (supposedly, by a mistake) from the light LoadingOverlay, this rendered both overlays at the same time with dark theme on.


rtivital:
Thanks!

PR #8820, opened 2026-04-07
fix: link checkbox errors to input with aria-describedby

### LINKED ISSUE #8340: Error message of an erroneous checkbox component is not accessible to voice readers

oozan:
## Summary

Fix Checkbox accessibility so the input is linked to description and error content with `aria-describedby`.

## Changes

- add stable ids to `InlineInput` description and error elements
- wire `Checkbox` input to those ids via `aria-describedby`
- add tests for description and error associations

Closes #8340


rtivital:
Thanks!

PR #8823, opened 2026-04-08
Fix Combobox dropdown z-index inside FloatingWindow

addu2612:
## Summary
- raise the Combobox dropdown z-index when it is rendered inside `FloatingWindow`
- provide `FloatingWindow` z-index through context so nested Combobox/Select instances can inherit it
- add a regression test covering `Select` opened inside `FloatingWindow`

## Testing
- Added focused regression coverage in `packages/@mantine/core/src/components/Select/Select.test.tsx`

Closes #8813


rtivital:
Thanks for submitting the PR! This is not the intended fix for the issue

PR #8826, opened 2026-04-08
[mantine.dev] AppShell: Fix useHeadroom destructuring in Headroom example

sdelpercio:
## Summary

`useHeadroom` returns an object `{ pinned, scrollProgress }`, but the Headroom example was assigning the return value directly to `pinned` (`const pinned = useHeadroom(...)`). This made `pinned` always truthy (an object reference), so the header's `collapsed` prop never toggled and the headroom behavior was completely broken.

## Changes

- Fix `const pinned` → `const { pinned }` in `Headroom.tsx`, `code.json`, and the inline snippet in `app-shell.mdx`


rtivital:
Thanks!

PR #8830, opened 2026-04-13
Fix form validate type inference

dkpark10:
## Problem

The useForm hook allows unknown keys in the validate option that are not present in initialValues, which is a type safety gap.

##  Solution

Unknown keys in validate are now inferred as never, causing a type error when keys not present in initialValues are used.

This should not affect runtime behavior, as unrecognized keys in the validate rules record are already silently ignored during validation.

```javascript
const form = useForm({
  initialValues: {
      orange: '',
      banana: '',
    },
    validate: {
      orange: (value) => value.length < 2 ? 'Too Short' : null,
      banana: (value) => value.length < 2 ? 'Too Short' : null,
      unknownProperty: () => null,
    },
);

form.validate().errors; //  { orange: 'Too Short', banana: 'Too Short' };
```

**As-is**
스크린샷 2026-04-13 오후 7 38 30

**To-be**
스크린샷 2026-04-13 오후 7 38 06

### DIFF
--- a/packages/@mantine/form/src/tests/use-form/validate-types.test.ts
+++ b/packages/@mantine/form/src/tests/use-form/validate-types.test.ts
@@ -1,0 +1,44 @@
+import { renderHook } from '@testing-library/react';
+import { useForm } from '../../use-form';
+
+describe('@mantine/form/validate-type-safety', () => {
+  it('accepts valid validate keys from initialValues', () => {
+    const hook = renderHook(() =>
+      useForm({
+        initialValues: { firstName: '', lastName: '' },
+        validate: {
+          firstName: (value) => (value.length < 2 ? 'Too short' : null),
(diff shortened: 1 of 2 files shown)


rtivital:
I'd rather not add this as it might break existing workflows and make code reuse harder. I do not see issues with having extra validations that do not run.

Example with validators that would break:
```tsx
const userValidators = {
    email: (v: string) => (isEmail(v) ? null : 'Invalid email'),
    name: (v: string) => (v.length < 2 ? 'Too short' : null),
    phone: (v: string) => (isPhone(v) ? null : 'Invalid phone'),
  };

  // Profile form — only name + email
  useForm({
    initialValues: { name: '', email: '' },
    validate: userValidators,   // breaks: 'phone' isn't in initialValues
  });
```

PR #8833, opened 2026-04-13
Reset useDebouncedValue `leading` trigger more consistently

noahsilas:
When the "leading" edge is fired, it sets the cooldown ref but doesn't start a timer to clear it. This means that the cooldown after the first "leading" edge was effectively infinite.

- Added a timer to the leading edge handler to clear the cooldown after the wait period

- Added test coverage for this hook, including a scenario covering this exact case.


rtivital:
Thanks!

PR #8841, opened 2026-04-17
[@mantine/core] Checkbox, Switch, Radio: link error message via aria-describedby

### LINKED ISSUE #8340: Error message of an erroneous checkbox component is not accessible to voice readers

akinshaywai:
## Problem

Fixes #8340

Screen readers (VoiceOver, NVDA) do not announce the error message when a `Checkbox`, `Switch`, or `Radio` has an `error` prop. This is because:

- The error element rendered by `InlineInput` has no `id`
- The `<input>` has no `aria-describedby` pointing to it

`TextInput` and other text-based inputs handle this correctly through `InputWrapper` → `InputWrapperProvider` → `describedBy` context, but inline inputs (Checkbox, Switch, Radio) use `InlineInput` which had no equivalent mechanism.

## Fix

Three-part change:

### 1. `InlineInput` — accept `errorId` and forward it to `Input.Error`
Added optional `errorId?: string` prop. When provided it is passed as `id` to the rendered `Input.Error` element so assistive technology can find it by id.

### 2. `Checkbox` — wire `errorId` to the input
Derives `errorId = error ? \`${uuid}-error\` : undefined` and passes it to:
- `InlineInput` as `errorId` (gives the error element its id)
- The `<input>` as `aria-describedby` (links input → error message)

### 3. `Switch` and `Radio` — same fix
Both use `InlineInput` and had the same gap. Applying the identical pattern keeps behaviour consistent across all inline form controls.

## Testing

New tests in `Checkbox.test.tsx`:
- `links error message to input via aria-describedby` — asserts both the input attribute and the error element id
- `does not set aria-describedby when no error` — asserts no spurious attribute when error is absent

### DIFF
--- a/packages/@mantine/core/src/components/Checkbox/Checkbox.tsx
+++ b/packages/@mantine/core/src/components/Checkbox/Checkbox.tsx
@@ -206,6 +206,7 @@ export const Checkbox = factory<CheckboxFactory>((_props) => {
 
   const { styleProps, rest } = extractStyleProps(others);
   const uuid = useId(id);
+  const errorId = error ? `${uuid}-error` : undefined;
 
   const withContextProps = {
     checked: ctx?.value.includes(rest.value as string) ?? checked,
@@ -241,6 +242,7 @@ export const Checkbox = factory<CheckboxFactory>((_props) => {
       label={label}
       description={description}
       error={error}
+      errorId={errorId}
       disabled={finalDisabled}
       classNames={classNames}
       styles={styles}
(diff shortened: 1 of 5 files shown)


packages/@mantine/core/src/components/Checkbox/Checkbox.tsx
```diff
@@ -206,6 +206,7 @@ export const Checkbox = factory<CheckboxFactory>((_props) => {
 
   const { styleProps, rest } = extractStyleProps(others);
   const uuid = useId(id);
+  const errorId = error ? `${uuid}-error` : undefined;
```
rtivital, line 209: Incomplete logic: `error` prop can be boolean. Setting it to true would lead to invalid aria-describedby id.

rtivital (changes requested):

rtivital:
Duplicate of https://github.com/mantinedev/mantine/pull/8820

PR #8842, opened 2026-04-17
[@mantine/core] Fix SemiCircleProgress not being displayed

### LINKED ISSUE #8840: SemiCircleProgress not displayed
### Dependencies check up

9.0.2

@mantine/core

Vite

Chrome

### Describe the bug

The SemiCircleProgress component is not displayed, its also not displayed in the documentation : https://mantine.dev/core/semi-circle-progress/

_No response_

### Possible fix

_No response_

### Self-service

semimikoh:
## What is the purpose of this pull request?

  Fixes #8840.

  `SemiCircleProgress` lost explicit SVG dimensions when `viewBox` was added. As a result, the component no longer guaranteed that the
  `size` prop controlled the rendered SVG size, which could make the progress indicator appear missing in some layouts.

  This PR restores the explicit `width` and `height` attributes on the SVG while keeping the `viewBox`.

  ## What was changed?

  - Added `width={size}` and `height={size / 2}` to the `SemiCircleProgress` SVG
  - Added a regression test to verify SVG `width`, `height`, and `viewBox` values are derived from the `size` prop

  ## Testing

  ```sh
  npx jest packages/@mantine/core/src/components/SemiCircleProgress/SemiCircleProgress.test.tsx --runInBand

  Result:

  Test Suites: 1 passed
  Tests: 47 passed

### DIFF
--- a/packages/@mantine/core/src/components/SemiCircleProgress/SemiCircleProgress.test.tsx
+++ b/packages/@mantine/core/src/components/SemiCircleProgress/SemiCircleProgress.test.tsx
@@ -54,4 +54,13 @@ describe('@mantine/core/SemiCircleProgress', () => {
     render(<SemiCircleProgress {...defaultProps} label="test-label" orientation="down" />);
     expect(screen.getByText('test-label')).toHaveAttribute('data-orientation', 'down');
   });
+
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #8845, opened 2026-04-17
[@mantine/core] Radio: Center icon via translate transform

stasyzon:
Radio icon was centered via `calc(50% - size/2)`, which renders off-center on displays with fractional scaling (125%, 150%)                  
                                                                                                    
Switched to `top/left: 50%` + `translate: -50% -50%` so the compositor anti-aliases sub-pixel offsets instead of pixel snapping


rtivital:
translate is already used in transform for animations

stasyzon:
`translate` is a separate CSS property from `transform`. They compose, not conflict. New `translate` isn't involved in the animation

rtivital:
Thanks!

PR #8849, opened 2026-04-19
[@mantine/core] Highlight | Fix wholeWord unicode matching

Israadaassi1:
I first reproduced the problem by using Highlight with wholeWord on a string that contains a non‑ASCII word like īdem, and confirmed it did not get highlighted. Next I traced the behavior into packages/@mantine/core/src/components/Highlight/highlighter/highlighter.ts and saw that when wholeWord is enabled it builds the regex with \b(${matcher})\b, which only treats ASCII letters/digits/underscore as “word characters”. 

Then I replaced those \b boundaries with Unicode-aware “letter boundaries” using Unicode property escapes, so the pattern becomes (?<!\p{L})(${matcher})(?!\p{L}), and I updated the RegExp flags to include the u flag (giu) when wholeWord is true so \p{L} is supported. 
After that, I added unit/component tests that highlight īdem correctly and make sure we don’t match inside a longer Unicode word (e.g. īdem should not match inside īdemque). Finally I ran typecheck, lint, formatting, and the relevant Jest tests to verify everything is fine.

### DIFF
--- a/packages/@mantine/core/src/components/Highlight/Highlight.test.tsx
+++ b/packages/@mantine/core/src/components/Highlight/Highlight.test.tsx
@@ -132,6 +132,18 @@ describe('@mantine/core/Highlight', () => {
       expect(marks[0].textContent).toBe('test');
       expect(marks[1].textContent).toBe('testing');
     });
+
+    it('supports unicode letters in wholeWord matching', () => {
+      const { container } = render(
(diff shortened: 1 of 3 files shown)


rtivital:
Thanks for submitting the PR, the issue was fixed in another PR merged before this one

PR #8853, opened 2026-04-19
[@mantine/core] Highlight: Fix wholeWord matching for non-ASCII characters

### LINKED ISSUE #8838: Highlight component: wholeWord and Unicode
### Dependencies check up

9.0.1

@mantine/core

React Router

All

### Describe the bug

I'm using the `Highlight` component to highlight words in a sentence. I don't want to highlight "then" if "the" is to be highlighted, so I'm using the new `wholeWord` property. This works fine when the words only use ASCII characters.

However, with a word like *īdem* in `ergō numerus syllabārum et vōcālium īdem est`, the word is not highlighted while it should be.

I traced this to the RegEx in highlighter.ts here: https://github.com/mantinedev/mantine/blob/e58fb66f021a6065caae5c3755f57afcb9f076f5/packages/%40mantine/core/src/components/Highlight/highlighter/highlighter.ts#L45

The `\b` class only works on ASCII strings, it's the transition of a letter to a non-letter. But in my example, `ī` is not considered a letter since it's not an ASCII letter, so it is considered a word boundary and thus the RegEx doesn't match.

https://codesandbox.io/p/sandbox/mantine-v7-forked-wmpcvl

### Possible fix

I believe a fix is to use Unicode-aware matchers as described here for example: https://stackoverflow.com/a/57290540

Suggestion for a fix: Change the pattern to `(?<!\\p{L})(${matcher})(?!\\p{L})` when `wholeWord` is enabled. This will also require the `u` flag for the RegEx constructor.

### Self-service

dfedoryshchev:
Closes #8838.

## Bug
`Highlight` with `wholeWord` fails to match words that contain non-ASCII letters (`īdem`, `café`, `тест`). Reporter traced this to `highlighter.ts:45`, where the regex uses
ASCII `\b`. Since `\w` in JavaScript regex is ASCII-only (`[A-Za-z0-9_]`), characters like `é` and `ī` are treated as non-word, producing spurious word boundaries inside words.

## Fix
Two changes in `packages/@mantine/core/src/components/Highlight/highlighter/highlighter.ts`:
1. **`wholeWord` pattern**: replace `\b...\b` with Unicode-aware lookarounds `(?<![\p{L}\p{N}_])...(?![\p{L}\p{N}_])` and add the `u` flag. The character class `[\p{L}\p{N}_]` mirrors `\b`'s notion of "word character" (`\w` = letters + digits + underscore) but in a Unicode-aware way: any Unicode letter or number, plus underscore.
2. **`escapeRegex`**: tightened to escape only the actual ECMAScript regex syntax characters (`\ ^ $ . * + ? ( ) [ ] { } |`). The previous version also escaped `-`, `,`, `#`, which are not regex syntax. In non-`u` mode this over-escaping was a no-op, but under the new `u` flag it would throw `SyntaxError` for any wholeWord search containing those characters. Behavior is unchanged for the existing test suite.

## Tests (TDD)
Added 5 new test cases under `wholeWord option` in `highlighter.test.ts`:
- Reporter's exact case: `īdem` matches in Latin text.
- Common accented case: `café` matches in `I love café and croissant`.
- Negative Unicode case: `café` does NOT match inside `cafés` (lookahead correctly sees `s` as a letter).
- Digit-adjacent case: `test` does NOT match inside `test1` (lookahead sees `1` as `\p{N}`).
- Underscore-adjacent case: `hello` does NOT match inside `hello_world` (lookahead sees `_`).

Verified TDD red -> green: with the original code, the first three Unicode tests fail; the underscore test fails after only adding `[\p{L}\p{N}]` (without `_`). All 74 Highlight tests pass after the full fix.
`npx oxlint` and `npm run format:write:files` both pass on the changed files.

### DIFF
--- a/packages/@mantine/core/src/components/Highlight/highlighter/highlighter.test.ts
+++ b/packages/@mantine/core/src/components/Highlight/highlighter/highlighter.test.ts
@@ -123,5 +123,46 @@ describe('@mantine/core/Highlight/highlighter', () => {
         { chunk: '-case testing', highlighted: false },
       ]);
     });
+
+    it('matches whole word containing non-ASCII letters', () => {
+      const text = 'ergō numerus syllabārum et vōcālium īdem est';
+      expect(highlighter(text, 'īdem', { wholeWord: true })).toStrictEqual([
+        { chunk: 'ergō numerus syllabārum et vōcālium ', highlighted: false },
+        { chunk: 'īdem', highlighted: true },
+        { chunk: ' est', highlighted: false },
+      ]);
+    });
+
+    it('matches whole word containing accented characters', () => {
+      const text = 'I love café and croissant';
+      expect(highlighter(text, 'café', { wholeWord: true })).toStrictEqual([
+        { chunk: 'I love ', highlighted: false },
+        { chunk: 'café', highlighted: true },
+        { chunk: ' and croissant', highlighted: false },
+      ]);
+    });
+
+    it('does not match partial word across Unicode letter boundary', () => {
+      const text = 'cafés are nice';
+      expect(highlighter(text, 'café', { wholeWord: true })).toStrictEqual([
+        { chunk: 'cafés are nice', highlighted: false },
+      ]);
+    });
+
+    it('does not match word adjacent to digit', () => {
+      const text = 'test test1 final';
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #8856, opened 2026-04-21
fix(Card): resolve CardSection padding in RSC

### LINKED ISSUE #8846: <CardSection> has different behavior if it is in client vs non-client component
### Dependencies check up

9.0.2

@mantine/core

Next.js

Firefox 149, Chrome 146. Did not try others

### Describe the bug

`CardSection` respects the padding of the card when it is used in a component without `"use client";` at the top. With `"use client";` it ignores the padding (as is the intention I believe).

I believe this was introduced in v9 as this was not the case in v8.

I don't have code sandbox account so here are manual reproduction steps:

1. Create a Next.JS app with `npx create-next-app@latest --typescript`
2. Install Mantine `npm install @mantine/core`
3. Create the following files as `app/1.tsx` and `app/2.tsx`:
```ts
"use client";
import { Card, Box, CardSection } from "@mantine/core";

export default function Page() {
  return (
    <Card padding="lg" withBorder w="10rem" >
      <CardSection bg="magenta">
        <Box h="10rem">
          Card 1
        </Box>
      </CardSection>
      <Box>
        Test
      </Box>
    </Card>
  );
}
```
```ts
import { Card, Box, CardSection } from "@mantine/core";

export default function Page() {
  return (
    <Card padding="lg" withBorder w="10rem">
      <CardSection bg="magenta">
        <Box h="10rem">
          Card 2
        </Box>
      </CardSection>
      <Box>
        Test
      </Box>
    </Card>
  );
}
```

4. Replace `app/page.tsx` with the following:
```ts
import { MantineProvider, Stack } from "@mantine/core"
import '@mantine/core/styles.css';
import Card1 from "./1.tsx"
import Card2 from "./2.tsx"

export default function Home() {
  return <MantineProvider>
    <Stack m="md" gap="md">
      <Card1 />
      <Card2 />
    </Stack>
  </MantineProvider>
}
```

5. Observe the difference

Image

_No response_

### Possible fix

_No response_

### Self-service

malekmimouna:
Fixes #8846

#### Root cause identified and fix proposed
After investigating, the bug is in Card.tsx in the children mapping loop. The original check child.type === CardSection fails when CardSection is used inside a React Server Component (no "use client"). In Next.js with Turbopack, RSC children are wrapped in a react.lazy object for SSR chunking, which means the direct reference comparison always returns false. As a result, cloneElement is never called and data-orientation is never added to the section, so the CSS negative margins that create the full-bleed effect never apply.
#### Fix:
tschild.type === CardSection ||
(child.type as any)?.displayName === "@mantine/core/CardSection" ||
(child.type as any)?._payload?.value?.[2] === "CardSection"
#### Reproduction environment:
Next.js 16.2.4 (Turbopack), Mantine 9.0.2, React 19.2.4.

### DIFF
--- a/packages/@mantine/core/src/components/Card/Card.tsx
+++ b/packages/@mantine/core/src/components/Card/Card.tsx
@@ -95,11 +95,13 @@ export const Card = polymorphicFactory<CardFactory>((_props) => {
 
   const _children = Children.toArray(children);
   const content = _children.map((child, index) => {
-    if (typeof child === 'object' &&
-  child &&
-  'type' in child &&
-  (child.type === CardSection ||
-    (child.type as any)?.displayName === '@mantine/core/CardSection')) {
+    if (
+      typeof child === 'object' &&
+      child &&
+      'type' in child &&
+      (child.type === CardSection ||
+        (child.type as any)?.displayName === '@mantine/core/CardSection')
+    ) {
       return cloneElement(child, {
         'data-orientation': orientation,
         'data-first-section': index === 0 || undefined,
--- a/packages/@mantine/core/src/components/Card/Card.tsx
+++ b/packages/@mantine/core/src/components/Card/Card.tsx
@@ -95,7 +95,9 @@ export const Card = polymorphicFactory<CardFactory>((_props) => {
 
   const _children = Children.toArray(children);
   const content = _children.map((child, index) => {
-    if (typeof child === 'object' && child && 'type' in child && child.type === CardSection) {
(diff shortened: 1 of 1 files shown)


malekmimouna:
Thank you for the review! I've addressed the feedback in the latest commit:

- Removed the brittle _payload check entirely
- Switched to single quotes for consistency with the codebase
- Reformatted the condition to the multi-line style

The fix now relies only on displayName as a stable fallback alongside the direct reference check.

rtivital:
Thanks!

PR #8858, opened 2026-04-22
[@mantine/core] Fix renderOption type

seanwu1105:
The `renderOption` type of `Select` and `MultiSelect` use the wrong default generic type. The PR fixes it by passing the generic `Value` type from the component to it.


rtivital:
Thanks!

PR #8874, opened 2026-05-05
[@mantine/schedule] MonthView: render multi-day events overlapping maxed-out days

liamdon:
I noticed an issue with multi-day events that overlap a day that has been truncated by `maxEventsPerDay` - the entire event disappears. This is particularly disconcerting when using drag/drop, because the invisible event can no longer be dragged. 

The **Before** GIF illustrates the problem.

We could sort all-day or multi-day events to the top of the list, but that would only mask the problem. And on a real calendar, I think all-day events are often lower priority than timed events, so we would probably prefer to retain the existing sort.

The solution is to:
- Split multi-day events into render-only fragments when they cross a day hidden behind `maxEventsPerDay`.
- Keep MonthView’s event data, grouping, and row assignment unchanged.
- Preserve original click and drag behavior by having each fragment still reference the original event.
- Mark clipped fragment edges with angled end caps so it is visually clear the event continues through a hidden day.

### Before
multiday-before

### After
multiday-after


rtivital:
Thanks!

PR #8890, opened 2026-05-13
Accent insensitive option for Highlight

sarioglu:
This PR introduces two props `caseInsensitive` and `accentInsentive` to Highlight component for better control over what to match. I chosen to make them true by default to keep the previous behavior with cases and have a comparable default for accents too.

A short explanation on why this is needed can be found in this discussion: https://github.com/orgs/mantinedev/discussions/8884


rtivital:
Thanks!

sarioglu:
Thanks for the fixes and improvements you made on top of this PR!

PR #8901, opened 2026-05-16
 [@mantine/core] Select: fix clear button not showing for falsy primitive values

### LINKED ISSUE #8894: Select with `clearable` doesn't have clear button in the right section when the value is `false`
### Dependencies check up

9.2.1

@mantine/core

Next.js

Chrome

### Describe the bug

Reproducable in Chrome, Edge, and FireFox. With or without Next.js.

When a Select with `clearable` has a boolean value of `false`, the clear button in the right section doesn't appear.

Image

 

https://codesandbox.io/p/sandbox/mantine-react-template-forked-r4fym7

### Possible fix

Maybe change that condition so that it can display the clear button when the selected value is not `null`, so it can appear for `false` as well.
Or add a new prop such as `renderClearButton?: (v: null | TPrimitive) => boolean`

### Self-service

hyeongjun6364:
Fixes #8894

## Problem

When `Select` has the `clearable` prop and the current value is a falsy primitive (`false`, `0`, `""`), the clear button does not appear even though a valid option is selected.

The condition `!!_value` incorrectly treats falsy values as "no value":

```tsx
// Before
const _clearable = clearable && !!_value && !disabled && !readOnly;
```

## Fix

Replaced `!!_value` with `_value != null` so the clear button is hidden only when the value is actually absent (`null`).

```tsx
// After
const _clearable = clearable && _value != null && !disabled && !readOnly;
```

Applied the fix to `Select` clear button visibility checks.

## Tests added

* [x] displays clear button when clearable is true and value is set
* [x] does not display clear button when clearable is true and value is null
* [x] displays clear button when clearable is true and value is an empty string
* [x] displays clear button when clearable is true and value is false
* [x] displays clear button when clearable is true and value is 0

### DIFF
--- a/packages/@mantine/core/src/components/Select/Select.test.tsx
+++ b/packages/@mantine/core/src/components/Select/Select.test.tsx
@@ -283,6 +283,54 @@ describe('@mantine/core/Select', () => {
     expect(input).toHaveStyle({ color: 'rgb(255, 0, 0)' });
   });
 
+  it('displays clear button when clearable is true and value is set', () => {
+    const { container } = render(
+      <Select data={['test-1', 'test-2']} value="test-1" clearable onChange={() => {}} />
+    );
+    expect(container.querySelector('.mantine-InputClearButton-root')).toBeInTheDocument();
+  });
+
+  it('does not display clear button when clearable is true and value is null', () => {
+    const { container } = render(
+      <Select data={['test-1', 'test-2']} value={null} clearable onChange={() => {}} />
+    );
+    expect(container.querySelector('.mantine-InputClearButton-root')).not.toBeInTheDocument();
+  });
+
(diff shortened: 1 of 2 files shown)


malekmimouna:
### Code Review: Logic and Test Suite look excellent! 

I checked out the branch locally to verify the fix. The change from `!!_value` to `_value != null` correctly solves the falsy primitive bug without any risk of regression.

The test coverage in `Select.test.tsx` is exceptionally thorough and directly validates all the edge cases:
* **Boolean:** `false` 
* **Numeric:** `0` 
* **Empty string:** `""` 
* **Omission validation:** Properly hidden when value is `null` 

Everything looks structurally sound, well-tested, and ready to merge.

rtivital:
Thanks!

PR #8905, opened 2026-05-18
[@mantine/core] Input: Fix sections misplaced when `dir` overrides parent direction (#8903)

### LINKED ISSUE #8903: rightsection missplace of ltr input on rtl page
### Dependencies check up

9.2.1

@mantine/core

Vite

All

### Describe the bug

Image

Image

Image
my html is rtl and i used mantine provider with direction.but if i want to have sppecific ltr textinput rigth and left section are broken

_No response_

### Possible fix

_No response_

### Self-service

semimikoh:
## Summary

  Closes #8903.

  When the host page is RTL (`<html dir="rtl">` or `MantineProvider direction="rtl"`) and a specific `TextInput`/`Input` opts back into LTR via
  `dir="ltr"`, the right/left sections used to remain visually mirrored.

  Two reasons:

  1. `Input` did not explicitly forward the `dir` prop to the wrapper. The wrapper inherited the RTL direction from the ancestor, so
  `inset-inline-start/end` resolved to the RTL positions.
  2. `Input.module.css` writes the RTL border-radius / text-align values via `:where([dir="rtl"]) .wrapper`. That selector matches whenever
  **any** ancestor is RTL, regardless of the wrapper's own `dir` attribute, so the CSS variables stayed RTL-flipped even after fixing (1).

  ## Changes

  - Add `dir` to `__InputProps` (typed via `React.ComponentProps<'input'>['dir']` to stay compatible with `ElementProps<'input', ...>` inheritance
   in `TextInput`, `Textarea`, `DateInput`, `TimeInput`, `TimePicker`, `SpotlightSearch`, etc.).
  - Forward `dir` to both the wrapper `Box` and the inner `<input>` in `Input.tsx`.
  - Add an `&[dir='ltr']` self-override block in `Input.module.css` (specificity `(0,2,0)` beats the `:where(...)` rule at `(0,1,0)`) so the LTR
  border-radius and `text-align` are restored when the wrapper itself is LTR inside an RTL ancestor.

  ## Test plan

  - New regression test in `Input.test.tsx`: `Input dir="ltr"` nested inside `<div dir="rtl">` keeps `dir="ltr"` on both wrapper and `<input>`.
  - Manually: render `<MantineProvider><html dir="rtl"><TextInput dir="ltr" rightSection="X" /></html></MantineProvider>` and confirm right
  section is on the right with correctly rounded corners.
  - `npm run typecheck` ✓
  - `npx jest packages/@mantine/core/src/components/Input/Input.test.tsx` ✓ (70 tests)
  - `npx oxlint` ✓

### DIFF
--- a/packages/@mantine/core/src/components/Input/Input.module.css
+++ b/packages/@mantine/core/src/components/Input/Input.module.css
@@ -152,6 +152,12 @@
     --left-section-border-radius: 0 var(--input-radius) var(--input-radius) 0;
     --right-section-border-radius: var(--input-radius) 0 0 var(--input-radius);
   }
+
+  &[dir='ltr'] {
+    --input-text-align: left;
+    --left-section-border-radius: var(--input-radius) 0 0 var(--input-radius);
+    --right-section-border-radius: 0 var(--input-radius) var(--input-radius) 0;
+  }
 }
 
 .input {
--- a/packages/@mantine/core/src/components/Input/Input.test.tsx
+++ b/packages/@mantine/core/src/components/Input/Input.test.tsx
@@ -159,6 +159,24 @@ describe('@mantine/core/Input', () => {
     expect(screen.getByRole('textbox')).toHaveAttribute('size', '5');
   });
 
+  it('sets dir attribute on the input wrapper and input element', () => {
+    const { container } = render(<Input dir="ltr" rightSection="right" />);
+
+    expect(getInputWrapper(container)).toHaveAttribute('dir', 'ltr');
+    expect(screen.getByRole('textbox')).toHaveAttribute('dir', 'ltr');
+  });
+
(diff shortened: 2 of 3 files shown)


rtivital:
Thanks!

mohammadj789:
on PasswordInput or Textinput with rightSection or leftSection it gets mixed up

semimikoh:
@mohammadj789 thanks for catching this! You're right — `PasswordInput` (and any component that spreads its rest props onto an inner `<input>` rather than the `Input` wrapper) wasn't covered by this PR, so `dir` never reached the wrapper that the `&[dir='ltr']` override targets.

I've opened #8935 to track it and submitted a fix in #8936. Could you check whether that resolves the cases you were seeing?

PR #8908, opened 2026-05-18
[@mantine/core] Expose safe polygon options in Menu.Sub

chbaefront:
### Summary

This PR exposes the underlying Floating UI `safePolygon` options on `Menu.Sub` via the `safeAreaPolygon` prop.

It keeps the current behavior by default with `safeAreaPolygon: true`, while allowing consumers to:

- disable the safe polygon behavior with `safeAreaPolygon={false}`
- tune the safe polygon behavior with options such as `buffer` and `requireIntent`
- account for submenu gaps introduced by positive `offset` values

### Testing

yarn jest packages/@mantine/core/src/components/Menu/Menu.test.tsx --runInBand

25 tests passed.


rtivital:
Thanks!

PR #8912, opened 2026-05-20
[core] sponsors: warn and skip when OpenCollective returns invalid lastTransactionAt

calebcgates:
### What

`scripts/docs/sponsors.ts` calls `dayjs(item.lastTransactionAt).isAfter(thirtyOneDaysAgo)` on data fetched live from `https://opencollective.com/mantinedev/members/all.json` (line 30). When OpenCollective returns a member whose `lastTransactionAt` is missing or unparseable, the existing `item.lastTransactionAt && dayjs(...)` short-circuit silently drops the sponsor from `apps/mantine.dev/src/.docgen/sponsors.json` — so the docs site renders without that sponsor and there is no signal in CI/log output that anything went wrong.

This PR splits the predicate into three explicit branches:

1. `lastTransactionAt` missing → drop the sponsor silently (unchanged behavior, just lifted out of the `&&` chain).
2. `lastTransactionAt` present but `dayjs(...)` returns an invalid Day.js object → drop the sponsor **and** emit `logger.error(...)` with the sponsor name and the offending value. So when OpenCollective changes their date format (they did once in 2020), the next docs deploy logs a clear diagnostic instead of failing silently.
3. Valid date → existing 31-day window logic, unchanged.

### Why

`item.lastTransactionAt` is the only `dayjs(...)` argument in the repo that comes from an external HTTP API (every other call site parses values already typed as `DateStringValue` / `Date` by `@mantine/dates`'s public surface, or is a `dayjs()` no-arg construction, or is a Storybook playground). This script is the one place where a malformed upstream payload silently degrades the deployed site.

### Test

Added `scripts/docs/sponsors.test.ts` with 4 cases covering: in-window, out-of-window, missing date, unparseable date. Run with `npx jest scripts/docs/sponsors.test.ts` — all 4 pass.

### Scope

One file + one test. No behavior change for valid OpenCollective responses. The early-return refactor matches the style of recent fixes such as `[@mantine/spotlight] Fix error thrown when listId is empty (#8863)`.

### Pre-flight

- [x] `npm run typecheck` clean (root `tsc --noEmit`)
- [x] `npx oxlint scripts/docs/sponsors.ts scripts/docs/sponsors.test.ts` clean
- [x] `npm run format:write:files` applied (oxfmt)
- [x] `npx jest scripts/docs/sponsors.test.ts` — 4 passing
- [x] No new dependencies; no `package.json` changes; `syncpack` not affected
- [x] Commit message follows `[core] sponsors: ...` convention from CLAUDE.md

### DIFF
--- a/scripts/docs/sponsors.test.ts
+++ b/scripts/docs/sponsors.test.ts
@@ -1,0 +1,97 @@
+import dayjs from 'dayjs';
+
+interface OpenCollectiveMember {
+  name: string;
+  role: string;
+  tier: string;
+  lastTransactionAt: string | null;
+}
+
+function isRecentSponsor(
+  item: OpenCollectiveMember,
+  thirtyOneDaysAgo: dayjs.Dayjs,
+  onInvalid: (msg: string) => void
+): boolean {
+  const isActiveSponsor = item.role === 'BACKER' && item.tier === 'sponsor';
+  if (!item.lastTransactionAt) {
+    return false;
+  }
+  const lastTransaction = dayjs(item.lastTransactionAt);
+  if (!lastTransaction.isValid()) {
+    onInvalid(
+      `Skipping sponsor "${item.name}" — OpenCollective returned unparseable lastTransactionAt: ${item.lastTransactionAt}`
+    );
+    return false;
+  }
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks for submitting the PR! The issue was fixed differently in another commit.

PR #8919, opened 2026-05-24
Docs/combobox readonly prop

maether44:
## What's missing
The `readOnly` prop is fully implemented in `Combobox.tsx` but 
completely absent from the documentation page.

## What was added
Added a `## Read only` section between "Control dropdown opened 
state" and "Popover props" explaining:
- What `readOnly` actually does (disables keyboard navigation)
- That `ComboboxOption` clicks still fire — so `readOnly` on 
  `<TextInput>` is also needed for full read-only UX
- An inline code example showing the idiomatic pattern


rtivital:
The PR contains extra unwanted changes and has merge conflict.

PR #8920, opened 2026-05-24
Fix/swatches datatype prop

maether44:
## Root Cause
In `Swatches.tsx`, `datatype` was explicitly destructured from props 
but never forwarded to the underlying `<div>` element.

`SwatchesProps extends ElementProps<'div'>`, so TypeScript accepts 
`<Swatches datatype="xsd:date" />` without error — but the prop was 
silently swallowed at runtime.

## Fix
Removed `datatype` from the destructure so it falls through into 
`...others` and gets forwarded to `<Box {...others}>` correctly.

## Change
One line removed in `Swatches.tsx` — no logic changes.

Closes #ISSUE_NUMBER


rtivital:
Not a real issue.

PR #8924, opened 2026-05-25
[@mantine/core] NavLink: Fix typo in story button label

dfedoryshchev:
Removes a duplicate "item" in the NavLink dynamic-nested story. Sibling button on the line above renders "Add item", so the counterpart should read "Remove item".


rtivital:
Thanks!

PR #8925, opened 2026-05-25
[@mantine/hooks] use-id: Keep id stable across effect cleanup (Activity)

### LINKED ISSUE #8898: useId violates its contract inside Activity
### Dependencies check up

9.2.1

@mantine/core

None

Firefox

### Describe the bug

[useId](https://mantine.dev/hooks/use-id/#usage)'s contract is that it generates an ID that will not change as long as a component is mounted. When used inside a React Activity, this is no longer accurate. When an Activity is hidden, all effects get cleaned up, and triggered again when the Activity is revealed. Since the ID is generated in an effect, that means useId returns a new ID every time an Activity is revealed.
https://github.com/mantinedev/mantine/blob/cd3bff379050b1587989d0fd75ff97812eaf5220/packages/%40mantine/hooks/src/use-id/use-id.ts#L9-L11

As the output of useId is used in `key` props, this results in the forced re-creation of some components every time their parent Activity is revealed, for example PinInput:
https://github.com/mantinedev/mantine/blob/cd3bff379050b1587989d0fd75ff97812eaf5220/packages/%40mantine/core/src/components/PinInput/PinInput.tsx#L192
https://github.com/mantinedev/mantine/blob/cd3bff379050b1587989d0fd75ff97812eaf5220/packages/%40mantine/core/src/components/PinInput/PinInput.tsx#L430

Please also note that useId uses React's hook of the same name, and the docs there explicitly say its output should _not_ be used as a `key`. React's useId also does not appear to have this issue, as replacing Mantine's useId with React's useId inside PinInput resolves the issue of all children getting recreated.

_No response_

### Possible fix

Do not use an effect to generate a static identifier.

### Self-service

spokodev:
## Problem

\`useId\` from \`@mantine/hooks\` should return a stable id for the lifetime of a component instance — that's the contract the docs advertise and that consumers like \`PinInput\` rely on (it uses the id as a child \`key\`).

Inside React 19's \`<Activity>\`, effects clean up on \`mode="hidden"\` and re-fire on \`mode="visible"\`, while the component instance and its state are preserved. Mantine's \`useId\` performs its random-id assignment inside \`useIsomorphicEffect(..., [])\`, so every reveal calls \`setUuid(randomId())\` again and the returned id changes — even though the component never unmounted.

The reported user-facing effect (from #8898): \`PinInput.tsx:192\` / \`:430\` use the \`useId\` output as a \`key\`. Every Activity reveal forces re-creation of all the children, including their internal state and focus, because the key changes.

## Repro

Bug is structural — visible by reading \`packages/@mantine/hooks/src/use-id/use-id.ts:9-11\`:

\`\`\`ts
useIsomorphicEffect(() => {
  setUuid(randomId());
}, []);
\`\`\`

The new test \`keeps the same id across Activity hide/reveal cycles\` reproduces it directly: capture id under \`<Activity mode=\"visible\">\`, toggle to \`mode=\"hidden\"\`, back to \`mode=\"visible\"\`, then re-capture. On \`master\` the second capture has a different \`mantine-xxx\` value than the first; the test fails.

## Fix

Add a \`useRef\`-based guard so the random-id swap fires at most once per component instance, no matter how many times the effect re-runs:

\`\`\`diff
- import { useId as useReactId, useState } from 'react';
+ import { useId as useReactId, useRef, useState } from 'react';
  import { useIsomorphicEffect } from '../use-isomorphic-effect/use-isomorphic-effect';
  import { randomId } from '../utils';

  export function useId(staticId?: string) {
    const reactId = useReactId();
    const [uuid, setUuid] = useState(\`mantine-\${reactId.replace(/:/g, '')}\`);
+   const hasInitializedRef = useRef(false);

    useIsomorphicEffect(() => {
+     if (hasInitializedRef.current) {
+       return;
+     }
+     hasInitializedRef.current = true;
      setUuid(randomId());
    }, []);
\`\`\`

Properties:

- **SSR safety preserved**: initial state is still the deterministic \`mantine-\${reactId-without-colons}\`, so the server-rendered HTML still matches the first client render (the 2025-07-10 \"Fix hydration mismatch\" commit's contract is untouched).
- **First-mount UX preserved**: the very first effect run still swaps to a short \`randomId()\`, exactly like today.
- **Activity stability restored**: \`useRef\` state persists across Activity hide/reveal cycles, so subsequent effect re-runs are no-ops and \`uuid\` stays stable for the lifetime of the component instance.
- **Instance isolation preserved**: a different \`<PinInput key=\"x\" />\` is a different component instance, with its own ref, so it still gets its own fresh \`randomId()\` on its first mount.

## Tests

Test file renamed \`use-id.test.ts\` → \`use-id.test.tsx\` to accept JSX. Existing two tests (\`returns static id\`, \`returns random id if static id is not provided\`) preserved verbatim. One new test added:

- **Activity stability**: renders a \`Capture\` component under \`<Activity mode=\"visible\">\`, toggles to \`hidden\`, back to \`visible\`, and asserts the id remained constant. **Fails on master** (the bug), **passes on this branch**.

Verified locally with \`npm run jest -- packages/@mantine/hooks/src/use-id\`: 3/3 pass on branch; the new Activity test produces two different \`mantine-*\` ids on master so the bug reproduces in CI.

\`oxlint\` on both changed files clean. \`tsc --noEmit\` at root clean.

Fixes #8898

## Note on disclosure

Bug surfacing, root-cause analysis (effect re-firing under Activity), and the two affected \`PinInput.tsx\` line references are from \`@invliD\` in the issue body. I confirmed the effect re-fires by reading the existing \`useIsomorphicEffect(..., [])\` site, applied the ref-guard fix, wrote the Activity-based test, and verified RED→GREEN. AI assistance was used while drafting the PR body; the fix and the test reflect my own engineering decisions.


rtivital:
Thanks!

PR #8933, opened 2026-05-29
[refactor] Fix grammar in test descriptions ("allows/allow to" -> "allows/allow you to")

dfedoryshchev:
Adds the missing direct object after "allow" in test descriptions across `@mantine` packages: `'allows to <verb>'` -> `'allows you to <verb>'` (and the same for the negative form `'does not allow to'`).

`'allow to <verb>'` lacks an explicit object in English; the convention is `'allow [someone] to <verb>'`.


rtivital:
Thanks for submitting the PR, I'd rather keep the current tests names.

PR #8936, opened 2026-06-02
[@mantine/core] PasswordInput: Fix sections misplaced when `dir` overrides parent direction

### LINKED ISSUE #8935: PasswordInput: sections misplaced when `dir` overrides parent direction
### Describe the bug

Follow-up to #8903 (fixed by #8905).

#8905 forwards `dir` to the `Input` wrapper so the `&[dir='ltr']` self-override correctly restores section positions/border-radius when a field opts back into LTR inside an RTL ancestor. This works for `TextInput` and other `InputBase`-based components.

However, **`PasswordInput`** is not covered. It renders `<Input component="div">` and routes its rest props (`{...rest}`) to the inner native `<input>` rather than to the `Input` wrapper, so `dir` never reaches the wrapper. As a result the `&[dir='ltr']` rule never triggers and the left/right sections (incl. the visibility toggle) stay mirrored.

Reported by @mohammadj789 in https://github.com/mantinedev/mantine/pull/8905#issuecomment-4597417197

### What package has an issue

@mantine/core

Other / I'm not sure

### In which browser did the issue occur?

All browsers

### Reproduction

```tsx
<div dir="rtl">
  <PasswordInput dir="ltr" rightSection="X" leftSection="L" placeholder="Your password" />
</div>
```

Expected: left section on the left, visibility toggle / right section on the right (LTR layout).
Actual: sections stay mirrored (RTL layout) — toggle ends up on the left.

semimikoh:
## Summary

Closes #8935. Follow-up to #8905.

#8905 fixed sections being misplaced when `dir` overrides an RTL ancestor by forwarding `dir` to the `Input` wrapper (so the `&[dir='ltr']` self-override applies). That covers `TextInput` and other `InputBase`-based inputs.

`PasswordInput` was not covered: it renders `<Input component="div">` and spreads its rest props onto the inner native `<input>`, so `dir` never reached the `Input` wrapper. The `&[dir='ltr']` rule therefore never triggered and the left/right sections (including the visibility toggle) stayed mirrored.

Reported in https://github.com/mantinedev/mantine/pull/8905#issuecomment-4597417197.

## Changes

- Destructure `dir` from `PasswordInput` props and forward it to the `Input` wrapper (`<Input component="div" dir={dir}>`) so the wrapper-level `&[dir='ltr']` override applies.
- Keep `dir` on the inner native `<input>` as well (it previously received it via `{...rest}` before being destructured out).

## Test plan

- New regression test in `PasswordInput.test.tsx`: `<PasswordInput dir="ltr" rightSection>` nested inside `<div dir="rtl">` keeps `dir="ltr"` on both the `Input` wrapper and the inner `<input>`.
- Verified visually in Storybook: `dir="ltr"` field inside an RTL container now renders the left section on the left and the visibility toggle on the right; plain inherited-RTL fields are unaffected.
- `npm run typecheck` ✓
- `npm run jest PasswordInput` ✓ (63 tests)
- `npx oxlint` ✓


rtivital:
The PR has conflicts. Please resolve them to merge.

rtivital:
Thanks!

PR #8937, opened 2026-06-02
[@mantine/core] Collapse: Fix keepMounted not defaulting to false

KasperiP:
`keepMounted` is documented as default false, but no default was set. When omitted it stayed undefined and the content logic kept children mounted while collapsed.

You had to explicitly pass keepMounted={false} to get the documented behavior.


rtivital:
The logic in the source code is correct, the PR does not change anything.

KasperiP:
Isnt it currently rendering children because there is no default value given. keepMounted is documented default false but its undefined?

KasperiP:
> Isnt it currently rendering children because there is no default value given. keepMounted is documented default false but its undefined? 

Change in rendering logic on this PR doesnt change behaviour. Just cleaner IMO. **But adding the default false is the fix.**

PR #8939, opened 2026-06-02
[@mantine/core] Tree: Fix arrow key navigation focusing hidden nodes

KasperiP:
When `keepMounted` is enabled, collapsed subtrees stay in the DOM inside an Activity component, which hides them by setting display: none on the wrapping `<ul>`. ArrowUp/ArrowDown navigation filtered treeitems by their own inline style.display, which never matched these nested `<li>` elements, so focus moved into hidden nodes (and stalled, since display:none elements are not focusable). Navigation now walks ancestors up to the tree root and skips any node inside a hidden subtree.

broken

### After this fix:
fixed


rtivital:
Thanks!

PR #8952, opened 2026-06-07
[DOCS] [NumberInput] value as a string: The standalone minus sign was missing from the list 

slk333:
- The standalone minus sign scenario was missing from the list (it was mentioned in the BigInt section but it should still be mentioned here)
- When validating the list, I noticed that the 4th bullet was inconsistent, since it didn't include the single quotes like the other bullets.


rtivital:
Thanks!

PR #8953, opened 2026-06-08
[@mantine/core] Input: Add keepErrorMounted prop for persistent error element (aria-live)

### LINKED ISSUE #8932: Select unmounts error element, making aria-live validation announcements unreliable
### Dependencies check up

9.1.0

@mantine/core

React Router

All

### Describe the bug

I am working on field-level validation accessibility and noticed that Mantine Select removes the error element from the DOM completely when there is no error.

This becomes a problem when trying to announce validation changes with aria-live.

For screen reader announcement to be reliable, the live region usually needs to remain mounted in the DOM and only have its text updated. With Select, the error container is unmounted entirely, so there is no stable node to attach aria-live to.

The result is that:

- the field error message is not represented by a persistent DOM node
- screen reader announcement becomes inconsistent or impossible if the app relies on Mantine’s built-in error rendering
- consumers have to render a second custom error element outside the component just to make validation accessible
- This is especially awkward because Mantine already has an error prop, so from the component API it looks like the accessibility path should already be supported.

### Expected behavior
One of these would make the component much easier to use accessibly:

1. keep the error container mounted in the DOM even when empty
2. provide an option like keepErrorMounted
3. provide a documented accessibility pattern / API for associating a persistent error live region with Select

### Actual behavior
When error becomes falsy, the error element is removed from the DOM entirely.

That means an app cannot depend on Mantine’s own error node as a persistent live region.

### Minimal example
A simplified example:

```
<Select
  label="Bank"
  data={[{ value: "1", label: "Bank A" }]}
  error={hasError ? "Required field" : null}
/>
```
When hasError changes from true to false, the error node disappears from the DOM.

When it changes back to true, a new node is mounted instead of updating an existing live region.

_No response_

### Possible fix

Possible options:

- keep the error element mounted and render an empty string when there is no error
- add a prop like keepErrorMounted

### Self-service

phamvutinh:
### Problem

`Input.Wrapper` (and therefore all inputs built on it — `Select`, `TextInput`, `Textarea`, etc.) **unmounts the error element entirely** when the `error` prop is falsy. This makes it impossible to use Mantine's built-in error node as a reliable `aria-live` region for field-level validation: each time an error appears a brand-new node is mounted instead of an existing one having its text updated, so screen reader announcements become inconsistent. Consumers currently have to render a second custom error element just to make validation accessible.

### Solution

Add an opt-in `keepErrorMounted` prop (default `false`, fully backwards compatible). When set, the error element stays mounted even when there is no error, rendering empty content. Its text is then updated in place when an error appears/disappears, giving consumers a stable live-region host:

```tsx
<TextInput
  label="Email"
  keepErrorMounted
  errorProps={{ 'aria-live': 'polite' }}
  error={hasError ? 'Required field' : null}
/>
```

### Changes

- `Input.Wrapper`: new `keepErrorMounted` prop on `__InputWrapperProps`; error node renders when `hasError || keepErrorMounted`, with empty content when there is no error. `aria-describedby` references the error id whenever the node is mounted.
- `use-input-props`: route `keepErrorMounted` to the wrapper so all inputs (`Select`, `TextInput`, …) inherit it.
- `get-input-offsets`: reserve the error slot spacing when `keepErrorMounted` so layout doesn't shift as the error toggles.
- Docs: new "keepErrorMounted prop" section in `input.mdx`.
- Tests added for `InputWrapper` and `get-input-offsets`.

### Open question for maintainer

I went with an opt-in prop (no behavior change, no layout regression). `aria-live` is left to the consumer via `errorProps` rather than applied automatically — happy to adjust the approach (e.g. a documented pattern only, or auto `aria-live`) based on your preference.

### Checks

`npm run typecheck`, `npx oxlint`, `npm run build`, and `npm run jest` for Input/Select all pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


rtivital:
Thanks for submitting the PR! Currently this feature is not planned to be implemeted.

PR #8957, opened 2026-06-09
[@mantine/core] Splitter: Add reset on double-click functionality

KasperiP:
Adds the ability to reset a `Splitter` handle by double-clicking it. Double-clicking a handle restores its two adjacent panes to their default ratio while preserving their combined size, so other panes are left untouched.

Double-click-to-reset is a near-universal convention for splitters/resizable panels (VS Code, browser devtools, most IDE layouts etc). I think this would be nice quality of life improvement to Mantine :) 

IMO this should be default behaviour and you can out-out via passing `resetOnDoubleClick={false}`

https://github.com/user-attachments/assets/537e0a0b-cceb-47ac-8e5b-991b4d5ea821

### Other libraries
Here are few other component libraries that have this behaviour:
- [Shadcn](https://ui.shadcn.com/docs/components/radix/resizable) (default behaviour)
- [And design](https://ant.design/components/splitter?utm_source=chatgpt.com#splitter-demo-reset) (supports via `onDraggerDoubleClick`)
- [Mantine split pane](https://gfazioli.github.io/mantine-split-pane/#reset-with-double-click) As a bonus (but would be nice to not need separate package)

### DIFF
--- a/packages/@mantine/core/src/components/Splitter/Splitter.test.tsx
+++ b/packages/@mantine/core/src/components/Splitter/Splitter.test.tsx
@@ -1,3 +1,6 @@
+import { createRef } from 'react';
+import { act, fireEvent } from '@testing-library/react';
+import { UseSplitterReturnValue } from '@mantine/hooks';
 import { render, screen, tests } from '@mantine-tests/core';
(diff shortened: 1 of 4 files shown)


rtivital:
Thanks!

PR #8963, opened 2026-06-10
[mantine.dev] Disable eye dropper in swatches-only color input demo

KasperiP:
This is a very minor nitpick 🤓 but this is supposed to demonstrate restricted color selection. However you can still choose any color using the color picker.


rtivital:
Thanks!

PR #8970, opened 2026-06-12
[@mantine/dates] TimePicker: Allow entering 01-09 hours over a selected 00 value

### LINKED ISSUE #8967: TimePicker bug with entry when 00:00
### Dependencies check up

9.3.1

@mantine/core

Next.js

Chrome

### Describe the bug

There is a bug in TimePicker component. You can even see the issue on the website demo.
https://mantine.dev/dates/time-picker/

By default it is: -- : --
Then you enter 00:00
Then you click on first two numbers (hours) and you want to type 02, when you press 0 it jumps to minutes - it's impossible to enter anytime from 01-09.

This bugs is only when time is 00:00
If there is any number different then 00 you can change it.

_No response_

### Possible fix

_No response_

### Self-service

spokodev:
Fixes #8967

### Problem

When the hours field shows `00` (e.g. the value is `00:00`), clicking the field selects its text, and pressing `0` as the first digit of `01`–`09` immediately moves focus to minutes: the `keydown` handler advances on `0` whenever the field's value is `0`, before the input event can apply the replacement. Times `01:xx`–`09:xx` cannot be typed over a zero value at all — the leading `0` always jumps away. Reproducible on the docs demo.

### Fix

Skip the advance when the entire field text is selected: in that state the keystroke replaces the value and starts a new entry, so the field keeps focus with `0` applied and the next digit completes the hour (`02` → focus moves to minutes through the regular `00`-prefix path). Pressing `0` again with no selection still completes `00` and advances, both for an empty field and after an over-selection restart.

The existing `moves focus to the next input when double zero is entered` test typed each `0` through a separate `userEvent.type()` call, and each call clicks the field first — re-selecting the text between keystrokes. The second keystroke now lands as `userEvent.keyboard('0')` on the already-focused field, which matches how a user actually types `00`; the asserted behavior (double zero advances) is unchanged.

### Tests

`allows entering 01-09 hours over a selected 00 value` fails on `master` (the `0` jumps to minutes, so hours stay `00` and the `2` lands in minutes) and passes with this change. `@mantine/dates` suite: 2110 passed / 0 failed. `tsc --noEmit`, `oxlint`, and `format:write:files` on the changed files are clean.

### DIFF
--- a/packages/@mantine/dates/src/components/SpinInput/SpinInput.tsx
+++ b/packages/@mantine/dates/src/components/SpinInput/SpinInput.tsx
@@ -67,7 +67,11 @@ export function SpinInput({
     }
 
     if (event.key === '0' || event.key === 'Num0') {
-      if (value === 0) {
+      const { selectionStart, selectionEnd, value: inputValue } = event.currentTarget;
+      const isEntireValueSelected =
+        inputValue.length > 0 && selectionStart === 0 && selectionEnd === inputValue.length;
+
+      if (value === 0 && !isEntireValueSelected) {
         event.preventDefault();
         onNextInput?.();
       }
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #8977, opened 2026-06-15
fix: allow editing midnight timepicker hours

### LINKED ISSUE #8967: TimePicker bug with entry when 00:00

pupuking723:
Fixes #8967

This keeps the hours input editable when the current value is `00:00`. When the whole `00` hours field is selected, typing the first `0` should replace the selection and keep focus in the hours input so users can enter values like `02`.

The existing double-zero auto-advance behavior is preserved for user-entered temporary zero values.

Verification:

- `yarn jest packages/@mantine/dates/src/components/TimePicker/TimePicker.test.tsx --runInBand`
- `yarn format:write:files packages/@mantine/dates/src/components/SpinInput/SpinInput.tsx packages/@mantine/dates/src/components/TimePicker/TimePicker.test.tsx`
- `yarn oxlint packages/@mantine/dates/src/components/SpinInput/SpinInput.tsx packages/@mantine/dates/src/components/TimePicker/TimePicker.test.tsx`
- `git diff --check`

### DIFF
--- a/packages/@mantine/dates/src/components/SpinInput/SpinInput.tsx
+++ b/packages/@mantine/dates/src/components/SpinInput/SpinInput.tsx
@@ -1,3 +1,4 @@
+import { useRef } from 'react';
 import { clamp } from '@mantine/hooks';
 import { padTime } from '../TimePicker/utils/pad-time/pad-time';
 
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks for submitting the PR, the issue is fixed in another PR #8970

PR #8978, opened 2026-06-15
[@mantine/core] Allow undefined className with `exactOptionalPropertyTypes`

bensaufley:
For projects with `exactOptionalPropertyTypes` and `noUncheckedIndexedAccess`, this change allows CSS Modules with basic typing (e.g. when types come from `vite/client`, `{ readonly [key: string]: string }`) not to need non-null assertion `!`, which is a code smell and disallowed by many linters. If the className is in fact undefined (and the `!` is incorrect), nothing breaks, so changing the typing does not make the behavior less safe.

Obviously there's an argument to be made that _if this change is to be made at all_, it should be made more broadly so that all relevant optional props are also marked as allowing `undefined`, but that expands scope of the change massively and at least for me, this is the place I encounter this issue constantly.


rtivital:
Thanks!

PR #8983, opened 2026-06-17
[@mantine/core] RollingNumber: Fix rendering "-0" for values that round to zero

AlexRixten:
## Problem

  `RollingNumber` renders `"-0"` when a negative value rounds to zero:

  ```tsx
  <RollingNumber value={-0.4} decimalScale={0} />                      
```

  Root cause

  In getDigitParts, the sign was derived from the raw value (negative: value < 0)
  while the digits were derived from the rounded magnitude. A value whose magnitude
  rounds to zero therefore kept its minus sign, and both render paths (buildValue,
  getRenderSlots) rendered a leading -.

  Fix

  Treat the value as negative only when the rounded magnitude is non-zero:

  const roundedToZero = !/[1-9]/.test(str);
  return { negative: value < 0 && !roundedToZero, /* ... */ };

  getDigitParts is the single source of truth for the sign and is consumed by both
  render paths, so the fix applies everywhere. Values that round to a non-zero
  magnitude are unaffected — e.g. -0.6 with decimalScale={0} still renders "-1".

### DIFF
--- a/packages/@mantine/core/src/components/RollingNumber/build-value.test.ts
+++ b/packages/@mantine/core/src/components/RollingNumber/build-value.test.ts
@@ -39,6 +39,11 @@ describe('buildValue', () => {
     expect(buildValue({ value: -42 })).toBe('-42');
   });
 
+  it('does not render "-0" for negative values that round to zero', () => {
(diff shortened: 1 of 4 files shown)


rtivital:
Thanks!

PR #8989, opened 2026-06-19
fix(form): export FormProviderProps from public entrypoint

### LINKED ISSUE #8985: FormProviderProps is not exported — causes TS2742 when re-exporting createFormContext result
### Dependencies check up

9.3.1

@mantine/form

Vite

### Describe the bug

I ran into a TS2742 error when trying to re-export the result of `createFormContext()` from a business component library that builds with `declaration: true` and `moduleResolution: "bundler"`.

The problem is that `createFormContext` returns a tuple where the first element is typed as `React.FC<FormProviderProps<...>>`, but `FormProviderProps` is not re-exported from `@mantine/form`'s public entrypoint. TypeScript can't write the `.d.ts` because it can't reference that interface through the public API.

```ts
import { createFormContext } from '@mantine/form';

interface MyFormData { name: string }

const [FormProvider, useFormContext, useForm] = createFormContext<MyFormData>();

// ❌ TS2742: The inferred type of 'FormProvider' cannot be named without a reference
//    to '@mantine/form/lib/FormProvider/FormProvider.js'. This is likely not portable.
export { FormProvider };

// ✅ These two are fine, their types resolve to publicly exported things
export { useFormContext, useForm };
```

Looking at `lib/index.d.ts`, only the function is exported but not the interface its return type depends on:

```ts
// index.d.ts
export { createFormContext } from './FormProvider/FormProvider.js';
export type * from './types';
// ^ FormProviderProps is NOT in ./types, it lives in FormProvider/FormProvider.d.ts
```

Right now I'm working around it by manually annotating the type, basically duplicating the interface shape:

```ts
const formContext = createFormContext<MyFormData>();

export const FormProvider: React.FC<{
  form: UseFormReturnType<MyFormData>;
  children: React.ReactNode;
}> = formContext[0];
```

It works but i don't think is the desire behaviour.

_No response_

### Possible fix

Add `export type { FormProviderProps } from './FormProvider/FormProvider` to packages/@mantine/form/src/index.ts

### Self-service

suva11144:
## What this fixes

`FormProviderProps` is used as the prop type for `FormProvider` (returned by `createFormContext`), but it was never re-exported from `@mantine/form`'s public entrypoint. This caused a TS2742 error for consumers building with `declaration: true` and `moduleResolution: "bundler"` who tried to re-export `FormProvider` from their own component library.

## Change

Added a single type export in `packages/@mantine/form/src/index.ts`:

\`\`\`ts
export type { FormProviderProps } from './FormProvider/FormProvider';
\`\`\`

## Testing

Ran `npm run typecheck` — passes cleanly across root, `mantine.dev`, and `help.mantine.dev`.

Fixes #8985

### DIFF
--- a/packages/@mantine/form/src/index.ts
+++ b/packages/@mantine/form/src/index.ts
@@ -1,5 +1,6 @@
 export { useForm } from './use-form.js';
 export { createFormContext } from './FormProvider/FormProvider.js';
+export type { FormProviderProps } from './FormProvider/FormProvider';
 export { createFormActions } from './actions/index.js';
 export { FORM_INDEX } from './form-index.js';
 export * from './validators/index.js';


rtivital:
Duplicate of https://github.com/mantinedev/mantine/pull/9009

PR #8990, opened 2026-06-20
[@mantine/core] Stack: Add responsive support for gap

### LINKED ISSUE #8984: Cannot use the base synthax for the responsive spacing not compatible with MantineSpacing type
### Dependencies check up

9.3.1

@mantine/core

Next.js

None

### Describe the bug

I can use the very practical base prop to size components from Mantine/core but not their spacing ?
I want to do : 
`            <Stack maw={{base : "1000px", md: "1300px"}} w="100%" align={{ base : "flex-start", md: "center")} gap={{ base: "24px", md: "32px" }}>`

but instead I have to use media queries for only half of the properties of the mantine/core component : 

`         
const isDesktop = useMediaQuery(`(min-width: ${theme.breakpoints.lg})`);
...
   <Stack maw={{base : "1000px", md: "1300px"}} w="100%" align={isDesktop ?  "flex-start" : "center"} gap={isDesktop ? "24px" : "32px"}>`

It does not really make sense to be allowed to use the responsive synthax only for half the properties of the same component.  

The error returned : 

Image

_No response_

### Possible fix

_No response_

### Self-service

kushalkumar1362:
[@mantine/core] Stack: Add responsive support for gap, align and justify props

gap, align and justify now accept StyleProp objects ({ base, md, ... })
like other style props. Values are resolved via parseStyleProps and
emitted as media-query rules instead of inline CSS variables, so they
can change per breakpoint. Closes #8984


rtivital:
Thanks for submitting the PR! This feature is not planned to be added to the library.

PR #9002, opened 2026-06-23
fix(dates): allow duration values greater than 9999 hours

### LINKED ISSUE #8995: Duration demo cannot be set to 10000+ hours

https://mantine.dev/dates/time-picker/#duration-type

### What is wrong?

The docs say there is no upper bound but I can't set hours to 10000 or higher.

### Self-service

itsayushnst13:
## What does this PR do?

Fixes #8995.

Duration values were being clamped to `9999:59:59` during paste handling even though the documentation states that duration values do not have an upper bound.

### Changes

* Removed the implicit `9999:59:59` maximum for `type="duration"` in `use-time-picker.ts`
* Added a regression test covering values greater than `9999` hours

### Before

10000:00:00 → 9999:59:59

### After

10000:00:00 → 10000:00:00

### Tests

* Added test: `allows values greater than 9999 hours in duration mode`
* All TimePicker tests pass

### DIFF
--- a/packages/@mantine/dates/src/components/SpinInput/SpinInput.tsx
+++ b/packages/@mantine/dates/src/components/SpinInput/SpinInput.tsx
@@ -33,8 +33,9 @@ export function SpinInput({
   disableAutoAdvance = false,
   ...others
 }: SpinInputProps) {
-  const maxDigit = getMaxDigit(max);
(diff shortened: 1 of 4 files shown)


rtivital:
Thanks!

PR #9003, opened 2026-06-23
Fix resource schedule multiday events

itsayushnst13:
## Summary

Fixes an issue where multi-day events were only displayed on their start date in `ResourcesDayView`.

## Problem

Events spanning multiple days were filtered based on whether their start date matched the currently displayed day. As a result, events that overlapped the current day but started earlier were not rendered.

## Solution

Updated event filtering to use date-range overlap checks instead of start-date matching.

Added a regression test that verifies an event spanning multiple days is displayed on an overlapping day.

## Testing

* Added regression test for multi-day overlapping events
* `get-resources-day-view-events` tests pass
* `ResourcesSchedule` tests pass


rtivital:
@codex review

PR #9006, opened 2026-06-24
[@mantine/core] Fix light-color CSS variable in dark mode

### LINKED ISSUE #8895: Subtle and transparent colors do not match version 8
### Dependencies check up

9.1.1

@mantine/core

Next.js

Chrome

### Describe the bug

_No response_

### Possible fix

_No response_

### Self-service

agustinanievas16:
Fixes [#8895](https://github.com/mantinedev/mantine/issues/8895)

## Problem

In dark mode, `--mantine-color-{color}-light-color` was hardcoded to shade 0 (near-white), causing the `light`, `subtle`, and `transparent` variants to lose color identity, so all colors looked the same washed-out white.

## Fix

Changed the dark mode formula from `shade-0` to `Math.max(primaryShade - 5, 0)` (shade-3 with the default `primaryShade: 8`), which gives a visible, distinguishable color without being too saturated for a dark theme.

### Subtle

Before
https://github.com/user-attachments/assets/99670767-edf4-4d0f-906c-0f5b65689859

After
https://github.com/user-attachments/assets/18766c8a-b075-46a5-981c-3df6738edb1f

### Light

Before
old-light

After
new-light

### Transparent

Before
old-trasparent

After
new-transparent

## Open questions

In the issue I raised some questions about whether `subtle` and `transparent` should be treated differently from `light`, since they have different background behavior. I'd love feedback on whether shade-3 is the right value or if a different approach is preferred.

Happy to make any changes based on your feedback!


rtivital:
Thanks for submitting a PR! This change is not planned to be implemented this way, light variant colors are not planned to be changed, once I figure out the better approach for colors I'll work on it myself.

PR #9013, opened 2026-06-25
fix(core): forward style and ref when transitionDuration is 0

### LINKED ISSUE #9010: Collapse does not apply the `style` and `ref` props when transitionDuration=0
### Dependencies check up

6.0.22

@mantine/core

Vite

All

### Describe the bug

When the style prop is set on Collapse or any component that uses it, and the duration is set to 0, the style is no longer applied. Reading the source, it seems ref is also not set in that [code branch](https://github.com/mantinedev/mantine/blob/v6/src/mantine-core/src/Collapse/Collapse.tsx#L58-L60).

https://stackblitz.com/edit/vitejs-vite-bslsxdft?file=src%2FApp.tsx

### Possible fix

1. Add [those props](https://github.com/mantinedev/mantine/blob/v6/src/mantine-core/src/Collapse/Collapse.tsx#L58-L60) to the short-circuited branch.
2. combine both branches to avoid further drift.

### Self-service

itsayushnst13:
Fixes an issue where Collapse did not forward style and ref when transitionDuration={0}.

What changed
Forward style and ref in the zero-duration render path.
Add regression tests to verify that style is applied and ref is forwarded when transitionDuration={0}.

Fixes #9010


rtivital:
@codex review

rtivital:
Thanks!

PR #9016, opened 2026-06-26
[@mantine/core] Collapse: Apply style and ref props when transitionDuration is 0

### LINKED ISSUE #9010: Collapse does not apply the `style` and `ref` props when transitionDuration=0

Hossam-Ismail:
## Fixes #9010

When `transitionDuration={0}` (or reduced motion forces the duration to `0`), `Collapse` short-circuits and renders a plain `Box`, but it does **not** forward the `style` and `ref` props. Both are destructured out of `...others`, so the early-return branches silently drop them. The normal (animated) branch forwards both via `getCollapseProps`.

This means consumers lose their inline styles and ref access whenever the collapse animation is disabled.

## Changes

- Forward `ref` and `style` (converted with `getStyleObject`) to the `Box` in both `duration === 0` return paths (the `Activity`-wrapped path and the plain conditional path).

## Test

Added a regression test asserting that, with `transitionDuration={0}`, the inline `style` is applied and the `ref` resolves to the rendered element.

### DIFF
--- a/packages/@mantine/core/src/components/Collapse/Collapse.test.tsx
+++ b/packages/@mantine/core/src/components/Collapse/Collapse.test.tsx
@@ -1,4 +1,5 @@
-import { tests } from '@mantine-tests/core';
+import { createRef } from 'react';
+import { render, tests } from '@mantine-tests/core';
 import { Collapse, CollapseProps } from './Collapse';
 
 const defaultProps: CollapseProps = {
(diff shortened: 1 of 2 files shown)


rtivital:
Duplicate of #9013

PR #9017, opened 2026-06-28
[@mantine/charts] Fix incorrect @default JSDoc values for several chart props

Hossam-Ismail:
## What

Several chart props document a `@default` value in their JSDoc that does not match the actual value defined in the component's `defaultProps`. Since the props tables on mantine.dev are generated from these JSDoc comments, the documentation currently shows incorrect defaults.

| Component | Prop | JSDoc `@default` | Actual `defaultProps` |
| --- | --- | --- | --- |
| `DonutChart` | `size` | `80` | `160` |
| `PieChart` | `size` | `80` | `160` |
| `Heatmap` | `monthsLabelsHeight` | `30` | `14` |
| `BarsList` | `minBarSize` | `100` | `0` |

## Changes

Updated each JSDoc `@default` to match the real runtime default. Documentation-only change, no behavior is affected.

### DIFF
--- a/packages/@mantine/charts/src/BarsList/BarsList.tsx
+++ b/packages/@mantine/charts/src/BarsList/BarsList.tsx
@@ -64,7 +64,7 @@ export interface BarsListProps
   /** Controls gap between bars @default 5 */
   barGap?: MantineSpacing;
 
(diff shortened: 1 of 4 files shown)


rtivital:
Thanks!

PR #9019, opened 2026-06-29
[@mantine/hooks] use-intersection: Use the last entry from a batched observer callback

mahirhir:
Follow-up to #8509, which fixed the same issue in `use-in-viewport`.

`useIntersection` reads only the first entry from the observer callback:

```ts
new IntersectionObserver(([_entry]) => {
  setEntry(_entry);
}, options);
```

When the callback is delayed across frames (busy main thread, fast scrolling), the browser can queue more than one `IntersectionObserverEntry` for the observed element and deliver them in a single call. The first entry then holds an older state and the last one holds the current state, so the hook can report a stale `entry`. This is the same situation #8509 describes for `use-in-viewport`.

The fix mirrors #8509: read the last entry.

```ts
new IntersectionObserver((entries) => {
  setEntry(entries[entries.length - 1]);
}, options);
```

I left `use-resize-observer` as it is even though it also reads `entries[0]`: a `ResizeObserver` reports the current size of each target once per delivery instead of accumulating one entry per frame, so its first entry is already the latest.

Added a regression test that feeds a mock observer a two-entry batch and checks the hook keeps the last entry. Without the change the test fails because the hook returns the stale first entry.

### DIFF
--- a/packages/@mantine/hooks/src/use-intersection/use-intersection.test.ts
+++ b/packages/@mantine/hooks/src/use-intersection/use-intersection.test.ts
@@ -1,0 +1,62 @@
+import { act, renderHook } from '@testing-library/react';
+import { useIntersection } from './use-intersection';
+
+describe('@mantine/hooks/use-intersection', () => {
+  let lastCallback: IntersectionObserverCallback;
+  const original = window.IntersectionObserver;
+
+  beforeAll(() => {
+    class MockIntersectionObserver {
+      constructor(callback: IntersectionObserverCallback) {
+        lastCallback = callback;
+      }
+
+      observe() {}
+      unobserve() {}
+      disconnect() {}
+      takeRecords() {
+        return [];
+      }
+    }
+
+    window.IntersectionObserver = MockIntersectionObserver as any;
+  });
+
+  afterAll(() => {
+    window.IntersectionObserver = original;
+  });
+
+  it('reports the latest entry when the observer delivers a batch of entries', () => {
+    const hook = renderHook(() => useIntersection());
+
+    act(() => {
+      hook.result.current.ref(document.createElement('div'));
+    });
+
+    const stale = { isIntersecting: false, intersectionRatio: 0 } as IntersectionObserverEntry;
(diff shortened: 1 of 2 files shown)


mahirhir:
Same check as on #9020: this is 244 commits behind `master`, so I looked at whether it still applies. It does, and the package already contains the correct form in a neighbouring hook.

`use-intersection.ts` on `master` today, line 27:

```ts
observer.current = new IntersectionObserver(([_entry]) => {
  setEntry(_entry);
}, options);
```

`use-in-viewport.ts`, same package, line 17:

```ts
observer.current = new IntersectionObserver((entries) => {
  const lastEntry = entries[entries.length - 1];
  setInViewport(lastEntry.isIntersecting);
});
```

Two hooks over the same browser API, one directory apart. One destructures the first element. The other takes the last, which is what the observer contract asks for.

The callback receives a list because the browser is allowed to batch. When a target crosses the threshold more than once inside a single delivery window, the entries arrive together in observation order, and only the final element describes the target's current state. Destructuring `[_entry]` reads the oldest of that batch, so `entry.isIntersecting` can report `true` for an element that already left the viewport in the same tick. Fast scrolling and programmatic scroll are where this shows up, which is also why it does not reproduce by hand.

This PR changes the one line and adds two tests against a mock observer. The first hands the callback a batch of two entries and asserts the hook exposes the second:

```ts
act(() => {
  lastCallback([stale, latest], {} as IntersectionObserver);
});
expect(hook.result.current.entry).toBe(latest);
```

Read against the current line, `([_entry]) => setEntry(_entry)` binds `_entry` to `stale`, so that assertion should not hold before the change. I have not executed it on `master` myself, so treat that as read rather than run. The second test delivers a single entry and asserts it still arrives, so the change cannot pass by always reaching past the end of a one-element list.

I am not claiming CI evidence here either:

```console
$ gh pr checks 9019 --repo mantinedev/mantine
no checks reported on the 'use-intersection-last-batched-entry' branch
```

Happy to rebase onto current `master`, or take the one-line change and close this. No attribution needed.

rtivital:
Thanks!

PR #9020, opened 2026-06-29
[@mantine/hooks] use-media-query: Catch matchMedia errors when reading the initial value synchronously

mahirhir:
#8189 fixed a Safari crash where `window.matchMedia` exists but throws when called inside a cross-origin iframe. That PR wrapped the call inside the effect in a `try/catch`, but the same call in `getInitialValue` was left unguarded:

```ts
if (typeof window !== 'undefined' && 'matchMedia' in window) {
  return window.matchMedia(query).matches;
}
```

`getInitialValue` runs during render when a consumer passes `{ getInitialValueInEffect: false }` (used to read the value synchronously and avoid a flash on first paint). The `'matchMedia' in window` check only confirms the method exists, which is exactly the case in the Safari iframe scenario, so the call still throws and crashes the render. The effect path no longer crashes after #8189, but this synchronous path does.

The fix mirrors #8189: wrap the call in `try/catch` and fall back to `false`.

Added a test that makes `matchMedia` throw and renders with `getInitialValueInEffect: false`. Without the change the render throws; with it the hook returns `false`.

### DIFF
--- a/packages/@mantine/hooks/src/use-media-query/use-media-query.test.ts
+++ b/packages/@mantine/hooks/src/use-media-query/use-media-query.test.ts
@@ -1,0 +1,43 @@
+import { renderHook } from '@testing-library/react';
+import { useMediaQuery } from './use-media-query';
+
+describe('@mantine/hooks/use-media-query', () => {
+  const original = window.matchMedia;
+
+  afterEach(() => {
+    window.matchMedia = original;
+  });
+
+  it('does not crash on render when matchMedia throws and the value is read synchronously', () => {
+    // Safari throws when matchMedia is called inside a cross-origin iframe (#8189)
+    window.matchMedia = (() => {
+      throw new Error('SecurityError');
+    }) as any;
+
+    const hook = renderHook(() =>
+      useMediaQuery('(min-width: 200px)', undefined, { getInitialValueInEffect: false })
+    );
+
+    expect(hook.result.current).toBe(false);
+  });
+
+  it('reads the initial value synchronously when matchMedia works', () => {
+    window.matchMedia = ((query: string) =>
+      ({
+        matches: true,
+        media: query,
+        onchange: null,
+        addEventListener: () => {},
+        removeEventListener: () => {},
+        addListener: () => {},
+        removeListener: () => {},
+        dispatchEvent: () => false,
+      }) as MediaQueryList) as any;
+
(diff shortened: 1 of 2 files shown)


mahirhir:
This is 244 commits behind `master`, so I checked whether it still applies before asking for a review. It does, and the file already contains the fix at the other call site.

`use-media-query.ts` on `master` today has two calls to `window.matchMedia`. One is guarded:

```ts
useEffect(() => {
  try {
    if ('matchMedia' in window) {
      const mediaQuery = window.matchMedia(query);
      ...
  } catch (e) {
    // Safari iframe compatibility issue
    return undefined;
  }
}, [query]);
```

The other, thirty lines above it, is not:

```ts
function getInitialValue(query: string, initialValue?: boolean) {
  ...
  if (typeof window !== 'undefined' && 'matchMedia' in window) {
    return window.matchMedia(query).matches;   // no try
  }
```

The guard on the first one arrived in #8189, "fix(hooks): Safari iframe matchMedia crash", merged 2025-08-22. That PR wrapped the effect and did not touch `getInitialValue`. So the project already knows this call throws, already decided what to do about it, and the decision was applied to one of the two places it happens.

The unguarded path is not dead code. It runs exactly when the caller passes:

```ts
useMediaQuery('(min-width: 200px)', undefined, { getInitialValueInEffect: false })
```

which is the option people set specifically to avoid a flash on first paint. Inside a cross-origin iframe in Safari, `matchMedia` throws, the throw happens during render rather than in an effect, and the component tree goes down. The configuration chosen by the people most careful about hydration is the one still exposed.

This PR adds the same `try` at the second call site, plus two tests:

```
does not crash on render when matchMedia throws and the value is read synchronously
reads the initial value synchronously when matchMedia works
```

The second one is there on purpose. A guard that returns `false` on every path would pass the first test perfectly, so the second asserts that a working `matchMedia` still returns `true` synchronously. Without it the first test proves only that nothing threw.

One thing I will not claim: no checks ran on this branch, so I have no CI evidence to point at.

```console
$ gh pr checks 9020 --repo mantinedev/mantine
no checks reported on the 'use-media-query-guard-initial-matchmedia' branch
```

Happy to rebase onto current `master` so a fresh run decides it, or take the five-line change directly and close this. No attribution needed.

rtivital:
Thanks!

PR #9021, opened 2026-06-29
[@mantine/core] Accordion/Collapse: Add keepMountedMode prop

### LINKED ISSUE #8934: `refs` in `keepMounted` Modals are set to null when modal closes
### Dependencies check up

9.2.2

@mantine/core

Vite

Chrome

### Describe the bug

When a `Modal` has `keepMounted`, I expect `refs` referring to elements within the Modal to keep their value. For example:

```
const ref = useRef(); // illustrative, sandbox has real code
...
<Modal keepMounted><TextInput ref={ref} /></Modal>
```

When using Mantine v9.0.0 onwards, these refs are being set to `null` when the modal closes. The element still exists in DOM.

I have debugged and found that this is a bug from `v9.0.0` and affects `v9.0.0`, `v9.1.0` and `v9.2.0`. However, `v9.0.0-alpha.0` is not affected. This happens on both development and production builds (so is unrelated to `StrictMode`).

I tried setting a callback function as the `ref=` attribute to debug its calls. Whereas in React 18 the ref callback is only called once upon initialization, in React 19 the ref callback is called on every modal open and close.

https://codesandbox.io/p/sandbox/modal-ref-reproducer-vn45sd

### Possible fix

Maybe the way `forwardRef` was removed might have something to do with it. However, I'm not very sure myself.

### Self-service

saadpocalypse:
### Problem

Under React 19, the built-in `<Activity>` component is used by default in `Collapse` to hide collapsed elements while keeping them mounted in the DOM. However, when `<Activity mode="hidden">` is active, React cleans up and suspends all side effects (`useEffect`, `useLayoutEffect`) inside the hidden subtree. 

This causes background tasks inside collapsed Accordion panels, such as polling, WebSocket connections, event listeners, or custom state-syncing effects, to be paused and deactivated. Prior to this PR, there was no configuration option to override this behavior and keep background effects active when panels are collapsed.

### Solution

Added a new `keepMountedMode` prop with values `"activity" | "display-none"` (defaulting to `"activity"`) to `Collapse`, `Accordion`, and `AccordionPanel` components.

- `keepMountedMode="activity"` (default): Hides the content using React 19's `<Activity>` component, preserving performance by pausing hidden effects.
- `keepMountedMode="display-none"`: Hides the content using standard `display: none` styling instead. This keeps the React component tree fully active in the reconciliation cycle, allowing all background effects and processes to run normally when collapsed.

### Tests

Extended the unit test suites for both components:
- **`Collapse.test.tsx`**: Added a test to assert that `keepMountedMode="display-none"` correctly applies the `display: none` style when the panel is collapsed (`expanded={false}`) instead of using the `<Activity>` component.
- **`Accordion.test.tsx`**: Added tests verifying that setting `keepMountedMode="display-none"` on the `Accordion` level and overriding it on individual `AccordionPanel` elements successfully propagates the prop down to the underlying `Collapse` component, ensuring the panel is hidden with CSS styles.

All tests compile, formatting and typechecking are clean, and the complete verification suite passes successfully.

This closes [issue 8934](https://github.com/mantinedev/mantine/issues/8934) and [discussion 9012](https://github.com/orgs/mantinedev/discussions/9012).

### DIFF
--- a/packages/@mantine/core/src/components/Accordion/Accordion.context.ts
+++ b/packages/@mantine/core/src/components/Accordion/Accordion.context.ts
@@ -17,6 +17,7 @@ export interface AccordionContextValue {
   variant: string | undefined;
   unstyled: boolean | undefined;
   keepMounted: boolean | undefined;
+  keepMountedMode: 'activity' | 'display-none' | undefined;
 }
 
 export const [AccordionProvider, useAccordionContext] = createSafeContext<AccordionContextValue>(
--- a/packages/@mantine/core/src/components/Accordion/Accordion.story.tsx
+++ b/packages/@mantine/core/src/components/Accordion/Accordion.story.tsx
@@ -1,3 +1,4 @@
+import { useEffect, useState } from 'react';
 import { TextInput } from '../TextInput';
 import { Accordion } from './Accordion';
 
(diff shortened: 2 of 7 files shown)


rtivital:
Thanks!

PR #9024, opened 2026-06-29
[@mantine/core] Button: Native tooltip for button

WojakGra:
This pull request introduces a new demo showcasing how to implement a native CSS tooltip for a button using modern CSS Anchor Positioning features. The new example is integrated into the documentation, complete with both TypeScript and CSS code, and is now available in the Button component demos. This provides users with an alternative to JavaScript-based tooltips, while also explaining browser support caveats.

**New Native CSS Tooltip Demo:**

* Added a new demo component, `Button.demo.nativeTooltip.tsx`, which demonstrates a tooltip positioned with CSS Anchor Positioning and includes both the React and CSS code needed for the example.
* Created a corresponding CSS module, `Button.demo.nativeTooltipStyles.module.css`, defining the tooltip's positioning, appearance, and show/hide behavior on hover using anchor positioning properties.

**Documentation and Integration:**

* Updated the Button documentation (`button.mdx`) to include a new section explaining the native CSS tooltip approach, its limitations, and when to use the built-in Tooltip component instead.
* Registered the new demo in the Button demos story file, making it available in the demo explorer.
* Exported the new demo from the Button demos index for use in documentation and stories.

### DIFF
--- a/apps/mantine.dev/src/pages/core/button.mdx
+++ b/apps/mantine.dev/src/pages/core/button.mdx
@@ -75,6 +75,16 @@ selectors:
 
 <Demo data={ButtonDemos.disabledStyles} />
 
+## Native CSS tooltip
+
+You can create a simple tooltip without JavaScript by using CSS Anchor Positioning. The button is used as an anchor,
+and the tooltip is positioned relative to it with `position-anchor` and `anchor()` CSS functions.
+
+Note that this approach is intended only for simple tooltips. For more advanced use cases, such as controlled state,
+delays, arrows, portals and better browser support, use the [Tooltip](/core/tooltip) component.
+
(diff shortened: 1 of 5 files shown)


rtivital:
Thanks for putting this together! Unfortunately, I cannot merge it – tooltips are already covered by the `Tooltip` component, and the native `title` attribute works for no-JS hints, so we'd rather not document a third approach on the Button page.

The CSS-only version also has limitations that users would copy: hard-coded `anchor-name` and `id` break with more than one instance on the page, the tooltip is not shown on keyboard focus and cannot be dismissed with Escape (WCAG 1.4.13), and it is positioned incorrectly in browsers without anchor positioning support. Closing for now – thanks again!

PR #9027, opened 2026-06-29
docs(hooks): update useIntersection root example

### LINKED ISSUE #9026: useIntersection: documented `root: ref.current` usage is a fragile init-order pattern
### What package has an issue

`@mantine/hooks` (v9.3.1) — `useIntersection`

### Describe the bug

The [documented usage example](https://mantine.dev/hooks/use-intersection/#usage) passes `root: containerRef.current` into the hook:

```tsx
const containerRef = useRef<HTMLDivElement>(null);
const { ref, entry } = useIntersection({
  root: containerRef.current,
  threshold: 1,
});

return (
  <div ref={containerRef} style={{ overflowY: "scroll", height: 300 }}>
    <div ref={ref}>…</div>
  </div>
);
```

`containerRef.current` is read **during render**, where it is `null` on the first render (the container hasn't committed yet), and a ref is not reactive — populating `.current` does not trigger a re-render. So the observer is initially created against the wrong root.

### Why this is fragile

Walking through the hook (`use-intersection.ts`), the returned callback `ref` is memoized via `useCallback` keyed on `[options?.rootMargin, options?.root, options?.threshold]`, and it constructs the `IntersectionObserver` with `options` when the element attaches:

1. **Render 1:** `containerRef.current` is `null` → `useIntersection({ root: null })`. Callback ref memoized with `root: null`.
2. **Commit:** the inner `ref` attaches → an observer is created with **`root: null` (the viewport)**, not the intended container.
3. `IntersectionObserver` fires once after `observe()` → `setEntry(...)` → triggers a re-render.
4. **Render 2:** now `containerRef.current` is the element, so `options.root` changed → `useCallback` returns a new ref identity → React detaches the old observer and creates a **second** observer with the correct container root.

So it only self-corrects because `setEntry` happens to force a re-render. The cost is a throwaway viewport-rooted observer, an extra render, and a brief `entry: null` flicker. More importantly, the self-heal only approximates the right behavior while the container is fully within the viewport — if the container is larger than the viewport, partially scrolled off-screen, or if the corrective re-render is suppressed, the result keys off the viewport instead of the intended container.

This is the same root cause as #9025 (`useScrollSpy` `scrollHost`): the hook wants a resolved `HTMLElement` for `root`, but the idiomatic source is a ref, and the documented pattern papers over the gap by reading `.current` during render.

### Proposed fix

Two parts, either/both:

1. **Docs:** update the usage example to drive the container via state (a callback ref into `useState`), which is deterministic and avoids the throwaway observer:

   ```tsx
   const [container, setContainer] = useState<HTMLDivElement | null>(null);
   const { ref, entry } = useIntersection({ root: container, threshold: 1 });

   return (
     <div ref={setContainer} style={{ overflowY: "scroll", height: 300 }}>
       <div ref={ref}>…</div>
     </div>
   );
   ```

2. **API (optional, mirrors #9025):** allow `root` to accept a ref object and resolve it internally:

   ```ts
   root?: Element | Document | RefObject<Element | null> | null;
   ```

   This is backwards-compatible and lets callers pass the ref they already have without the state dance.

### Related

- #9025 — same class of issue for `useScrollSpy`'s `scrollHost`.

itsayushnst13:
## Summary

Updates the `useIntersection` documentation example to avoid reading
`containerRef.current` during render.

The example now uses a callback ref with `useState`, ensuring the
`IntersectionObserver` is created with the correct root after the container
element is available.

Fixes #9026

### DIFF
--- a/packages/@docs/demos/src/demos/hooks/use-intersection/use-intersection.demo.usage.tsx
+++ b/packages/@docs/demos/src/demos/hooks/use-intersection/use-intersection.demo.usage.tsx
@@ -1,22 +1,23 @@
-import { useRef } from 'react';
+import { useState } from 'react';
 import { Box, Paper, Text } from '@mantine/core';
 import { useIntersection } from '@mantine/hooks';
 import { MantineDemo } from '@mantinex/demo';
 
 const code = `
-import { useRef } from 'react';
+import { useState } from 'react';
 import { useIntersection } from '@mantine/hooks';
 import { Text, Paper, Box } from '@mantine/core';
 
 function Demo() {
-  const containerRef = useRef<HTMLDivElement>(null);
+  const [container, setContainer] = useState<HTMLDivElement | null>(null);
+
   const { ref, entry } = useIntersection({
-    root: containerRef.current,
+    root: container,
     threshold: 1,
   });
 
   return (
-    <Paper ref={containerRef} h={300} style={{ overflowY: 'scroll' }}>
+    <Paper ref={setContainer} h={300} style={{ overflowY: 'scroll' }}>
       <Box pt={260} pb={280}>
         <Paper
           ref={ref}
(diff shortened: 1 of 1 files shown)


rtivital:

PR #9030, opened 2026-06-30
[@mantine/core] ComboboxPopover: Fix incorrect @default JSDoc for maxDropdownHeight

Hossam-Ismail:
## What

The `maxDropdownHeight` prop on `ComboboxPopover` documents `@default 250`, but the actual fallback used in the code is `220`:

```tsx
mah={maxDropdownHeight ?? 220}
```

For comparison, `TreeSelect` documents the same prop correctly as `@default 220`, matching its `?? 220` fallback.

## Changes

Updated the JSDoc from `@default 250` to `@default 220` so the generated props table on mantine.dev reflects the real default. Documentation-only change, no behavior affected.


rtivital:
Thanks!

PR #9031, opened 2026-06-30
[@mantine/hooks] use-scroll-spy: Accept a ref object for scrollHost option

### LINKED ISSUE #9025: useScrollSpy: accept a RefObject<HTMLElement> for scrollHost, not just a resolved HTMLElement
### What package has an issue

`@mantine/hooks` (v9.3.1) — `useScrollSpy`

### Describe the feature

Currently `scrollHost` only accepts a resolved `HTMLElement`:

```ts
export interface UseScrollSpyOptions {
  /** Host element to attach scroll event listener, if not provided, `window` is used */
  scrollHost?: HTMLElement;
  // ...
}
```

The idiomatic way to obtain a DOM node in React is a ref (`useRef` / callback ref). But a ref's `.current` isn't populated on first render and mutating it doesn't trigger a re-render, so passing `scrollHostRef.current` to the hook gives `undefined` on mount and the hook never re-initializes when the node attaches.

The current workaround is to convert the element into state via a callback ref, purely so the hook re-runs once the host exists:

```tsx
const [scrollHost, setScrollHost] = useState<HTMLElement | null>(null);

useScrollSpy({ scrollHost: scrollHost ?? undefined, selector });

return <form ref={setScrollHost}>…</form>;
```

This works but is awkward boilerplate, especially when you already have a ref to that element for other purposes.

### Proposed API

Let `scrollHost` accept either a resolved element **or** a ref object, with the hook resolving `.current` internally and re-initializing once it's available:

```ts
scrollHost?: HTMLElement | RefObject<HTMLElement | null>;
```

Usage would then collapse to the natural form:

```tsx
const scrollHostRef = useRef<HTMLDivElement>(null);

useScrollSpy({ scrollHost: scrollHostRef, selector });

return <div ref={scrollHostRef}>…</div>;
```

This is backwards-compatible — existing `HTMLElement` callers are unaffected — and mirrors patterns elsewhere in Mantine where hooks accept refs.

### Possible implementation

Internally, narrow the prop: if it's a ref (`'current' in scrollHost`), read `.current`; otherwise use it directly. The existing initialization `useEffect` already keys off `scrollHost`, so it would just need the resolved node in its dependency/closure.

Hossam-Ismail:
## Closes #9025

Currently `useScrollSpy`'s `scrollHost` option only accepts a resolved `HTMLElement`. The idiomatic way to obtain a DOM node in React is a ref, but a ref's `.current` is `undefined` on first render, so users have to convert the element into state via a callback ref just to make the hook re-initialize once the host exists:

```tsx
const [scrollHost, setScrollHost] = useState<HTMLElement | null>(null);
useScrollSpy({ scrollHost: scrollHost ?? undefined, selector });
return <form ref={setScrollHost}>…</form>;
```

This change lets `scrollHost` accept either a resolved element **or** a ref object, collapsing the usage to the natural form:

```tsx
const scrollHostRef = useRef<HTMLDivElement>(null);
useScrollSpy({ scrollHost: scrollHostRef, selector });
return <div ref={scrollHostRef}>…</div>;
```

Because the listener is attached inside `useEffect` (which runs after the DOM is committed), `ref.current` is already populated by the time the host is resolved, so no extra state/re-render is required. This mirrors the existing pattern of other hooks such as `use-scroll-into-view`, which already accept a `RefObject`.

## Changes

- Widen `scrollHost` type to `HTMLElement | RefObject<HTMLElement | null>`.
- Resolve `.current` inside the effect when a ref object is passed; fall back to `window` when the resolved host is `null`/`undefined` (unchanged behavior for the existing `HTMLElement` and no-arg cases).
- Update the JSDoc for the option.

## Tests

Added `use-scroll-spy.test.ts` covering: default `window` host, a resolved `HTMLElement` host, a ref-object host, and a ref whose `current` is `null` (falls back to `window`).

### DIFF
--- a/packages/@mantine/hooks/src/use-scroll-spy/use-scroll-spy.test.ts
+++ b/packages/@mantine/hooks/src/use-scroll-spy/use-scroll-spy.test.ts
@@ -1,0 +1,34 @@
+import { renderHook } from '@testing-library/react';
+import { useScrollSpy } from './use-scroll-spy';
+
+describe('@mantine/hooks/use-scroll-spy', () => {
+  it('attaches scroll listener to window by default', () => {
+    const addSpy = jest.spyOn(window, 'addEventListener');
+    renderHook(() => useScrollSpy());
+    expect(addSpy).toHaveBeenCalledWith('scroll', expect.any(Function));
+    addSpy.mockRestore();
+  });
+
+  it('attaches scroll listener to scrollHost element', () => {
+    const host = document.createElement('div');
+    const addSpy = jest.spyOn(host, 'addEventListener');
+    renderHook(() => useScrollSpy({ scrollHost: host }));
+    expect(addSpy).toHaveBeenCalledWith('scroll', expect.any(Function));
+  });
+
+  it('resolves scrollHost passed as a ref object', () => {
+    const host = document.createElement('div');
+    const addSpy = jest.spyOn(host, 'addEventListener');
+    const ref = { current: host };
+    renderHook(() => useScrollSpy({ scrollHost: ref }));
+    expect(addSpy).toHaveBeenCalledWith('scroll', expect.any(Function));
+  });
+
+  it('falls back to window when ref host is null', () => {
+    const addSpy = jest.spyOn(window, 'addEventListener');
+    const ref = { current: null };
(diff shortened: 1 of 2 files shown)


rtivital:

PR #9033, opened 2026-07-02
[@mantine/schedule] Multi-hour intervalMinutes and align time grid to…

lt-jbrand:
… whole hours

Allow whole-hour intervals >60 and floor startTime to the interval grid so slots, events and the time indicator stay clock-aligned.


rtivital:
Thanks for submitting the PR! This feature was implemented differently.

PR #9035, opened 2026-07-03
[@mantine/hooks] Fix unstable `scrollIntoView` for some bundlers

justinsmid:
See discord discussion [here](https://discord.com/channels/854810300876062770/861521000985264138/1522580881938845756)

Assigning a default function value during destructuring leads to unstable references with certain bundlers, causing `scrollIntoView` to become unstable, which may cause errors for users that rely on this being stable, for example when it is used in dependency arrays of `useEffect`.


rtivital:
Thanks!

PR #9042, opened 2026-07-07
Fix appshell breakpoint 0

### LINKED ISSUE #9039: AppShell: navbar/aside with breakpoint: 0 in static mode overlaps main content
### Dependencies check up

9.4.1

@mantine/core

React Router

All

### Describe the bug

When an AppShell is in mode="static" and a navbar (or aside) is configured with breakpoint={0}, the navbar is rendered as a fixed element painted over the main section, and AppShell.Main stretches across the full width of the screen instead of sitting next to the navbar.

The breakpoint prop is typed to accept a number, and breakpoint={0} is used in public examples, so it is a valid breakpoint value.

https://codesandbox.io/p/sandbox/mantine-react-template-forked-vx2j5g

### Possible fix

CSS variables are set with a truthiness check that will not set certain values when breakpoint is 0 and AppShell is static, as defined [here with navbar](https://github.com/mantinedev/mantine/blob/714b17e7811ce2ebdf010141376deda94efa4935/packages/%40mantine/core/src/components/AppShell/AppShellMediaStyles/assign-navbar-variables/assign-navbar-variables.ts#L29) and [here with aside](https://github.com/mantinedev/mantine/blob/714b17e7811ce2ebdf010141376deda94efa4935/packages/%40mantine/core/src/components/AppShell/AppShellMediaStyles/assign-aside-variables/assign-aside-variables.ts). Changing these to accept the valid breakpoint=0 value would likely fix the issue.

### Self-service

itsayushnst13:
Fixes #9039.

`breakpoint={0}` is a valid value, but it was ignored because `assignNavbarVariables` and `assignAsideVariables` used truthiness checks (`if (breakpoint)`) before generating CSS variables. As a result, in `mode="static"` the navbar/aside did not switch to the static layout and overlapped the main content.

This PR replaces those truthiness checks with explicit `undefined` checks so that `breakpoint={0}` is handled correctly while preserving the existing behavior for all other breakpoint values.

## Changes

* Replace truthiness checks with `breakpoint !== undefined` in:

  * `assignNavbarVariables`
  * `assignAsideVariables`
* Add regression tests for both navbar and aside to verify that `breakpoint={0}` generates the expected static layout CSS variables.

## Testing

* Added regression tests covering `breakpoint={0}` in static mode for:

  * `assign-navbar-variables.test.ts`
  * `assign-aside-variables.test.ts`
* Ran:

  * `yarn jest packages/@mantine/core/src/components/AppShell/AppShellMediaStyles/assign-navbar-variables/assign-navbar-variables.test.ts`
  * `yarn jest packages/@mantine/core/src/components/AppShell/AppShellMediaStyles/assign-aside-variables/assign-aside-variables.test.ts`

### DIFF
--- a/packages/@mantine/core/src/components/AppShell/AppShellMediaStyles/assign-aside-variables/assign-aside-variables.test.ts
+++ b/packages/@mantine/core/src/components/AppShell/AppShellMediaStyles/assign-aside-variables/assign-aside-variables.test.ts
@@ -2,7 +2,7 @@ import { DEFAULT_THEME, px, rem } from '../../../../core';
 import { AppShellProps } from '../../AppShell';
 import { assignAsideVariables } from './assign-aside-variables';
 
-function getTestObject(aside: AppShellProps['aside']) {
+function getTestObject(aside: AppShellProps['aside'], mode: 'fixed' | 'static' = 'fixed') {
   const baseStyles = {};
   const minMediaStyles = {};
   const maxMediaStyles = {};
@@ -13,7 +13,7 @@ function getTestObject(aside: AppShellProps['aside']) {
     maxMediaStyles,
     theme: DEFAULT_THEME,
     aside,
-    mode: 'fixed',
+    mode,
   });
 
   return { baseStyles, minMediaStyles, maxMediaStyles };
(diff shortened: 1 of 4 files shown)


itsayushnst13:
Added regression tests covering breakpoint={0} in static mode for both navbar and aside to prevent this behavior from regressing.

rtivital:
Thanks!

PR #9043, opened 2026-07-07
fix(dates): preserve time format on paste

### LINKED ISSUE #9041: TimePicker paste emits HH:MM:SS through the clamp path even when withSeconds is false
### Dependencies check up

9.3.0 (bug verified present in master source as well)

@mantine/dates

Next.js

All

### Describe the bug

With `withSeconds={false}` (the default), typed/spinner input calls `onChange` with `"HH:MM"` — but **pasting** a time calls `onChange` with `"HH:MM:00"`. The two input paths disagree about the value format.

**Reproduce:**

1. Render `<TimePicker onChange={console.log} />` (no `withSeconds`).
2. Type a time via the segmented inputs → `onChange` receives e.g. `"08:30"`. ✅
3. Copy the text `08:30` (or `08:30:45`) and paste it into the picker → `onChange` receives `"08:30:00"`. ❌

**Why (source trace, `use-time-picker.ts`):** `onPaste` builds the value with `getTimeString`, which correctly omits the seconds segment when `withSeconds` is false (a pasted `08:30:45` becomes `"08:30"`). But the paste handler — uniquely among the input paths — then routes the value through `clampTime` → `secondsToTime`, and `secondsToTime` unconditionally formats three segments:

```ts
// utils/time-to-seconds/time-to-seconds.ts
timeString: `${padTime(hours)}:${padTime(minutes)}:${padTime(secs)}`,
```

```ts
// use-time-picker.ts onPaste
const clamped = clampTime(timeString.value, min || '00:00:00', max || defaultMax);
onChange?.(clamped.timeString); // "08:30:00" even though withSeconds is false
```

Since `getTimeString` drops the pasted seconds first, the emitted suffix is always `:00` — so this silently widens the value format rather than producing wrong times, but it breaks any consumer that validates or stores the documented no-seconds shape (in our app, a strict `HH:MM` Zod pattern rejects the pasted value with a confusing field error).

This appears to contradict the intended contract stated in #8607, where a request to always append `:00` for consistency was declined: *"It makes more sense for me not to include seconds if `withSeconds={false}`."* The paste path is the one place that still includes them.

### Possible fix

Re-format the clamped value according to `withSeconds` before calling `onChange` — e.g. have `clampTime` take `withSeconds` (or slice the seconds segment off `clamped.timeString` in `onPaste` when `withSeconds` is false).

### Self-service

itsayushnst13:
Fixes #9041.

When `withSeconds={false}`, typing a time emits values in `HH:MM` format, but pasting a time emitted `HH:MM:00` because the paste handler forwarded the clamped value from `clampTime`, which always includes seconds.

This change preserves the expected value format by removing the seconds segment from the pasted value when `withSeconds` is disabled, while keeping the existing behavior unchanged when `withSeconds` is enabled.

## What changed

- Preserve `HH:MM` format for pasted values when `withSeconds={false}`.
- Keep existing `HH:MM:SS` behavior when `withSeconds={true}`.
- Add a regression test covering the paste behavior.

### DIFF
--- a/packages/@mantine/dates/src/components/TimePicker/TimePicker.test.tsx
+++ b/packages/@mantine/dates/src/components/TimePicker/TimePicker.test.tsx
@@ -280,6 +280,20 @@ describe('@mantine/dates/TimePicker', () => {
     expect(spy).toHaveBeenLastCalledWith('13:34:00');
   });
 
+  it('does not include seconds in pasted value when withSeconds is false', async () => {
+    const spy = jest.fn();
+    render(<TimePicker {...defaultProps} format="24h" onChange={spy} />);
+
+    await userEvent.click(screen.getByLabelText('test-hours'));
+    expect(screen.getByLabelText('test-hours')).toHaveFocus();
+
+    await userEvent.paste('13:34:56');
+
+    expect(screen.getByLabelText('test-hours')).toHaveValue('13');
+    expect(screen.getByLabelText('test-minutes')).toHaveValue('34');
+    expect(spy).toHaveBeenLastCalledWith('13:34');
+  });
+
   it('calls onChange function when the value is valid (24h format)', async () => {
     const spy = jest.fn();
     render(<TimePicker {...defaultProps} withSeconds format="24h" onChange={spy} />);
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks for submitting the PR, it is now a full fix for this issue.

PR #9049, opened 2026-07-09
[@mantine/core] Tree: Fix keyboard navigation highlight on Safari

### LINKED ISSUE #9045: Tree keyboard navigation doesn't work on Safari
### Dependencies check up

9.4.1

@mantine/core

Vite

Safari

### Describe the bug

Keyboard navigation doesn't work on Safari both on basic and virtualized one.

Steps to reproduce:
1. Open `https://mantine.dev/core/tree/` on Safari
2. Click on a sample tree root
3. Navigate using arrow down/arrow up tree

Expectation: focus box will be visible and move according to keyboard event
Reality: Nothing happen when pressing arrow up/down key. But when pressing arrow right key the focus box visible

The navigation works without problem on Chrome

_No response_

### Possible fix

_No response_

### Self-service

saadpocalypse:
### Problem

On Safari, the focus ring outline on tree nodes does not display correctly during keyboard navigation. This is because Safari does not trigger the `:focus-visible` pseudo-class for programmatically focused (`.focus()`) elements that have `tabIndex={-1}`. Safari only triggers the focus ring programmatically for elements that are part of the tab order (`tabIndex >= 0`).

Additionally, since child tree nodes are DOM descendants of parent tree nodes (residing inside nested `<ul>` groups), moving focus from a child node back up to its parent triggers a bubbling `blur` event on the parent node. This causes the parent's `onBlur` handler to trigger and prematurely clear its focus ring state.

### Solution

Implemented a target-focused custom attribute style polyfill to resolve the Safari focus ring styling quirk:
- Prior to calling `.focus()` on a node during keyboard events, we set the custom attribute `data-focus-ring="true"`.
- Updated the styles in `Tree.module.css` to render the outline ring when the node matches `:focus-visible` OR when it matches `[data-focus-ring]:focus`. Scoping the selector with `:focus` guarantees only the active focused element shows the highlight.
- Managed the `onBlur` handler to only remove `data-focus-ring` when focus actually leaves the node's bounds (`!event.currentTarget.contains(event.relatedTarget)`), resolving the child-to-parent focus bubbling issue.

### Tests

- **Unit Tests**: All 63 tests in `Tree.test.tsx` pass successfully.
- **Manual Verification**: Verified inside Safari using Storybook that the focus ring moves continuously and matches expected `:focus-visible` behavior (hidden on mouse click, visible on arrow key navigation, only one active highlighted node, handles parent node focus transitions correctly).

This resolves [issue 9045.](https://github.com/mantinedev/mantine/issues/9045)

### DIFF
--- a/packages/@mantine/core/src/components/Tree/FlatTreeNode.tsx
+++ b/packages/@mantine/core/src/components/Tree/FlatTreeNode.tsx
@@ -99,7 +99,9 @@ export const FlatTreeNode = memo(function FlatTreeNode({
           : [];
         const index = nodes.indexOf(event.currentTarget as HTMLElement);
         if (index !== -1) {
-          nodes[index + 1]?.focus();
+          const nextNode = nodes[index + 1];
+          nextNode?.setAttribute('data-focus-ring', 'true');
+          nextNode?.focus();
         }
       } else if (hasChildren) {
         tree.expand(node.value);
@@ -117,6 +119,7 @@ export const FlatTreeNode = memo(function FlatTreeNode({
         const parentElement = root?.querySelector<HTMLElement>(
           `[role=treeitem][data-value="${CSS.escape(parent)}"]`
         );
+        parentElement?.setAttribute('data-focus-ring', 'true');
         parentElement?.focus();
       }
     }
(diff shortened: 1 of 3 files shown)


jamalsa:
I tested this and it works on Safari. Good job @saadpocalypse .

rtivital:
Thanks!

PR #9050, opened 2026-07-10
[@mantine/core] Menubar: Fix menu requiring two clicks to dismiss after switching menus with hover

### LINKED ISSUE #9037: Menubar Component requires 2 mouse clicks to dismiss a menu
### Dependencies check up

9.4.1

@mantine/core

Vite

All

### Describe the bug

https://mantine.dev/core/menubar/

The Menubar component requires 2 mouse clicks to dismiss when you hover between menu items.

First, to see the desired behavior, use a Mac menu. On a Mac:
- Click `File` at the top.
- While the `File` menu is active, hover over to `Edit`.
- Now while `Edit` is active, click `Edit`. The menu will dismiss.

Now let's try the Mantine Menubar at the link above.

- Click `File`
- While the `File` menu is active, hover your mouse over to `Edit`.
- Click `Edit`. It will not dismiss.
- Click `Edit` a second time. It will dismiss.

Could the behavior be changed so that the menubar mimics a desktop/Mac menubar more, so that a single click after changing menu cover dismisses the menu?

_No response_

### Possible fix

_No response_

### Self-service

Arman-Luthra:
Fixes #9037

With `trigger="click"`, after opening a menu with a click and switching to a sibling menu by hovering its target, clicking that target did not dismiss the menu — a second click was required. Desktop menubars (e.g. macOS) dismiss on the first click.

**Root cause:** `MenubarTarget` skips closing on target click when the open source is `'hover'` — the guard exists so that with `trigger="hover"` the click that usually follows a hover-open does not immediately close the menu. But hover-switching between menus with `trigger="click"` also recorded `'hover'` as the open source, so the first click on the switched-open target fell into the reopen branch (only re-marking the source as `'click'`) and the second click closed it.

**Fix:** hover-switching with `trigger="click"` continues a click-driven interaction, so record `'click'` as the open source. The next click on the open target now closes the menu. `trigger="hover"` behavior is unchanged.

Note on the existing test: `userEvent.click` fires hover before clicking, like a real pointer, so in "keeps only one menu open at a time and switches on click" the second click now dismisses the menus (same as macOS: click File, move to Edit, click Edit — the menubar closes). Updated that test to use `fireEvent.click` so it still covers switching for clicks without a preceding hover (keyboard/assistive technology activation), and added a regression test for the hover-then-click dismissal.

### DIFF
--- a/packages/@mantine/core/src/components/Menubar/Menubar.test.tsx
+++ b/packages/@mantine/core/src/components/Menubar/Menubar.test.tsx
@@ -107,12 +107,14 @@ describe('@mantine/core/Menubar', () => {
     expectAllClosed();
   });
 
-  it('keeps only one menu open at a time and switches on click', async () => {
+  it('keeps only one menu open at a time and switches on click', () => {
     render(<TestContainer />);
-    await userEvent.click(getTarget('File'));
+    // fireEvent.click does not fire mouseover: covers clicks that are not
+    // preceded by hover (keyboard/assistive technology activation)
+    fireEvent.click(getTarget('File'));
     expectMenuOpen('New file');
 
-    await userEvent.click(getTarget('Edit'));
+    fireEvent.click(getTarget('Edit'));
     expectMenuOpen('Undo');
     expect(screen.queryByText('New file')).not.toBeInTheDocument();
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #9051, opened 2026-07-10
[@mantine/core] Input: Lighter placeholder color for error state

radui1398:
**Summary**
When an input is in an error state, the placeholder and an actual invalid value are displayed in the same red color. This makes an empty field with a placeholder look very similar to a field that contains an invalid value, which can be confusing.

https://discord.com/channels/854810300876062770/861521000985264138/1525067554891104267

**Changes**
- New css variable: --mantine-color-placeholder-error
- Placeholder color is set to var(--mantine-color-placeholder-error) when data-error is present. 

**Considerations**
As the documentation states: 
```
If it is not possible to convert the color to rgba format (for example, if the color is a CSS variable), the function will use [color-mix](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-mix). Note that color-mix is not supported in some older browsers.
``` 

**Screenshots**
image
image

### DIFF
--- a/packages/@mantine/core/src/components/Input/Input.module.css
+++ b/packages/@mantine/core/src/components/Input/Input.module.css
@@ -146,7 +146,7 @@
     }
 
     --input-color: var(--mantine-color-error);
-    --input-placeholder-color: var(--mantine-color-error);
+    --input-placeholder-color: var(--mantine-color-placeholder-error);
     --input-section-color: var(--mantine-color-error);
   }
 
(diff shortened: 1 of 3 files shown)


radui1398:
This PR also fixes a formatting error added probably here:
https://github.com/mantinedev/mantine/commit/2494b3eec8bddb3269d39dda0ea999bec1e3f98c#diff-84221a8e40eab8ee9fb467a05d99de05a4b563c668e13e70b320427f6cf78b60

rtivital:
Thanks for the PR and for the screenshots! Unfortunately, I cannot accept it – error placeholder color is a design default, and colors are being revamped in the upcoming 10.0 version, so we'd rather revisit placeholder contrast there instead of changing it in 9.x.

Note for future reference: `default-css-variables.css` is generated from `default-css-variables-resolver.ts`, so a variable added only to the CSS file would be removed on the next codegen run.

In the meantime, you can get this look by overriding `--input-placeholder-color` for inputs with `[data-error]` attribute with Styles API.

PR #9054, opened 2026-07-12
[@mantine/hooks] use-intersection: Accept a ref object for the root option

### LINKED ISSUE #9026: useIntersection: documented `root: ref.current` usage is a fragile init-order pattern

h-harsh:
## Closes #9026

`useIntersection`'s `root` option previously only accepted a resolved `Element`, so the idiomatic React source — a ref — couldn't be used directly. `ref.current` is `null` on the first render and reading it during render is fragile, which is exactly what the documented usage example did (`root: containerRef.current`).

This mirrors #9031, which added ref support to `useScrollSpy`'s `scrollHost`.

## The catch (why this isn't a one-liner)

`useScrollSpy` attaches its listener inside a `useEffect`, so `ref.current` is already populated when it runs. `useIntersection` is different: it built the `IntersectionObserver` **synchronously inside the target's callback ref**. Because React attaches child refs before their parent's, a `root` ref is still `null` at that moment — so naively resolving `root.current` there would reproduce the exact bug, just relocated.

The fix moves observer creation into a `useEffect` (runs after all refs are committed), tracking the observed element in state via the callback ref. As a bonus this removes the old throwaway viewport-rooted observer + extra render that the previous `root: ref.current` pattern caused.

## Changes

- Widen the option type via a new, exported `UseIntersectionOptions` (`root?: IntersectionObserverInit['root'] | React.RefObject<Element | null>`).
- Resolve a ref object to its element before constructing the observer; passing a resolved `Element` is unchanged.
- Add the hook's **first test file** — root resolution (element / ref / null-current / viewport default), option forwarding, entry updates, cleanup on detach/unmount, and an end-to-end render proving the ref root resolves correctly despite child-before-parent ref attachment order.
- Update the docs and usage demo to pass the ref directly (`root: containerRef`).

## Notes

- Backwards compatible — existing `Element` / no-arg callers are unaffected.
- The usage-demo update overlaps with the docs-only PR #9027; this supersedes it by demonstrating the new ref API directly.

## Verification

- `use-intersection` tests: 11 passing
- `tsc --noEmit` clean, `oxlint` clean, formatted
- `@mantine/hooks` builds (incl. `.d.ts` generation)

### DIFF
--- a/apps/mantine.dev/src/pages/hooks/use-intersection.mdx
+++ b/apps/mantine.dev/src/pages/hooks/use-intersection.mdx
@@ -25,6 +25,27 @@ useIntersection({
 });
 ```
 
+The `root` option also accepts a ref object. This is usually more convenient than
+a resolved element: the hook reads `ref.current` after the element is mounted, so
+you can pass the ref you already have without reading `.current` during render
+(which is `null` on the first render):
+
+```tsx
+import { useRef } from 'react';
+import { useIntersection } from '@mantine/hooks';
+
+function Demo() {
+  const containerRef = useRef<HTMLDivElement>(null);
+  const { ref, entry } = useIntersection({ root: containerRef, threshold: 1 });
+
+  return (
+    <div ref={containerRef} style={{ overflowY: 'scroll', height: 300 }}>
+      <div ref={ref} />
+    </div>
+  );
+}
+```
+
 The hook returns a `ref` function that should be passed to the observed element, and the latest entry, as returned by the `IntersectionObserver`'s callback.
 See the [Intersection Observer API](https://developer.mozilla.org/en-US/docs/Web/API/Intersection_Observer_API) documentation to learn everything about the options.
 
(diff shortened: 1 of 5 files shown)


rtivital:

PR #9055, opened 2026-07-12
[@mantine/core] MantineProvider: Restore dark mode light-color to match v8

### LINKED ISSUE #8895: Subtle and transparent colors do not match version 8

h-harsh:
## Closes #8895

In dark mode, `--mantine-color-{color}-light-color` resolved to shade `0` (near-white), so the `light`, `subtle`, and `transparent` variants lost their color identity and rendered near-white text/icons. In v8 it resolved to `shade[primaryShade - 5]` (a visible tint of the color) — this is the v8→v9 regression reported in the issue (e.g. an orange `subtle` ActionIcon rendered white in dark mode).

## Root cause

The shared text color was flattened to shade `0` as an unintended side effect of 2cdebb631a, whose actual goal was to make the `-light` **backgrounds** solid instead of translucent:

```diff
- [`--mantine-color-${name}-light-color`]: `var(--mantine-color-${name}-${Math.max(primaryShade - 5, 0)})`,
+ [`--mantine-color-${name}-light-color`]: `var(--mantine-color-${name}-0)`,
```

`--mantine-color-{color}-light-color` is the text/icon color shared by the `light`, `subtle`, and `transparent` variants, so this affected all three.

## Fix

Restore the v8 value (`shade[max(primaryShade - 5, 0)]`, i.e. shade 3 for the default dark `primaryShade` of 8) while keeping the solid, no-transparency backgrounds from the original commit:

- **subtle / transparent** — transparent background, so text now sits on the app background with its color restored (matches v8).
- **light** — background is `darken(shade9, 0.5)` (very dark), so shade-3 colored text still contrasts well while regaining color identity.

## Changes

- Resolve dark-mode `-light-color` to `shade[max(primaryShade - 5, 0)]` in `get-css-color-variables.ts`.
- Regenerate `default-css-variables.css` and update the resolver snapshot — the only diff is the 14 default dark `-light-color` values changing `0 → 3`.

## Note for maintainers

This changes the appearance of `light` / `subtle` / `transparent` variants in dark mode for everyone, so it's worth a design check. I went with exact v8 parity since that's what the issue asks; happy to target a different shade (e.g. aligning with `outline`'s `shade[primaryShade - 4]`) if you'd prefer.

## Verification

- MantineProvider tests: 69 passing (snapshot updated)
- `tsc --noEmit` clean, `oxlint` clean, `stylelint` clean, formatted
- `@mantine/core` builds (incl. `.d.ts` generation)


rtivital:
Thanks for the PR! Unfortunately, I cannot accept it – colors are being revamped in the upcoming 10.0 version and #8895 will be addressed as part of that work, so we'd rather not change color variables defaults before the new color system lands.

PR #9056, opened 2026-07-13
[@mantine/core] Modal/Drawer: Add keepMountedMode prop

saadpocalypse:
### Problem

[Issue #8934](https://github.com/mantinedev/mantine/issues/8934) was closed by the merge of [PR #9021](https://github.com/mantinedev/mantine/pull/9021), which introduced the `keepMountedMode` prop to Accordion and Collapse. However, the issue was closed prematurely since `Modal` and `Drawer` were not yet updated with the new prop and still had no option to bypass React 19's `<Activity>` component. 

As a result, refs referring to elements within `Modal` and `Drawer` components with `keepMounted` still get torn down and set to `null` when closed, and background side-effects inside them are paused/reset.

### Solution

Extended the `keepMountedMode` prop with values `"activity" | "display-none"` (defaulting to `"activity"`) to `ModalBase` and propagated it to `ModalRoot`, `Modal`, `DrawerRoot`, and `Drawer`.

* `keepMountedMode="activity"` (default): Hides the content using React 19's `<Activity>` component, preserving performance by pausing hidden effects.
* `keepMountedMode="display-none"`: Hides the content using standard CSS `display: none` styling instead. This keeps the React component tree fully active in the reconciliation cycle, preventing refs from being set to `null` and allowing all background effects and processes to run normally when closed.

### Tests

* Added unit tests to `Modal.test.tsx` and `Drawer.test.tsx` asserting that setting `keepMounted={true}` and `keepMountedMode="display-none"` correctly applies the CSS `display: none` style on the content wrapper when closed.
* Added comparative storybook demos to `Modal.story.tsx` and `Drawer.story.tsx` featuring active timers and ref state counters to easily verify both modes visually.
* All tests compile, formatting and typechecking are clean, and the complete verification suite passes successfully.

Refers to [issue #8934](https://github.com/mantinedev/mantine/issues/8934).

### DIFF
--- a/packages/@mantine/core/src/components/Drawer/Drawer.story.tsx
+++ b/packages/@mantine/core/src/components/Drawer/Drawer.story.tsx
@@ -1,8 +1,10 @@
+import { useEffect, useRef, useState } from 'react';
 import { useDisclosure } from '@mantine/hooks';
 import { Button } from '../Button';
 import { useModalsStack } from '../Modal';
 import { ScrollArea } from '../ScrollArea';
 import { Tabs } from '../Tabs';
+import { Stack as MantineStack } from '../Stack';
 import { Drawer } from './Drawer';
 
 export default { title: 'Drawer' };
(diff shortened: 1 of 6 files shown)


rtivital:
The test renderer always sets env="test" (packages/@mantine-tests/core/src/render.tsx:11), and Transition has an early-return for that env (Transition.tsx:88-90) that applies display: none for any keepMounted, ignoring keepMountedMode entirely. I ran the PR's exact test locally against the current codebase without the ModalBase change — both it and a variant with only keepMounted (no mode at all) pass. So the tests assert pre-existing test-env behavior, not the new prop. The (transitionDuration is 0) note in the test name is also misleading — duration never comes into play in test env.

saadpocalypse:
Sorry about that, completely missed it. Pushed a fix.

rtivital:
Thanks!

PR #9058, opened 2026-07-14
[@mantine/core] PinInput: Fix incorrect @default JSDoc for gap prop

Hossam-Ismail:
The `gap` prop JSDoc documents `@default 'md'`, but the component's `defaultProps` sets `gap: 'sm'`.

```ts
// packages/@mantine/core/src/components/PinInput/PinInput.tsx
const defaultProps = {
  gap: 'sm',
  // ...
} satisfies Partial<PinInputProps>;
```

The `@default` annotation feeds the auto-generated props table on mantine.dev, so this corrects the documented default to match the actual value (`'sm'`). Docs-only change, no runtime behavior affected.


rtivital:
Thanks!

PR #9059, opened 2026-07-14
[@mantine/charts] PieChart: Fix incorrect @default JSDoc for withTooltip prop

Hossam-Ismail:
The `withTooltip` prop JSDoc documents `@default true`, but the component's `defaultProps` sets `withTooltip: false`.

```ts
// packages/@mantine/charts/src/PieChart/PieChart.tsx
const defaultProps = {
  withTooltip: false,
  // ...
} satisfies Partial<PieChartProps>;
```

The `@default` annotation feeds the auto-generated props table on mantine.dev, so this corrects the documented default to match the actual value (`false`). Docs-only change, no runtime behavior affected.


rtivital:
Thanks!

PR #9061, opened 2026-07-14
[@mantine/core] ColorInput: Add support for fullWidth prop

### LINKED ISSUE #9060: ColorInput allows full width ColorPicker
### Dependencies check up

9.4.1

@mantine/core

Vite

### Describe the bug

The `ColorPicker` component is able to be rendered with the full width using the prop [fullWidth](https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/components/ColorPicker/ColorPicker.tsx#L91).

Unfortunately there is no way to set/pass the `fullWidth` prop to the `ColorPicker` from the `ColorInput`. ([Source](https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/components/ColorInput/ColorInput.tsx#L294-L310)).

It would be great if this would be possible.

_No response_

### Possible fix

_No response_

### Self-service

saadpocalypse:
### Problem

The `ColorPicker` component supports taking 100% width of its container using the `fullWidth` prop. However, when rendering `ColorPicker` within a `ColorInput` dropdown, there was no way to pass the `fullWidth` prop from `ColorInput` down to the picker. Additionally, because the popover dropdown has `width: 'max-content'` by default, passing a 100% width constraint directly to the child color picker caused the layout to collapse.

### Solution

*   Moved the `fullWidth` prop from `ColorPickerProps` to `__ColorPickerProps` so that it is properly inherited by `ColorInputProps`.
*   Forwarded `fullWidth` from `ColorInput` to the internal `<ColorPicker />` component.
*   Passed `width={fullWidth ? 'target' : undefined}` to the `<Popover />` wrapper inside `ColorInput` so that the dropdown matches the target input's width and prevents the picker layout from collapsing.

### Tests

*   **Unit Tests**: Added a test case in `ColorInput.test.tsx` to verify that passing `fullWidth` to `ColorInput` correctly applies the `data-full-width` attribute to the underlying picker wrapper inside the dropdown.
*   **Visual Stories**: Added a `FullWidth` story in `ColorInput.story.tsx` with `popoverProps={{ opened: true }}` to demonstrate the color picker expanding to the full width of the input dropdown.

Resolves [#9060](https://github.com/mantinedev/mantine/issues/9060)

### DIFF
--- a/packages/@mantine/core/src/components/ColorInput/ColorInput.story.tsx
+++ b/packages/@mantine/core/src/components/ColorInput/ColorInput.story.tsx
@@ -21,3 +21,11 @@ export function Unstyled() {
     </div>
   );
 }
+
+export function FullWidth() {
+  return (
+    <div style={{ padding: 40 }}>
+      <ColorInput label="Full width color picker" fullWidth popoverProps={{ opened: true }} />
+    </div>
+  );
+}
--- a/packages/@mantine/core/src/components/ColorInput/ColorInput.test.tsx
+++ b/packages/@mantine/core/src/components/ColorInput/ColorInput.test.tsx
@@ -1,4 +1,6 @@
+import { render } from '@testing-library/react';
 import { inputDefaultProps, inputStylesApiSelectors, tests } from '@mantine-tests/core';
+import { MantineProvider } from '../../core';
 import { __InputStylesNames } from '../Input';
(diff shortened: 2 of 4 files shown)


rtivital:
Thanks!

PR #9063, opened 2026-07-15
[@mantine/core] Marquee: Fix duration default in JSDoc

dfedoryshchev:
The `duration` prop is documented as `@default 40000`, but the component applies `duration: 100_000` from its `defaultProps` (via `useProps`), so the effective default is `100000`ms. Updated the JSDoc `@default` to match `defaultProps`.


dfedoryshchev:
Hi, any comments on this one? It is a one-line JSDoc correction and `test_pull_request` is green. Happy to rebase if it has drifted.

rtivital:
Thanks!

PR #9066, opened 2026-07-16
feat: support RefObject for useScrollSpy scrollHost

deshna1810:
## Summary

This PR adds support for passing a React `RefObject` as the `scrollHost` option to `useScrollSpy`, in addition to the existing `HTMLElement` support.

## Changes

- Extended the `scrollHost` type to accept `HTMLElement | RefObject<HTMLElement | null>`.
- Resolved the underlying DOM element internally when a ref is provided.
- Preserved the existing behavior for `HTMLElement` inputs, ensuring full backward compatibility.
- Updated the hook implementation to work seamlessly with React refs.
- Added/updated tests to cover both `HTMLElement` and `RefObject` usage.

## Motivation

Previously, `scrollHost` only accepted a resolved `HTMLElement`, which made using React refs awkward because `ref.current` is `null` during the initial render and doesn't trigger a re-render when populated. Supporting `RefObject` provides a more idiomatic React API and removes the need for callback ref or state-based workarounds while maintaining existing functionality.

### DIFF
--- a/packages/@mantine/hooks/src/use-scroll-spy/use-scroll-spy.ts
+++ b/packages/@mantine/hooks/src/use-scroll-spy/use-scroll-spy.ts
@@ -1,4 +1,4 @@
-import { useEffect, useEffectEvent, useRef, useState } from 'react';
+import { RefObject, useEffect, useEffectEvent, useRef, useState } from 'react';
 import { randomId } from '../utils';
 
 function getHeadingsData(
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks for the PR! Support for passing a `RefObject` as `scrollHost` to `useScrollSpy` has already been added in 9.5.1 (see #9025) with a nearly identical implementation. Since this change is already on master and the PR now conflicts with it, I'm closing it. Thanks again!

PR #9068, opened 2026-07-16
[@mantine/form] Fix async validation with debounce on initial keystroke of empty input

### LINKED ISSUE #9067: Async validation with debounce not working on initial keystroke of an empty input
### Dependencies check up

9.2.1

@mantine/form

Vite

All

### Describe the bug

I had some custom debounce code, and I'm trying to simplify the logic by using the new async/debounce validation in Mantine 9.

However, I have noticed that when I start typing on an empty input, the debounce seems to not be working. I don't get a request on every keystroke, but maybe once every 2 keystrokes.

Once the input has some text, if I delete some characters, or add more, the debounce works properly and only triggers once I stop typing. If I delete all characters in the input, and then start typing again, I get the issue again.

https://codesandbox.io/p/sandbox/mantine-react-template-forked-y6n8gv

### Possible fix

_No response_

### Self-service

saadpocalypse:
### Problem
When using `validateDebounce` with async validation on change (`validateInputOnChange: true`), the debounce mechanism fails to cancel/clear the initial validation timer on subsequent rapid keystrokes if typing started from an empty input.
This occurs because:
* On the first keystroke of an empty input, the field transitions from clean to dirty.
* This transition triggers a state update (`setDirtyState`) that causes the component to re-render.
* During the re-render, if the `validate` rules are defined as an inline object (common practice), the `useMemo` for `debouncedValidateField` in `use-form.ts` re-runs because its dependencies changed.
* Re-running the `useMemo` discards the local `timers` object and creates a new one, leaving the first timer ($T_1$) orphaned in the old object.
* On subsequent rapid keystrokes (which do not trigger re-renders since the field is already dirty), the new `debouncedValidateField` attempts to clear the previous timer but looks in the new, empty `timers` object, failing to clear $T_1$.
* Consequently, both the initial keystroke's timer and the final keystroke's timer execute, resulting in duplicate/unwanted API requests.

### Solution
Refactored the validation timer management in `packages/@mantine/form/src/use-form.ts`:
* Moved the `timers` dictionary from a local variable inside the `useMemo` callback of `debouncedValidateField` into a shared React `useRef` (`timers.current`) at the `useForm` hook level.
* Storing timers in a ref ensures that the timer references survive any re-renders or `useMemo` updates, allowing subsequent keystrokes to successfully locate and clear previous timers.
* Added a `useEffect` cleanup hook to clear all active validation timers on unmount.
* Added a new story `ValidateDebounce` to `packages/@mantine/form/src/stories/Form.validateOnChange.story.tsx` to showcase and easily test debounced change validations.

### Tests
* **Automated Tests**: Added a new test suite `packages/@mantine/form/src/tests/use-form/validateDebounce.test.ts` to verify that rapid value changes within the debounce window correctly clear preceding timers. Verified that all `@mantine/form` tests pass successfully (`npm run jest packages/@mantine/form`).
* **Build & Typecheck**: Successfully verified that the codebase typechecks (`npm run typecheck`) and builds without errors (`npm run build`).
* **Manual Verification**: Verified visually using the new `ValidateDebounce` story in Storybook.

Resolves [#9067](https://github.com/mantinedev/mantine/issues/9067)

### DIFF
--- a/packages/@mantine/form/src/stories/Form.validateOnChange.story.tsx
+++ b/packages/@mantine/form/src/stories/Form.validateOnChange.story.tsx
@@ -81,3 +81,39 @@ export function ValidateOnChangeUncontrolled() {
     </FormBase>
   );
 }
+
+import { useState } from 'react';
+import { Button } from '@mantine/core';
+
+export function ValidateDebounce() {
+  const [requestCount, setRequestCount] = useState(0);
+
+  const form = useForm({
+    mode: 'uncontrolled',
+    validateInputOnChange: true,
+    validateDebounce: 500,
+    initialValues: {
+      name: '',
+    },
+    validate: {
+      name: async (value) => {
+        if (value.trim().length === 0) {
+          return 'Name is required';
+        }
+        setRequestCount((c) => c + 1);
+        await new Promise((resolve) => setTimeout(resolve, 300));
+        return value.length < 5 ? 'Must be at least 5 characters' : null;
+      },
+    },
+  });
+
+  return (
+    <FormBase form={form}>
+      <div>Request Count: {requestCount}</div>
+      <TextInput label="Name" {...form.getInputProps('name')} />
+      <Button onClick={() => setRequestCount(0)} mt="xs">
+        Reset Count
+      </Button>
+    </FormBase>
+  );
+}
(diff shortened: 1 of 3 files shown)


rtivital:
Thanks!

PR #9079, opened 2026-07-24
[@mantine/hooks] Fix useEffectEvent-based hooks not updating inside forwardRef and memo components

### LINKED ISSUE #9078: useHotkeys (and other useEffectEvent-based hooks) silently stop updating inside forwardRef / memo components
### What happens

Since v9, `useHotkeys` subscribes to `keydown` once on mount and keeps its handler current via React's `useEffectEvent`. Inside a `forwardRef` component — or a `memo(fn)` component — that handler is **pinned to the first render forever**. It keeps firing, but always with the values it closed over on mount.

Nothing warns, and the hotkey still "works", so this reads as a logic bug in application code rather than a stale closure.

This is a v8 → v9 regression. v8 depended on `hotkeys`:

```ts
// 8.3.0 — new array literal every render, so the effect re-ran and
// re-registered a fresh closure. Immune regardless of component type.
useEffect(() => {
  const keydownListener = (event) => { hotkeys.forEach(...) };
  document.documentElement.addEventListener('keydown', keydownListener);
  return () => document.documentElement.removeEventListener('keydown', keydownListener);
}, [hotkeys]);
```

```ts
// 9.4.1 — registers once, delegates freshness to useEffectEvent
const handleKeydown = useEffectEvent((event) => { hotkeys.forEach(...) });
useEffect(() => {
  document.documentElement.addEventListener('keydown', handleKeydown);
  return () => document.documentElement.removeEventListener('keydown', handleKeydown);
}, []);
```

### Minimal reproduction

```tsx
import { forwardRef, useCallback, useState } from 'react';
import { useHotkeys } from '@mantine/hooks';

function Plain({ value }: { value: number }) {
  useHotkeys([['ArrowDown', useCallback(() => console.log('plain      sees', value), [value])]]);
  return null;
}

const Forwarded = forwardRef<HTMLDivElement, { value: number }>(({ value }, ref) => {
  useHotkeys([['ArrowDown', useCallback(() => console.log('forwardRef sees', value), [value])]]);
  return <div ref={ref} />;
});

  const [value, setValue] = useState(0);
  return (
    <>
      <button onClick={() => setValue((v) => v + 1)}>increment (value = {value})</button>
      <Plain value={value} />
      <Forwarded value={value} />
    </>
  );
}
```

Click increment a few times, then press ArrowDown:

```
plain      sees 3
forwardRef sees 0     ← never changes
```

### Why

React applies `useEffectEvent` impl updates in exactly one place, and that site only handles fiber tag `0` (`FunctionComponent`). Tags `11` (`ForwardRef`) and `15` (`SimpleMemoComponent`) fall straight through, so `ref.impl` is never reassigned for those components:

```js
switch (root.tag) {
  case 0:
    if ((flags & 4) && root.updateQueue?.events) {
      for (...) payload.ref.impl = payload.nextImpl;   // only here
    }
    break;
  case 11:   // ForwardRef
  case 15:   // SimpleMemoComponent
    break;   // ← never applied
  ...
}
```

Verified behaviour, which lines up with that switch exactly:

| Wrapper | Fiber tag | Handler |
| --- | --- | --- |
| plain function component | 0 | stays fresh |
| `memo(fn)` | 15 SimpleMemoComponent | **frozen at mount** |
| `memo(fn, compare)` | 14 MemoComponent → inner tag 0 | stays fresh |
| `forwardRef(fn)` | 11 ForwardRef | **frozen at mount** |

`memo(fn, compare)` escapes only because the hooks live on a separate inner `FunctionComponent` fiber.

### Impact beyond useHotkeys

Every hook in `@mantine/hooks@9.4.1` that pairs `useEffectEvent` with a subscribe-once effect has the same exposure inside `forwardRef` / `memo(fn)`:

`use-hotkeys`, `use-window-event`, `use-click-outside`, `use-page-leave`, `use-headroom`, `use-scroll-direction`, `use-scroll-spy`, `use-text-selection`, `use-collapse`

`useClickOutside` looks especially likely to bite, since outside-click handlers are commonly written in `forwardRef` wrappers and close over props like `opened` or `onClose`.

### How this surfaced

A virtualised list component (a `forwardRef`) used `useHotkeys` for ArrowUp/ArrowDown row navigation. It mounts with an empty `items` array because the data has not been fetched yet, so the pinned handler could only ever compute `[].indexOf(activeId) === -1` and bail. Arrow keys were dead for the lifetime of the page, and only a remount *after* the data arrived brought them back. Everything looked correct in the fiber — props were current, the `useCallback` cells held current deps — because the only stale thing was `ref.impl` inside the effect event.

### Suggested fix

Either restore a dependency so the listener re-registers when the hotkeys change, or hold the array in a ref that each commit refreshes rather than relying on `useEffectEvent`. A ref keeps the single-subscription behaviour that v9 was presumably after:

```ts
export function useHotkeys(hotkeys, tagsToIgnore = ['INPUT', 'TEXTAREA', 'SELECT'], triggerOnContentEditable = false) {
  const latest = useRef(hotkeys);
  useEffect(() => { latest.current = hotkeys; });

  useEffect(() => {
    const handleKeydown = (event) => { latest.current.forEach(...) };
    document.documentElement.addEventListener('keydown', handleKeydown);
    return () => document.documentElement.removeEventListener('keydown', handleKeydown);
  }, []);
}
```

Whether React should also be applying effect-event updates for tags 11/15 is arguably a separate upstream question, but the v8 → v9 change is what turned a working hook into a silently broken one for consumers.

### Environment

- `@mantine/hooks` 9.4.1
- react / react-dom 19.2.3
- Reproduced in both a browser (production build) and jsdom; independent of React Compiler (reproduces with the compiler entirely absent).

dudusotero:
Fixes #9078

## Problem

React only applies `useEffectEvent` implementation updates for plain function component fibers (tag 0). Inside `forwardRef` (tag 11) and `memo(fn)` (tag 15) the update is never applied, so any hook that pairs `useEffectEvent` with a subscribe-once effect keeps calling the closure from the first render. The listener still fires, but it always sees the initial props and state.

Reproduced with react-dom 19.2.7 in the browser and in jsdom. The regression tests in this PR fail on master.

Affected hooks: `use-hotkeys`, `use-window-event`, `use-click-outside`, `use-page-leave`, `use-headroom`, `use-scroll-direction`, `use-scroll-spy`, `use-text-selection`, `use-collapse`, `use-horizontal-collapse`.

## Solution

Replaced React's `useEffectEvent` with the existing `useCallbackRef` util in these hooks. It gives the same stable identity and single subscription, but keeps the callback fresh through a ref that is updated on every commit, and that works for all fiber types.

There is one supporting change: `useCallbackRef` now updates its ref in `useIsomorphicEffect` instead of `useEffect`. This matches the commit timing of `useEffectEvent` and is needed for `use-headroom`, which calls the wrapped callbacks inside layout effects. With a passive update those callbacks would be one commit behind.

## Tests

- `useHotkeys`: latest handler is called after rerender inside `forwardRef` and inside `memo` components, both fail on master
- `useClickOutside`: latest handler is called after rerender inside a `forwardRef` component, fails on master

`npm test` passes.

## Note

`useEffectEvent` from react is also imported in `@mantine/core` (ScrollArea), `@mantine/notifications` and `@mantine/schedule`, and those components are created with `forwardRef`, so the same staleness can happen there. I can send the same change for those packages in a separate PR if you prefer to keep this one scoped to `@mantine/hooks`.

### DIFF
--- a/packages/@mantine/hooks/src/use-click-outside/use-click-outside.test.tsx
+++ b/packages/@mantine/hooks/src/use-click-outside/use-click-outside.test.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { forwardRef, useState } from 'react';
 import { render, screen } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { useClickOutside } from './use-click-outside';
@@ -239,5 +239,40 @@ describe('@mantine/hooks/use-click-outside', () => {
     const event = handler.mock.calls[0][0];
     expect(event).toHaveProperty('type', 'mousedown');
     expect(event).toHaveProperty('target', outsideTarget);
+  });
+
+  it('calls the latest handler after rerender inside a forwardRef component', async () => {
+    const ForwardedTarget = forwardRef<HTMLButtonElement, UseClickOutsideProps>(
+      ({ handler }, buttonRef) => {
+        const ref = useClickOutside<HTMLDivElement>(handler);
+        return (
+          <div data-testid="target" ref={ref}>
+            <button type="button" ref={buttonRef} />
+          </div>
+        );
+      }
+    );
+
+    const first = jest.fn();
+    const second = jest.fn();
+
+    const { rerender } = render(
+      <>
+        <ForwardedTarget handler={first} />
+        <div data-testid="outside-target" />
+      </>
+    );
+
+    rerender(
+      <>
+        <ForwardedTarget handler={second} />
+        <div data-testid="outside-target" />
+      </>
+    );
+
+    await userEvent.click(screen.getByTestId('outside-target'));
+
+    expect(first).not.toHaveBeenCalled();
+    expect(second).toHaveBeenCalledTimes(1);
   });
 });
--- a/packages/@mantine/hooks/src/use-click-outside/use-click-outside.ts
+++ b/packages/@mantine/hooks/src/use-click-outside/use-click-outside.ts
@@ -1,4 +1,5 @@
-import { useEffect, useEffectEvent, useRef } from 'react';
+import { useEffect, useRef } from 'react';
+import { useCallbackRef } from '../utils';
 
 type EventType = MouseEvent | TouchEvent;
 
@@ -13,7 +14,7 @@ export function useClickOutside<T extends HTMLElement = any>(
   const ref = useRef<T>(null);
   const eventsList = events || DEFAULT_EVENTS;
 
-  const listener = useEffectEvent((event: Event) => {
+  const listener = useCallbackRef((event: Event) => {
     const { target } = event ?? {};
     const shouldIgnore =
       !document.body.contains(target as Node) && (target as Element)?.tagName !== 'HTML';
--- a/packages/@mantine/hooks/src/use-collapse/use-collapse.ts
+++ b/packages/@mantine/hooks/src/use-collapse/use-collapse.ts
@@ -1,7 +1,8 @@
-import React, { CSSProperties, useEffectEvent, useRef, useState } from 'react';
+import React, { CSSProperties, useRef, useState } from 'react';
 import { flushSync } from 'react-dom';
 import { useDidUpdate } from '../use-did-update/use-did-update';
 import { mergeRefs } from '../use-merged-ref/use-merged-ref';
+import { useCallbackRef } from '../utils';
 
 function getAutoHeightDuration(height: number | string) {
   if (!height || typeof height === 'string') {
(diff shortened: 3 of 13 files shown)


codeCraft-Ritik (commented):
The reproduction tests in useHotkeys and useClickOutside wrapping both forwardRef (tag 11) and memo (tag 15) are spot on.

Did we also add a quick assertion verifying the layout-effect timing for use-headroom (or any consumer reading the callback inside useLayoutEffect) to guarantee that useCallbackRef updates synchronously before passive effects fire?

rtivital:
Thanks for the thorough investigation and the regression tests! The stale callback behavior comes from a bug in React's `useEffectEvent` for `forwardRef`/`memo` components (https://github.com/react/react/issues/34818). It was fixed upstream in https://github.com/react/react/pull/34831 and released in React 19.3. We'd rather keep using `useEffectEvent` than work around a bug that React has already fixed – the recommended fix is to update React to 19.3 (see #9078). The branch also conflicts with master now, so I'm closing this PR.

PR #9087, opened 2026-07-27
[@mantine/core] FloatingWindow: Fix default z-index

DanielDostal1:
Fixes #8813

## Problem

`FloatingWindow` defaults to the overlay z-index tier (`400`), while portalled dropdowns and tooltips use the popover tier (`300`). Since the default portals share a body-level target, the window and its portalled overlays become sibling stacking contexts and the window covers the overlays.

This affects `Select` and every other Popover- or Tooltip-based surface rendered from a `FloatingWindow`.

## Solution

- change the default `FloatingWindow` z-index from the overlay tier (`400`) to the modal tier (`200`)
- preserve explicit `zIndex` overrides
- add real-portal regression coverage for `Select`, `Tooltip`, and `Tooltip.Floating`
- add a Storybook reproduction with an opened `Select` dropdown

This fixes the layer assignment at the container instead of coupling individual overlay components to `FloatingWindow`. It follows the existing hierarchy where application containers use the modal tier and transient floating surfaces use the popover tier.

The issue is labeled Deferred because changing the documented default is observable. This PR is intended for consideration with the next major release. Consumers that need the previous layer can continue to set `zIndex={400}` explicitly.

## Testing

- `npm run setup`
- `npm run build`
- `npm test` (647 suites, 15,788 tests, 1 snapshot)

### DIFF
--- a/packages/@mantine/core/src/components/FloatingWindow/FloatingWindow.story.tsx
+++ b/packages/@mantine/core/src/components/FloatingWindow/FloatingWindow.story.tsx
@@ -1,3 +1,4 @@
+import { Select } from '../Select';
 import { FloatingWindow } from './FloatingWindow';
 
 export default { title: 'FloatingWindow' };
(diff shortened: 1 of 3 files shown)


rtivital:
Thanks for the PR! Changing the default `zIndex` of FloatingWindow changes the layering of existing applications (for example, FloatingWindow would no longer be displayed above an open Modal), so it is not planned to be changed in 9.x (#8813 is deferred to the next major version).

Until then, you can set `zIndex` on FloatingWindow or pass `comboboxProps={{ zIndex }}` to Select and other combobox-based components to display dropdowns above the window.

PR #9091, opened 2026-07-29
fix(PasswordInput): allow keyboard navigation for visibility toggle

### LINKED ISSUE #9090: Can't reach the visibility toggle of PasswordInput with keyboard navigation
### Dependencies check up

9.5.0

@mantine/core

Tanstack Start

All

### Describe the bug

I'm simply using the doc site for this
https://mantine.dev/core/password-input

I expect to be able to navigate to the toggle visibility button with `tab` `shift + tab`, but it skips over it.

I can avoid this for right now with the following prop `visibilityToggleButtonProps={{ tabIndex: 0 }}`

https://mantine.dev/core/password-input

### Possible fix

[Remove this line](https://github.com/mantinedev/mantine/blob/953192e2207722a43568468d0db8d20eceb21307/packages/%40mantine/core/src/components/PasswordInput/PasswordInput.tsx#L168).

However, given that not having a `tabIndex` applied would have made this keyboard accessible, the fact that it's written here suggests that it was added to avoid something else. I don't know what that would be though.

### Self-service

Animediscussion:
Fixes #9090

Removed tabIndex={-1} from the PasswordInput visibility toggle button so it can be reached with keyboard navigation.

Added a regression test to verify the toggle button can be focused and activated with keyboard input.

### DIFF
--- a/packages/@mantine/core/src/components/PasswordInput/PasswordInput.test.tsx
+++ b/packages/@mantine/core/src/components/PasswordInput/PasswordInput.test.tsx
@@ -122,6 +122,22 @@ describe('@mantine/core/PasswordInput', () => {
     expect(input.type).toBe('password');
   });
 
+    it('allows visibility toggle button to be reached with keyboard navigation', async () => {
+    const user = userEvent.setup();
+    render(<PasswordInput label="Password" />);
+
+      await user.tab();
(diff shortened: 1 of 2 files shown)


rtivital:

PR #9093, opened 2026-07-31
Add mantine-map into 3rd party extensions

sarioglu:
I recently created a map component based on my needs. This PR adds it into 3rd party extensions list so that more people can benefit from it


rtivital:
Thanks!

PR #9101, opened 2026-08-07
@mantine/core] Portal: cache the shared portal node lookup (O(portals × DOM nodes) → O(1))

vintagemachine:
```markdown
With the default `reuseTargetNode: true`, `Portal`'s `getTargetNode` runs
`document.querySelector('[data-mantine-shared-portal-node]')` on **every
portal mount**, uncached. The shared node is appended as the **last child of
`document.body`**, so each lookup traverses essentially the whole document
before finding it — mounting N portals into a D-node document is **O(N × D)**.

This bites hard on data-dense pages because `Tooltip`, `Menu`/`PopoverDropdown`
and `ModalBase` mount their `OptionalPortal` **while closed**: thousands of
portal mounts accumulate without the user ever opening anything.

**Real-world case** (production React app, Chrome, M-series MacBook): a page
mounting ~40k closed tooltips/menus/modals into a 616k-node document froze for
168 s — **145.8 s (86.8 % of total CPU) was this single `querySelector`**
(V8 profile attribution: `getTargetNode` ← Portal mount effect ←
`commitHookEffectListMount`). With this fix the same page loads in ~21 s.

**Fix**: cache the shared node in a module-level variable, guarded by
`document.body.contains(...)` so a removed node (tests, jsdom teardown,
`document.body` replacement) falls back to the original lookup path.
Semantics are unchanged — the node was already shared and reused; it is now
found once instead of once per mount.

**Isolated benchmark** (jsdom, bare `<Portal>`s, stock vs patched, same
process — repro linked below):

| document nodes | portals | stock mount | patched mount | querySelector calls |
|---:|---:|---:|---:|---:|
| 10,000 | 500 | 2,536 ms | 37 ms | 1,000 → 1 |
| 50,000 | 500 | 31,499 ms | 90 ms | 1,000 → 1 |
| 50,000 | 2,000 | 120,910 ms | 137 ms | 4,000 → 1 |
| 150,000 | 2,000 | 389,543 ms | 296 ms | 4,000 → 1 |

Repro (run `npm install && npm run repro`): https://gist.github.com/vintagemachine/06247a6c6c2db22ef38145cbce64163c

Present from at least 8.x through current master.
```

## D. Gist contents

Four files from this folder (NOT `node_modules`, NOT `package-lock.json`,
NOT this file):

- `repro.mjs`
- `README.md`
- `package.json`
- `mantine-core-portal.patch` (referenced by the README's Files section)


MILLERMARRU (commented):
I checked Portal.tsx on master and `getTargetNode` really does run a full `document.querySelector` on every mount when `reuseTargetNode` is on, which is the default, so an app rendering many tooltips/modals/popovers pays a DOM query on every single one just to find the same shared node. The cache invalidation via `document.body.contains(cachedSharedPortalNode)` before trusting the cache is the right safety net, without it, a cached reference to a node removed by unrelated code would silently return a detached node. One thing I'd flag for consideration: the cache is a bare module-level variable, so in environments with multiple concurrent React roots on the same page (micro-frontends, or request-scoped rendering reused across requests in some non-browser test environments) a cache invalidated by one root's node removal could produce query churn for another. That's a pretty narrow scenario given `document.body.contains` already handles the common case correctly, and I don't think it should block the fix.

vintagemachine:
> I checked Portal.tsx on master and `getTargetNode` really does run a full `document.querySelector` on every mount when `reuseTargetNode` is on, which is the default, so an app rendering many tooltips/modals/popovers pays a DOM query on every single one just to find the same shared node. The cache invalidation via `document.body.contains(cachedSharedPortalNode)` before trusting the cache is the right safety net, without it, a cached reference to a node removed by unrelated code would silently return a detached node. One thing I'd flag for consideration: the cache is a bare module-level variable, so in environments with multiple concurrent React roots on the same page (micro-frontends, or request-scoped rendering reused across requests in some non-browser test environments) a cache invalidated by one root's node removal could produce query churn for another. That's a pretty narrow scenario given `document.body.contains` already handles the common case correctly, and I don't think it should block the fix.

Thanks for the review! On the multi-root concern: I think it's even narrower than described: invalidation can never produce more query work than the current code, since stock already runs `querySelector` on every mount.

Worst case the cache degrades to exactly stock behavior plus one `document.body.contains()` check.

Separate `@mantine/core` copies (micro-frontend bundles) each cache independently but converge on the same shared node via the attribute selector, so they don't conflict.

And the non-browser/request-scoped case can't occur, since `getTargetNode` only runs inside `useIsomorphicEffect`, which never executes during SSR.

rtivital:
Thanks!

PR #9109, opened 2026-08-11
[mantine.dev] Fix prevent docs navbar focus outline clipping

### LINKED ISSUE #9108: Focus outline is clipped in the desktop documentation navbar

https://mantine.dev/getting-started/

### Description

When navigating the documentation navbar with the `Tab` key on desktop, the inline-start portion of the focused link’s outline is clipped.

This happens because the desktop navigation links are positioned directly against the edge of a `ScrollArea` viewport, which clips overflowing content.

The issue does not occur in the mobile and tablet navbar because its content has additional inline padding inside the scroll viewport.

### Steps to reproduce

1. Open https://mantine.dev/getting-started/ at a desktop viewport width.
2. Press the `Tab` key until a link in the left navigation receives focus.
3. Observe the inline-start edge of the focus outline.

### Expected behavior

The complete keyboard focus outline should be visible.

### Actual behavior

The inline-start portion of the focus outline is clipped by the scroll viewport.

### Proposed solution

Add a small desktop-only inline-start gutter to the categories container:

```css
.navbar .categories {
  padding-inline-start: 4px;
}
```
### Environment

- Browser: Google Chrome 151.0.7922.76 (Official Build) (arm64)
- Operating system: macOS 26.6.1

### Screenshot

Image

### Self-service

locphamnice:
Hello! Thanks for taking the time to review this PR.

If this approach doesn't match the intended direction for the project, I'm happy to adjust it or try an alternative approach based on your feedback.

Fixes #9108

## Description

Adds a small inline-start gutter to the desktop documentation navbar so keyboard focus outlines are not clipped by the `ScrollArea` viewport.

The change is scoped to the desktop navbar, leaving the existing mobile and tablet layout unchanged. It uses a logical CSS property to preserve RTL support.

## Testing

- `npm run stylelint`
- `npm run build`

### DIFF
--- a/apps/mantine.dev/src/components/Shell/DocsNavbar/DocsNavbar.module.css
+++ b/apps/mantine.dev/src/components/Shell/DocsNavbar/DocsNavbar.module.css
@@ -28,6 +28,10 @@
   }
 }
 
+.navbar .categories {
+  padding-inline-start: 4px;
+}
+
 .link {
   display: block;
   font-size: var(--mantine-font-size-sm);


rtivital:
Thanks for showing interest in contributing to Mantine! The associated issue was closed as not planned, so this PR is no longer needed.

Issue #9072, opened 2026-07-20
Tooltip: content is not hoverable — tooltip closes before the pointer can reach it (WCAG 1.4.13 violation)


andrewborstein:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

9.4.1

### What package has an issue?

@mantine/core

### What framework do you use?

React Router

### In which browsers you can reproduce the issue?

All

### Describe the bug

Tooltip content is not hoverable, meaning it only appears when the target is hovered, not when the tooltip itself it hovered. I consider this a bug and not a feature request as it's an accessibility conformance issue. The [existing accessibility section in the docs](https://mantine.dev/core/tooltip/#accessibility) seems essentially incomplete in that regard, and the WAI-ARIA link is a 404. I don't expect that any consumer of Mantine would prefer that tooltip content is _not_ hoverable, so it seems like a fair default. Not all, but many popular component libraries already offer this behavior by default, like [shadcn/ui](https://ui.shadcn.com/docs/components/base/tooltip) and [react aria](https://react-aria.adobe.com/Tooltip) and [material ui](https://mui.com/material-ui/react-tooltip/), etc. 

### References
- [WAI-ARIA APG Tooltip pattern](https://www.w3.org/WAI/ARIA/apg/patterns/tooltip)
  - "If the tooltip is invoked when a pointing cursor moves over the trigger element, then it remains open as long as the cursor is over the trigger or the tooltip."
- [USWDS Tooltip accessibility tests](https://designsystem.digital.gov/components/tooltip/accessibility-tests/) 
  - "You can hover over tooltip content. When you move the pointer over an active tooltip, the tooltip stays open."
- [Sarah Higley, "Tooltips in the time of WCAG 2.1"](https://sarahmhigley.com/writing/tooltips-in-wcag-21/#best-practices-summary)
  - "Allow the mouse to easily move over the tooltip without dismissing it." 
- [Scott O'Hara, accessible tooltips](https://scottaohara.github.io/a11y_tooltips/)
  - "The tooltip should dismiss upon mouse hover leaving the element which invoked the tooltip, **unless** the mouse is moved to hover the tooltip itself, per [WCAG success criterion for 1.4.13: Content on Hover or Focus](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html)."

### If possible, include a link to a codesandbox with a minimal reproduction

[...]

PR #9110, opened 2026-08-11
[@mantine/core] Tooltip: Rename `interactive` prop to `hoverable` and fix interactive-content guidance

andrewborstein:
Follow-up to #9072 / the `interactive` prop shipped in 9.5.1.

@rtivital I'm sorry if there was confusion or I could have been more clear in [my original issue description](https://github.com/mantinedev/mantine/issues/9072), but the goal of #9072 was only to make the tooltip **hoverable** — so the pointer can travel from the target onto the tooltip without it closing (WCAG 2.1 SC 1.4.13, _Hoverable_). And at the very most, allow users to select/copy the contents of the tooltip, not to enable users to put interactive content in there. We obviously can't stop them but we shouldn't encourage them either.

### 1. The docs encourage putting interactive content in a tooltip

[The updated docs](https://mantine.dev/core/tooltip/#interactive-tooltip) say 

> Set `interactive` if the tooltip contains content that the user must be able to reach with a pointer, for example a link.

and the demo renders an `<Anchor href>` inside the tooltip. 

However that wasn't the intent of the original fix because interactive/focusable content in a tooltip isn't accessible: a tooltip is exposed to assistive technology only as an `aria-describedby` description, isn't in the focus order, and is dismissed on `Escape`, so keyboard and screen-reader users can't reliably reach it. 

Example guidance against interactive tooltip content:
- [Inclusive Components](https://inclusive-components.design/tooltips-toggletips/) — "Don't put interactive content such as close and confirm buttons or links in tooltips or toggletips."
- [Sarah Higley, _Tooltips in the time of WCAG 2.1_](https://sarahmhigley.com/writing/tooltips-in-wcag-21/#best-practices-summary) — "No interactive content."
- [CSS-Tricks - _Tooltips Best Practices_](https://css-tricks.com/tooltip-best-practices/) — "Tooltips must not contain interactive content."

This PR updates the docs to make that more clear, suggesting that interactive content be put in a `Popover` instead (an accessible non-modal dialog reachable by mouse, keyboard, and touch — which is what the WAI-ARIA APG recommends). It also updates the demo and also corrects a wording bug: 1.4.13 was described as applying _only_ when the tooltip contains interactive content, but it applies to any content shown on hover (and a tooltip should never contain interactive content).

### 2. The prop name `interactive` contradicts that guidance

Naming the prop `interactive` is slightly misleading because it sounds like it's recommending the thing the accessibility guidance says not to do — the docs for `interactive` have to warn against interactive content, which feels odd. This PR renames the prop to **`hoverable`**, which names the actual behavior (and matches the WCAG 1.4.13 _hoverable_ condition it satisfies). Default stays `false` for smallest impact on users, even  though I think you could make a case it should default to `true`. Renaming the prop now, about a week after it shipped, keeps the breaking change as small as possible.

### Commits (intentionally separable)

1. **Clarify that tooltips must not contain interactive content** — docs/demo/JSDoc guidance fix. Non-breaking. Stands on its own.
2. **Rename `interactive` prop to `hoverable`** — breaking rename on top, so it can be `git revert`ed independently if you'd prefer to keep only the guidance fix, or take the guidance now and rename later.

The 9.5.0 changelog prose is left as the historical record of that release; only its live `<Demo>` reference was updated to the renamed export.

Happy to adjust the name, split differently, or drop the rename commit — whatever you prefer.

### DIFF
--- a/apps/mantine.dev/src/pages/changelog/9-5-0.mdx
+++ b/apps/mantine.dev/src/pages/changelog/9-5-0.mdx
@@ -156,4 +156,4 @@ the tooltip contains content that the user must be able to reach with a pointer,
 It is also required by [WCAG 2.1 success criterion 1.4.13](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html),
 which states that content that appears on hover must remain visible while the pointer is over it.
 
-<Demo data={TooltipDemos.interactive} />
+<Demo data={TooltipDemos.hoverable} />
--- a/apps/mantine.dev/src/pages/core/tooltip.mdx
+++ b/apps/mantine.dev/src/pages/core/tooltip.mdx
@@ -170,7 +170,9 @@ By default, the tooltip closes as soon as the pointer leaves the target element,
 
 <Demo data={TooltipDemos.interactive} />
 
-Set `interactive` if the tooltip contains content that the user must be able to reach with a pointer, for example a link. It is also required by [WCAG 2.1 success criterion 1.4.13: Content on Hover or Focus](https://www.w3.org/WAI/WCAG21/Understanding/content-on-hover-or-focus.html), which states that content that appears on hover must remain visible while the pointer is over it.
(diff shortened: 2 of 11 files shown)


rtivital:
Thanks for the PR! I've kept the documentation changes and reverted the `interactive` → `hoverable` rename – the `interactive` prop has been released since 9.5.1, so renaming it (and the `data-interactive` attribute) would be a breaking change for existing users. I've also removed the link from the interactive tooltip story to match the new guidance.

rtivital:
Thanks!

PR #9114, opened 2026-08-13
fix(ScrollArea): replace useEffectEvent with useCallbackRef to prevent React 19.2 rAF crash (Fixes #9113)

### LINKED ISSUE #9113: [@mantine/core] ScrollArea passes useEffectEvent trampolines to requestAnimationFrame, throwing React #440 and wedging the root
### Dependencies check up

9.5.1

@mantine/core

Vite

Firefox

### Describe the bug

`ScrollArea` passes React 19.2's strict `useEffectEvent` trampolines to `requestAnimationFrame` and to a `document` wheel listener. Neither is an effect, so React throws error #440 — *"A function wrapped in useEffectEvent can't be called during rendering"* — whenever one of those callbacks lands while React has `executionContext & RenderContext` set.

The clearest violation is in `packages/@mantine/core/src/components/ScrollArea/use-resize-observer.ts`, where the trampoline **is** the rAF callback:

```ts
const handleResize = useEffectEvent(onResize);

useIsomorphicEffect(() => {
  let rAF = 0;
  if (element) {
    const resizeObserver = new ResizeObserver(() => {
      cancelAnimationFrame(rAF);
      rAF = window.requestAnimationFrame(handleResize);   // ← called outside an effect
    });
    ...
```

React only guarantees an Effect Event is callable from inside an effect or another Effect Event. A rAF callback is neither, so this is a contract violation by construction — React's own guard is `if (executionContext & RenderContext) throw Error(440)`.

Two more sites in the same component are on the same path:

| Site | Invoked from |
|---|---|
| `use-resize-observer.ts` → `handleResize` | `requestAnimationFrame` |
| `ScrollArea.tsx` → `handleOverflowCheck` | passed *into* `useResizeObserver`, so also reached from rAF |
| `ScrollAreaScrollbar/Scrollbar.tsx` → `handleWheelScroll` | `document` `wheel` listener |

`handleOverflowCheck` matters independently: it is itself a `useEffectEvent` trampoline that `useResizeObserver` then wraps a second time, so fixing only `use-resize-observer` moves the throw one frame inward instead of removing it.

**Production evidence.** We hit this on `@mantine/core` 9.5.1 with React 19.2.8. The error is reported by our error tracker with:

```
mechanism:       auto.browser.browserapierrors.requestAnimationFrame
extra.arguments: [3085.62]
```

`3085.62` is the `DOMHighResTimeStamp` that `requestAnimationFrame` passes its callback, and the mechanism tag is our tracker's rAF wrapper — together they place the throw inside a rAF callback, which matches `requestAnimationFrame(handleResize)` exactly.

The consequence is worse than a single caught error. Throwing mid-render leaves React's `executionContext` dirty, so the *next* scheduler task fails its entry guard with **"Should not already be working."** That second throw happens outside any component render, so no error boundary can catch it — the React root stops updating and the tab is stuck until the user reloads.

Observed on Firefox 153 / macOS, 7 occurrences across 4 users since 2026-06-13. It is timing-dependent (it needs the rAF callback to land while React is rendering), so it is rare per user but persistent in aggregate.

I do not have a standalone minimal reproduction — the interleave is a race and I have not been able to force it deterministically outside production. I am reporting it because the contract violation is unambiguous from the source alone, independent of how often the race is won.

Note this is **not** the same as #9078 / #9079. That one is a stale-handler bug, needs `forwardRef`/`memo`, and is confined to `@mantine/hooks`; #9079 does not touch `@mantine/core`, so it would not fix this. Also worth noting: nine hooks in `@mantine/hooks` pair `useEffectEvent` with subscribe-once effects, but as far as I can tell none of them reaches a rAF — `use-collapse` contains both, but calls its trampoline in the effect body before the rAF blocks, which is legal. `ScrollArea` looks like the only rAF path.

### Possible fix

Use the existing ref-based `useCallbackRef` from `@mantine/hooks` instead of React's strict `useEffectEvent` at all three sites. A rAF callback or DOM listener may legally call it, and it keeps the single-subscription behaviour these hooks want.

`Scrollbar.tsx` already uses `useCallbackRef` for its four sibling handlers (`onThumbChange`, `onThumbPointerUp`, `onThumbPositionChange`, `onThumbPointerDown`) — `onWheelScroll` is the odd one out, so this also makes the component internally consistent.

```diff
-import { useEffectEvent } from 'react';
-import { useIsomorphicEffect } from '@mantine/hooks';
+import { useCallbackRef, useIsomorphicEffect } from '@mantine/hooks';

 export function useResizeObserver(element: HTMLElement | null, onResize: () => void) {
-  const handleResize = useEffectEvent(onResize);
+  const handleResize = useCallbackRef(onResize);
```

One caveat if you take this route: `useCallbackRef` currently refreshes its ref in a passive `useEffect`, so a wheel event landing between commit and the passive flush would use the previous render's closure for one tick. #9079 already proposes moving `useCallbackRef` to `useIsomorphicEffect` for exactly that reason, which would close the window for these sites too.

We are running this as a `pnpm` patch against 9.5.1 and it resolves the crash for us.

### Self-service

waterWang:
## Description

`ScrollArea` passes React 19.2's strict `useEffectEvent` trampolines to `requestAnimationFrame` and to a `document` wheel listener. Neither is an effect, so React throws error #440 — *"A function wrapped in useEffectEvent can't be called during rendering"* — whenever one of those callbacks lands while React has `executionContext & RenderContext` set.

The consequence is worse than a single caught error. Throwing mid-render leaves React's `executionContext` dirty, so the *next* scheduler task fails its entry guard with **"Should not already be working."** That second throw happens outside any component render, so no error boundary can catch it — the React root stops updating and the tab is stuck until the user reloads.

## Changes

Replace `useEffectEvent` with `useCallbackRef` from `@mantine/hooks` at all three sites:

1. **`use-resize-observer.ts`** — `handleResize` was passed to `requestAnimationFrame`
2. **`ScrollArea.tsx`** — `handleOverflowCheck` (itself a `useEffectEvent`) was passed into `useResizeObserver`, which then wraps it a second time
3. **`Scrollbar.tsx`** — `handleWheelScroll` was a `document` `wheel` listener

`Scrollbar.tsx` already uses `useCallbackRef` for its four sibling handlers (`onThumbChange`, `onThumbPointerUp`, `onThumbPositionChange`, `onThumbPointerDown`) — `onWheelScroll` was the odd one out, so this also makes the component internally consistent.

Fixes #9113

### DIFF
--- a/packages/@mantine/core/src/components/ScrollArea/ScrollArea.tsx
+++ b/packages/@mantine/core/src/components/ScrollArea/ScrollArea.tsx
@@ -1,6 +1,6 @@
-import { useCallback, useEffectEvent, useRef, useState } from 'react';
+import { useCallback, useRef, useState } from 'react';
 import { useMergeRefs } from '@floating-ui/react';
-import { useIsomorphicEffect } from '@mantine/hooks';
+import { useCallbackRef, useIsomorphicEffect } from '@mantine/hooks';
 import {
   Box,
   BoxProps,
@@ -387,7 +387,7 @@ export const ScrollAreaAutosize = factory<ScrollAreaAutosizeFactory>((props) => 
   const overflowingRef = useRef(false);
   const didMountRef = useRef(false);
 
-  const handleOverflowCheck = useEffectEvent(() => {
+  const handleOverflowCheck = useCallbackRef(() => {
     const el = viewportObserverRef.current;
     if (!el || !onOverflowChange) {
       return;
--- a/packages/@mantine/core/src/components/ScrollArea/ScrollAreaScrollbar/Scrollbar.tsx
+++ b/packages/@mantine/core/src/components/ScrollArea/ScrollAreaScrollbar/Scrollbar.tsx
@@ -1,4 +1,4 @@
-import { useEffect, useEffectEvent, useRef, useState } from 'react';
+import { useEffect, useRef, useState } from 'react';
 import { useCallbackRef, useDebouncedCallback, useMergedRef } from '@mantine/hooks';
 import { useScrollAreaContext } from '../ScrollArea.context';
 import { Sizes } from '../ScrollArea.types';
@@ -42,7 +42,7 @@ export function Scrollbar(props: ScrollbarProps) {
   const prevWebkitUserSelectRef = useRef<string>('');
   const { viewport } = context;
   const maxScrollPos = sizes.content - sizes.viewport;
-  const handleWheelScroll = useEffectEvent(onWheelScroll);
+  const handleWheelScroll = useCallbackRef(onWheelScroll);
   const handleThumbPositionChange = useCallbackRef(onThumbPositionChange);
   const handleResize = useDebouncedCallback(onResize, 10);
 
--- a/packages/@mantine/core/src/components/ScrollArea/use-resize-observer.ts
+++ b/packages/@mantine/core/src/components/ScrollArea/use-resize-observer.ts
@@ -1,8 +1,8 @@
-import { useEffectEvent } from 'react';
+import { useCallbackRef } from '@mantine/hooks';
 import { useIsomorphicEffect } from '@mantine/hooks';
 
 export function useResizeObserver(element: HTMLElement | null, onResize: () => void) {
-  const handleResize = useEffectEvent(onResize);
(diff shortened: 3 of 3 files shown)


rtivital:

PR #9122, opened 2026-08-15
[@mantine/core] Rating: Fix allowClear not working on touch devices

dkpark10:
## Problem

On touch devices, tapping the same star does not clear the rating even with `allowClear` enabled, unlike mouse clicks. `handleTouchStart` calls `setValue` directly, bypassing the `allowClear` check.

Expected the same behavior on touch devices as on PC when `allowClear` is enabled.

https://github.com/user-attachments/assets/d65364ce-37d6-482a-9d95-44718c138616

## Solve

Use `handleChange` instead of `setValue` in `handleTouchStart` so touch goes through the same `allowClear` logic.

https://github.com/user-attachments/assets/270693fa-275b-44f0-b0be-a68f083e7ccc

### DIFF
--- a/packages/@mantine/core/src/components/Rating/Rating.test.tsx
+++ b/packages/@mantine/core/src/components/Rating/Rating.test.tsx
@@ -1,3 +1,4 @@
+import { fireEvent } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { render, screen, tests } from '@mantine-tests/core';
 import { Rating, RatingProps, RatingStylesNames } from './Rating';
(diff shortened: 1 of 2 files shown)


MILLERMARRU (commented):
I compared `handleTouchStart` against `handleChange` in Rating.tsx and the divergence is exactly what the PR title describes: `handleChange` (used by click and keyboard interactions) has the `allowClear` branch that resets to 0 when the tapped value equals the currently rounded value, but `handleTouchStart` calls `setValue` directly and skips that branch entirely. So `allowClear` silently only worked for mouse/keyboard users and never for touch, which is a meaningful gap given Rating is a UI element commonly used on mobile. Routing `handleTouchStart` through `handleChange` instead of `setValue` is the minimal correct fix since it reuses the exact same logic path rather than duplicating the allowClear condition. The new test mocks `getBoundingClientRect` and fires a real touchStart event at the midpoint of a 3-star-wide rating to hit the same star that's already selected, which is a faithful reproduction of the real user gesture.

RobertPakko:
Hi @dkpark10 and @MILLERMARRU , I have also noticed some issues with the Rating component on mobile.

In addition to this issue that you've noticed, scrolling while allowClear is true causes duplicate clicks which result in ratings spontaneously clearing themselves.

I have a separate PR which should fix both the issue you've mentioned here as well as this other issue but they do conflict with each other:
https://github.com/mantinedev/mantine/pull/9159

Just to be transparent the PR is AI generated but I use the forked logic in my own app and it seems to resolve both issues cleanly (https://github.com/RobertPakko/star-voting)

PR #9123, opened 2026-08-16
fix(@mantine/hooks): useDebouncedValue with leading:true emits stale values (#9119)

### LINKED ISSUE #9119: useDebouncedValue with `leading: true` fires multiple times per burst and can emit a stale value
### Dependencies check up

9.4.2

@mantine/hooks

Vite

All

### Describe the bug

When `useDebouncedValue` is used with `{ leading: true }`, a single burst of rapid value changes produces **more than the expected two emissions** (one leading, one trailing), and the **final emission can be a stale, out-of-order value**.

Typing `abc` — three changes, each 100ms apart, with `wait: 500` — emits:

```
'a'  →  'abc'  →  'ab'
```

The last emission is `'ab'`, a value that was superseded 400ms earlier. For the common case of debouncing a search box, that means the last request fired is for a query the input no longer contains, so the UI settles on results for the wrong term. It also fires 3 requests where 2 are expected.

The longer the burst, the more often it repeats — every third change re-triggers the leading edge.

### Cause

In [`use-debounced-value.ts`](https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/hooks/src/use-debounced-value/use-debounced-value.ts):

```ts
const cancel = useCallback(() => {
  window.clearTimeout(timeoutRef.current);
  timeoutRef.current = null;
  cooldownRef.current = false;   // (1)
}, []);

useEffect(() => {
  if (mountedRef.current) {
    if (!cooldownRef.current && options.leading) {
      cooldownRef.current = true;
      setValue(value);
      timeoutRef.current = window.setTimeout(() => {   // (2)
        cooldownRef.current = false;
      }, wait);
    } else {
      cancel();
      timeoutRef.current = window.setTimeout(() => {
        cooldownRef.current = false;
        setValue(value);
      }, wait);
    }
  }
}, [value, options.leading, wait]);
```

Two defects interact:

1. **`cancel()` resets `cooldownRef`.** That is correct for its unmount/abort role, but the trailing branch reuses `cancel()` purely to clear the timer. As a side effect it clears the cooldown flag — the only thing preventing the leading edge from re-firing mid-burst.

2. **The leading branch overwrites `timeoutRef.current` without clearing it first.** Once (1) has let control back into the leading branch, the pending trailing timer is orphaned: nothing else holds a reference to it, so it is never cancelled, and it later fires `setValue(value)` with the `value` captured from an earlier render.

### Reproduction

```tsx
function Demo() {
  const [value, setValue] = useState('');
  const [debounced] = useDebouncedValue(value, 500, { leading: true });

  useEffect(() => {
    console.log('emitted:', debounced);
  }, [debounced]);

  return <input value={value} onChange={(e) => setValue(e.currentTarget.value)} />;
}
```

Type `abc` quickly (each keystroke under 500ms apart). Console logs `a`, then `abc`, then `ab`.

can be reproduced on the[ mantine.dev site](https://mantine.dev/hooks/use-debounced-value/#leading-update):  

### Possible fix

I believe these changes would fix the issue and ensure the stale / redundant call isn't executed. 

1. extract clearTimer() so clearing a timer no longer resets the cooldown
2. call it unconditionally at the top of the effect
3. drop the now-redundant cancel() from the trailing branch.

### Self-service

waterWang:
## Problem
When `useDebouncedValue` is used with `{ leading: true }`, a single burst of rapid value changes produces more than the expected two emissions (one leading, one trailing), and the final emission can be a stale, out-of-order value.

Typing `abc` — three changes, each 100ms apart, with `wait: 500` — emits `'a' → 'abc' → 'ab'`. The last emission is `'ab'`, a value that was superseded.

## Root Cause
Two defects in `use-debounced-value.ts`:

1. **`cancel()` resets `cooldownRef`.** The trailing branch reused `cancel()` just to clear the timer, but as a side effect it cleared the cooldown flag — the only thing preventing the leading edge from re-firing mid-burst.

2. **The leading branch overwrites `timeoutRef.current` without clearing it first.** Once the cooldown is cleared, the pending trailing timer is orphaned and later fires `setValue()` with a stale `value` from an earlier render.

## Fix
- Replace `cancel()` in the trailing branch with `window.clearTimeout()` only — do not reset `cooldownRef`, so the leading edge does not re-fire mid-burst.
- Also clear any lingering timeout in the leading branch before setting a new one, so an orphaned timer cannot later fire with a stale value.
- `cancel()` is unchanged for its unmount/abort/`flush` role.

Fixes #9119

### DIFF
--- a/packages/@mantine/hooks/src/use-debounced-value/use-debounced-value.test.ts
+++ b/packages/@mantine/hooks/src/use-debounced-value/use-debounced-value.test.ts
@@ -126,5 +126,81 @@ describe('useDebouncedValue', () => {
       rerender({ value: 'c' });
       expect(result.current[0]).toBe('c');
     });
+
+    test('emits only the leading and the final value during a long burst', () => {
+      const emitted: string[] = [];
+      const { rerender } = renderHook(
+        ({ value }) => {
+          const [debounced] = useDebouncedValue(value, 500, { leading: true });
+          if (emitted[emitted.length - 1] !== debounced) {
+            emitted.push(debounced);
+          }
+          return debounced;
+        },
+        {
+          initialProps: { value: '' },
+        }
+      );
+
+      ['a', 'ab', 'abc', 'abcd', 'abcde'].forEach((value) => {
+        rerender({ value });
+        act(() => jest.advanceTimersByTime(100));
+      });
+
+      act(() => jest.advanceTimersByTime(500));
+
+      // the leading edge must not re-fire mid-burst, and the trailing edge
+      // must emit the latest value rather than a superseded one
+      expect(emitted).toEqual(['', 'a', 'abcde']);
+    });
+
+    test('the trailing update is scheduled from the last change', () => {
+      const { result, rerender } = renderHook(
+        ({ value }) => useDebouncedValue(value, 200, { leading: true }),
+        {
+          initialProps: { value: 'a' },
+        }
+      );
+
+      rerender({ value: 'b' });
+      expect(result.current[0]).toBe('b');
+
+      act(() => jest.advanceTimersByTime(150));
+      rerender({ value: 'c' });
+
+      act(() => jest.advanceTimersByTime(199));
+      expect(result.current[0]).toBe('b');
+
+      act(() => jest.advanceTimersByTime(1));
+      expect(result.current[0]).toBe('c');
+    });
+
+    test('flush is not followed by a stale update', () => {
+      const emitted: string[] = [];
+      const { result, rerender } = renderHook(
+        ({ value }) => {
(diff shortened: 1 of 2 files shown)


Dylan-Brandenburg:
This is essentially how I patched a local repository, thank you for picking this up!. The only difference is I had moved the timer resets into a function call, essentially just guarding on the timeoutRef


```
const clearTimer = useCallback(() => {
  if (timeoutRef.current !== null) { // <- ensure current exists
    window.clearTimeout(timeoutRef.current);
    timeoutRef.current = null;
  }
}, []);

```

rtivital:

PR #9125, opened 2026-08-16
fix(Calendar): always use current date as fallback month instead of minDate

### LINKED ISSUE #9117: DatePicker shows different initial month compared to v7
### Dependencies check up

9.5.1

@mantine/dates

Tanstack Start

All

### Describe the bug

Back in Mantine v7, DatePicker defaulted to the current date (`new Date()`) if you don't specify a `date`
In v8/v9 it seems to use the month of `minDate` instead, if that is given.

v7: https://codesandbox.io/p/sandbox/suspicious-cookies-sdrfcc?file=%2Fsrc%2FApp.tsx%3A13%2C1
v8: https://codesandbox.io/p/sandbox/quirky-waterfall-2nxqry?file=%2Fsrc%2FApp.tsx%3A16%2C18
v9: https://codesandbox.io/p/devbox/trusting-kirch-g2r5l4?file=%2Fsrc%2FApp.tsx

Discord thread: https://discord.com/channels/854810300876062770/1341862410096607383

https://codesandbox.io/p/devbox/trusting-kirch-g2r5l4?file=%2Fsrc%2FApp.tsx

### Possible fix

_No response_

### Self-service

waterWang:
## Summary

In Mantine v7, `DatePicker` (and other date-based components built on `Calendar`) defaulted to the current date (`new Date()`) when no `date`/`defaultDate` was provided. Since v8, when a `minDate` is set and today is after `minDate`, the fallback displayed date incorrectly uses `minDate` instead of the current date — so the calendar opens showing a different month than expected.

### Root cause

In `packages/@mantine/dates/src/components/Calendar/Calendar.tsx`, the fallback date ref was built as:

```ts
fallbackDateRef.current =
  minDate && dayjs(now).isAfter(minDate) ? minDate : dayjs(now).format('YYYY-MM-DD');
```

This picks `minDate` whenever it is in the past, which is the common case for date inputs with `minDate` set (e.g. "select a future date"). The displayed month then jumps to `minDate`'s month instead of the current month.

### Fix

Always use the current date as the fallback, matching v7 behavior:

```ts
fallbackDateRef.current = dayjs(now).format('YYYY-MM-DD');
```

### Test plan

- Existing `Calendar.test.tsx` suite still passes (fallback date assertions unchanged for the no-`minDate` case).
- Manual: render `<DatePicker minDate={dayjs().subtract(1, 'year').toDate()} />` with no `date` — the calendar now opens on the current month, as in v7, and `minDate` is still respected for selection boundaries.

Fixes #9117

### DIFF
--- a/packages/@mantine/dates/src/components/Calendar/Calendar.tsx
+++ b/packages/@mantine/dates/src/components/Calendar/Calendar.tsx
@@ -304,9 +304,7 @@ export const Calendar = factory<CalendarFactory>((_props) => {
   const fallbackDateRef = useRef<DateStringValue | null>(null);
   if (fallbackDateRef.current === null) {
     const now = new Date();
-    fallbackDateRef.current = (
-      minDate && dayjs(now).isAfter(minDate) ? minDate : dayjs(now).format('YYYY-MM-DD')
-    ) as DateStringValue;
+    fallbackDateRef.current = dayjs(now).format('YYYY-MM-DD') as DateStringValue;
   }
   const currentDate = _date || fallbackDateRef.current;
 


rtivital:

PR #9131, opened 2026-08-18
fix(TooltipFloating): guard against NaN in top/left position style

### LINKED ISSUE #9130: TooltipFloating: NaN is an invalid value for top/left CSS when y/x is NaN from floating-ui
## Bug description

`Tooltip.Floating` sets `top: NaN` on the tooltip element when `y` from `useFloating` resolves to `NaN`. This triggers a React warning:

```
NaN is an invalid value for the `top` css style property.
```

## Root cause

In `TooltipFloating.tsx`, the position style is computed as:

```js
top: (y && Math.round(y)) ?? "",
left: (x && Math.round(x)) ?? "",
```

The guard `(y && ...)` uses a **falsy check**, but `NaN` is falsy in JavaScript. This means:

- `NaN && Math.round(NaN)` → `NaN` (short-circuits, returns `NaN`)
- `NaN ?? ""` → `NaN` (nullish coalescing only catches `null`/`undefined`, not `NaN`)

So `top: NaN` is set on the DOM element.

## When does y become NaN?

In `use-floating-tooltip.ts`, `handleMouseMove` builds a virtual bounding rect using `clientX`/`clientY` from the event:

```js
const handleMouseMove = useCallback(({ clientX, clientY }) => {
  refs.setPositionReference({ getBoundingClientRect() {
    return {
      left: clientX + horizontalOffset,
      top: clientY + verticalOffset,
      ...
    };
  } });
}, [elements.reference]);
```

On **mobile/touch devices**, the synthesized `mouseenter` event can have `clientX`/`clientY` as `undefined` in some browsers. This causes:

- `undefined + offset = NaN` in the bounding rect
- floating-ui computes `y = NaN`
- `top: NaN` is set → React warning

## Reproduction

1. Use `Tooltip.Floating` wrapping any interactive element (e.g. a Recharts `BarChart`)
2. Open on a mobile device or Chrome DevTools touch emulation
3. Tap the element — observe the console warning:

```
NaN is an invalid value for the `top` css style property.
```

## Expected behavior

`top` and `left` should fall back to `""` when the computed value is `NaN`.

## Suggested fix

```diff
- top: (y && Math.round(y)) ?? "",
- left: (x && Math.round(x)) ?? "",
+ top: (y != null && !Number.isNaN(y)) ? Math.round(y) : "",
+ left: (x != null && !Number.isNaN(x)) ? Math.round(x) : "",
```

## Additional notes

- `disabled={true}` on `Tooltip.Floating` does **not** prevent this — the component still mounts and injects `onMouseEnter` via `cloneElement`, so the event still fires and position tracking still runs. Only conditional rendering (not mounting `Tooltip.Floating` at all) avoids it.
- Mantine version: latest (reproduced in `@mantine/core` with the `es6` build)

sakal-s:
## Problem

`Tooltip.Floating` sets `top: NaN` on the tooltip element when `useFloating` returns `NaN` for `x`/`y`, triggering a React warning:

```
NaN is an invalid value for the `top` css style property.
```

This happens on **mobile/touch devices** where the synthesized `mouseenter` event can have `clientX`/`clientY` as `undefined`. In `use-floating-tooltip.ts`, `handleMouseMove` uses these to build a virtual bounding rect — `undefined + offset = NaN` — so floating-ui computes `y = NaN`.

The existing guard in `TooltipFloating.tsx` does not catch this:

```ts
top: (y && Math.round(y)) ?? '',
```

- `NaN` is **falsy**, so `NaN && Math.round(NaN)` → `NaN`
- `NaN` is **not** `null`/`undefined`, so `NaN ?? ''` → `NaN`
- Result: `top: NaN` on the DOM element

## Fix

Replace the falsy check with `Number.isFinite`, which correctly rejects `null`, `undefined`, `NaN`, and `Infinity` while preserving `0` (which `||` would incorrectly coerce to `''`):

```ts
// before
top: (y && Math.round(y)) ?? '',
left: (x && Math.round(x)) ?? '',

// after
top: Number.isFinite(y) ? Math.round(y!) : '',
left: Number.isFinite(x) ? Math.round(x!) : '',
```

Closes #9130

### DIFF
--- a/packages/@mantine/core/src/components/Tooltip/TooltipFloating/TooltipFloating.test.tsx
+++ b/packages/@mantine/core/src/components/Tooltip/TooltipFloating/TooltipFloating.test.tsx
@@ -1,0 +1,30 @@
+import { render, screen, userEvent } from '@mantine-tests/core';
+import { TooltipFloating } from './TooltipFloating';
+
+function TargetWithoutMouseCoordinates({ onMouseEnter, ...others }: any) {
+  return (
+    <button {...others} type="button" onClick={() => onMouseEnter({ chartX: 10, chartY: 20 })}>
+      target
+    </button>
+  );
+}
+
+describe('@mantine/core/TooltipFloating', () => {
+  it('does not set NaN position when children call onMouseEnter without mouse coordinates', async () => {
+    const spy = jest.spyOn(console, 'error').mockImplementation(() => {});
+
+    render(
+      <TooltipFloating label="test-tooltip" withinPortal={false}>
+        <TargetWithoutMouseCoordinates />
+      </TooltipFloating>
+    );
+
+    await userEvent.click(screen.getByRole('button'));
+
+    const nanWarnings = spy.mock.calls.filter((call) => String(call[0]).includes('NaN'));
+    spy.mockRestore();
+
+    expect(nanWarnings).toHaveLength(0);
+    expect(screen.getByText('test-tooltip').getAttribute('style')).not.toContain('NaN');
+  });
+});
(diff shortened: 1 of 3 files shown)


rtivital:
Thanks!

PR #9134, opened 2026-08-19
[@mantine/core] Notification: Allow CloseButton props in closeButtonProps type

### LINKED ISSUE #9133: [@mantine/core] Notification: closeButtonProps type does not accept CloseButton props
### Describe the bug

`NotificationProps['closeButtonProps']` is typed as `ElementProps<'button'> & DataAttributes`, which only permits plain DOM `<button>` attributes and `data-*` attributes. As a result, none of the `CloseButton`-specific props type-check, even though all of them work at runtime:

```tsx
<Notification closeButtonProps={{ iconSize: 20 }} />
// Object literal may only specify known properties, and 'iconSize' does not exist
// in type 'ElementProps<"button"> & DataAttributes'. ts(2353)
```

The same happens for `size`, `radius` and `icon`.

The props do take effect: `Notification` spreads `closeButtonProps` **after** its own `iconSize={16}` / `color="gray"` defaults, which is what #2205 changed on purpose so that these values could be overridden. `CloseButton`'s vars resolver then turns `iconSize` into `--cb-icon-size`, so the icon really does resize. Only the type disagrees.

`Modal` and `Drawer` are unaffected: both type `closeButtonProps` as `ModalBaseCloseButtonProps`, which extends `__CloseButtonProps`. `Notification` is the only component taking a `closeButtonProps` prop that does not include the `CloseButton` props.

Related but not the same: #4793 (fixed, was about `Modal`), #4893 (about `NotificationsProps` on `@mantine/notifications`, a different type).

Reproduction without a sandbox (a sandbox editor does not reliably surface `tsc` diagnostics, so here is the exact compiler check instead). In a project with `@mantine/core@9.5.1` installed:

```ts
// check.ts
import type { NotificationProps } from '@mantine/core';

const a: NotificationProps['closeButtonProps'] = { iconSize: 20 };
void a;
```

```bash
npx tsc --noEmit --strict --module esnext --moduleResolution bundler \
  --target es2020 --lib es2024,dom,dom.iterable --jsx react-jsx check.ts
```

```
check.ts(4,52): error TS2353: Object literal may only specify known properties, and 'iconSize'
does not exist in type 'ElementProps<"button"> & DataAttributes'.
```

### Possible fix

Widen the type to include `__CloseButtonProps`, which is already exported from `@mantine/core`:

```diff
- closeButtonProps?: ElementProps<'button'> & DataAttributes;
+ closeButtonProps?: __CloseButtonProps & ElementProps<'button'> & DataAttributes;
```

This is a pure type widening — no runtime change and not breaking. `ButtonHTMLAttributes` declares none of `iconSize`, `size`, `radius` or `icon`, so nothing in the intersection collides. After the change, unknown props and wrongly-typed `iconSize` values are still rejected, so the type is widened without being loosened.

Happy to open a PR.

amansachdev:
Closes #9133

### Problem

`NotificationProps['closeButtonProps']` was typed as `ElementProps<'button'> & DataAttributes`, so `iconSize`, `size`, `radius` and `icon` were rejected by TypeScript despite working at runtime — `Notification` spreads `closeButtonProps` after its own `iconSize={16}` / `color="gray"` defaults, which is what #2205 made possible.

`Modal` and `Drawer` are not affected: both type the same prop as `ModalBaseCloseButtonProps`, which extends `__CloseButtonProps`.

### Solution

Add `__CloseButtonProps` (already exported from `@mantine/core`) to the intersection, matching how `Modal` and `Drawer` type the same prop:

```diff
- closeButtonProps?: ElementProps<'button'> & DataAttributes;
+ closeButtonProps?: __CloseButtonProps & ElementProps<'button'> & DataAttributes;
```

- Types only — no runtime, CSS, or export changes
- Not breaking — the type is strictly widened, every currently valid `closeButtonProps` object stays valid
- `ButtonHTMLAttributes` declares none of `iconSize`, `size`, `radius` or `icon`, and `__CloseButtonProps['data-disabled']` (`boolean`) is assignable to `DataAttributes`' `string | number | boolean | undefined` index signature, so no member of the intersection collides
- Verified that unknown props and `iconSize: true` are still rejected after the change, so this widens the type without loosening it

### Tests

Type-only change, no runtime behavior modified. `Notification.test.tsx` (48 tests) passes unchanged. Also verified locally: `npm run typecheck`, `npm run lint`, `npm run format:test`, `npm run syncpack` and `npm run docs:docgen` (no generated output changes).

### DIFF
--- a/packages/@mantine/core/src/components/Notification/Notification.tsx
+++ b/packages/@mantine/core/src/components/Notification/Notification.tsx
@@ -14,7 +14,7 @@ import {
   useProps,
   useStyles,
 } from '../../core';
-import { CloseButton } from '../CloseButton';
+import { __CloseButtonProps, CloseButton } from '../CloseButton';
 import { Loader, LoaderProps } from '../Loader';
 import classes from './Notification.module.css';
 
@@ -60,7 +60,7 @@ export interface NotificationProps
   withCloseButton?: boolean;
 
   /** Props passed down to the close button */
-  closeButtonProps?: ElementProps<'button'> & DataAttributes;
+  closeButtonProps?: __CloseButtonProps & ElementProps<'button'> & DataAttributes;
 
   /** Props passed down to the `Loader` component */
   loaderProps?: LoaderProps;


rtivital:
Thanks!

PR #9136, opened 2026-08-19
[@mantine/core] RollingNumber: Fix off-screen digits captured on copy

### LINKED ISSUE #9132: RollingNumber copy-and-paste includes off-screen numbers
### Dependencies check up

9.5.1

@mantine/core

React Router

Firefox

### Describe the bug

Thanks for the RollingNumber component! It's a great visual addition to my app.

Sadly, when you select and copy the number displayed by the component, the result is something like:

```
1234567890101234567890101234567890101234
```

This is from the first example on the documentation page, copied while it was displaying "1234".

https://mantine.dev/core/rolling-number/

### Possible fix

Making the off-screen numbers non-selectable via `user-select`, possibly? I'm no expert on this.

### Self-service

Sriram-Nambiar:
Fixes #9132

### Summary of changes
- Only render real DOM `#text` nodes for active digits in `DigitColumn`; inactive/off-screen strip cells receive `data-value` attributes without child text nodes.
- Render off-screen strip digits via CSS `&::before { content: attr(data-value); }`, preserving smooth vertical roll animations while preventing browser clipboard serialization of off-screen numbers.
- Removed duplicate `<VisuallyHidden>` text nodes and suppressed empty/collapsed char slots.
- Added unit test asserting clean rendered `textContent` on formatted numbers.

### DIFF
--- a/packages/@mantine/core/src/components/RollingNumber/DigitColumn.tsx
+++ b/packages/@mantine/core/src/components/RollingNumber/DigitColumn.tsx
@@ -34,7 +34,7 @@ export function DigitColumn({
   const columnStyles = getStyles('digitColumn');
 
   return (
-    <span {...digitStyles} data-empty={empty || undefined} aria-hidden="true">
+    <span {...digitStyles} data-empty={empty || undefined}>
       <span
         key={digit}
         {...columnStyles}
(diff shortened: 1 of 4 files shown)


rtivital:

PR #9139, opened 2026-08-20
fix(@mantine/dates): prevent SpinInput auto-advance when appending to 00 value

### LINKED ISSUE #9128: DateTimePicker Time mis-focuses on minutes after a single '0' '1' or '2' input for hours
### Dependencies check up

8.5.1

@mantine/dates

Vite

Chrome

### Describe the bug

Reproduction:

1. Head to https://mantine.dev/dates/date-time-picker/
2. Open the picker modal
3. Select a date first
4. Press on the hours input, and enter '1'

Expected behaviour:

The focus remains on the hour section, as is the case for the TimePicker and the DateTimePicker when a date is not selected beforehand.

Actual behaviour:

The focus jumps to minutes, requiring re-focus on hours to complete correct input.

_No response_

### Possible fix

This issue might have been addressed for the TimePicker previously: https://github.com/mantinedev/mantine/issues/8967 

### Self-service

waterWang:
fix(@mantine/dates): prevent SpinInput auto-advance when appending to 00 value

Closes #9128

## Problem

When the DateTimePicker shows `00:00` (e.g. after selecting a date with default time `00:00`), clicking the hours input and typing a single digit like `1`, `2` or `0` moves focus to the minutes input prematurely.

The user intended to enter a value like `01`, `02` or `12` — the focus jump happens after the first digit, forcing the user to re-focus the hours input.

## Root cause

In `SpinInput`, the auto-advance condition:

```ts
if (!disableAutoAdvance && (clampedValue > maxDigit || value.startsWith('00'))) {
  onNextInput?.();
}
```

`value.startsWith('00')` is too broad: when the hours input already displays `00` and the user types at the end of the field, the new value becomes `001` / `002` / `003`... which also starts with `00` and triggers `onNextInput()`, jumping to minutes.

## Fix

Narrow the condition to only match an exact `00` value (2 characters), which is the intended auto-advance case — when the user has typed a complete `00`:

```ts
if (!disableAutoAdvance && (clampedValue > maxDigit || value === '00')) {
  onNextInput?.();
}
```

## Tests

- Added a DateTimePicker test: select a date (with default `00:00`), type `1` on the hours input — expect the hours input to keep focus instead of jumping to minutes.
- Updated the TimePicker test from #8967 ("allows entering 01-09 hours over a selected 00 value") to reflect the corrected behavior: after typing `02`, the hours input retains focus (the value is correctly set to `02`).

## Verification

- `TimePicker.test.tsx`: 101/101 passed
- `DateTimePicker.test.tsx`: 111/111 passed

### DIFF
--- a/packages/@mantine/dates/src/components/DateTimePicker/DateTimePicker.test.tsx
+++ b/packages/@mantine/dates/src/components/DateTimePicker/DateTimePicker.test.tsx
@@ -275,6 +275,25 @@ describe('@mantine/dates/DateTimePicker', () => {
     expect(getTimePicker()).toHaveFocus();
   });
 
+  it('keeps focus on hours input when typing a single digit after selecting a date', async () => {
+    const { container } = render(<DateTimePicker {...defaultProps} />);
+    await clickInput(container);
+    // Select a date — this triggers handleDefaultDateChange, which focuses hours input
+    // and also triggers re-render via useDidUpdate (startTimeValue update)
+    await userEvent.click(container.querySelector('table button')!);
+
+    const hoursInput = getTimePicker();
+    expect(hoursInput).toHaveFocus();
+    // After re-render, focus is preserved but selection may be lost.
+    // Type '1' as if the cursor is at the end of "00" (the default time value).
+    // Without the fix, "001" triggers startsWith('00') and jumps to minutes.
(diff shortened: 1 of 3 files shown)


rtivital:

PR #9140, opened 2026-08-20
[@mantine/code-highlight] CodeHighlight: Keep indentation of the first line of the code block

mokevnin:
`CodeHighlight` passes `code.trim()` to the highlighter, so the leading whitespace of the **first** line is dropped while every other line keeps its indentation. Any block whose first line is intentionally indented comes out misaligned:

```tsx
<CodeHighlight
  language="text"
  code={'              [  6   7 ]\nA · B = [ 0  1  2 ] · [  8   9 ]\n        [ 3  4  5 ]   [ 10  11 ]\n'}
/>
```

renders as

```text
[  6   7 ]
A · B = [ 0  1  2 ] · [  8   9 ]
        [ 3  4  5 ]   [ 10  11 ]
```

Same for a wrapped shell command, where the continuation lines then hang under nothing:

```text
    docker exec mongo mongodump \
      --authenticationDatabase admin \
      --out /tmp/mongo-backup
```

We hit this rendering markdown course content: a fenced block carries its text verbatim (first line's indentation included) and always ends with a trailing `\n`, so `trim()` does more than drop that newline. GitHub/GitLab markdown preview of the same source renders those blocks correctly. About 2000 blocks in our content start with an indented first line.

## Fix

`normalizeCode()` strips leading blank lines and trailing whitespace, but not the indentation of the first line. The trailing half has to stay: markdown appends a trailing `\n` and `withLineNumbers` counts `split('\n')`, so dropping it would add a phantom last line number.

Applied to every place that trimmed before: the highlighter input, the non-highlighted fallback (this is also the first paint with an async adapter such as shiki), line numbers, `CopyCodeButton` (the copied snippet had a broken first line too — matters for Python/YAML) and `createHighlightJsAdapter`.

`InlineCodeHighlight` goes through the same path, so inline snippets keep their leading whitespace as well.

Dedenting by the common indentation would not be equivalent: in the matrix example above the second line starts at column 0, so the common indent is 0 while the first line still loses its 14 spaces.

## Test

`CodeHighlight` keeps the indentation of the first line, strips the trailing newline and renders 2 line numbers for a 2-line block. Asserted on `textContent` rather than `toHaveTextContent`, which normalizes whitespace and would pass either way.

### DIFF
--- a/apps/mantine.dev/src/pages/x/code-highlight.mdx
+++ b/apps/mantine.dev/src/pages/x/code-highlight.mdx
@@ -222,6 +222,18 @@ Set `withLineNumbers` prop to display line numbers alongside the code:
 
 <Demo data={CodeHighlightDemos.lineNumbers} />
 
+## First line indentation
+
+`CodeHighlight` trims the code before rendering it, so the indentation of the **first** line
+is removed while every other line keeps it. Set `withFirstLineIndentation` prop to preserve it –
+this is useful for blocks whose first line is intentionally indented, for example column-aligned
+text or a wrapped shell command:
+
+<Demo data={CodeHighlightDemos.firstLineIndentation} />
+
+Blank lines at the start and trailing whitespace are removed in both cases. The prop also
+controls the code that is copied by the copy button.
+
 ## Expandable code
 
 If the code snippet is too long, you can make it expandable with `withExpandButton`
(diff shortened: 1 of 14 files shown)


rtivital:
Thanks!

PR #9141, opened 2026-08-21
[@mantine/code-highlight] Add support for lazy languages loading

mokevnin:
## Problem

Shiki loads grammars of every language given to the highlighter up front. In an application that highlights code in many languages this means all of those grammars are downloaded before the first code block can be highlighted — in our case (~60 languages passed to `createHighlighterCore`) that is ~2.4 MB of grammar source, and a single `cpp` grammar alone is 524 KB.

Lazy loading is not expressible with the current adapter contract: `getHighlighter` is synchronous, so there is no point at which an adapter can await a grammar for the language it is asked to highlight.

## What this adds

`CodeHighlightAdapter` gets an optional `loadLanguage(ctx, language)`:

- `CodeHighlight` reports the language it renders via a new `useLoadLanguage` hook.
- The provider calls `adapter.loadLanguage` once per language (deduplicated, so N blocks with the same language load it once) and re-renders when the returned promise settles.
- Adapters that do not implement `loadLanguage` behave exactly as before.

For the shiki adapter this is opt-in via the `resolveLanguage` option, whose return value is passed to shiki's `highlighter.loadLanguage`:

```tsx
// bundled highlighter — resolves grammars by name itself
createShikiAdapter(loadShiki, { resolveLanguage: (language) => language });

// shiki/core — grammars are not bundled
const grammars = { python: () => import('@shikijs/langs/python') };
createShikiAdapter(loadShiki, { resolveLanguage: (language) => grammars[language] });
```

## Behavior change (second commit, separate on purpose)

Until a grammar is loaded the language is unavailable, so the shiki adapter needs to be able to render that state. `createShikiAdapter` now renders code as plain text when the language is not loaded, instead of letting shiki throw `Language ... not found` from render. This affects existing setups too: a code block with a language missing from `langs` used to crash, now it renders unhighlighted. It felt like a strict improvement, but it is isolated in its own commit in case you would rather not take it.

`getLoadedLanguages` is called defensively (`typeof ... === 'function'`) — `loadShiki` is typed as `() => Promise<any>`, so a context that does not expose it keeps its current behavior. Shiki's special languages (`text`, `plaintext`, `txt`, `ansi`) are always treated as available.

## Docs

Added a "Lazy languages loading" section to `x/code-highlight` with both the bundled and `shiki/core` setups.

## Verification

- `npx jest packages/@mantine/code-highlight` — 136 tests pass, including new provider tests (plain → highlighted transition, one `loadLanguage` call for two blocks of the same language) and shiki adapter tests.
- `npx oxlint -c oxlint.config.mjs packages/@mantine/code-highlight/src` — clean.
- `npm run format:write:files` on the changed files.
- `npm run typecheck` / `npm run build` could not run in my environment (`yarn install` fails to build `sharp`, so workspace type declarations are missing); instead I typechecked the package with `tsc` and paths mapped to workspace sources — clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/apps/mantine.dev/src/pages/x/code-highlight.mdx
+++ b/apps/mantine.dev/src/pages/x/code-highlight.mdx
@@ -140,6 +140,16 @@ const shikiAdapter = createShikiAdapter(loadShiki, {
 If `resolveLanguage` returns `null` or `undefined`, the language is considered unsupported and
 its code is displayed as plain text.
 
+## Unavailable languages
+
+If a code block uses a language that is not loaded in the highlighter and cannot be loaded on
+demand, its code is displayed as plain text instead of throwing an error. In development, a
+warning is logged to the console once per language – either add the language to the `langs`
+option of the highlighter, or load it on demand with `resolveLanguage`.
+
+If a grammar fails to load (for example, a dynamic import fails), the code stays plain text and
+the language is loaded again the next time a code block with that language is rendered.
+
 ## Usage with highlight.js
 
 [Highlight.js](https://highlightjs.org/) provides less accurate highlighting compared to shiki,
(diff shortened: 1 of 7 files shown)


rtivital:
Thanks!

PR #9146, opened 2026-08-23
[@mantine/hooks] use-throttled-callback: Fix callback becoming unresponsive after clearTimeout

vansh917917:
## Description
Fixes an issue where `useThrottledCallback` becomes permanently unresponsive after invoking `clearTimeout`.

Previously, `clearTimeout` only cancelled `timeoutRef.current` without resetting `active.current = true` or clearing internal argument refs. When the throttled function was invoked subsequently, it was trapped in the `else` branch (`active.current === false`), updating `latestInArgsRef.current` without scheduling a timer or resetting `active.current = true`. This caused the throttled callback to permanently discard all future invocations.

## Changes
- Reset `active.current = true`, `timeoutRef.current = -1`, `latestInArgsRef.current = null`, and `latestOutArgsRef.current = null` inside `clearTimeout`.
- Added unit test in `use-throttled-callback.test.ts` verifying that `clearTimeout()` permits subsequent calls to fire immediately.

## Checklist
- [x] Ran unit tests with `npm run jest packages/@mantine/hooks` (All 536 tests pass)
- [x] Ran `oxlint` and `oxfmt`
- [x] Ran `tsc --noEmit`
- [x] No unnecessary inline comments added

### DIFF
--- a/packages/@mantine/hooks/src/use-throttled-callback/use-throttled-callback.test.ts
+++ b/packages/@mantine/hooks/src/use-throttled-callback/use-throttled-callback.test.ts
@@ -1,5 +1,8 @@
 import { act, renderHook } from '@testing-library/react';
-import { useThrottledCallback } from './use-throttled-callback';
+import {
+  useThrottledCallback,
+  useThrottledCallbackWithClearTimeout,
+} from './use-throttled-callback';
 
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #9153, opened 2026-08-26
[@mantine/schedule] MobileMonthView: Fix multi-day events not showing when they start in the previous month

dkpark10:
## Problem
Multi-day events starting in the previous month were filtered out entirely, since the filter only checked if `event.start` was in the displayed month.

For example, the following event data:

```tsx
const date: ScheduleEventData[] = [
  {
    id: 1,
    title: 'Team Meeting',
    start: `2026-07-30`,
    end: `2026-08-03`,
    color: 'blue',
  },
];
```

would not show up at all when viewing August, even on August 1–3, which the event actually overlaps.

스크린샷 2026-08-26 오후 4 04 51

## Solve
Filter by month range overlap instead of start-month equality, matching `MonthView`'s existing approach.

스크린샷 2026-08-26 오후 3 28 06


rtivital:
Thanks!

PR #9156, opened 2026-08-28
fix(Popover): update placement dynamically when position prop changes while open

### LINKED ISSUE #9144: Menu Dropdown Position doesn't appear to work currently.
### Dependencies check up

9.5.2

@mantine/core

Vite

### Describe the bug

Browsers Tested: Firefox, Chrome

Menu dropdown position does not appear to be working in this version. Works as expected in latest 8x version.

First noticed in my application, confirmed not working on the mantine.dev site: https://mantine.dev/core/menu/#dropdown-position

_No response_

### Possible fix

_No response_

### Self-service

AdarshYadav9:
## **Description**

Fixes a bug where the Popover/Menu dropdown position stopped updating when the `position` prop changed while the dropdown was open, for example when using the interactive controls in the Menu **Dropdown Position** demo.

## **How I Fixed It**

1. Reset the locked placement when the `position` prop changes so the new position is applied immediately.
2. Trigger `floating.update()` when the `position` prop changes to recalculate the dropdown placement.
3. Prevent the placement from being re-locked until the new position has been calculated.

## **Related Issues**

Fixes #9144

## **Steps to Test**

1. Open the **Menu → Dropdown Position** demo.
2. Open the Menu dropdown.
3. While it remains open, change **Position** between:
   - `bottom-end`
   - `top-end`
   - `left-end`
   - `right-end`
4. Verify the dropdown moves to the selected position without reopening.
5. Change **Arrow position** between `side` and `center`.
6. Verify the arrow updates correctly while the dropdown remains open.
7. Run the regression tests and verify they pass.

### Before 

Before: Menu dropdown position does not update dynamically

### After Changes 

After: Menu dropdown position updates dynamically

## **Self-service**

- [x] I implemented a fix for this issue.

### DIFF
--- a/packages/@docs/demos/src/demos/core/Menu/Menu.demo.positionConfigurator.tsx
+++ b/packages/@docs/demos/src/demos/core/Menu/Menu.demo.positionConfigurator.tsx
@@ -32,8 +32,8 @@ export const positionConfigurator: MantineDemo = {
     {
       prop: 'position',
       type: 'select',
-      initialValue: 'bottom',
-      libraryValue: 'bottom',
+      initialValue: 'bottom-end',
+      libraryValue: 'bottom-end',
       data: [
         { label: 'bottom', value: 'bottom' },
         { label: 'bottom-start', value: 'bottom-start' },
--- a/packages/@docs/demos/src/demos/core/Menu/Menu.demo.positionConfigurator.tsx
+++ b/packages/@docs/demos/src/demos/core/Menu/Menu.demo.positionConfigurator.tsx
@@ -32,8 +32,8 @@ export const positionConfigurator: MantineDemo = {
     {
       prop: 'position',
       type: 'select',
-      initialValue: 'bottom-end',
-      libraryValue: 'bottom-end',
+      initialValue: 'bottom',
+      libraryValue: 'bottom',
       data: [
         { label: 'bottom', value: 'bottom' },
         { label: 'bottom-start', value: 'bottom-start' },
--- a/packages/@mantine/core/src/components/Popover/Popover.test.tsx
+++ b/packages/@mantine/core/src/components/Popover/Popover.test.tsx
@@ -129,6 +129,21 @@ describe('@mantine/core/Popover', () => {
     expect(container.querySelectorAll('.mantine-Popover-arrow')).toHaveLength(0);
   });
 
+  it('updates placement when position prop changes dynamically', async () => {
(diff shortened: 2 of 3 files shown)


MILLERMARRU (commented):
Repro from #9144 confirmed, and the approach (drop the locked placement + re-measure when `position` changes) is the right idea. Some concerns with how the reset is wired:

1. The reset runs in the render body and advances refs there:
   ```
   if (previousPositionRef.current !== options.position) {
     previousPositionRef.current = options.position;
     measuredAfterShowRef.current = false;
     currentLockedPlacement = null;
     if (lockedPlacement !== null) setLockedPlacement(null);
   }
   ```
   This extends the existing `wasOpenedRef` pattern, so it is house style, but it now also applies to `position`. If React discards a render (concurrent, offscreen, a Suspense retry), `previousPositionRef.current` has already moved to the new value while the paired `setLockedPlacement(null)` never commits, so the next real render skips the block and the position change is silently dropped. Doing this in the `useDidUpdate(..., [options.position])` you already added (reset there, not in render) would be tear-safe.

2. `useDidUpdate(() => { floating.update(); }, [options.position])` closes over `floating` before it is declared. It works because the callback runs post-commit, but it is fragile next to the render-phase reset that runs first; a reader has to know the ordering.

3. The `arrowPosition` regression test: nothing in the diff targets `arrowPosition`, and a plain `arrowPosition` change (with `position` unchanged) does not hit the new `useDidUpdate([options.position])`. Does that test actually fail against `main` without this patch, or is it green either way because the fresh `middleware` array on every render already makes `@floating-ui/react` recompute? If it is the latter it is not guarding anything.

4. The bottom `useIsomorphicEffect` guard changed from `if (!lockEnabled || lockedPlacement !== null) return;` to `if (!lockEnabled) return;`, so once locked it no longer early-returns and instead falls through to the `floating.isPositioned && lockedPlacement === null` check each run. That is fine as written, but it means the effect body now executes on every `floating.placement` change for the whole open lifetime rather than just until the first lock. Worth a comment on why that is intended.

AdarshYadav9:
@MILLERMARRU
Thanks for the detailed review. I've addressed the concerns:

- Moved the position reset out of the render phase and into useDidUpdate.
- Adjusted the floating.update() handling.
- Updated the regression test as suggested.
- Added a comment explaining the useIsomorphicEffect behavior.

I've pushed the changes to the existing PR. Please take another look.

rtivital:
Thanks!

PR #9157, opened 2026-08-29
[@mantine/schedule] WeekView, DayView: Fix same-day events ending at exclusive midnight being dropped

### LINKED ISSUE #9147: @mantine/schedule: WeekView/DayView drop same-day events that end at exclusive midnight (00:00:00)
### Dependencies check up

9.5.2 (also reproduced on 9.5.1; `isEventInTimeRange` is unchanged between the two)

@mantine/schedule

Vite

All

### Describe the bug

Same-day timed events that use an **exclusive end at next-day `00:00:00`** (i.e. “runs until midnight”) are silently dropped from `WeekView` / `DayView`. They never appear in the timed grid **or** the all-day row.

This matches the usual calendar contract (`DTEND` exclusive, RFC 5545) and Mantine’s own encoding for a full day (`today 00:00:00` → `tomorrow 00:00:00` in the v9.0.0 changelog). `@mantine/schedule` already treats exclusive midnight correctly in `isMultidayEvent` (subtract a day when `end` is `00:00:00`), but `isEventInTimeRange` does not.

**Minimal events**

```ts
const events = [
  {
    id: 'until-midnight',
    title: 'Until midnight (missing)',
    start: '2026-08-20 17:15:00',
    end: '2026-08-21 00:00:00', // exclusive midnight
    color: 'gray',
  },
  {
    id: 'until-2359',
    title: 'Until 23:59 (visible)',
    start: '2026-08-20 17:15:00',
    end: '2026-08-20 23:59:00',
    color: 'blue',
  },
  {
    id: 'full-day',
    title: 'Full day (all-day row)',
    start: '2026-08-20 00:00:00',
    end: '2026-08-21 00:00:00',
    color: 'red',
  },
];
```

Render `<Schedule date="2026-08-20" defaultView="week" events={events} />` with default `startTime` / `endTime` (`00:00:00`–`23:59:59`).

**Actual**

- `until-midnight` is not rendered at all.
- `until-2359` shows as a timed block on 20 Aug.
- `full-day` shows in the all-day row.

**Expected**

`until-midnight` should render as a timed event on 20 Aug from 17:15 through the end of the day (same occupancy as `[17:15, 24:00)`). It is not multi-day and not all-day.

**Why it disappears**

`filterWeekViewEvents` keeps events only if `isEventInTimeRange` is true. For non-multiday, non-full-day events that helper compares **clock minutes only**:

```ts
eventEnd.hour() * 60 + eventEnd.minute() <= timeWindowStartMinutes
```

For `end: '2026-08-21 00:00:00'`, that is `0 <= 0` (default `startTime` is `00:00:00`), so the event is treated as ending before the visible window.

Meanwhile `isMultidayEvent` already special-cases exclusive midnight:

```ts
rawEnd.hour() === 0 && rawEnd.minute() === 0 && rawEnd.second() === 0
  ? rawEnd.startOf('day').subtract(1, 'day')
  : rawEnd.startOf('day')
```

so `17:15` → next `00:00` is classified as **same-day**. `spansFullDay` also requires start at `00:00`, so this event is neither multi-day nor full-day, and never reaches layout. Setting `withAllDaySlots` does not help.

**Possible fix**

Apply the same exclusive-midnight rule in `isEventInTimeRange` (and the day-view equivalent): if `end` is exactly `00:00:00`, treat the event as occupying through `24:00` / `23:59:59` of the previous local day when comparing to `startTime`/`endTime`. Overlap should use actual instants (or exclusive end), not “end clock minutes vs window start minutes”.

I am not volunteering a PR from this report.

giaBaoJS:
Fixes #9147.

A same-day timed event whose `end` is the next day at `00:00:00` never renders in `WeekView` or `DayView`, neither in the timed grid nor in the all-day row.

## Cause

`isEventInTimeRange` compares clock minutes, and midnight collapses to `0`.

`packages/@mantine/schedule/src/utils/is-event-in-time-range/is-event-in-time-range.ts:38`

```ts
const eventEndMinutes = eventEnd.hour() * 60 + eventEnd.minute();
```

For `end: '2026-08-21 00:00:00'` that is `0`, so line 40 reads the event as ending before the window opens (`0 <= 0` with the default `startTime` of `00:00:00`) and filters it out.

Nothing else picks the event up. `isMultidayEvent` already subtracts a day when `end` is midnight (`is-multiday-event.ts:7-8`), so the event is classified as same-day, and `spansFullDay` requires a start at `00:00` (`is-event-in-time-range.ts:14`), so it is not all-day either.

Exclusive midnight is the encoding the package already uses everywhere else:

* `is-multiday-event.ts:7-8` and `get-event-end-date.ts:7-8` both treat a midnight `end` as belonging to the previous day.
* `spansFullDay` accepts `end.isSame(nextDayStart)` (`is-event-in-time-range.ts:18`).
* The published docs demo encodes a one day holiday that way: `AgendaView.demo.allDayAndMultiday.tsx:18-19` runs from `day+1 00:00:00` to `day+2 startOf('day')`.

`isEventInTimeRange` is the only helper in the chain that does not.

## Fix

Count an end that falls exactly on the start day's next midnight as `24:00` rather than `00:00`:

```ts
const endsAtNextDayStart = eventEnd.isSame(eventStart.startOf('day').add(1, 'day'));
const eventEndMinutes = endsAtNextDayStart ? 24 * 60 : eventEnd.hour() * 60 + eventEnd.minute();
```

Comparing against the start day's next midnight, rather than testing `hour()`/`minute()` directly, leaves a degenerate `00:00` to `00:00` same-day event on the old path.

Layout needed no change. `getDayPosition` already clips `eventEnd` to `literalEnd` (`get-day-position.ts:50`), so the event renders from its start time through the end of the day.

## Call sites

`isEventInTimeRange` has four call sites, all inside this package.

* `filter-week-view-events.ts:38` and `get-day-view-events.ts:47` are the reported paths. Both are covered by new tests.
* `get-resources-day-view-events.ts:82` is the same shape as the `DayView` call.
* `get-resources-day-view-events.ts:110` receives a clipped event, so I checked it by instrumenting the call. Everything reaching it arrives with `end` clipped to `dayEnd` (`23:59:59.999`) or to an instant strictly inside the day, so `endsAtNextDayStart` is always `false` there and the behaviour is unchanged.

## Tests

* `is-event-in-time-range.test.ts`: a same-day event ending at exclusive midnight is kept; an event ending at `23:59` is still kept; an event ending at exclusive midnight but starting after the window end is still rejected.
* `filter-week-view-events.test.ts` and `get-day-view-events.test.ts`: the reported event survives the filter.
* `is-multiday-event.test.ts`: one assertion locking in that a `17:15` to next `00:00` event stays same-day, since the fix depends on that classification.

Reverting only the source change turns the three positive tests red with `Expected: true / Received: false`. The controls stay green, so they are not passing for the wrong reason.

`npx jest packages/@mantine/schedule`: 1698 tests before, 1703 after, 88 suites passing in both runs.
Full suite: 16021 tests before, 16026 after, 659 suites passing in both runs.
`tsc --noEmit`, `oxlint`, `oxfmt` and `npm run build` are clean.

### DIFF
--- a/packages/@mantine/schedule/src/components/DayView/get-day-view-events/get-day-view-events.test.ts
+++ b/packages/@mantine/schedule/src/components/DayView/get-day-view-events/get-day-view-events.test.ts
@@ -56,6 +56,21 @@ describe('@mantine/schedule/get-day-view-events', () => {
     );
   });
 
+  it('keeps same-day events that end at exclusive midnight', () => {
+    const events = [
+      testUtils.createEvent({ id: 1, start: `${testDate} 17:15:00`, end: '2024-01-16 00:00:00' }),
+    ];
+
+    expect(
+      getDayViewEvents({ events, date: testDate, startTime: '00:00:00', endTime: '23:59:59' })
+    ).toStrictEqual({
+      allDayEvents: [],
+      regularEvents: [{ ...events[0], position: expect.any(Object) }],
+      backgroundTimedEvents: [],
+      backgroundAllDayEvents: [],
+    });
+  });
+
   it('filters events based on startTime and endTime', () => {
     const events = [
       testUtils.createEvent({ id: 1, start: `${testDate} 10:00:00`, end: `${testDate} 11:00:00` }),
--- a/packages/@mantine/schedule/src/components/WeekView/get-week-view-events/filter-week-view-events.test.ts
+++ b/packages/@mantine/schedule/src/components/WeekView/get-week-view-events/filter-week-view-events.test.ts
@@ -36,6 +36,21 @@ describe('@mantine/schedule/filter-week-view-events', () => {
     ).toStrictEqual([events[1]]);
   });
 
+  it('keeps same-day events that end at exclusive midnight', () => {
+    const events = [
+      testUtils.createEvent({ id: 1, start: '2024-01-10 17:15:00', end: '2024-01-11 00:00:00' }),
+    ];
+
+    expect(
+      filterWeekViewEvents({
+        date: '2024-01-08',
+        events,
+        startTime: '00:00:00',
+        endTime: '23:59:59',
+      })
+    ).toStrictEqual(events);
+  });
+
   it('throws error when duplicated event ids are found', () => {
     const events = [
       testUtils.createEvent({ id: 1, start: '2024-01-10 10:00:00', end: '2024-01-10 11:00:00' }),
--- a/packages/@mantine/schedule/src/utils/is-event-in-time-range/is-event-in-time-range.test.ts
+++ b/packages/@mantine/schedule/src/utils/is-event-in-time-range/is-event-in-time-range.test.ts
@@ -34,6 +34,30 @@ describe('@mantine/schedule/is-event-in-time-range', () => {
     expect(isEventInTimeRange({ event, startTime: '10:00', endTime: '22:00' })).toBe(false);
   });
 
+  it('returns true for same-day event that ends at exclusive midnight', () => {
+    const event = testUtils.createEvent({
+      start: `${testUtils.testDate} 17:15:00`,
+      end: '2025-01-16 00:00:00',
+    });
+    expect(isEventInTimeRange({ event, startTime: '00:00:00', endTime: '23:59:59' })).toBe(true);
+  });
+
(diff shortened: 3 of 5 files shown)


rtivital:
Thanks!

PR #9158, opened 2026-08-29
fix(hooks): stop use-interval restarting when fn identity changes

### LINKED ISSUE #9155: useInterval: stops if `fn` changes during a render
I had an issue in an app where an interval unexpectedly would stop. It seemed to be triggered by usage of certain React components (some buttons). It turns out, Mantine's `useInterval()` was being called with an inline arrow function which caused the following effect's cleanup to run and stop the interval:

https://github.com/mantinedev/mantine/blob/8a284e2c2c53a9cb6f39f5dc389bf41b7a2073f8/packages/@mantine/hooks/src/use-interval/use-interval.ts#L66-L70

Can this be fixed? I easily fixed my usage to use a stable reference but it'd be nice if `fn` were wrapped in a `useEffectEvent()` or similar internally (e.g. via the current `useRef()`) but that if a new reference were passed in on a render that it woudln't impact the already running interval… unless this is intended behavior in which case this being called our in the document would be beneficial.

WhoamiI00:
Closes #9155.

## Problem

The restart effect listed `fn` as a dependency:

```ts
useEffect(() => {
  fnRef.current = fn;
  active && start();
  return stop;
}, [fn, active, interval]);
```

so passing an inline arrow — the usual way to call this hook — tore the interval down and recreated it on every render. The timer restarted from zero each time, and a component re-rendering faster than `interval` never reached a tick at all. That matches the report: the interval appeared to stop, and it correlated with unrelated components re-rendering.

## Why `fn` was a dependency

It is not gratuitous, which is the part worth flagging. The hook already keeps a `fnRef`, but the timer is created as:

```ts
window.setInterval(fnRef.current!, intervalValueRef.current)
```

That passes the function **by value**, so the timer keeps whatever `fnRef` held at creation and later writes to the ref have no effect. Restarting the interval was the only way to swap in a new callback. Simply dropping `fn` from the dependency array would have frozen the interval on the first render's closure — a stale-callback bug in place of a restarting one.

## Fix

Call through the ref instead, so each tick reads the current callback:

```ts
window.setInterval(() => fnRef.current(), intervalValueRef.current)
```

With that, the interval always runs the latest `fn` and `fn` no longer needs to be a dependency. `interval` remains one, since changing the period genuinely does require a new timer.

`fnRef` is now seeded with `fn` and assigned during render, mirroring how `intervalValueRef` already tracks `interval` two lines above.

## Tests

Two added:

| test | fails without the fix |
| --- | --- |
| interval is not restarted when `fn` identity changes — advance half the period, re-render, advance the other half, expect one tick | **yes** — `Expected 1, Received 0` |
| the latest `fn` is called after a re-render | no — guard |

The second passes either way by design: today the restart swaps the callback, and after this change the ref does. It is there so a fix that decoupled `fn` *without* the ref indirection — freezing the first render's closure — would fail it.

```
jest packages/@mantine/hooks   -> 63 suites, 537 passed
oxfmt / oxlint on both files   -> clean
tsc --noEmit                   -> 0 errors
```

## Note

This is a behaviour change, though I believe it is the intended one: the hook already carries a `fnRef` whose purpose is precisely to decouple the callback from the timer's lifetime — it just could not work while the function was passed by value. If you would rather keep the restart semantics and document them instead, as the issue offers as an alternative, I am happy to close this and send a docs change.

AI was used to help write this change; I have verified the behaviour and the tests myself.

### DIFF
--- a/packages/@mantine/hooks/src/use-interval/use-interval.test.ts
+++ b/packages/@mantine/hooks/src/use-interval/use-interval.test.ts
@@ -95,6 +95,50 @@ describe('@mantine/hooks/use-interval', () => {
     expect(callback).toHaveBeenCalledTimes(1);
   });
 
+  it('should not restart the interval when fn identity changes between renders', () => {
+    const fn = jest.fn();
+    // A fresh inline arrow on every render, which is how the hook is usually called.
+    const { result, rerender } = renderHook(() => useInterval(() => fn(), defaultTimeout));
+
+    act(() => {
+      result.current.start();
+    });
+
+    act(() => {
+      jest.advanceTimersByTime(defaultTimeout / 2);
+    });
+    rerender();
+    act(() => {
+      jest.advanceTimersByTime(defaultTimeout / 2);
+    });
+
+    // Restarting on the re-render would have reset the timer half way, so the tick
+    // would never arrive for a component that re-renders faster than `interval`.
+    expect(fn).toHaveBeenCalledTimes(1);
+    expect(result.current.active).toBe(true);
+  });
+
+  it('should call the latest fn after a re-render', () => {
+    const first = jest.fn();
+    const second = jest.fn();
+    const { result, rerender } = renderHook(({ fn }) => useInterval(fn, defaultTimeout), {
+      initialProps: { fn: first },
+    });
+
+    act(() => {
+      result.current.start();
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #9159, opened 2026-08-29
[@mantine/core] Rating: Fix value being changed by scroll gestures on touch devices

RobertPakko:
## Problem

`Rating` commits a new value from `touchstart`, reading it off the finger's x coordinate against the root's bounding box:

```tsx
const handleTouchStart = (event) => {
  ...
  if (!readOnly) {
    const touch = touches[0];
    setValue(getRatingFromCoordinates(touch.clientX));
  }
};
```

Two things follow from that, neither of which reproduces with a mouse or in a desktop browser's device toolbar (that emulates pointer/mouse events, not real touch events):

**1. A scroll that starts on the component changes the value.** Putting a finger on the stars to scroll the page sets a rating on the way past — the value is committed on finger-down, before the browser has decided the gesture is a scroll. On a form with several `Rating`s this reads to the user as "scrolling changed an answer I already gave", and with `allowClear` it silently zeroes one out, because landing on the star that was already selected is treated as a re-select.

**2. A single tap can be applied twice.** After the touch, the browser synthesises a click on the label, which calls `onChange` again with the same value. `handleTouchEnd` tries to swallow that click with an unconditional `event.preventDefault()`; where that does not hold, the tap is applied a second time.

On `master` the second one is usually invisible, because re-applying the same value changes nothing you can see. It becomes visible as soon as the clear logic runs on the click path — see below.

## Relationship to #9122

#9122 fixes a third, separate symptom: with `allowClear`, tapping the already-selected star did not clear it, because `handleTouchStart` called `setValue` directly and so bypassed the clear logic. Its one-line change routes that call through `handleChange`.

It does not address either defect above. The value is still committed from `touchstart`, so a scroll that starts on the stars still changes the rating, and the emulated click still applies the tap a second time.

For the second defect the one-line change makes things worse. Once the emulated click goes through `handleChange`, it meets the `allowClear` branch and resets the value to 0. So on a rating with `allowClear`, every tap whose emulated click the `preventDefault` in `touchend` fails to swallow will set the score and immediately clear it. (This is not hypothetical — it is what happens today to anyone who has worked around #9122's symptom by reimplementing the clear in their own `onChange`.)

This PR fixes #9122's symptom too, as a side effect: moving the commit to `touchend` routes the touch path through the same clear logic as clicks. That does make the two branches conflict textually — happy to rebase on top of #9122 if it lands first.

## Solution

- The value is committed on `touchend`, from `changedTouches[0]`, instead of on `touchstart`.
- The touch is discarded — no value change — if it moves further than `TOUCH_MOVE_THRESHOLD` (10px, roughly the tap slop devices use) from where it landed, if `touchcancel` arrives (what the browser fires once it takes the gesture over for scrolling), or if a second finger lands. A swipe that begins on the stars now scrolls and changes nothing.
- `handleChange` ignores changes arriving within `EMULATED_CLICK_TIMEOUT` of a handled touch, so the emulated click cannot apply the tap a second time even where `preventDefault` is not honoured.
- `preventDefault()` is now called only when the tap was actually handled, rather than on every `touchend`, so a `readOnly` rating no longer swallows the click.
- The `allowClear` logic moved into `applyValue`, so the touch path and the click path share it.

Behaviour with a mouse and with the keyboard is unchanged.

## Reproducing it

On a phone, open <https://mantine.dev/core/rating>, put a finger on the stars of the first (Usage) demo and swipe up to scroll the page. The rating jumps to wherever your finger landed.

It also reproduces on the desktop docs site without any device emulation, by dispatching the touch events directly. Paste this into the console on <https://mantine.dev/core/rating>:

```js
function gesture(root, fromFraction, dy = 0) {
  const b = root.getBoundingClientRect();
  const x = b.left + b.width * fromFraction;
  const y = b.top + b.height / 2;
  const mk = (cy) => new Touch({ identifier: 1, target: root, clientX: x, clientY: cy });
  const fire = (type, cy) => {
    const list = type === 'touchend' ? [] : [mk(cy)];
    root.dispatchEvent(new TouchEvent(type, {
      bubbles: true, cancelable: true,
      touches: list, targetTouches: list, changedTouches: [mk(cy)],
    }));
  };
  fire('touchstart', y);
  if (dy) { fire('touchmove', y + dy); }
  fire('touchend', y + dy);
}

const value = (root) => root.querySelector('input:checked')?.value ?? null;
const ratings = document.querySelectorAll('.mantine-Rating-root');

// Defect 1 — a scroll gesture that starts on the "Usage" demo
const usage = ratings[0];
console.log('before:', value(usage));
gesture(usage, 0.9, 150);
setTimeout(() => console.log('after scroll gesture:', value(usage)), 100);
```

On `master` this prints `before: 2` then `after scroll gesture: 5` — the swipe set the rating. With this PR the value is unchanged.

The same helper shows #9122's symptom on the "Allow clear" demo, which starts at 3:

```js
const allowClear = ratings[2];
gesture(allowClear, 0.5); // tap the star that is already selected
setTimeout(() => console.log('after tap on selected star:', value(allowClear)), 100);
```

On `master` this prints `3` (nothing cleared). With this PR, or with #9122, it prints `0`.

## Verifying the fix

`npm run storybook`, then open **Rating → Controlled**, which renders the current value next to the component. Running the same helper there (against `.mantine-Rating-root`, reading the value from the label rather than the checked input) gives:

| Gesture | Before | After |
| --- | --- | --- |
| `touchstart` alone | value jumps immediately | no change |
| swipe down starting on the stars | value stuck at where the finger landed | no change |
| clean tap on star 4 | 4 | 4 |
| tap with a few px of drift | 1 | 1 |
| emulated click straight after a tap | applied a second time | ignored |
| mouse click well after a touch | works | works |

## Tests

Seven tests added in `Rating.test.tsx`. Four of them fail on `master` and pass with the fix:

- `does not change value when the touch moves before it ends`
- `does not change value when the touch is cancelled`
- `ignores the click emulated by the browser after a touch tap`
- `clears value on touch tap on the same value with allowClear=true` (the case #9122 covers)

The other three (`sets value on touch tap`, `does not change value on multi touch gestures`, `does not change value on touch tap when readOnly`) pass either way and are there to pin down the behaviour that should not regress.

`npx jest packages/@mantine/core/src/components/Rating` — 74 passed. `tsc --noEmit`, `oxlint` and `oxfmt` are clean on the changed files.

### DIFF
--- a/packages/@mantine/core/src/components/Rating/Rating.test.tsx
+++ b/packages/@mantine/core/src/components/Rating/Rating.test.tsx
@@ -1,8 +1,31 @@
+import { fireEvent } from '@testing-library/react';
 import userEvent from '@testing-library/user-event';
 import { render, screen, tests } from '@mantine-tests/core';
 import { Rating, RatingProps, RatingStylesNames } from './Rating';
 
 const defaultProps: RatingProps = {};
+
+function getRoot(container: HTMLElement) {
+  const root = container.querySelector('[class*="root"]') as HTMLElement;
+  root.getBoundingClientRect = () =>
+    ({
+      left: 0,
+      right: 500,
+      width: 500,
+      top: 0,
+      bottom: 20,
+      height: 20,
+      x: 0,
+      y: 0,
+      toJSON: () => ({}),
+    }) as DOMRect;
+
+  return root;
+}
+
+function touchList(...points: [number, number][]) {
+  return points.map(([clientX, clientY]) => ({ clientX, clientY }));
+}
 
 describe('@mantine/core/Rating', () => {
   tests.itSupportsSystemProps<RatingProps, RatingStylesNames>({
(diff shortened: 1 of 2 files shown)


rtivital:
Thanks!

PR #9162, opened 2026-08-31
[@mantine/core] Textarea: Fix ResizeObserver loop error when autosize textarea width changes

### LINKED ISSUE #9161: Textarea: autosize triggers "ResizeObserver loop completed with undelivered notifications" when the textarea width changes (since 9.4.0)
### Dependencies check up

9.5.2

@mantine/core

Vite

Chromium (Chrome 140 / Playwright headless-shell); the same pattern was reported for Firefox in #9121.

### Describe the bug

Since **9.4.0** (the release that replaced `react-textarea-autosize` with the in-house `TextareaAutosize`, 841fea7), `<Textarea autosize />` raises

```
ResizeObserver loop completed with undelivered notifications
```

as a window `error` event whenever the textarea's **width** changes — e.g. it lives in a resizable sidebar, a `Splitter` pane, or the viewport is resized.

Cause, in `packages/@mantine/core/src/components/Textarea/Autosize.tsx`: the `ResizeObserver` observes the `<textarea>` itself and, when its `offsetWidth` changed, calls `resizeTextarea()` synchronously inside the callback. `resizeTextarea()` writes `node.style.height`, which changes the size of the very element whose notification is being delivered. Per spec the browser cannot deliver that second observation in the same cycle and reports the loop error (layout settles on the next frame, so nothing is visibly wrong — but it's an uncaught error for `window.onerror`, error-tracking tools, and test runners that fail on page errors).

Bisect with only `@mantine/*` varied in an otherwise identical app: 9.1.1 ✅, 9.2.0 ✅, 9.3.0 ✅ (still `react-textarea-autosize`), **9.4.0 ❌**, 9.5.2 ❌.

Minimal reproduction (no codesandbox — the error only surfaces on width change, which the snippet triggers with a CSS-resizable box; drag the handle horizontally with the console open):

```tsx
import { Box, Textarea } from '@mantine/core';

export function Demo() {
  return (
    <Box style={{ resize: 'horizontal', overflow: 'auto', width: 400, padding: 16 }}>
      <Textarea
        autosize
        minRows={2}
        maxRows={8}
        placeholder="A placeholder long enough to wrap onto a second line once the box gets narrower"
      />
    </Box>
  );
}
```

Steps: open DevTools console → drag the box's resize handle left until the placeholder wraps → `ResizeObserver loop completed with undelivered notifications` is logged (Chrome shows it as an error; `window.addEventListener('error', …)` receives it).

_No response_ (inline reproduction above)

### Possible fix

Defer the height write out of the observer callback, the same way `ScrollArea/use-resize-observer.ts` already does:

```ts
let frame = 0;
const observer = new ResizeObserver(() => {
  if (libRef.current && libRef.current.offsetWidth !== widthRef.current) {
    widthRef.current = libRef.current.offsetWidth;
    cancelAnimationFrame(frame);
    frame = requestAnimationFrame(resizeTextarea);
  }
});
observer.observe(node);
return () => {
  cancelAnimationFrame(frame);
  observer.disconnect();
};
```

Verified in our app: with this change applied to the built `Autosize.mjs`, a Playwright test that resizes a sidebar containing an autosize `Textarea` (which failed deterministically on stock 9.4.0–9.5.2) passes with no page errors.

### Self-service

nathanpond:
Fixes #9161

## Summary

Since 9.4.0 (`TextareaAutosize` replacing `react-textarea-autosize`), `<Textarea autosize />` raises `ResizeObserver loop completed with undelivered notifications` as a window `error` whenever the textarea's width changes (resizable sidebar, `Splitter` pane, viewport resize).

The `ResizeObserver` in `Autosize.tsx` observes the `<textarea>` itself and, when its width changed, calls `resizeTextarea()` synchronously inside the callback. That writes `style.height` on the very element whose notification is being delivered, which the browser reports as the loop error (the second observation is delivered next frame, so there is no visible defect — but it is an uncaught error for `window.onerror`, error trackers, and test runners that fail on page errors).

## Fix

Defer the height write to `requestAnimationFrame`, cancelled on cleanup — the same pattern `ScrollArea/use-resize-observer.ts` already uses. Width tracking (`widthRef`) is unchanged, so the observer still only reacts to width changes.

## Verification

- Bisect with only `@mantine/*` varied: 9.3.0 ✅ → 9.4.0 ❌ → 9.5.2 ❌.
- With this change applied to the built `Autosize.mjs`, a Playwright test that keyboard-resizes a sidebar containing an autosize `Textarea` — which failed deterministically on stock 9.4.0–9.5.2 with the loop error — passes with no page errors; autosize behaviour (min/max rows, growth on typing) is unchanged.

### DIFF
--- a/packages/@mantine/core/src/components/Textarea/Autosize.tsx
+++ b/packages/@mantine/core/src/components/Textarea/Autosize.tsx
@@ -212,15 +212,26 @@ export function TextareaAutosize({
 
     widthRef.current = node.offsetWidth;
 
+    // The observer watches the textarea itself; writing its height inside
+    // the callback changes the size of an element whose notification is
+    // being delivered, which browsers report as
+    // "ResizeObserver loop completed with undelivered notifications"
+    // (a window `error` event). Defer the write to the next frame instead.
+    let frame = 0;
+
     const observer = new ResizeObserver(() => {
       if (libRef.current && libRef.current.offsetWidth !== widthRef.current) {
         widthRef.current = libRef.current.offsetWidth;
-        resizeTextarea();
+        cancelAnimationFrame(frame);
+        frame = requestAnimationFrame(resizeTextarea);
       }
     });
 
     observer.observe(node);
-    return () => observer.disconnect();
+    return () => {
+      cancelAnimationFrame(frame);
+      observer.disconnect();
+    };
   }, []);
 
   useEffect(() => {
--- a/packages/@mantine/core/src/components/Textarea/Autosize.tsx
+++ b/packages/@mantine/core/src/components/Textarea/Autosize.tsx
@@ -212,11 +212,6 @@ export function TextareaAutosize({
 
     widthRef.current = node.offsetWidth;
 
-    // The observer watches the textarea itself; writing its height inside
-    // the callback changes the size of an element whose notification is
-    // being delivered, which browsers report as
-    // "ResizeObserver loop completed with undelivered notifications"
-    // (a window `error` event). Defer the write to the next frame instead.
     let frame = 0;
 
     const observer = new ResizeObserver(() => {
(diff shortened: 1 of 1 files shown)


rtivital:
Thanks!

PR #9164, opened 2026-09-01
[@mantine/hooks] use-hotkeys: Fix preventDefault not being applied when other hotkey options are set

giaBaoJS:
## Problem

`useHotkeys` and `getHotkeyHandler` resolve hotkey item options with a destructuring default on the whole object:

```ts
([hotkey, handler, options = { preventDefault: true, usePhysicalKeys: false }]) => {
```

That default only applies when the third tuple element is absent. As soon as any options object is passed, `options.preventDefault` is `undefined` and `event.preventDefault()` is never called, even though `preventDefault` is meant to default to `true` (the option was added in #3044 to opt out of `preventDefault` always running).

The `use-hotkeys` documentation is affected by this. It shows `{ preventDefault: false }` as the way to opt out, and in the same example passes `{ usePhysicalKeys: true }` on its own:

```tsx
const hotkeys: HotkeyItem[] = [
  ['mod+J', () => {}, { preventDefault: false }],
  ['D', () => {}, { usePhysicalKeys: true }],
];
```

The second item silently loses `preventDefault`, so the browser default action for the key still runs.

## Solution

Resolve each option separately instead of defaulting the whole object:

```ts
const { preventDefault = true, usePhysicalKeys = false } = options || {};
```

Applied in both `useHotkeys` and `getHotkeyHandler`.

## Testing

Added tests to `use-hotkeys.test.tsx` and `parse-hotkey.test.ts` covering all three cases for both entry points: options omitted, a partial options object (`{ usePhysicalKeys: true }`), and an explicit `{ preventDefault: false }`.

Reverting the source change with the tests in place fails exactly the two partial options cases and leaves the other four passing.

- `npx jest packages/@mantine/hooks` - 63 suites, 551 tests passing
- `npx jest packages/@mantine/spotlight` - 44 tests passing (`Spotlight` is the in-repo `useHotkeys` consumer)
- `npm run typecheck`, `npm run build`, `npx oxlint`, `npm run format:write:files`

### DIFF
--- a/packages/@mantine/hooks/src/use-hotkeys/parse-hotkey.test.ts
+++ b/packages/@mantine/hooks/src/use-hotkeys/parse-hotkey.test.ts
@@ -1,4 +1,4 @@
-import { getHotkeyMatcher, parseHotkey } from './parse-hotkey';
+import { getHotkeyHandler, getHotkeyMatcher, parseHotkey } from './parse-hotkey';
 
 describe('@mantine/hooks/use-hot-key/parse-hotkey', () => {
   it('should parse hotkey correctly', () => {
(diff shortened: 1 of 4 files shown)


rtivital:
Thanks!

PR #9167, opened 2026-09-03
[@mantine/lightbox] Lightbox: Fix ESM import throwing on a circular dependency between Lightbox and LightboxProvider

### LINKED ISSUE #9166: [@mantine/lightbox] Circular dependency causes ESM import to throw
### Dependencies check up

9.6.0

@mantine/lightbox

Tanstack Start

Chrome

### Describe the bug

Importing `@mantine/lightbox@9.6.0` in a TanStack Start application throws before rendering:

```text
ReferenceError: Cannot access 'LightboxProviderComponent' before initialization
```

It can also be reproduced with the published ESM build:

```bash
node --input-type=module -e "import('@mantine/lightbox')"
```

The package contains this runtime dependency cycle:

```text
Lightbox.mjs
→ LightboxProvider.mjs
→ Lightbox.mjs
```

During module initialization, `Lightbox.mjs` accesses the provider before it has been initialized:

```js
Lightbox.Provider = LightboxProviderComponent;
```

Expected: the package can be imported normally.

Actual: the ESM module throws before the application renders.

_No response_

### Possible fix

Prefer removing the runtime cycle. For example, attach `Lightbox.Provider` in the public entry module after both components have initialized, or extract a base Lightbox component that does not import the provider.

A lazy getter avoids the TDZ as a minimal workaround, but keeps the circular dependency:

```js
Object.defineProperty(Lightbox, 'Provider', {
  get: () => LightboxProviderComponent,
});
```

### Self-service

solomonaustin:
Fixes #9166.

## Problem

`@mantine/lightbox@9.6.0` cannot be imported at all through its ESM build. The reporter's repro fails on a clean install:

```console
$ node --input-type=module -e "import('@mantine/lightbox')"
ReferenceError: Cannot access 'LightboxProviderComponent' before initialization
```

The package has a runtime import cycle:

```
Lightbox.tsx  ──▶  LightboxProvider.tsx  ──▶  Lightbox.tsx
```

`LightboxProvider` needs `Lightbox` to render it, and `Lightbox` needs `LightboxProviderComponent` to attach `Lightbox.Provider`. That second edge is read **at module scope**, so whichever module the bundler enters first decides whether the package works.

In the published `esm/index.mjs` the emitted import order puts `LightboxProvider` before `Lightbox`, so evaluation goes `LightboxProvider` → `Lightbox` → back into the still-initializing `LightboxProvider`, and `Lightbox.Provider = LightboxProviderComponent` hits the TDZ. Source order in `index.ts` happens to be the other way around, which is why tests and dev never caught it — the bug only exists in the built artifact.

## Fix

The cycle is structurally forced as long as one module both defines `Lightbox` and attaches `.Provider`, so this moves the component into `LightboxBase.tsx` and leaves `Lightbox.tsx` as the composition point:

```
Lightbox.tsx ──▶ LightboxProvider.tsx ──▶ LightboxBase.tsx
      │                                        ▲
      └────────────────────────────────────────┘
```

No cycle, and evaluation order no longer matters. `LightboxBase.tsx` keeps its type-only import of `LightboxProviderComponent` for the `staticComponents` entry, which is erased at build time.

I kept the `Lightbox.Provider` assignment inside `Lightbox.tsx` rather than moving it to `index.ts` or to the provider module, for two reasons: `sideEffects` is `["*.css"]`, so a cross-module mutation could be tree-shaken away and silently leave `Lightbox.Provider` undefined; and the existing tests, stories and demos import `Lightbox` from `./Lightbox` directly, so the static has to be attached by that module.

Public API, types and `displayName`s are unchanged.

## Tests

Two module-initialization tests in `Lightbox.test.tsx`. The first evaluates the provider module before `Lightbox`, reproducing the bundler's order — reverting the fix makes it fail with the exact production error:

```
● attaches Lightbox.Provider when the provider module is evaluated first
  ReferenceError: Cannot access 'LightboxProviderComponent' before initialization
    at Object.<anonymous> (packages/@mantine/lightbox/src/Lightbox.tsx:64:45)
```

## Verification

- `npx jest packages/@mantine/lightbox` — 182 tests, 8 suites, all pass
- Regression test confirmed to fail with the fix reverted and pass with it restored
- `npm run typecheck` — 399 errors before and after on this tree, none in lightbox (the tree's docs-app errors are pre-existing, from packages not being prebuilt)
- `npx oxlint` and `oxfmt` clean on the changed files
- `npm run build` — rebuilt the package and reran the reporter's command against the real artifact:

```console
✅ ESM OK — Lightbox.Provider = function | displayName = @mantine/lightbox/LightboxProvider
✅ CJS OK — Lightbox.Provider = function | displayName = @mantine/lightbox/LightboxProvider
```

All statics still present on the built export: `Caption, CloseButton, Navigation, Provider, Root, Slide, Slides, Thumbnails, Toolbar, classes, close, displayName, extend, next, open, prev, setIndex, withProps`. Walking the built `esm/` import graph now reports 0 cycles.


rtivital:

PR #9170, opened 2026-09-04
[@mantine/hooks] use-mask: Fix transform not being applied to the initial input value

giaBaoJS:
## Problem

`useMask` applies the `transform` option when it formats a raw value (`applyMaskToRaw`), but not when it formats the value that is already on the input element at the moment the ref is attached (`processInput`). `processInput` never accepted a `transform` parameter at all.

Because `initializeInputValue` uses `processInput`, any character that only matches the mask pattern *after* `transform` runs is silently dropped on mount, even though typing the exact same character works.

```tsx
// Renders an empty input
<MaskInput mask="AAA" transform={(char) => char.toUpperCase()} defaultValue="abc" />

// Typing "abc" into the same input renders "ABC"
```

Partial values are truncated rather than dropped:

```tsx
// Renders "(123) " instead of "(123) ABC"
<MaskInput mask="(999) AAA" transform={(char) => char.toUpperCase()} defaultValue="(123) abc" />
```

This affects uncontrolled forms in particular, where `defaultValue` is the only way to seed the field.

## Fix

`processInput` now takes the same optional `transform` argument as `applyMaskToRaw` and applies it at the same point, right before the slot pattern test. `initializeInputValue` passes it through.

The parameter is only passed from `initializeInputValue`, the one call site whose input is untransformed DOM text. The other call sites receive values that `applyMaskToRaw` has already transformed, so they are left untouched and there is no risk of a non-idempotent `transform` running twice.

## Tests

Three tests were added and each was verified to fail before the fix:

- `use-mask.test.ts`: `applies transform to the initial input value` (`""` before the fix, `"ABC"` after)
- `use-mask.test.ts`: `applies transform to the initial input value that contains literals` (`"(123) "` before, `"(123) ABC"` after)
- `MaskInput.test.tsx`: `applies transform to defaultValue on mount` (empty before, `"ABC-123"` after)


giaBaoJS:
Rebased onto master, conflict resolved.

#9196 reshaped `initializeInputValue` with the `isOwnValue` / `parse` split, so the fix moved into that shape: `transform` is resolved alongside `initialSlots` and passed only to the `processInput` branch of `parse`. The `applyMaskToRaw(rawValueRef.current, ...)` branch is left as #9196 wrote it, since that path reads a raw value the hook already transformed and running `transform` again would double-apply a non-idempotent one.

The bug is still present on current master: `displayValueRef` is empty on mount, so `isOwnValue` is false for a `defaultValue` and the `processInput` path is the one that runs.

All three tests still fail without the source change and pass with it. Full jest suite: 16764 passed.

rtivital:
Thanks!

PR #9173, opened 2026-09-05
fix(notifications): avoid transition shorthand in stacked layout

### LINKED ISSUE #9168: Stacked notifications mix transition shorthand with transitionDelay
### Dependencies check up

9.6.0

@mantine/notifications

Tanstack Start

Firefox

### Describe the bug

When a Notification is using `layout="stacked"`, sliding it gives this error in the console:

> Updating a style property during rerender (transition) when a conflicting property is set (transitionDelay) can lead to styling bugs. To avoid this, don't mix shorthand and non-shorthand properties for the same value; instead, replace the shorthand with separate values.

_No response_

### Possible fix

`NotificationContainer` sets both on the same inline style object when `layout === "stacked"`:

```
transition: isDragging
  ? "none"
  : `transform ${transitionDuration}ms cubic-bezier(.51,.3,0,1.21), opacity ${transitionDuration}ms ease`,
transitionDelay: isDragging || isExiting ? "0ms" : `${staggerDelay}ms`,
```

React treats transition as the shorthand for transitionDelay. Updating one while the other is already set is exactly what that warning is for.

Default layout already does this the right way in get-notification-state-styles.ts: transitionProperty, transitionDuration, and transitionTimingFunction as separate longhands. Stacked layout should do the same and keep transitionDelay next to them.

### Self-service

kavyansh18:
Fixes #9168

The stacked notification layout currently uses the `transition` shorthand together with `transitionDelay`, which triggers a React warning.

This replaces the shorthand with the corresponding transition properties while preserving the existing transition and staggered delay behavior.

Added test coverage for the stacked notification transition styles.

### DIFF
--- a/packages/@mantine/notifications/src/NotificationContainer.tsx
+++ b/packages/@mantine/notifications/src/NotificationContainer.tsx
@@ -337,9 +337,11 @@ export function NotificationContainer({
           zIndex: (stackSize || 5) - (stackIndex ?? 0),
           pointerEvents: isCollapsed ? ('none' as const) : undefined,
           alignSelf: stackDirection === 1 ? ('start' as const) : ('end' as const),
-          transition: isDragging
-            ? 'none'
-            : `transform ${transitionDuration}ms cubic-bezier(.51,.3,0,1.21), opacity ${transitionDuration}ms ease`,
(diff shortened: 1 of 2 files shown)


rtivital:

PR #9174, opened 2026-09-05
[@mantine/notifications] Add stacked drag-cancel regression coverage

ahamSel:
Regression coverage for #9168.

The transition warning was fixed upstream in a0ad393896a101503dc95e93ca33375ae247619b and released in 9.6.1. I rebased this PR and removed the overlapping implementation changes. The diff now only adds a drag/cancel regression test; it does not change production code or the existing exit-animation test.

The test starts a drag on a stacked notification, cancels it, checks that the offset resets without dismissing either notification, and checks that React does not report a transition style conflict. I verified that it fails with the original transition shorthand and passes with the upstream fix.

## Testing

- Notification package: 96 tests passed.
- Full Jest suite with `TZ=UTC`: 16,704 tests passed across 690 suites; one snapshot passed.
- Package and documentation typechecks passed.
- Package build and changed-file lint/format checks passed.
- Separate Codex CLI review found no actionable issues.

Local checks used Node 24.11.0. The full suite used UTC because unrelated schedule tests depend on that timezone. This update is test-only and does not add new visual behavior.

I used AI to help write and review the regression test.

### DIFF
--- a/packages/@mantine/notifications/src/Notifications.test.tsx
+++ b/packages/@mantine/notifications/src/Notifications.test.tsx
@@ -601,6 +601,59 @@ describe('@mantine/core/Notifications', () => {
     expect(Number(behind.style.zIndex)).toBeLessThan(Number(front.style.zIndex));
   });
 
+  it('preserves stacked transitions without style conflicts when dragging is canceled', () => {
+    jest.useFakeTimers();
+    const store = createNotificationsStore();
+    const consoleError = jest.spyOn(console, 'error');
+
+    render(
+      <Notifications
+        store={store}
+        withinPortal={false}
+        autoClose={false}
+        transitionDuration={100}
+        layout="stacked"
+      />
+    );
+
+    act(() => {
+      notifications.show({ id: 'drag-stacked-a', message: 'First stacked' }, store);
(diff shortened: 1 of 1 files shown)


ahamSel:
Thanks for fixing this in 9.6.1. I've rebased the PR onto current master and kept the upstream implementation unchanged. The remaining diff only adds the drag/cancel regression test.

I checked that the test fails with the original transition shorthand and passes with the upstream fix. All 96 notification tests and the full 16,704-test suite pass, along with typechecks, the package build, and lint/format checks. I've updated the title and description to reflect the test-only scope.

rtivital:
Thanks for putting this together and for the careful investigation! Unfortunately, I cannot merge it – the new test does not catch anything that the current test suite does not already catch:

- The #9168 regression is already covered on master: the `keeps stacked styling while a stacked notification is exiting` test asserts longhand `transitionProperty`/`transitionDuration` values and fails if the shorthand `transition` is reintroduced.
- The `transitionDelay === '0ms'` assertions can never fail – the dragged notification is the front one (`stackIndex` 0), so its stagger delay is always `0ms` regardless of drag state.
- The remaining assertions (offset resets and nothing is dismissed on cancel) go through the same end-of-drag path as the existing short-drag test.

Thanks again, and also for spotting the spacing issue in #9176!

Issue #7349, opened 2025-01-13
Unrecognized prop error in @mantine/charts


paddyvypr:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

^7.15.1

### What package has an issue?

@mantine/core

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

When using a BarChart and passing in `textColor` props I receive the following error from React:

Warning: React does not recognize the textColor prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase textcolor instead. If you accidentally passed it from a parent component, remove it from the DOM element.

I suspect this is likely the same problem seen in [here](https://github.com/mantinedev/mantine/issues/6800) [and here](https://github.com/mantinedev/mantine/pull/6810) in which props were being unintentionally spread across an element that can't accept them.

### If possible, include a link to a codesandbox with a minimal reproduction

https://codesandbox.io/p/sandbox/mantine-react-template-forked-yqwscq

### Possible fix

The same fix applied for `gridColor` [here](https://github.com/mantinedev/mantine/issues/7288) 

### Self-service

- [ ] I would be willing to implement a fix for this issue

rtivital:
It was fixed in 7.15.3

dcentrih:
There's also the gridColor prop with the same issue. Package version 7.16.0.

Issue #7288, opened 2024-12-19
Unrecognized Prop Error in @mantine/charts


r-walsh:
### Dependencies check up

- [X] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

^7.15.1

### What package has an issue?

@mantine/charts

### What framework do you use?

Next.js

### In which browsers you can reproduce the issue?

All

### Describe the bug

When using a `BarChart` and passing in `gridColor` and/or `textColor` props I receive the following error from React:

> Warning: React does not recognize the `gridColor` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `gridcolor` instead. If you accidentally passed it from a parent component, remove it from the DOM element.

I suspect this is likely the same problem seen in [this issue](https://github.com/mantinedev/mantine/issues/6800) (and its [related PR](https://github.com/mantinedev/mantine/pull/6810)) in which props were being unintentionally spread across an element that can't accept them.

I was originally on version `7.7.2`, upgraded all mantine packages to `7.15.1` while looking for a fix and was still able to reproduce.

### If possible, include a link to a codesandbox with a minimal reproduction

https://codesandbox.io/p/sandbox/mantine-react-template-forked-yqwscq

### Possible fix

Assuming my intuition about the cause is correct, I expect a fix will follow the same pattern as the [PR](https://github.com/mantinedev/mantine/pull/6810) linked above on a similar issue.

### Self-service

- [ ] I would be willing to implement a fix for this issue

rtivital:
Fixed in [7.15.2](https://github.com/mantinedev/mantine/releases/tag/7.15.2)

r-walsh:
@rtivital Thanks for the quick response! However it looks like `textColor` is still showing the issue

PR #9181, opened 2026-09-07
[@mantine/charts] LineChart, CompositeChart, BubbleChart: Fix `textColor` and `gridColor` props being passed as attributes to the DOM node

coldsmirk:
`textColor` (and `gridColor` in `CompositeChart` and `BubbleChart`) is read by the vars resolver but never taken out of `props`, so it stays in `...others` and lands on the root `div` as an attribute. React logs this on every render of a chart that sets it:

```
React does not recognize the `textColor` prop on a DOM element. If you intentionally want it to appear in the DOM as a custom attribute, spell it as lowercase `textcolor` instead.
```

Same thing as #7288 / #7349, which fixed `BarChart` in 7.15.2 and 7.15.3, and #7378 for `AreaChart`. `LineChart`, `CompositeChart` and `BubbleChart` still have it; every other chart already pulls the two props out before spreading the rest. Ran into it with `LineChart` on 9.6.0.

The fix is the same one-liner as before. I also added a small test to each of the three files that renders with both props and checks the root element does not carry them – it fails on `master` without the change.


rtivital:
Thanks!

Issue #7615, opened 2025-03-22
[core] SegmentedControl: adding&selecting a new value causes FloatingIndicator to disappear


BATCOH:
### Dependencies check up

- [x] I have verified that I use latest version of all @mantine/* packages

### What version of @mantine/* packages do you have in package.json?

7.17.2

### What package has an issue?

@mantine/core

### What framework do you use?

Vite

### In which browsers you can reproduce the issue?

Chrome

### Describe the bug

Steps to reproduce:
1. Use controlled SegmentedControl
2. Add a new item to data list
3. Set this new item as SegmentedControl value at the same time
FloatingIndicator indicator will disappear

![Image](https://github.com/user-attachments/assets/877298bf-ca91-446e-93e4-4f08329e0201)

### If possible, include a link to a codesandbox with a minimal reproduction

https://codesandbox.io/p/sandbox/d6485g

### Possible fix

As a workaround data change and value change can be separated (using timeout), as I did in codesandbox example. In this case, everything works correctly.
Or a `key` property with `data.length` can be added to the `SegmentedControl` component.

I think it can be fixed in mantine's code by re-creating SegmentedControl `refs` reference when the new `val` appears ([here](https://github.com/mantinedev/mantine/blob/master/packages/%40mantine/core/src/components/SegmentedControl/SegmentedControl.tsx#L180-L184)).
In any other case `refs` reference should be preserved to avoid endless update loop.
Something like that:

```typescript
const setElementRef = (element: HTMLElement | null, val: string) => {
  if (refs[val]) {
    refs[val] = element;
    setRefs(refs);
  } else {
    setRefs((refs) => ({ ...refs, [val]: element }));
  }
};
```

I'm willing to do a PR with a fix if that approach seems appropriate.

### Self-service

- [x] I would be willing to implement a fix for this issue

rtivital:
Fixed in [7.17.3](https://github.com/mantinedev/mantine/releases/tag/7.17.3)

PR #9182, opened 2026-09-07
fix(SegmentedControl): fix indicator not updating when data changes with a newly selected value

### LINKED ISSUE #9175: SegmentedControl loses its indicator when data is replaced by a list of the same length
### Dependencies check up

9.6.0

@mantine/core

Vite

All

### Describe the bug

A controlled `SegmentedControl` loses its indicator when it is given a different `data` array that happens to have the same number of items as the previous one (with `value` set to something from the new array). The selected label is drawn in the selected text colour with nothing behind it, and it stays like that until something unrelated re-renders the component. In my app that "something" was a websocket message a second or two later, so it looked like the indicator was animating in late.

I ran into it with a header that shows a "last 10 / 30 / 60 minutes" control on one tab and a "last 7 / 30 / 90 days" control on another. Both lists have three entries, so React keeps the same `SegmentedControl` instance and just feeds it the other list.

Minimal version:

```tsx
const MINUTES = ['10', '30', '60'];
const DAYS = ['7', '30', '90'];

function Demo() {
  const [days, setDays] = useState(false);
  return (
    <>
      <Button onClick={() => setDays((v) => !v)}>swap</Button>
      <SegmentedControl data={days ? DAYS : MINUTES} value={days ? '7' : '10'} onChange={() => {}} />
    </>
  );
}
```

Click "swap": the indicator is gone. Trigger any other re-render of `Demo` and it comes back. Make one of the lists four entries long instead of three and it works, because that is the case #7615 covered.

What I think is going on, from `SegmentedControl.tsx`: the indicator gets remounted through `setKey` in a `useShallowEffect` that depends on `data.length` only (the #7615 fix), and label refs are collected with

```ts
const setElementRef = (element, val) => {
  refs[val] = element;
  setRefs(refs);
};
```

`setRefs(refs)` hands React the same object, so it bails out and nothing re-renders. The render that swaps `data` passes `target={refs['7']}` to `FloatingIndicator`, which is `undefined` at that point, and `FloatingIndicator` returns `null` without a target. The new labels then register `refs['7']` from their ref callbacks, but since that setState is a no-op there is no render in which the target is present — until some other state change happens to cause one.

No sandbox, the snippet above is the whole reproduction. It also shows up in a jsdom test: right after the swap `.mantine-SegmentedControl-indicator` is not in the DOM, after a forced re-render it is.

### Possible fix

Either of these works locally:

- key the indicator on the values instead of the count: `useShallowEffect(() => setKey(randomId()), [_data.map((item) => item.value)])` — `useShallowEffect` already compares that array element by element;
- or make registering a new value a real state update, something like `if (!(val in refs)) setRefs({ ...refs, [val]: element })`, which is roughly what #7615 suggested in the first place.

On the consumer side, giving the control a `key` that changes with the data set works around it.

### Self-service

shahharis98:
## What this fixes

Fixes #9175

When `SegmentedControl` is controlled and its `data` prop is replaced with a
new array of the same length, the `FloatingIndicator` can fail to update or
disappear entirely — particularly when a newly-added value is selected in
the same update as the data change.

## Root cause

`setElementRef` mutated the `refs` object in place and then called
`setRefs(refs)` with that same object reference:

\`\`\`js
const setElementRef = (element: HTMLElement | null, val: string) => {
  refs[val] = element;
  setRefs(refs);
};
\`\`\`

Because React bails out of re-rendering when `setState` is called with a
value that's reference-equal to the current state, this mutation never
actually triggered a re-render. Ref registrations for newly-mounted items
silently updated the underlying object but never propagated to a new render,
so `FloatingIndicator`'s `target` prop could point at a stale or missing
element.

## Fix

`setElementRef` now only calls `setRefs` when a ref's value has actually
changed, and does so with a genuinely new object so React re-renders:

\`\`\`js
const setElementRef = (element: HTMLElement | null, val: string) => {
  if (refs[val] !== element) {
    setRefs((current) => ({ ...current, [val]: element }));
  }
};
\`\`\`

I also found that simply fixing the mutation surfaced a second issue: the
`ref` callback passed to each label was an inline arrow function
(`ref={(node) => setElementRef(node, val)}`), which gets a new identity on
every render. React detaches and reattaches refs whenever the callback's
identity changes, even if the underlying DOM node hasn't — this caused an
infinite render loop once the mutation bug above was fixed.

To fix this, each label's ref callback is now cached per value using
`useRef`, so the same function instance is reused across renders unless the
value itself changes:

\`\`\`js
const refCallbacks = useRef<Record<string, (node: HTMLElement | null) => void>>({});

const getElementRef = (val: string) => {
  if (!refCallbacks.current[val]) {
    refCallbacks.current[val] = (node: HTMLElement | null) => {
      setRefs((current) => {
        if (current[val] === node) return current;
        return { ...current, [val]: node };
      });
    };
  }
  return refCallbacks.current[val];
};
\`\`\`

## Testing

- Manually reproduced the bug in Storybook using a controlled
  `SegmentedControl` where a data swap simultaneously selects a
  newly-added value — indicator failed to position correctly before this
  fix, works correctly after.
- Verified this doesn't regress the related fix in #7615 (adding + selecting
  a new item), tested via the existing `SelectedItemRemoved` story.
- Added a regression test in `SegmentedControl.test.tsx` covering the
  data-swap + new-value-selection scenario.

## Type of change

- [x] Bug fix (non-breaking change which fixes an issue)

### DIFF
--- a/packages/@mantine/core/src/components/SegmentedControl/SegmentedControl.story.tsx
+++ b/packages/@mantine/core/src/components/SegmentedControl/SegmentedControl.story.tsx
@@ -184,3 +184,25 @@ export function ScaledAncestor() {
     </div>
   );
 }
+export function BugRepro() {
+  const [value, setValue] = useState('a');
+  const [dataSet, setDataSet] = useState(0);
+
+  const dataSets = [
+    [{ value: 'a', label: 'Apple' }, { value: 'b', label: 'Banana' }, { value: 'c', label: 'Cherry' }],
+    [{ value: 'a', label: 'Apricot' }, { value: 'x', label: 'Xigua' }, { value: 'c', label: 'Cranberry' }],
+  ];
+
+  return (
+    <div style={{ padding: 40 }}>
+      <SegmentedControl
+        value={value}
+        onChange={setValue}
+        data={dataSets[dataSet]}
+      />
+      <button onClick={() => setDataSet((d) => (d + 1) % 2)} style={{ marginTop: 20 }}>
+        Swap data (same length, different values)
+      </button>
+    </div>
+  );
+}
--- a/packages/@mantine/core/src/components/SegmentedControl/SegmentedControl.tsx
+++ b/packages/@mantine/core/src/components/SegmentedControl/SegmentedControl.tsx
@@ -1,4 +1,4 @@
-import { useState } from 'react';
+import { useState , useRef } from 'react';
 import {
   randomId,
   useId,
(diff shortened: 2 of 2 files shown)


rtivital:

PR #9188, opened 2026-09-09
[@mantine/core] Modal, Drawer: Handle Escape once per stack

### LINKED ISSUE #9185: Drawer.Stack / Modal.Stack: one Escape closes two layers when the top layer was rendered before the one under it
### Dependencies check up

9.6.0 (`@mantine/core`, `@mantine/hooks`), React 19.2.8

@mantine/core

Other, I will specify in the bug description: a plain `createRoot` app bundled by Bun. Nothing framework-specific is involved; the reproduction below drops into the codesandbox template's `App.tsx`.

Chrome (tested in headless Chrome). The mechanism is standard event dispatch order, so I expect all browsers.

### Describe the bug

The docs for `Drawer.Stack` and `Modal.Stack` say the stack "disables focus trap and `Escape` key handling for all drawers except the one that is currently opened" and that z-index is managed "disregarding their order in the DOM". In one common layout a single `Escape` closes **two** stacked layers instead of the top one:

1. The layer on top was rendered (mounted) **before** the layer under it. Typical case: an app-level drawer rendered once by the root, opened from a button inside a page-level drawer.
2. The top layer's close commit schedules another React update. Rendering the body only while the layer is open does that as soon as the body holds anything with a state-setter ref, for example Mantine's own `ScrollArea`.

With both, pressing `Escape` closes the top layer and the layer under it. Without 1, the docs demo order, it works. Without 2, it happens to work because of React's flush timing (details below).

Steps with the code below:

1. Click "Open b".
2. Inside drawer b, click "Open a on top". Drawer a is now on top, drawer b is hidden under it.
3. Press `Escape`.

Expected: `stack.state` is `{"a":false,"b":true}`, drawer b is back on top.
Actual: `stack.state` is `{"a":false,"b":false}`, both drawers are closed.

Doing the same in the docs' order ("Open a", then "Open b on top", `Escape`) closes only b. Replacing `Drawer` with `Modal` and `useDrawersStack` with `useModalsStack` gives the same result.

```tsx
import { Button, Drawer, Group, ScrollArea, useDrawersStack } from '@mantine/core';

// The Drawer.Stack docs demo with two changes: each drawer renders a ScrollArea while it is
// open, and drawer "b" can open drawer "a" on top of itself, so "a" (rendered first) is opened last.
  const stack = useDrawersStack(['a', 'b']);
  return (
    <>
      <Drawer.Stack>
        <Drawer {...stack.register('a')} title="Drawer a">
          {stack.state.a ? <ScrollArea.Autosize mah="50vh">Drawer a body</ScrollArea.Autosize> : null}
          <Button onClick={() => stack.open('b')}>Open b on top</Button>
        </Drawer>
        <Drawer {...stack.register('b')} title="Drawer b">
          {stack.state.b ? <ScrollArea.Autosize mah="50vh">Drawer b body</ScrollArea.Autosize> : null}
          <Button onClick={() => stack.open('a')}>Open a on top</Button>
        </Drawer>
      </Drawer.Stack>
      <Group>
        <Button onClick={() => stack.open('a')}>Open a</Button>
        <Button onClick={() => stack.open('b')}>Open b</Button>
      </Group>
      <pre>{JSON.stringify(stack.state)}</pre>
    </>
  );
}
```

What happens, as far as I could trace it:

- Every `Modal`/`Drawer` closes itself from its own `window` capture `keydown` listener (`useModal` via `useWindowEvent`), registered when the component mounts. Listeners on the same target and phase run in registration order, so the layer rendered first handles the key first, whichever layer is on top.
- The stack gates that listener with `closeOnEscape: ctx.currentId === stackId`, a value from the last render.
- When the top layer (rendered first) closes, React commits the `opened={false}` render in the microtask checkpoint the browser runs between two listeners of the same event. When that commit schedules another update (here the `ScrollArea` viewport's ref callback runs with `null` because the body unmounts), React also flushes the pending passive effects, so `Drawer`'s `useEffect` calls `ctx.removeModal(stackId)`, the stack re-renders, and the layer underneath becomes `currentId`.
- That layer's listener runs next, on the same `keydown`, now with `closeOnEscape: true`, and closes it too.

I verified the timing with a capture listener registered before React mounts and one registered after: by the time the second Mantine listener runs, the lower layer has already lost `data-hidden`. A synthetic `dispatchEvent` of the same key (no microtask checkpoints between listeners) closes only the top layer.

No sandbox account; the file above is the whole reproduction and replaces `App.tsx` in the template.

### Possible fix

Decide who handles a given `Escape` per event rather than per render. For example, the stack context could remember the event a member consumed (a `WeakSet<Event>`), and `useModal` could skip a `keydown` already consumed by another member of the same stack. Deferring `removeModal` out of the key dispatch would also work but is less direct.

Workaround in user land: close a stacked layer on the next tick, `onClose={() => setTimeout(close, 0)}`, so one key dispatch reaches every listener with the stack it started with.

### Self-service

mnkprs:
Fixes #9185

When the top modal or drawer mounts before the layer underneath it, closing it can flush React updates between the window's Escape listeners. The lower layer then becomes active and closes on the same keypress.

Each stack now tracks handled Escape events and records the event before calling `onClose`. One Escape closes only the top layer, and the next keypress closes the layer beneath it. Separate stacks handle events independently, and the event continues to reach other listeners.

Added regression tests for both opening orders, successive keypresses, separate stacks, event propagation, and the existing composition/stop-propagation guards. The tests flush close effects between native listeners to reproduce browser timing in jsdom; eight cases fail when the fix is removed.

Testing:

- Reproduced the original conditional `ScrollArea.Autosize` example in Chromium for both Modal and Drawer. Both opening orders pass with the fix.
- All 566 Modal and Drawer tests pass.
- `TZ=UTC npm test` passes: 16,699 tests across 689 suites, plus typechecks for packages and docs, linting, formatting, and dependency consistency checks.
- All 30 packages build successfully with `npm run build`.

### DIFF
--- a/packages/@mantine/core/src/components/Drawer/Drawer.tsx
+++ b/packages/@mantine/core/src/components/Drawer/Drawer.tsx
@@ -94,6 +94,7 @@ export const Drawer = factory<DrawerFactory>((_props) => {
           closeOnEscape: ctx.currentId === stackId,
           trapFocus: ctx.currentId === stackId,
           zIndex: ctx.getZIndex(stackId),
+          __handledEscapeEvents: ctx.handledEscapeEvents,
         }
       : {};
 
--- a/packages/@mantine/core/src/components/Drawer/DrawerStack.test.tsx
+++ b/packages/@mantine/core/src/components/Drawer/DrawerStack.test.tsx
@@ -1,23 +1,23 @@
 import { act, fireEvent, screen } from '@testing-library/react';
 import { render } from '@mantine-tests/core';
-import { Modal } from './Modal';
-import { useModalsStack } from './use-modals-stack';
+import { useDrawersStack } from '../Modal/use-modals-stack';
+import { Drawer } from './Drawer';
 
 function TestContainer() {
-  const stack = useModalsStack(['a', 'b']);
+  const stack = useDrawersStack(['a', 'b']);
 
   return (
     <>
-      <Modal.Stack>
+      <Drawer.Stack>
         {(['a', 'b'] as const).map((id) => (
-          <Modal
+          <Drawer
             key={id}
             {...stack.register(id)}
             onClose={() => act(() => stack.close(id))}
             title={id}
           />
         ))}
-      </Modal.Stack>
+      </Drawer.Stack>
       <button type="button" onClick={() => stack.open('a')}>
         Open a
       </button>
@@ -33,7 +33,7 @@ function pressEscape() {
   document.body.dispatchEvent(new KeyboardEvent('keydown', { key: 'Escape', bubbles: true }));
 }
 
-describe('@mantine/core/Modal.Stack', () => {
+describe('@mantine/core/Drawer.Stack', () => {
   it.each([
     ['a', 'b'],
     ['b', 'a'],
--- a/packages/@mantine/core/src/components/Drawer/DrawerStack.tsx
+++ b/packages/@mantine/core/src/components/Drawer/DrawerStack.tsx
@@ -8,6 +8,7 @@ interface DrawerStackContext {
   getZIndex: (id: string) => string;
   currentId: string;
   maxZIndex: string | number;
+  handledEscapeEvents: WeakSet<KeyboardEvent>;
 }
 
 export const DrawerStackContext = createContext<DrawerStackContext | null>(null);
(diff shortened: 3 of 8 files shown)


rtivital:
Thanks!

svyatov:
Thank you!

PR #9193, opened 2026-09-16
[@mantine/core] NumberInput: Fix clamping with min and max when decim…

### LINKED ISSUE #9192: NumberInput clamping with `min` and `max` values does not work when `decimalScale` is set
### Dependencies check up

9.6.1

@mantine/core

Next.js

All

### Describe the bug

If the `decimalScale` prop on NumberInput is set to a value other than 0 (e.g. 2), the input will no longer clamp the values within the set min and max on blur. 

I assume it may be related to values being treated as strings when `decimalScale` is set.

Edit: The above only occurs when combined with the `trimLeadingZeroesOnBlur` prop equal to `false`.

https://codesandbox.io/p/devbox/mantine-react-template-forked-fn5k8k?workspaceId=ws_AHRzzGSwfeHFAZY6aSFMD1

### Possible fix

_No response_

### Self-service

Adityazzzzz:
Fixes #9192

### What was the issue?
When `NumberInput` values are kept as strings (which happens when `decimalScale` is used with `fixedDecimalScale`, trailing zeros, or formatted inputs), `handleBlur` was only running the clamp check when `typeof sanitizedValue === 'number'`. Because of this, values stored as strings bypassed clamping to `min` and `max` on blur.

### What changed?
- In `handleBlur`, added handling for string values when `clampBehavior === 'blur'` by parsing the value with `parseFloat` and clamping it against `min` and `max`.
- Added unit tests in `NumberInput.test.tsx` verifying blur clamping behavior with `decimalScale` and `fixedDecimalScale`.

### Testing
- `npm run jest packages/@mantine/core/src/components/NumberInput/NumberInput.test.tsx` passes.
- Linting and formatting verified.


codeCraft-Ritik (commented):
Great catch finding that typeof sanitizedValue === 'number' was bypassing string values during blur events.

Is this same typeof check present in the clampBehavior === 'always' or onChange pipeline? It would be good to verify that strings with fixed decimals or trailing zeros don't bypass clamping under clampBehavior === 'always' as well

rtivital:
Thanks for the PR! The underlying issue (#9192) has already been fixed on master in d68920e564 and released in 9.6.2. That fix goes through a shared blur-sanitizing helper, so it also covers trailing decimal separator and `clampBehavior="none"` cases.

For reference: the bug only reproduces with `trimLeadingZeroesOnBlur={false}` – the tests in this PR do not set it, so they pass with or without the source change. The PR also conflicts with the refactored `handleBlur` now.

Closing as already resolved – thanks again for looking into it!

PR #9194, opened 2026-09-16
[@mantine/core] Fix active click hit area at top edge (#9064)

### LINKED ISSUE #9064: Clicking at the very top edge of the button doesn't trigger the button's action

Adityazzzzz:
Fixes #9064

### What was the issue?
When elements with `.mantine-active` enter the `:active` state, `transform: translateY(rem(1px))` moves the element downward by 1px. When clicking at the topmost boundary pixel ($y = 0$), this 1px shift causes the element to move out from underneath the cursor. Consequently, `mouseup` occurs outside the element, causing browsers (notably Firefox and Safari) to cancel the `click` event.

### What changed?
- Added an invisible `&::after` pseudo-element on `.mantine-active:active` bridging the 1px shift gap at the top (`top: rem(-1px)`, `height: rem(1px)`).
- Added `overflow: visible` during `&:active` on `Button` and `ActionIcon` so the active `::after` hit area is not clipped by `overflow: hidden`.
- Verified `mouseup` registers on the element at the top edge and dispatches the `click` event properly.

### Testing
- `npm run stylelint` passes with 0 errors.
- `npm run typecheck` passes across all packages and apps.
- `yarn jest` passed for Button and ActionIcon test suites.


rtivital:
Thanks for the PR and for investigating this! Unfortunately, I cannot merge it – extending the hit area with a global `::after` pseudo-element and toggling `overflow` while pressed introduces side effects (content spilling out of fixed-width buttons, children losing rounded corners, overriding users' own `::after` styles).

I've closed #9064 as not planned and posted a workaround there – `theme.activeClassName` can be used to replace the default translate with an active animation that does not move the element.

PR #9195, opened 2026-09-16
[@mantine/core] Fix subtle and transparent variants color in dark mod…

### LINKED ISSUE #8895: Subtle and transparent colors do not match version 8

Adityazzzzz:
Fixes #8895

### What was the issue?
In Mantine v9, `--mantine-color-{color}-light-color` was updated for dark mode to resolve to `shade-0` (near white) to ensure contrast against the dark background of `variant="light"`. 

However, `subtle` and `transparent` variants have transparent backgrounds in their rest state. Because `defaultVariantColorsResolver` was using `light-color` for them, icons and text in components like `ActionIcon`, `Button`, `Badge`, and `NavLink` displayed as white in dark mode, losing their theme color identity (e.g. `orange` appearing white).

### What changed?
- Updated `defaultVariantColorsResolver` so that `subtle` and `transparent` variants (when no shade is specified) use `var(--mantine-color-${color}-text)` instead of `light-color`.
- In light mode, `--mantine-color-{color}-text` resolves to filled (`shade-6`). In dark mode, it resolves to `shade-4`, matching `outline`'s vibrant color identity and legibility on dark backgrounds.
- Added regression unit tests for `defaultVariantColorsResolver`.

### Testing
- `yarn jest packages/@mantine/core/src/core/MantineProvider/color-functions/default-variant-colors-resolver/default-variant-colors-resolver.test.ts` passes.
- `npx oxlint` and `npm run format:write:files` applied.
- `npm run typecheck` passed with 0 errors across all packages and apps.


rtivital:
Thanks for the PR! Unfortunately, I cannot accept it – colors are being revamped in the upcoming 10.0 version and #8895 will be addressed as part of that work. Changing variant colors now would also affect light mode (for example, `subtle` gray goes from gray-9 to gray-6 with lower contrast), so we'd rather not change defaults before the new color system lands.

PR #9204, opened 2026-09-18
[@mantine/core] use-mantine-color-scheme: Preserve FloatingIndicator transitions during theme change (#9203)

### LINKED ISSUE #9203: SegmentedControl does not animate when switching color schema
### Dependencies check up

9.6.1

@mantine/core

Next.js

All

### Describe the bug

I have a SegmentedControl based component to let user decide on the color schema. I realized it doesn't animate the transition while all the other SegmentedControls work as expected. Here is the code I used:

```typescript
"use client";

import { Center, SegmentedControl, useComputedColorScheme, useMantineColorScheme } from "@mantine/core";
import { IconMoonStars, IconSun } from "@tabler/icons-react";
import type { FC } from "react";

export const ColorSchemaSwitch: FC = () => {
  const { setColorScheme } = useMantineColorScheme();
  const colorScheme = useComputedColorScheme("light", { getInitialValueInEffect: true });

  return (
    <SegmentedControl
      value={colorScheme}
      onChange={(value) => setColorScheme(value as "light" | "dark")}
      data={[
        {
          value: "light",
          label: (
            <Center>
              <IconSun />
            </Center>
          ),
        },
        {
          value: "dark",
          label: (
            <Center>
              <IconMoonStars />
            </Center>
          ),
        },
      ]}
    />
  );
};

```

_No response_

### Possible fix

I'm not sure what's the root cause and what does a potential solution look like but I can take a look if needed.

### Self-service

GabrielBaiano:
Hello there!

Fixes #9203

### Problem
When `SegmentedControl` is used to toggle color schemes (calling `setColorScheme` in `onChange`), the `FloatingIndicator` jumps to the active item without animation.

This happens because `disableTransition` injects `*, *::before, *::after { transition: none !important; }` into `<head>` to prevent color flashes during theme change, which also kills `FloatingIndicator`'s CSS `transform` transition on the same frame.

### Solution
- Added `preserve-transition` mod to `FloatingIndicator` (`data-preserve-transition="true"`).
- Updated `disableTransition` in `use-mantine-color-scheme` to exempt `[data-preserve-transition]` elements and their pseudo-elements from `transition: none !important`.
- Added unit tests in `FloatingIndicator.test.tsx` and `use-mantine-color-scheme.test.tsx`.

### Tests
- `yarn jest packages/@mantine/core/src/components/FloatingIndicator`
- `yarn jest packages/@mantine/core/src/core/MantineProvider/use-mantine-color-scheme`
- `yarn jest packages/@mantine/core/src/components/SegmentedControl`
- `npx oxlint` & `npx tsc --noEmit` pass with 0 errors.


rtivital:
Thanks for the PR! Unfortunately, I cannot merge it – disabling all transitions during color scheme change is intended behavior (see my reply in #9203). It prevents inconsistent animations and color flashes while CSS variables are switched.

If you want transitions to run during color scheme change, use the supported option – `useMantineColorScheme({ keepTransitions: true })` in the component that calls `setColorScheme`, see https://mantine.dev/theming/color-schemes/#transitions-during-color-scheme-change. With it, the SegmentedControl indicator animates as expected.

Adding an exception only for FloatingIndicator would introduce a new data attribute and a more complex global selector for a single component, which makes the documented rule inconsistent when there is already an option for this.

PR #9207, opened 2026-09-19
[@mantine/form] use-form: Use structural sharing in setPath and skip window listeners on unnamed forms

qorvr:
## Problem

1. In `@mantine/form`, `setPath` previously relied on `klona/full` to deep-clone the entire form values object on every field update. For larger forms or forms with complex/nested state trees, cloning the entire object graph on each keystroke adds memory and garbage-collection overhead while breaking object reference stability for untouched subtrees.
2. `useFormActions` registered 15 global `window` event listeners unconditionally for every form instance, even when `name` was not provided. When rendering forms without a `name`, dozens of unused event listeners were attached to `window`.

## Fix

1. Replaced the full object clone in `setPath` with path-based structural sharing: only objects along the updated path are shallow-copied, while untouched sibling properties and branches keep their existing references.
2. Made `useFormActions` skip attaching `window` listeners when `name` is not set.

## Tests

- Added test in `set-path.test.ts` verifying that untouched subtrees maintain reference equality after path updates.
- Added tests in `packages/@mantine/form/src/tests/use-form/actions.test.ts` verifying that unnamed forms do not register `window` listeners, and named forms continue to register and handle actions properly.
- All 69 `@mantine/form` test suites pass (428 tests).


rtivital:
Thanks for the PR! Unfortunately, I cannot merge it – replacing `klona/full` with spread-based structural sharing in `setPath` is a regression for non-plain values. `klona` preserves types of everything it copies, while `{ ...current }` converts any container on the path into a plain object:

| Values | Path | master | this PR |
| --- | --- | --- | --- |
| `new FormModel()` (class instance as form values) | `name` | `FormModel` | `Object` |
| `{ price: new Money(5) }` | `price.amount` | `Money` (methods preserved) | `Object` (prototype lost) |
| `{ date: new Date(0) }` | `date.foo` | `Date` | `Object` |
| `{ tags: new Set([1]) }` | `tags.foo` | `Set` | `Object` |
| `{ map: new Map([['a', 1]]) }` | `map.foo` | `Map` | `Object` |

It also changes other behavior – missing intermediate paths are now created instead of ignored, and paths through primitives spread strings into objects.

The unnamed forms listener change is a valid fix though – forms without `name` currently subscribe to `mantine-form:undefined:*` events, so `createFormActions('undefined')` affects all of them. Feel free to submit it as a separate PR without the `setPath` changes.

